PR #4, opened 2018-09-07
Logstore work to create secret and CR for elasticsearch-operator

ewolinetz:


pkg/k8shandler/logstore.go
```diff
@@ -6,20 +6,88 @@ import (
...
+    Type: "Opaque",
+    Data: map[string][]byte{
+      "elasticsearch.key": getFileContents("/tmp/_working_dir/elasticsearch.key"),
+      "elasticsearch.crt": getFileContents("/tmp/_working_dir/elasticsearch.crt"),
+      "logging-es.key": getFileContents("/tmp/_working_dir/logging-es.key"),
+      "logging-es.crt": getFileContents("/tmp/_working_dir/logging-es.crt"),
+      "admin-key": getFileContents("/tmp/_working_dir/system.admin.key"),
+      "admin-cert": getFileContents("/tmp/_working_dir/system.admin.crt"),
+      "admin-ca": getFileContents("/tmp/_working_dir/ca.crt"),
+      "passwd.yml": createPasswdFile("prometheus", getRandomWord(16)),
```
ewolinetz: should the username be configurable?

ewolinetz: @t0ffel how can we get this info to ES?

pkg/k8shandler/logstore.go
```diff
@@ -6,20 +6,88 @@ import (
...
+  "io/ioutil"
+
   metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
   sdk "github.com/operator-framework/operator-sdk/pkg/sdk"
   logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
 )
 
-func CreateOrUpdateLogStore(logging *logging.ClusterLogging)(error) {
-  return nil
+func CreateOrUpdateLogStore(logging *logging.ClusterLogging) error {
```
jcantrill: I'm a little late to evaluating this code, but I think we should consider restructuring to be less functional and more OO.   The method might be attached to the `ClusterLogging`  allowing its use like `logging.CreateOrUpdateLogStore` which I believe is inline with origin code a facilitates mocking and testing.  Not sure fully who to speak to about that.

pkg/k8shandler/logstore.go
richm: yes, but perhaps not for the initial commit

pkg/k8shandler/logstore.go
ewolinetz: I'm not following.. we are calling this in such a way from `stub/handler.go`:
`k8shandler.CreateOrUpdateLogStore(logging)` 
where `k8shandler` is the package and name we are using when importing.

jcantrill: I meant something like this: https://github.com/openshift/origin/blob/release-3.9/pkg/oc/bootstrap/docker/openshift/ansible.go#L146

where the 'thing' being acted upon is 'clusterlogging' instead of 'ansibleRunner' .  It's semantics but you can, for instance, with things that require certs may have the same similar abstractions akin to an interface allowing you do do something like:

` collector.addCerts` or `logstore.addCerts`

I think this will become more useful when we add an alternate collector where kinds need similar functionality but are slightly tweaked.  This example is contrived, and I'm not suggesting we need to modify this PR to address my suggestion but we should consider in future how a functional design vs OO may benefit us.

ewolinetz: ah, i think get what you're saying, ack.

yeah I think the reason I just have it like this now is because my intention is we eventually move away from managing our own certs and leverage the cert signing service. If that isn't feasible I could see is refactoring things to have the different components have a function to request/generate their certs.

pkg/k8shandler/logstore.go
ewolinetz: @jcantrill for now I'm going to leave this generation in here and if you're able to update the es plugin to do prometheus auth with a token then we can remove this.

Thoughts?

PR #5, opened 2018-09-10
Kibana work

ewolinetz:


pkg/k8shandler/visualization.go
```diff
@@ -0,0 +1,49 @@
...
+
+  proxySecret := utils.Secret(
+    "logging-kibana-proxy",
+    logging.Namespace,
+    map[string][]byte{
+      "oauth-secret": utils.GetRandomWord(64),
+      "session-secret": utils.GetRandomWord(32),
+      "server-key": utils.GetFileContents("/tmp/_working_dir/kibana-internal.key"),
+      "server-cert": utils.GetFileContents("/tmp/_working_dir/kibana-internal.crt"),
+//      "server-tls.json": utils.GetFileContents("/tmp/_working_dir/server-tls.json"),
```
ewolinetz: @josefkarasek does the oauth-proxy still make use of this file?

jcantrill: I don't believe so.  The configuration for the proxy is here and I don't see a similar file. https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_kibana/templates/kibana.j2#L98-L110

pkg/k8shandler/certificates.go
```diff
@@ -13,8 +13,7 @@ import (
   metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-func extractMasterCertificate(namespace string, secretName string)(error) {
-
+func extractFromSecret(namespace string, secretName string, key string, toFile string) error {
```
jcantrill: consider renaming to something like `extractSecretToFile` since we are not directly returning the value for our use

pkg/k8shandler/certificates.go
```diff
@@ -27,37 +26,39 @@ func extractMasterCertificate(namespace string, secretName string)(error) {
...
 
   // check to see if the map value exists
   if ! ok {
-    logrus.Fatalf("No secret data \"masterca\" found")
+    logrus.Fatalf("No secret data \"%s\" found", key)
     return nil
   }
 
-  err = ioutil.WriteFile("/tmp/_working_dir/ca.crt", caValue, 0644)
+  err = ioutil.WriteFile("/tmp/_working_dir/" + toFile, value, 0644)
```
jcantrill: * refactor `/tmp/_working_dir/` to a constant.  
* Consider using [path](https://golang.org/pkg/path/) lib to join file name and parent

pkg/k8shandler/logstore.go
```diff
@@ -6,20 +6,56 @@ import (
...
+  err := sdk.Create(esSecret)
+  if err != nil && !errors.IsAlreadyExists(err) {
+    logrus.Fatalf("Failure constructing Elasticsearch secret: %v", err)
+  }
+
   return nil
 }
 
-func CreateOrUpdateCR(logging *logging.ClusterLogging)(error) {
+func createOrUpdateESCR(logging *logging.ClusterLogging) error {
```
jcantrill: for readability consider renaming to expanding `CR` to CustomResource

pkg/utils/utils.go
```diff
@@ -0,0 +1,157 @@
...
+
+  "k8s.io/api/core/v1"
+  apps "k8s.io/api/apps/v1"
+  metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+  route "github.com/openshift/api/route/v1"
+)
+
+func GetFileContents(filePath string) []byte {
+  contents, err := ioutil.ReadFile(filePath)
+  if err != nil {
```
jcantrill: Do we need to return the err or is it sufficient to return a nil contents?

pkg/utils/utils.go
```diff
@@ -0,0 +1,157 @@
...
+
+func GetRandomWord(wordSize int) []byte {
+  b := make([]rune, wordSize)
+  for i := range b {
+    b[i] = letters[rand.Intn(len(letters))]
+  }
+  return []byte(string(b))
+}
+
+func Secret(secretName string, namespace string, data map[string][]byte) *v1.Secret {
```
jcantrill: Consider naming these functions like `newSecret` which I believe is the pattern to identify a 'constructor' like method.

pkg/utils/utils.go
```diff
@@ -0,0 +1,157 @@
...
+  }
+}
+
+// loggingComponent = kibana
+// component = kibana{,-ops}
+func Deployment(deploymentName string, namespace string, loggingComponent string, component string, podSpec v1.PodSpec) *apps.Deployment {
+  return &apps.Deployment{
+    TypeMeta: metav1.TypeMeta{
+      Kind: "Deployment",
+      APIVersion: "extensions/v1beta1",
```
jcantrill: Deployments still in beta for 4.0?

pkg/k8shandler/logstore.go
ewolinetz: I think to follow the naming pattern I've reused I'll expand `ES` to `Elasticsearch`

pkg/utils/utils.go
ewolinetz: We could do that, I'll be honest that I didn't focus much on error handling for this first pass

ewolinetz:
/hold

waiting on #4

pkg/utils/utils.go
ewolinetz: I would argue that this isn't a constructor though, rather we are returning an structure.

pkg/utils/utils.go
ewolinetz: according to the API, yes

pkg/utils/utils.go
ewolinetz: Going to defer a solution on this until after all component PRs have merged in

ewolinetz:
@jcantrill PTAL, i added e2e scaffolding. I'd like to keep it a separate commit since it is different than the Kibana work feature.

PR #6, opened 2018-09-11
fix LOG-175. Add Dockerfile and make for building operator

jcantrill:
Adds a makefile for building:

* binary
* image
*cleaning


Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ENV APP_HOME=/opt/app/src \
+    GOPATH=/go \
+    PATH="${PATH}:${GOPATH}/bin:/usr/local/bin" \
+    INSTALL_PKGS="golang make git"
+
+RUN yum install -y ${INSTALL_PKGS} --setopt=tsflags=nodocs
+
+WORKDIR /tmp/cluster-logging-operator
+COPY vendor/ /tmp/cluster-logging-operator/vendor
+COPY scripts/* /usr/local/bin
```
ewolinetz: this is called with a relative path to the operator, I don't think we should move it to the bin path nor make it executable (its called with `bash cert_generation.sh`

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+
+ENV APP_HOME=/opt/app/src \
+    GOPATH=/go \
+    PATH="${PATH}:${GOPATH}/bin:/usr/local/bin" \
+    INSTALL_PKGS="golang make git"
+
+RUN yum install -y ${INSTALL_PKGS} --setopt=tsflags=nodocs
+
+WORKDIR /tmp/cluster-logging-operator
+COPY vendor/ /tmp/cluster-logging-operator/vendor
```
ewolinetz: do we need to copy these? won't they be in our current directory when we `RUN make`?

Dockerfile
jcantrill: The operator binary is moved to `/usr/local/bin` so it will be relative to the binary.  I can remove executable bit if you think that's better; I don't have a strong opinion

Dockerfile
jcantrill: The build happens in context of the Dockerfile so all the code needs to be copied there.  We are able to `make` outside of the context of Docker and `make image` inside the context of Docker.  The instructions in the Dockerfile will remove all build sources and build tools.

Dockerfile
ewolinetz: ack, so can we copy it to `/usr/local/bin/scripts/` then? so this way I don't need to update the operator code to use a different path from what it does currently which is `scripts/cert_generation.sh`.

ewolinetz: Also, I'm alright with leaving the permissions change off. This way someone isn't tempted to just run it?

jcantrill: Updated

PR #8, opened 2018-09-17
[WIP] Fluentd work

ewolinetz:


pkg/k8shandler/collection.go
```diff
@@ -0,0 +1,129 @@
...
+        VolumeMounts: []v1.VolumeMount{
+          {Name: "runlogjournal", MountPath: "/run/log/journal"},
+          {Name: "varlog", ReadOnly: true, MountPath: "/var/log"},
+          {Name: "varlibdockercontainers", ReadOnly: true, MountPath: "/var/lib/docker"},
+          {Name: "config", ReadOnly: true, MountPath: "/etc/fluent/configs.d/user"},
+          {Name: "certs", ReadOnly: true, MountPath: "/etc/fluent/keys"},
+          {Name: "dockerhostname", ReadOnly: true, MountPath: "/etc/docker-hostname"},
+          {Name: "localtime", ReadOnly: true, MountPath: "/etc/localtime"},
+          {Name: "dockercfg", ReadOnly: true, MountPath: "/etc/sysconfig/docker"},
+          {Name: "originnodecfg", ReadOnly: true, MountPath: "/etc/origin/node"},
```
richm: we will not need this anymore

ewolinetz:
/hold

waiting on #5

ewolinetz:
/hold cancel
@richm @jcantrill I tested this and verified we see the api objects created correctly, this is ready for review again.

pkg/k8shandler/collection.go
```diff
@@ -0,0 +1,166 @@
+package k8shandler
```
jcantrill: do we want to use 'collector' instead of 'collection' to distinguish it from confusion with set/list/collection

pkg/k8shandler/collection.go
```diff
@@ -0,0 +1,166 @@
...
+  // Currently we cannot do validation checking on this due to version issues with the scheduling api and the version we create the priorityclass with
+  /*err := sdk.Create(collectionPriorityClass)
+  if err != nil && !errors.IsAlreadyExists(err) {
+    logrus.Fatalf("Failure creating Collection priority class: %v", err)
+  }*/
+
+  return nil
+}
+
+func createOrUpdateFluentdServiceAccount(logging *logging.ClusterLogging) error {
```
jcantrill: createOrUpdateCollectorServiceAccount ?

pkg/k8shandler/collection.go
```diff
@@ -0,0 +1,166 @@
...
+
+  fluentdServiceAccount := utils.ServiceAccount("aggregated-logging-fluentd", logging.Namespace)
```
jcantrill: collector in lieu of aggregated-logging-fluentd  ?

pkg/k8shandler/collection.go
```diff
@@ -0,0 +1,166 @@
...
+
+  err := sdk.Create(fluentdConfigMap)
+  if err != nil && !errors.IsAlreadyExists(err) {
+    logrus.Fatalf("Failure constructing Fluentd configmap: %v", err)
+  }
+
+  return nil
+}
+
+func createOrUpdateFluentdSecret(logging *logging.ClusterLogging) error {
```
jcantrill: createOrUpdateCollectorSecret  would this secret be different between fluent and rsyslog?

pkg/k8shandler/collection.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func fluentdPodSpec(logging *logging.ClusterLogging) v1.PodSpec {
+  return v1.PodSpec{
+    NodeSelector: map[string]string {
+      "logging-infra-fluentd": "true",
+    },
+    PriorityClassName: "cluster-logging",
+    Containers: []v1.Container{
+      {Name: "fluentd", Image: "openshift/origin-logging-fluentd:latest", ImagePullPolicy: v1.PullIfNotPresent,
```
jcantrill: Maybe not this PR, but doesnt this need to come from a manifest... or possibly someplace we can override?

pkg/k8shandler/collection.go
ewolinetz: potentially, yes. its just following the naming pattern of our other functions as we could be creating a secret for either fluentd or rsyslog

pkg/k8shandler/collection.go
ewolinetz: this is addressed a bit more in #11

pkg/k8shandler/collection.go
ewolinetz: per the discussion with @richm on IRC, we would want to be explicit about the service account name, as it is more straight forward and easier when it comes to docs/support

pkg/k8shandler/collection.go
ewolinetz: its specific to fluentd

pkg/k8shandler/collection.go
ewolinetz: given the context, i dont think it will be confused with it. It matches with our spec struct name.

PR #11, opened 2018-09-19
Api rework

ewolinetz:
Depends on #10


pkg/apis/logging/v1alpha1/types.go
```diff
@@ -24,46 +24,66 @@ type ClusterLogging struct {
 
 type ClusterLoggingSpec struct {
   AllInOne          bool `json:allinone`
```
jcantrill: We are removing this right?

ewolinetz: Did we come to a conclusion on how we would accomplish telling the cluster to send logs to two different destinations (and stand up the required LogStores if necessary)? To me, this says we need to have multiple Visualizations, LogStores, and Curations and the Collection is going to need to use different end points (as opposed to them being the same)

ewolinetz: If we are unable to do a separate operations deployment for 4.0 though, would that be considered a regression? 

The reason I don't think being able to have named pipelines would be enough is that we don't have a way to say "give me two different ES clusters" and that was what `AllInOne` was meant to accomplish.

ewolinetz: Per the discussion on slack, going to make this functionality be provided by an annotation instead

deploy/operator.yaml
```diff
@@ -2,6 +2,8 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: cluster-logging-operator
+  annotations:
+    clusterlogging.alpha/allinone: ""
```
jcantrill: I think this needs to be namespaced similar to other annotations (i.e build) like: `io.openshift.clusterlogging.alpha` or somethign like that.

ewolinetz:
Also need to add owner refs to our api objects

Gopkg.toml
```diff
@@ -14,38 +14,35 @@ required = [
...
 [[constraint]]
   name = "github.com/operator-framework/operator-sdk"
   # The version rule is used for a specific release and the master branch for in between releases.
   branch = "master"
   # version = "=v0.0.5"
 
-[[constraint]]
-  branch = "master"
-  name = "github.com/ViaQ/elasticsearch-operator"
+[[override]]
```
ewolinetz: this will need to be changed back after https://github.com/openshift/elasticsearch-operator/pull/4 merges in

pkg/utils/utils.go
```diff
@@ -249,6 +382,11 @@ func GetInt64(value int64) *int64 {
   return &i
 }
 
+func GetString(value string) *string {
```
ewolinetz: i can remove this... its no longer needed

pkg/k8shandler/collection.go
```diff
@@ -101,65 +116,71 @@ func createOrUpdateFluentdDaemonset(logging *logging.ClusterLogging) error {
...
+func getFluentdPodSpec(logging *logging.ClusterLogging, elasticsearchAppName string, elasticsearchInfraName string) v1.PodSpec {
+
+  fluentdContainer := utils.Container("fluentd", v1.PullIfNotPresent, logging.Spec.Collection.FluentdSpec.Resources)
+
+  fluentdContainer.Env = []v1.EnvVar{
+    {Name: "MERGE_JSON_LOG", Value: "true"},
+    {Name: "K8S_HOST_URL", Value: "https://kubernetes.default.svc"},
+    {Name: "ES_HOST", Value: elasticsearchAppName},
+    {Name: "ES_PORT", Value: "9200"},
+    {Name: "ES_CLIENT_CERT", Value: "app-cert"},
```
ewolinetz: these need to be updated to be full path from where the secret is mounted

PR #13, opened 2018-10-05
LOG-198. Add manifests

jcantrill:
This PR adds manifests to the image to resolve: https://jira.coreos.com/browse/LOG-198

### DIFF
--- a/manifests/01-namespace.yaml
+++ b/manifests/01-namespace.yaml
@@ -1,0 +1,8 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-logging
+  annotations:
+    openshift.io/node-selector: ""
+  labels:
+    openshift.io/cluster-logging: "true"
--- a/manifests/image-references
+++ b/manifests/image-references
@@ -1,0 +1,20 @@
+kind: ImageStream
+apiVersion: image.openshift.io/v1
+spec:
+  tags:
+  - name: cluster-logging-operator
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/cluster-logging-operator:latest
+  - name: logging-kibana5
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/logging-kibana5:latest
+  - name: logging-curator5
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/logging-curator5:latest
+  - name: logging-fluentd
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/logging-fluentd:latest
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,8 @@ FROM openshift/origin-release:golang-1.10 AS builder
 
 WORKDIR /tmp/cluster-logging-operator
 COPY vendor/ /tmp/cluster-logging-operator/vendor
+COPY manifests/ /manifests/
+COPY scripts/* /usr/local/bin/scripts/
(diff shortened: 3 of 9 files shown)


jcantrill:
cc @smarterclayton look correct?

manifests/01-namespace.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: openshift-logging
+  annotations:
+    openshift.io/node-selector: ""
+  labels:
+    openshift.io/cluster-logging: "true"
```
ewolinetz: should we add an empty node selector here? The daemonset doesn't play nice if there are additional node selector labels from the project (per @thoraxe )

ewolinetz:
looks like test failed because it didn't have the reference image for the cluster logging operator

jcantrill:
@ewolinetz This was from last week but I don't know what exactly it means.

```
Clayton Coleman [4:19 PM]
@jcantril you asked to replace “quay.io/openshift/origin-cluster-logging-operator:latest” in your manifests, but when we looked at your manifests, we didn’t see that string


Jeff Cantrill [4:20 PM]
https://github.com/openshift/cluster-logging-operator/pull/13/files#diff-fa8eaa615c05d654eac0582e117f1986R8

Jeff Cantrill [4:25 PM]
@Clayton Coleman im not sure from 'where' i asked for that replacement

Clayton Coleman [4:26 PM]
image-references tells the payload to look at the other files in the directory and perform a find and replace on the value of the “name” field on the image stream tag at build time with whatever the latest pull spec should be
let me look at your PR more closely
oh
no input image
have you published cluster-logging-operator yet?
```

manifests/01-namespace.yaml
deads2k: > should we add an empty node selector here? The daemonset doesn't play nice if there are additional node selector labels from the project (per @thoraxe )

Specify the same selector on your daemonset and you'll have no trouble.  And the controller should no longer destroy the cluster when it's denied (since 3.10 I think)

jcantrill: @deads2k  You mean here [1]? or apply the same annotation to the Daemonset?

[1] https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/collection.go#L181

ewolinetz: @deads2k could that potentially be problematic though for customers? if the daemonset lives in the `openshift-logging` namespace and it has a default node selector, unless they have that node selector on all nodes, they'll have a gap in log collection...

manifests/image-references
```diff
@@ -0,0 +1,28 @@
+kind: ImageStream
+apiVersion: image.openshift.io/v1
+spec:
+  tags:
+  - name: cluster-logging-operator
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/cluster-logging-operator:latest
+  - name: elasticsearch-operator
```
ewolinetz: lets drop the elasticsearch-operator stream from this manifest

manifests/image-references
```diff
@@ -0,0 +1,28 @@
...
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-elasticsearch-operator:latest
+  - name: kibana
```
ewolinetz: lets add the version to this name, and prefix with `logging-` perhaps for image stream groupings

PR #15, opened 2018-10-05
use multi-stage builds for images

jcantrill:
To satisfy @sosiouxme  https://jira.coreos.com/browse/ART-162?focusedCommentId=50773&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-50773

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,28 +1,26 @@
-FROM openshift/origin-base
-
-ENV APP_HOME=/opt/app/src \
-    GOPATH=/go \
-    PATH="${PATH}:${GOPATH}/bin:/usr/local/bin" \
-    INSTALL_PKGS="golang make git"
-
-RUN yum install -y ${INSTALL_PKGS} --setopt=tsflags=nodocs
+# first stage
+FROM openshift/origin-release:golang-1.10 AS builder
 
 WORKDIR /tmp/cluster-logging-operator
 COPY vendor/ /tmp/cluster-logging-operator/vendor
-COPY scripts/* /usr/local/bin/scripts/
 COPY cmd/ /tmp/cluster-logging-operator/cmd
 COPY pkg/ /tmp/cluster-logging-operator/pkg
 COPY Makefile /tmp/cluster-logging-operator
 
-RUN make && \
-    cp _output/bin/cluster-logging-operator /usr/local/bin/ && \
-    rm -rf /tmp/cluster-logging-operator && \
-    yum erase -y ${INSTALL_PKGS} && \
-    yum clean all
+RUN make
+
+# second stage
+
+FROM openshift/origin-base
+
+ENV PATH="/usr/local/bin:${PATH}"
+
+COPY scripts/* /usr/local/bin/scripts/
+COPY --from=builder /tmp/cluster-logging-operator/_output/bin/cluster-logging-operator /usr/local/bin/
+
+ENTRYPOINT ["/usr/local/bin/cluster-logging-operator"]
 
 LABEL io.k8s.display-name="OpenShift cluster-logging-operator" \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,28 +1,28 @@
...
+# second stage
+
+FROM openshift/origin-base
+
+ENV PATH="${GOPATH}/bin:/usr/local/bin:${PATH}"
+
+COPY scripts/* /usr/local/bin/scripts/
+COPY --from=builder /tmp/cluster-logging-operator/_output/bin/cluster-logging-operator /usr/local/bin/
+
+ENTRYPOINT ["/usr/local/bin/cluster-logging-operator"]
```
ewolinetz: does it make a difference the layer order for `ENTRYPOINT`?

Makefile
```diff
@@ -5,6 +5,7 @@ GOBUILD=go build
 GOPATH=$(TARGET_DIR):$(TARGET_DIR)/vendor:$(CURPATH)/cmd
 
 DOCKER_OPTS=
+IMAGE_BUILD=docker build
```
ewolinetz: shouldn't this be in quotes? `IMAGE_BUILD="docker build"`

ewolinetz:
questions, otherwise LGTM

Dockerfile
jcantrill: I don't think it matters [1] but I can move to the bottom if you have a preference

[1] https://docs.docker.com/develop/develop-images/dockerfile_best-practices/

Makefile
jcantrill: i don't 'think' it matters.  my version of docker doesnt support multistage so i had to test using image builder... but the 'docker build' command otherwise failed...

jcantrill: @ewolinetz per [1] not required

[1] https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html#SEC64

Dockerfile
sosiouxme: should not matter

Dockerfile
```diff
@@ -1,28 +1,28 @@
...
-
-ENV APP_HOME=/opt/app/src \
-    GOPATH=/go \
-    PATH="${PATH}:${GOPATH}/bin:/usr/local/bin" \
-    INSTALL_PKGS="golang make git"
+# first stage
+FROM openshift/origin-release:golang-1.10 AS builder
 
-RUN yum install -y ${INSTALL_PKGS} --setopt=tsflags=nodocs
+ENV GOPATH=/go
```
sosiouxme: this should already be set in the builder image, and FWIW, downstream includes /go as well as something scl-specific. I don't know for sure if this will cause any issues in the downstream compile. if there's not a specific need for this I'd leave it out.

I'm also curious why all your source goes in /tmp below instead of /go.

Dockerfile
```diff
@@ -1,28 +1,28 @@
...
-    rm -rf /tmp/cluster-logging-operator && \
-    yum erase -y ${INSTALL_PKGS} && \
-    yum clean all
+RUN make
+
+# second stage
+
+FROM openshift/origin-base
+
+ENV PATH="${GOPATH}/bin:/usr/local/bin:${PATH}"
```
sosiouxme: FWIW `$GOPATH` isn't defined here so this is probably not doing anything. That first segment should probably be removed.

sosiouxme:
some oddities there but looks like it should work.

Dockerfile
jcantrill: In the current incarnation, this allows you to build the binary outside of the image.  Not sure how this would play out otherwise

jcantrill:
@sosiouxme updated

Dockerfile
sosiouxme: The build depends on the /tmp path? *shrug* seems odd but if it works for you...

PR #22, opened 2018-10-11
Component restructure

ewolinetz:
Working to restructure the creation/updating of components to more match the sdk example of:

1. Create
2. Get
3. Update Object if necessary
4. Update Status if necessary


pkg/k8shandler/logstore.go
```diff
@@ -6,18 +6,43 @@ import (
...
 	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
 	sdk "github.com/operator-framework/operator-sdk/pkg/sdk"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 func CreateOrUpdateLogStore(logging *logging.ClusterLogging) (err error) {
-	if err = createOrUpdateElasticsearchSecret(logging); err != nil {
-		return
+
+	if logging.Spec.LogStore.Type == "elasticsearch" {
```
josefkarasek: Can you source "elasticsearch" from a common conststants file?

pkg/k8shandler/logstore.go
```diff
@@ -97,51 +122,69 @@ func getElasticsearchCR(logging *logging.ClusterLogging, elasticsearchName strin
...
+	}
+
+	return nil
+}
+
+func isElasticsearchCRDifferent(current *v1alpha1.Elasticsearch, desired *v1alpha1.Elasticsearch) (*v1alpha1.Elasticsearch, bool) {
+
+	different := false
+
+	//TODO: populate this
```
josefkarasek: Is this something that should be implemented?

pkg/k8shandler/collection.go
```diff
@@ -3,31 +3,56 @@ package k8shandler
...
+	if logging.Spec.Collection.Type == "fluentd" {
+		if err = createOrUpdateFluentdServiceAccount(logging); err != nil {
+			return
+		}
 
-	if err = createOrUpdateFluentdConfigMap(logging); err != nil {
-		return
-	}
+		if err = createOrUpdateCollectionPriorityClass(logging); err != nil {
+			return
```
josefkarasek: What happens if the operation fails?

pkg/k8shandler/collection.go
```diff
@@ -3,31 +3,56 @@ package k8shandler
...
-	if err = createOrUpdateFluentdServiceAccount(logging); err != nil {
-		return
-	}
 
-	if err = createOrUpdateCollectionPriorityClass(logging); err != nil {
-		return
-	}
+	if logging.Spec.Collection.Type == "fluentd" {
+		if err = createOrUpdateFluentdServiceAccount(logging); err != nil {
+			return
```
jcantrill: Don't we need to return `err` here? Can we continute to modify anything else?

pkg/k8shandler/logstore.go
ewolinetz: This will be implemented in 4.1
For now this is just a placeholder for elasticsearch

pkg/k8shandler/collection.go
ewolinetz: It will try to recreate it again the next time it goes through. The error handling for this should be  handled within the function `createOrUpdateCollectionPriorityClass`

pkg/k8shandler/collection.go
ewolinetz: no need since we're using named returns!
We want to return out since something failed, it will then retry after the interval time has passed (5 sec)

pkg/k8shandler/logstore.go
josefkarasek: Enum might be sufficient for this
```go
const (
  Elasticsearch = 0
  Fluentd = 1
  // ...
)
```

ewolinetz: Addressed by creating additional types

ewolinetz:
The change to the `types.go` to allow eventual implementation of event collection gives a status section that looks like this:
```
status:
  collection:
    eventCollection: {}
    logCollection:
      fluentdStatus:
        daemonSet: fluentd
        pods:
          fluentd-bl5gx: ip-172-18-8-188.ec2.internal
    normalizerStatus: {}
 . . .
```

ewolinetz:
Addressed comments and placeholder for event routing. PTAL

deploy/cr.yaml
```diff
@@ -19,5 +19,6 @@ spec:
     type: "curator"
     schedule: "30 3 * * *"
   collection:
-    type: "fluentd"
-    nodeSelector: "logging-infra-fluentd=true"
+    logCollection:
+      type: "fluentd"
+      nodeSelector: "logging-infra-fluentd=true"
```
richm: How will this work for rsyslog?  Will we need to change the cr to use `type: rsyslog` and `nodeSelector: "logging-infra-rsyslog=true"`?
Would it be possible to have a map or list here e.g.
```
     logCollection: 
       - type: "fluentd"
         nodeSelector: "logging-infra-fluentd=true"
       - type: "rsyslog"
         nodeSelector: "logging-infra-rsyslog=true"
```

pkg/apis/logging/v1alpha1/types.go
```diff
@@ -60,11 +60,20 @@ type ElasticsearchSpec struct {
...
 type CollectionSpec struct {
-	Type        string `json:"type"`
+	LogCollection   LogCollectionSpec   `json:"logCollection,omitempty"`
+	EventCollection EventCollectionSpec `json:"eventCollection,omitempty"`
+	Normalizer      NormalizerSpec      `json:"normalizer,omitempty"`
+}
+
+type LogCollectionSpec struct {
+	Type        LogCollectionType `json:"type"`
 	FluentdSpec `json:"fluentd,omitempty"`
```
richm: so for rsyslog, we'll have both the FluentdSpec and
```
 	RsyslogSpec `json:"rsyslog,omitempty"`
```
?

pkg/apis/logging/v1alpha1/types.go
```diff
@@ -122,8 +131,16 @@ type ElasticsearchStatus struct {
...
 type CollectionStatus struct {
-	FluentdStatus    FluentdCollectorStatus `json:"fluentdStatus,omitempty"`
-	NormalizerStatus NormalizerStatus       `json:"normalizerStatus,omitempty"`
+	LogCollection    LogCollectionStatus   `json:"logCollection,omitempty"`
+	EventCollection  EventCollectionStatus `json:"eventCollection,omitempty"`
+	NormalizerStatus NormalizerStatus      `json:"normalizerStatus,omitempty"`
+}
+
+type LogCollectionStatus struct {
+	FluentdStatus FluentdCollectorStatus `json:"fluentdStatus,omitempty"`
```
richm: Same here - we'll have both FluentdStatus and RsyslogStatus?

deploy/cr.yaml
ewolinetz: Yes, it would instead be `type: "rsyslog"`. Would we want to support both fluentd and rsyslog running at the same time?

pkg/apis/logging/v1alpha1/types.go
ewolinetz: Yes

pkg/apis/logging/v1alpha1/types.go
ewolinetz: Correct

deploy/cr.yaml
richm: Not sure - how do I change this to deploy rsyslog instead of fluentd?

jcantrill: Maybe we should consider changing to something like `logging-infra-collector=[fluentd|rsyslog]` or `logging-infra=<componentname>` for flexibility

ewolinetz: this is user defined in this case. so they could update it to match whatever they would prefer but we could have a different default value for each component.

ewolinetz: @richm this cr is actually missing `nodeSelector` being under `fluentd:` to match the log collection spec.
```
logCollection:
  type: "fluentd"
  fluentd:
    nodeSelector: "logging-infra-fluentd=true"
```
I'll update this.

PR #23, opened 2018-10-17
Expand e2e tests

ewolinetz:


deploy/role_binding.yaml
```diff
@@ -0,0 +1,11 @@
+kind: RoleBinding
```
jcantrill: Does the manifest need to be updated?

README.md
```diff
@@ -5,7 +5,11 @@ Operator to support OKD cluster logging
 
 ### E2E Testing
 To run the e2e tests, from the repo directory, run:
-`operator-sdk test local ./test/e2e/`
+```
```
jcantrill: Consider adding a 'make test' target to support this

deploy/role_binding.yaml
ewolinetz: I don't believe so. the e2e test framework just expected different file names

Dockerfile
```diff
@@ -19,8 +19,13 @@ ENV PATH="/usr/local/bin:${PATH}"
 
 COPY scripts/* /usr/local/bin/scripts/
 COPY --from=builder /tmp/cluster-logging-operator/_output/bin/cluster-logging-operator /usr/local/bin/
+RUN mkdir /tmp/_working_dir && \
+    chmod og+w /tmp/_working_dir && \
+    yum install -y openssl
```
ewolinetz: @sosiouxme sanity check that we don't need to do anything for image healthiness, please?

deploy/role_binding.yaml
ewolinetz: After running the tests it seems like there are some rbac definitions we're missing... we'll have to update the manifests as well. I'll include those changes once I can get them all in place.

Dockerfile
```diff
@@ -5,6 +5,7 @@ WORKDIR /tmp/cluster-logging-operator
 COPY vendor/ /tmp/cluster-logging-operator/vendor
 COPY manifests/ /manifests/
 COPY scripts/* /usr/local/bin/scripts/
+COPY files/* /usr/local/bin/files/
```
ewolinetz: I need to confirm if this fixed the errors i was seeing in the CLO logs when trying to build the curator configmap...

pkg/utils/utils.go
```diff
@@ -256,7 +256,7 @@ func Deployment(deploymentName string, namespace string, loggingComponent string
 	return &apps.Deployment{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       "Deployment",
-			APIVersion: "extensions/v1beta1",
+			APIVersion: apps.SchemeGroupVersion.String(),
```
ewolinetz: TODO: do this for the other objects

test/e2e/clusterlogging_test.go
```diff
@@ -101,7 +101,7 @@ func clusterLoggingFullClusterTest(t *testing.T, f *framework.Framework, ctx *fr
 			Curation: logging.CurationSpec{
 				Type: "curator",
 				CuratorSpec: logging.CuratorSpec{
-					Schedule: "30 3 * * *",
+					Schedule: "* * * * *",
```
richm: will this value always be replaced?  If not, `"30 3 * * *"` is a good default value

pkg/utils/utils.go
josefkarasek: I'll put it on my list too.

deploy/role.yaml
```diff
@@ -19,6 +19,7 @@ rules:
   - events
   - configmaps
   - secrets
+  - serviceaccounts
```
ewolinetz: the manifest needs the updates in this file...

Dockerfile
sosiouxme: I don't see any new problems with the docker build (the /tmp/_working_dir thing still looks weird to me)

ewolinetz:
For some reason I'm seeing that fluentd keeps crashing without logs when running `exec fluentd` in the container. So the tests pass on occasion, but incorrectly.
That is the last thing I want to get done for this PR.

Dockerfile
ewolinetz: That dir is used by the operator for extracting secrets into... probably doesn't need to be there so i'll add a comment to clean it up. It would also need to be addressed in code. Thanks for the sanity check

ewolinetz:
Depends on https://github.com/openshift/origin-aggregated-logging/pull/1433

ewolinetz:
Going to manually test to ensure we aren't getting false positives...

ewolinetz:
Verified we don't see false positives with that image fix. PTAL again

ewolinetz:
@jcantrill should i add the manifest updates to this pr?

manifests/03-role.yaml
```diff
@@ -31,3 +32,29 @@ rules:
...
+  - "*"
+- apiGroups:
+  - batch
+  resources:
+  - cronjobs
+  verbs:
+  - "*"
+
+---
+kind: ClusterRole
```
ewolinetz: @jcantrill @smarterclayton for cluster role and role bindings should we be specifying that in our manifest or communicating that with the privileged operator?

jcantrill: I defer to @smarterclayton and @derekwaynecarr  but i presume if the role is privileged we need to defer to the privileged operator

pkg/utils/utils.go
```diff
@@ -325,7 +332,7 @@ func DaemonSet(daemonsetName string, namespace string, loggingComponent string,
 						"logging-infra": loggingComponent,
 					},
 					Annotations: map[string]string{
-						"scheduler.alpha.kubernetes.io/critical-pod": "''",
+						"scheduler.alpha.kubernetes.io/critical-pod": "",
```
jcantrill: Is still needed in 4.0?

ewolinetz: @aveshagarwal

manifests/03-role.yaml
smarterclayton: In your manifest if you think only static changes are necessary.  It’s easy to move from manifest to operator if necessary.

ewolinetz: Okay that makes sense. 
@jcantrill the role isn't privileged but it is a cluster role definition to create priority classes. So long as OLM (or whomever it is that does operations with our manifest files) can create a cluster role and cluster role binding then we should be all set here.

PR #25, opened 2018-10-22
initial rsyslog support

richm:
first pass at rsyslog support


pkg/k8shandler/collection.go
```diff
@@ -105,6 +157,66 @@ func createOrUpdateFluentdConfigMap(logging *logging.ClusterLogging) error {
...
+	// need three configmaps
+	// - one for rsyslog run.sh script - rsyslog-bin
+	// - one for main rsyslog.conf file - rsyslog-main
+	// - one for the actual config files - rsyslog
+	rsyslogConfigMaps := make(map[string]*v1.ConfigMap)
+	rsyslogBinConfigMap := utils.ConfigMap(
+		"rsyslog-bin",
+		logging.Namespace,
+		map[string]string{
+			"rsyslog.sh": string(utils.GetFileContents("files/rsyslog/rsyslog.sh")),
```
ewolinetz: I was actually wondering why we had the `rsyslog.sh` file there. Is there a reason this isn't baked into the image instead?

pkg/k8shandler/collection.go
```diff
@@ -224,6 +386,82 @@ func getFluentdPodSpec(logging *logging.ClusterLogging, elasticsearchAppName str
...
+			{Name: "certs", VolumeSource: v1.VolumeSource{Secret: &v1.SecretVolumeSource{SecretName: "rsyslog"}}},
+			{Name: "dockerhostname", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/etc/hostname"}}},
+			{Name: "localtime", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/etc/localtime"}}},
+			{Name: "filebufferstorage", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/var/lib/rsyslog"}}},
+		},
+	)
+
+	rsyslogPodSpec.PriorityClassName = "cluster-logging"
+
+	rsyslogPodSpec.NodeSelector = map[string]string{
```
ewolinetz: I think I made this change in #23 but we likely want this to come from the CR instead of being hardcoded

pkg/k8shandler/collection.go
richm: Yes - the image, including its built-in run script, is designed to be used as a replacement for the system rsyslog.  Our requirements are very different, so we should provide our own run.sh script.  Basically, we are treating the container as a bag of packages and a runtime, with no scripts or configuration for us to use - we will provide/replace all scripts/configuration in the image.

pkg/k8shandler/collection.go
richm: Ok.  Likely #23 will merge first and I will rebase this on top of it.

richm:
Passes logging CI when using https://github.com/openshift/origin-aggregated-logging/pull/1442
This PR does not handle upgrading from fluentd to rsyslog.  There are still issues with fluentd file buffers, journald position file, and json-file/crio position files.  Basically, we need to have a procedure in place for fluentd to flush all file buffers, and to convert the position files from fluentd format to rsyslog format, in order to ensure there are no duplicates and no missing records.

jcantrill:
@richm I can squash and merge what i have... which is functional.. .might give you a bases to build upon

ewolinetz:
@richm I'm not too worried about updating between the two at this time. I think for 4.0 we can advise customers they will need to make the decision at installation time and it won't be possible to switch between the two (yet)

richm:
ok - or we can advise customers if they want to switch that they'll have to flush all of the fluentd buffers, then lose some number of records as rsyslog will read from the tail of the logs

deploy/cr.yaml
```diff
@@ -25,3 +25,6 @@ spec:
       fluentd:
         nodeSelector:
           logging-infra-fluentd: "true"
+      rsyslog:
```
ewolinetz: I was actually wondering about scenarios like this... 
For this case, we would want someone to configure to install with either `rsyslog` **or** `fluentd`. So should we just provide two different CR examples?

pkg/k8shandler/collection.go
```diff
@@ -51,6 +53,42 @@ func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
 			}
 		}
 	}
+	if cluster.Spec.Collection.LogCollection.Type == logging.LogCollectionTypeRsyslog {
+		if err = createOrUpdateRsyslogServiceAccount(cluster); err != nil {
+			return
+		}
+
+		if err = createOrUpdateCollectionPriorityClass(cluster); err != nil {
```
ewolinetz: since this is common for both fluentd and rsyslog, can we just move it up a level out of both of their if statements?

deploy/cr.yaml
richm: I suppose, although the `type:` is the thing which controls which one is used.

ewolinetz:
That's true... let's leave it as is then

ewolinetz:
Can you add an e2e test that deploys rsyslog? It can do just that and not worry about any other components.
Otherwise LGTM

richm:
> Can you add an e2e test that deploys rsyslog? It can do just that and not worry about any other components.

done

files/rsyslog/55-elasticsearch.conf
```diff
@@ -0,0 +1,94 @@
...
+            queue.spoolDirectory=`echo $RSYSLOG_SPOOLDIRECTORY`
+            queue.type=`echo $ES_QUEUE_TYPE`
+            queue.maxDiskSpace=`echo $ES_QUEUE_MAXDISKSPACE`
+            queue.maxFileSize=`echo $ES_QUEUE_MAXFILESIZE`
+            queue.checkpointInterval=`echo $ES_QUEUE_CHECKPOINTINTERVAL`
+            queue.syncqueuefiles="on"
+            )
+    } else {
+        action(
+            name="elasticsearch-ops"
```
ewolinetz: for consistency can we call this "elasticsearch-infra" ?

files/rsyslog/55-elasticsearch.conf
```diff
@@ -0,0 +1,94 @@
...
+    if strlen($.omes!_index) > 0 then {
+        # retry
+        set $.viaq_index_name = $.omes!_index;
+    } else {
+        set $.viaq_index_name = exec_template("prefix_index_template");
+    }
+    if ($.viaq_index_prefix startswith "project.") or
+       ((`echo $ES_HOST` == `echo $OPS_HOST`) and (`echo $ES_PORT` == `echo $OPS_PORT`)) then {
+        action(
+            name="elasticsearch-apps"
```
ewolinetz: for consistency "elasticsearch-app"

ewolinetz:
just two tiny nits, otherwise LGTM

richm:
@ewolinetz ptal

pkg/apis/logging/v1alpha1/types.go
```diff
@@ -148,6 +155,11 @@ type FluentdCollectorStatus struct {
 	Pods      map[string]string `json:"pods"`
 }
 
+type RsyslogCollectorStatus struct {
+	DaemonSet string            `json:"daemonSet"`
+	Pods      map[string]string `json:"pods"`
```
ewolinetz: todo in follow up: update the status to match what #36 added

richm:
@ewolinetz rebased and added new status fields

PR #26, opened 2018-10-24
hardcode SAN DNS names for now

richm:
hardcode SAN DNS names for now
we'll need to get the SAN DNS names from the elasticsearch
service names in the future

### DIFF
--- a/scripts/cert_generation.sh
+++ b/scripts/cert_generation.sh
@@ -226,5 +226,6 @@ generate_certs 'system.logging.kibana'
 generate_certs 'system.logging.curator'
 generate_certs 'system.admin'
 
-generate_certs 'elasticsearch' "$(generate_extensions true logging-es{,-ops})"
-generate_certs 'logging-es' "$(generate_extensions false logging-es{,-ops}{,-cluster}{,.${NAMESPACE}.svc.cluster.local})"
+# TODO: get es SAN DNS, IP values from es service names
+generate_certs 'elasticsearch' "$(generate_extensions true elasticsearch elasticsearch-infra elasticsearch-apps)"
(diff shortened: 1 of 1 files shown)


scripts/cert_generation.sh
```diff
@@ -226,5 +226,6 @@ generate_certs 'system.logging.kibana'
 generate_certs 'system.logging.curator'
 generate_certs 'system.admin'
 
-generate_certs 'elasticsearch' "$(generate_extensions true logging-es{,-ops})"
-generate_certs 'logging-es' "$(generate_extensions false logging-es{,-ops}{,-cluster}{,.${NAMESPACE}.svc.cluster.local})"
+# TODO: get es SAN DNS, IP values from es service names
+generate_certs 'elasticsearch' "$(generate_extensions true elasticsearch elasticsearch-infra elasticsearch-apps)"
```
ewolinetz: shouldn't these be comma delimited?

richm: I don't think so
```bash
> echo logging-es{,-ops}
logging-es logging-es-ops
> foo() { echo $1; echo $2;}
> foo logging-es{,-ops}
logging-es
logging-es-ops
> generate_extensions true logging-es{,-ops}
IP.1:127.0.0.1,DNS.1:localhost,DNS.2:logging-es,DNS.3:logging-es-ops,RID.1:1.2.3.4.5.5
> generate_extensions true elasticsearch elasticsearch-infra elasticsearch-apps
IP.1:127.0.0.1,DNS.1:localhost,DNS.2:elasticsearch,DNS.3:elasticsearch-infra,DNS.4:elasticsearch-apps,RID.1:1.2.3.4.5.5
```
```

PR #29, opened 2018-10-24
LOG-220. Add fluentd ServiceMonitor definition

jcantrill:
Resolve https://jira.coreos.com/browse/LOG-220

by adding the fluentd servicemonitor definition

### DIFF
--- a/manifests/10-service-monitor-fluentd.yaml
+++ b/manifests/10-service-monitor-fluentd.yaml
@@ -1,0 +1,14 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: fluentd
+  namespace: openshift-logging
+spec:
+  endpoints:
+  - port: 24231
+    scheme: https
+    tlsConfig:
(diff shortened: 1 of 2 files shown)


jcantrill:
@openshift/sig-logging  Starting a discussion here.  I would like to add a test.  Note sure how that is realized.  Is it an e2e of some kind (but the actual SM will be deployed by OLM right) ?  Is there some sort of unit test?  not really sure if there is a test to be had.  Do we start adding tests that would normally be found in origin-aggregated-logging?

ewolinetz:
I'm not sure there's much testing we can do from within here. Unless we had a prometheus installed that could scrape and we could verify. Even then, it being a manifest file says to me that OLM will create it so we would end up manually creating it as part of our e2e tests but again, this is only something prometheus acts on so I'm not sure what we would verify.

/lgtm

jcantrill:
Maybe once we get the origin-aggregated-logging tests working with the operator we can pull the endpoint info from the SM to test the fluent metrics endpoint in lieu of hard coding them into our test

PR #31, opened 2018-10-26
add make targets for deploying operator

jcantrill:
This PR provides make targets to:
* deploy pre reqs
* deploy full image
of the cluster-logging-operator


hack/undeploy.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+
+source "$(dirname $0)/common"
+
+${OC} login ${API_SERVER} -u ${ADMIN_USER} -p ${ADMIN_PSWD}
+	
+${OC} delete project openshift-logging ||:
```
ewolinetz: we can just `oc delete -f` keep it consistent with the deploy pattern

manifests/05-crd.yaml
```diff
@@ -2,6 +2,7 @@ apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
   name: clusterloggings.logging.openshift.io
+  namespace: openshift-logging
```
ewolinetz: i think these are non-namespaced, just fyi

hack/install-deps.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
```
ewolinetz: i think the name of this file (and the make target) is too vague for whats being done here. This is actually installing the operator sdk. "install deps" seems like it would be for the operator itself, which is unnecessary.

ewolinetz: why not call this like "install-sdk" instead?

jcantrill: The script is meant to be a placeholder for all dependencies but I think I will remove for the time being since I don't think its fully flushed out.

hack/common
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+if [ -n "${DEBUG:-}" ]; then
+    set -x
+fi
+
+alias oc=${OC:-oc}
+
+repo_dir="$(dirname $0)/.."
+ELASTICSEARCH_OP_REPO=${ELASTICSEARCH_OP_REPO:-${repo_dir}/../elasticsearch-operator}
```
richm: so this means `ELASTICSEARCH_OP_REPO=../../elasticsearch-operator`?

hack/deploy-setup.sh
```diff
@@ -0,0 +1,42 @@
...
+# This script inits a cluster to allow cluster-logging-operator
+# to deploy logging.  It assumes it is capable of login as a
+# user who has the cluster-admin role
+
+set -e
+
+source "$(dirname $0)/common"
+
+repo_dir="$(dirname $0)/.."
+ELASTICSEARCH_OP_REPO=${ELASTICSEARCH_OP_REPO:-${repo_dir}/../elasticsearch-operator}
```
richm: you already do this in `common` - do you need to do it here too?

hack/deploy-setup.sh
```diff
@@ -0,0 +1,42 @@
...
+    for m in $(ls); do
+      if [ "$(echo ${EXCLUSIONS[@]} | grep -o ${m} | wc -w)" == "0" ] ; then
+        oc create -f ${m} ${namespace:-}
+      fi
+    done
+  popd
+}
+
+load_manifest ${repo_dir}
+load_manifest ${ELASTICSEARCH_OP_REPO} openshift-logging
```
richm: es needs a namespace but not clo?

hack/deploy.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+source "$(dirname $0)/common"
+
+registry_ip=$(oc get service docker-registry -n default -o jsonpath={.spec.clusterIP})
+cat manifests/05-deployment.yaml | \
+    sed -e "s/quay\.io\/openshift\/cluster-logging-operator/${registry_ip}:5000\/openshift\/cluster-logging-operator/" | \
```
richm: Is there a template-like feature e.g. `oc process -p IMAGE_PREFIX=$registry_ip:5000/openshift/cluster-logging-operator`

You can specify a different sed delimiter like this, if you want to avoid the "leaning toothpick syndrome":
```
sed -e "s,quay\.io/openshift/cluster-logging-operator,${registry_ip}:5000/openshift/cluster-logging-operator," | \ 
```

hack/test-e2e.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+set -e
+
+source "$(dirname $0)/common"
+
+repo_dir="$(dirname $0)/.."
+if [ -z "$(oc get project openshift-logging -o name --ignore-not-found)" ] ; then
```
richm:
```bash
if oc get project openshift-logging > /dev/null 2>&1 ; then
```

hack/deploy-setup.sh
jcantrill: I believe this was to try and satisfy the desire to keep the elasticsearch-operator generic allowing us to load all the resources into a given namespace where CLO is expected to already exist in the desired namespace by defining it in the manifest

hack/common
jcantrill: I was trying to achieve an absolute path to the repo assumed to be parallel to this one.

PR #32, opened 2018-10-26
Creating the oauth client for kibana redirects

ewolinetz:
To address the lack of a created oauthclient
Current gaps: no top level config to pull cluster configuration from to get the default dns

### DIFF
--- a/pkg/k8shandler/visualization.go
+++ b/pkg/k8shandler/visualization.go
@@ -10,6 +10,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"reflect"
 
+	oauth "github.com/openshift/api/oauth/v1"
 	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
 	sdk "github.com/operator-framework/operator-sdk/pkg/sdk"
 	apps "k8s.io/api/apps/v1"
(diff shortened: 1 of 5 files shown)


pkg/k8shandler/visualization.go
```diff
@@ -122,6 +129,51 @@ func createOrUpdateKibanaDeployment(logging *logging.ClusterLogging) (err error)
...
+			Name:      "kibana-proxy",
+			Namespace: logging.Namespace,
+			Labels: map[string]string{
+				"logging-infra": "support",
+			},
+		},
+		Secret: oauthSecret,
+		// TODO: don't just use string concatination here...
+		RedirectURIs: []string{
+			"https://kibana." + clusterDomain,
```
jcantrill: Isn't this an issue if user's customize their route?  (i.e. operations use of `logs.somedomain.com`).  Do we need to add a field to the CRD?

ewolinetz: I started having a discussion about this on slack with @richm. I think we would need pm to resolve this, I'm of the opinion that we should just control that, I don't see a benefit to letting them choose what their Kibana route is and it increases our complexity.

richm: I think for now, let's merge this PR, and revisit the issue in another PR

PR #33, opened 2018-11-01
Fix image streams

jcantrill:


manifests/05-deployment.yaml
```diff
@@ -28,3 +28,13 @@ spec:
...
           - name: OPERATOR_NAME
             value: "cluster-logging-operator"
+          - name: ELASTICSEARCH_IMAGE
+            value: "docker.io/openshift/origin-logging-elasticsearch5:latest"
+          - name: FLUENTD_IMAGE
+            value: "docker.io/openshift/origin-logging-fluentd:latest"
+          - name: KIBANA_IMAGE
+            value: "docker.io/openshift/origin-logging-kibana5:latest"
+          - name: CURATOR_IMAGE
+            value: "docker.io/openshift/origin-logging-curator:latest"
```
richm: curator5

pkg/utils/utils.go
```diff
@@ -121,16 +51,21 @@ func AddOwnerRefToObject(object metav1.Object, ownerRef metav1.OwnerReference) {
...
-	return image.String()
+	env_var_name, ok := COMPONENT_IMAGES[component]
+	if !ok {
+		logrus.Errorf("Environment variable name mapping missing for component: %s", component)
+		return ""
+	}
+	imageTag := os.Getenv(env_var_name)
+	if imageTag != "" {
+		logrus.Errorf("No image tag defined for component '%s' vi environment variable '%s'", component, env_var_name)
+		return ""
```
ewolinetz: there isn't a reason to have this return since the one at the end of the function would return `""` in this case

jcantrill: Updating to remove the return and noted the condition is in correct.  We should log error if the value is empty

pkg/utils/utils.go
```diff
@@ -121,16 +51,21 @@ func AddOwnerRefToObject(object metav1.Object, ownerRef metav1.OwnerReference) {
...
+	env_var_name, ok := COMPONENT_IMAGES[component]
+	if !ok {
+		logrus.Errorf("Environment variable name mapping missing for component: %s", component)
+		return ""
+	}
+	imageTag := os.Getenv(env_var_name)
+	if imageTag == "" {
+		logrus.Errorf("No image tag defined for component '%s' by environment variable '%s'", component, env_var_name)
+	}
+	logrus.Debugf("Setting component image for '%s' to: '%s'", component, imageTag)
```
ewolinetz: is this still needed?

jcantrill:
@smarterclayton Is this how we should expect to ensure we are using released versions of component images?  `ci-operator` will modify manifest files to update operator deployments?

pkg/utils/utils.go
jcantrill: I think maybe not... but i left it for sanity purposes.  Its only visible at debut level so I thought that would be harmless

smarterclayton:
Yes. The release process will perform a `sed` replacement equivalent from the value you add in `spec.tags.from.name` to the correct image pull spec in your manifests.

PR #34, opened 2018-11-02
Component image changes

ewolinetz:
Addresses image changes and needing to update the deployment objects in cluster logging operator accordingly.


pkg/k8shandler/collection.go
```diff
@@ -243,7 +265,75 @@ func isFluentdDaemonsetDifferent(current *apps.DaemonSet, desired *apps.DaemonSe
...
+
+	if strings.HasPrefix(currVersion, "v") {
+		currVersion = strings.Split(currVersion, "v")[1]
+	}
+
+	if strings.HasPrefix(desVersion, "v") {
+		desVersion = strings.Split(desVersion, "v")[1]
+	}
+
+	return (currVersion == "3.11" && desVersion == "4.0.0")
```
ewolinetz: @richm are we providing fluentd 1.x as our 4.0 image?

richm: yes

pkg/k8shandler/collection.go
```diff
@@ -501,7 +523,10 @@ func isFluentdDaemonsetDifferent(current *apps.DaemonSet, desired *apps.DaemonSe
 
 	different := false
 
-	// TODO: populate this
+	if isImageDifference(current, desired) {
+		logrus.Infof("Fluentd image change found, updating %q", current.Name)
```
jcantrill: In the logging PR from @josefkarasek I thought we had these at debug messages.  I actually prefer info since an image change is more of a major event IMO.

pkg/k8shandler/collection.go
```diff
@@ -501,7 +523,10 @@ func isFluentdDaemonsetDifferent(current *apps.DaemonSet, desired *apps.DaemonSe
...
+		return desired, true
```
jcantrill: Why do we need to return `desired` here?

pkg/k8shandler/visualization.go
```diff
@@ -446,5 +446,43 @@ func isKibanaDifferent(current *apps.Deployment, desired *apps.Deployment) (*app
...
+	if isImageDifference(current, desired) {
+		current = updateCurrentImages(current, desired)
+		logrus.Infof("Kibana image(s) change found, updating %q", current.Name)
+		different = true
+	}
+
 	return current, different
 }
+
+func isImageDifference(current *apps.Deployment, desired *apps.Deployment) bool {
```
jcantrill: Isn't this really a util function?  Nothing visualization specific here.

test/e2e/clusterlogging_test.go
```diff
@@ -279,3 +285,49 @@ func ClusterLoggingClusterRsyslog(t *testing.T) {
...
+	if err != nil {
+		return fmt.Errorf("failed to get currentOperator: %v", err)
+	}
+
+	currentEnv := currentOperator.Spec.Template.Spec.Containers[0].Env
+	newEnv := []v1.EnvVar{
+		{Name: "WATCH_NAMESPACE", ValueFrom: &v1.EnvVarSource{FieldRef: &v1.ObjectFieldSelector{FieldPath: "metadata.namespace"}}},
+		{Name: "OPERATOR_NAME", Value: "cluster-logging-operator"},
+		{Name: "ELASTICSEARCH_IMAGE", Value: "docker.io/openshift/origin-logging-elasticsearch5:upgraded"},
+		{Name: "FLUENTD_IMAGE", Value: "docker.io/openshift/origin-logging-fluentd:upgraded"},
```
jcantrill: This wont allow you to evaluate the need to upgrade the version.  Do we need a var to explicitly tell the operator to explicitly flush the buffers?

pkg/k8shandler/collection.go
ewolinetz: No, this is a new message. IMO it should be at `info` because it is used to notify users that we need to adjust something.

pkg/k8shandler/collection.go
ewolinetz: I'm going to update this to set the image in current and return that later.

pkg/k8shandler/visualization.go
ewolinetz: Yes... but currently the only `Deployment` is from Visualization

ewolinetz:
/hold
need to add logic for ES CR

test/e2e/clusterlogging_test.go
ewolinetz: This will tell it to upgrade the version, it doesn't cause it to do a buffer flush though. That may be outside the scope of this actually since that is part of the adoption work.
@richm thoughts?

richm: Will this upgrade the image from 3.11 to 4.0?

ewolinetz: We aren't advertising being able to upgrade from `3.11` to `4.0` currently, but yes, once we say we can do that if the current DS is at `3.11` based on the image name and the image provided on the `ENV` for the operator is `4.0` then it should trigger that it requires the buffer flush

richm: then I guess triggering the buffer flush it outside the scope of this work, and in a future PR we'll need to have some sort of logic like this:
- user wants to upgrade the image
- `oc set env ds/logging-fluentd FLUSH_AT_SHUTDOWN=true`
- restart fluentd on all nodes (the set env ds trigger usually restarts fluentd)
- shutdown fluentd on all nodes - will flush the buffers at shutdown
- change the image to the new version and set FLUSH_AT_SHUTDOWN=false
- start fluentd

richm: note that step `shutdown fluentd on all nodes - will flush the buffers at shutdown` may take a long, long time . . . depending on how much log buffer data there is

ewolinetz: So the way I have this written is that first we update the `ENV` to add `FLUSH_AT_SHUTDOWN=true` then we wait to have full availability again of the DS pods. After that we make the upgrade to `4.0` (this change would also remove the added `ENV` entry) which would cause another rollout, the operator doesn't wait on the pods after that.

PR #36, opened 2018-11-05
Updating our status to be useful for the podStatus console descriptor

ewolinetz:
Part of the work for the OLM is to create descriptors that the console can then use.
The `urn:alm:descriptor:com.tectonic.ui:podStatuses` x-descriptor takes advantage of a `map[string][]string` structure for formatting its output[1] where the key is the status used and the array is the list of pods in that status.

We will want to also update Elasticsearch to follow this pattern and then update this PR to use that updated structure. We can organize by Role and then pod status after.
Following that, for an all-in-one elasticsearch cluster with 3 nodes we would see:
```
status:
  pods:
  - master:
    ready:
    - podA
    - podB
    - podC
    notReady: []
    failed: []
  - client:
    ready:
    - podA
    - podB
    - podC
    notReady: []
    failed: []
  - data:
    ready:
    - podA
    - podB
    - podC
    notReady: []
    failed: []
```

[1] https://github.com/openshift/console/tree/master/frontend/public/components/operator-lifecycle-manager/descriptors#example


richm:
How does OLM know to look for the specific string values "ready" "notReady" and "failed"?  Are these defined in the Kubernetes spec somewhere?

ewolinetz:
@richm it doesn't look for those specific values, it just cares for the structure `map[string][]string` any of the keys that the map has it will use though when displaying.

pkg/apis/logging/v1alpha1/types.go
```diff
@@ -127,7 +127,8 @@ type ElasticsearchStatus struct {
 	ReplicaSets  []string `json:"replicaSets"`
 	Deployments  []string `json:"deployments"`
 	StatefulSets []string `json:"statefulSets"`
-	Pods         []string `json:"pods"`
+	Pods			   []string `json:"pods"`
```
richm: formatting looks off?

pkg/apis/logging/v1alpha1/types.go
```diff
@@ -145,13 +146,14 @@ type EventCollectionStatus struct {
 
 type FluentdCollectorStatus struct {
 	DaemonSet string            `json:"daemonSet"`
-	Pods      map[string]string `json:"pods"`
+	Nodes     map[string]string `json:"nodes"`
+	Pods			PodStateMap `json:"pods"`
```
richm: formatting?

ewolinetz: +1 i didn't run `gofmt` yet

ewolinetz:
Depends on https://github.com/openshift/elasticsearch-operator/pull/26

ewolinetz:
with latest update in status we see the following:
```
status:
  collection:
    eventCollection: {}
    logCollection:
      fluentdStatus:
        daemonSet: fluentd
        nodes:
          fluentd-q2h8w: ip-172-18-10-196.ec2.internal
        pods:
          failed: []
          notReady:
          - fluentd-q2h8w
          ready: []
    normalizerStatus: {}
  curation:
    curatorStatus:
    - cronJobs: curator
      schedules: 30 3 * * *
      suspended: false
  logStore:
    elasticsearchStatus:
    - clusterName: elasticsearch
      deployments:
      - elasticsearch-clientdatamaster-0-1
      pods:
        client:
          failed: []
          notReady:
          - elasticsearch-clientdatamaster-0-1-79d5c457db-gzz49
          ready: []
        data:
          failed: []
          notReady:
          - elasticsearch-clientdatamaster-0-1-79d5c457db-gzz49
          ready: []
        master:
          failed: []
          notReady:
          - elasticsearch-clientdatamaster-0-1-79d5c457db-gzz49
          ready: []
      replicaSets:
      - elasticsearch-clientdatamaster-0-1-79d5c457db
      replicas: 1
      statefulSets:
      - ""
    - clusterName: elasticsearch-app
      deployments:
      - elasticsearch-app-clientdatamaster-0-1
      pods:
        client:
          failed: []
          notReady:
          - elasticsearch-app-clientdatamaster-0-1-64ff7c85b8-v6jw7
          ready: []
        data:
          failed: []
          notReady:
          - elasticsearch-app-clientdatamaster-0-1-64ff7c85b8-v6jw7
          ready: []
        master:
          failed: []
          notReady:
          - elasticsearch-app-clientdatamaster-0-1-64ff7c85b8-v6jw7
          ready: []
      replicaSets:
      - elasticsearch-app-clientdatamaster-0-1-64ff7c85b8
      replicas: 1
      statefulSets:
      - ""
  visualization:
    kibanaStatus:
    - deployment: kibana
      pods:
        failed: []
        notReady: []
        ready:
        - kibana-57f79f6b64-pk4r9
      replicaSets:
      - kibana-57f79f6b64
      replicas: 1
```

richm:
The elasticsearch output is not consistent with the output from the other services.  es has `pods.client.failed`, `pods.master.failed`, etc. but the others have e.g. `pods.failed`.   I think to make the output consistent you need to have `pods.failed` for elasticsearch too.  That is - shouldn't they all report `pods.$status` instead of `pods.somethingelse.$status`?  As for the node type, `client`, `master`, `data`, that should be a field under the pod?  Not sure where that should go, or even if it should be reported here.

ewolinetz:
@richm my thought with nesting it like that is because then when we present the information for the console we can see a breakdown by role what the pod state is for each ES cluster otherwise if we just show all the `ready`, `notReady`, and `failed` pods we lose the detail of what type of pod that is.

The cluster is in a worse state to have failed `data` nodes rather than just failed `client` nodes.

The ES status is unique compared to the other components in that way.

richm:
ok - as long as the UI doesn't show the pod status as "master"

ewolinetz:
Correct, it would just group all the masters together but still have the possible status states as `ready` `notReady` `failed`

ewolinetz:
We will need to rerun `dep ensure` after https://github.com/openshift/elasticsearch-operator/pull/26 merges in

PR #38, opened 2018-11-07
several e2e test cleanups

richm:
- use `imagebuilder` instead of `docker build`
- do not set `GOPATH` for all tests - instead, set only
  `BUILD_GOPATH` for building the image
- add dependencies on `imagebuilder` and `operator-sdk`
- oauthclient does not have a namespace - just add the
  namespace as a label
- fix some gofmt issues
- add the scc for fluentd
- it is apparently not required to push the image to the
  local registry (which wasn't working anyway)


manifests/05-deployment.yaml
```diff
@@ -28,3 +28,7 @@ spec:
                 fieldPath: metadata.namespace
           - name: OPERATOR_NAME
             value: "cluster-logging-operator"
+          - name: REPO_PREFIX
```
ewolinetz: I'm not sure this is something we want to always have for normal deployments

ewolinetz:
comment otherwise lgtm

manifests/05-deployment.yaml
richm: I have no idea how it is supposed to work for normal deployments

ewolinetz: Scratch this comment, I thought I remembered that the e2e tests also created the deployment within the test. I had forgotten that it relies on this file.

ewolinetz:
/lgtm
thanks @richm

PR #46, opened 2018-11-30
Creating shared-config for console to consume

ewolinetz:
Addresses https://jira.coreos.com/browse/LOG-233

@spadgett PTAL and confirm this would work. There will always be `kibanaAppHost` and `kibanaInfraHost` entries, even if it is an "all-in-one" configuration where we only create a single `route`.

e.g.
```
oc get configmap -o yaml sharing-config
apiVersion: v1
data:
  kibanaAppHost: kibana-app
  kibanaInfraHost: kibana-infra
kind: ConfigMap
metadata:
  creationTimestamp: 2018-11-30T18:21:05Z
  name: sharing-config
  namespace: openshift-logging
  ownerReferences:
  - apiVersion: logging.openshift.io/v1alpha1
    controller: true
    kind: ClusterLogging
    name: example-cluster-logging
    uid: b27ce418-f4cc-11e8-a9ff-0e3a960e086c
  resourceVersion: "7727"
```

### DIFF
--- a/pkg/k8shandler/visualization.go
+++ b/pkg/k8shandler/visualization.go
@@ -188,6 +188,7 @@ func createOrUpdateKibanaRoute(logging *logging.ClusterLogging) error {
 			logging.Namespace,
 			"kibana",
 			"kibana",
+			"/tmp/_working_dir/ca.crt",
 		)
 
 		utils.AddOwnerRefToObject(kibanaRoute, utils.AsOwner(logging))
@@ -197,7 +198,7 @@ func createOrUpdateKibanaRoute(logging *logging.ClusterLogging) error {
 			return fmt.Errorf("Failure creating Kibana route: %v", err)
 		}
 
-		sharedConfig := createSharedConfig(logging, "kibana", "kibana")
+		sharedConfig := createSharedConfig(logging, "https://kibana", "https://kibana")
 		utils.AddOwnerRefToObject(sharedConfig, utils.AsOwner(logging))
 
 		err = sdk.Create(sharedConfig)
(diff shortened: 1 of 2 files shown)


spadgett:
Do we need to say if it's http/https?

spadgett:
Here is what the monitoring team did if we want to be consistent:

https://github.com/openshift/cluster-monitoring-operator/pull/155
https://github.com/openshift/cluster-monitoring-operator/pull/169
https://github.com/openshift/cluster-monitoring-operator/pull/170

ewolinetz:
@spadgett that is a good question. What did we have before? It would likely be best to avoid regressions from that.

spadgett:
We had the full URL in the config, but we did require https since we passed
the user’s token.

pkg/k8shandler/visualization.go
```diff
@@ -488,3 +507,14 @@ func updateCurrentImages(current *apps.Deployment, desired *apps.Deployment) (*a
 
 	return current
 }
+
+func createSharedConfig(logging *logging.ClusterLogging, kibanaAppHost, kibanaInfraHost string) *v1.ConfigMap {
+	return utils.ConfigMap(
+		"sharing-config",
+		logging.Namespace,
+		map[string]string{
+			"kibanaAppHost":   kibanaAppHost,
```
richm: Is this a host or a url?  The arguments to `createSharedConfig` are urls

ewolinetz: updated name to be more appropriate

ewolinetz:
Also added in missing `TLS` section for route

Issue #39, opened 2018-11-10
Use RetryOnConflict for updates to existing objects #28 


richm:
First take a look at https://github.com/kubernetes/client-go/blob/master/examples/create-update-delete-deployment/main.go#L102 which describes why `RetryOnConflict` is needed.

There are several patterns in our code like this:
```
  client.Get(object)
  object.somefield = "new value"
  client.Update(object)
```
The problem is that the object can be updated by another client between the `Get` and the `Update` and the `Update` will return a `Conflict` error.  Instead, we need to wrap all such places in our code with `RetryOnConflict`

I've already seen cases running e2e tests where we get errors from conflicts.

PR #47, opened 2018-11-30
Improved error handling

ewolinetz:
Addresses https://github.com/openshift/cluster-logging-operator/issues/39

### DIFF
--- a/pkg/k8shandler/collection.go
+++ b/pkg/k8shandler/collection.go
@@ -10,12 +10,14 @@ import (
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
 	"strings"
 	"time"
 
 	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
 	sdk "github.com/operator-framework/operator-sdk/pkg/sdk"
 	apps "k8s.io/api/apps/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 var (
@@ -47,18 +49,26 @@ func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
 		}
 
 		fluentdStatus, err := getFluentdCollectorStatus(cluster.Namespace)
-
 		if err != nil {
 			return fmt.Errorf("Failed to get status of Fluentd: %v", err)
 		}
 
-		if !reflect.DeepEqual(fluentdStatus, cluster.Status.Collection.LogCollection.FluentdStatus) {
-			logrus.Infof("Updating status of Fluentd")
-			cluster.Status.Collection.LogCollection.FluentdStatus = fluentdStatus
-
-			if err = sdk.Update(cluster); err != nil {
-				return fmt.Errorf("Failed to update Cluster Logging Fluentd status: %v", err)
+		printUpdateMessage := true
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
+				if !reflect.DeepEqual(fluentdStatus, cluster.Status.Collection.LogCollection.FluentdStatus) {
+					if printUpdateMessage {
+						logrus.Info("Updating status of Fluentd")
+						printUpdateMessage = false
+					}
+					cluster.Status.Collection.LogCollection.FluentdStatus = fluentdStatus
+					return sdk.Update(cluster)
+				}
 			}
+			return nil
+		})
+		if retryErr != nil {
+			return fmt.Errorf("Failed to update Cluster Logging Fluentd status: %v", retryErr)
 		}
 	}
 	if cluster.Spec.Collection.LogCollection.Type == logging.LogCollectionTypeRsyslog {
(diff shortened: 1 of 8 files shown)


ewolinetz:
@richm I had trouble timing it to recreate the case where an outside entity updated the CR right before the operator wanted to update it. But this now gracefully handles cases where the CR is deleted and objects are missing during execution. We also have the `retry` logic in there when we would do any updates.

pkg/k8shandler/collection.go
```diff
@@ -47,18 +49,22 @@ func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
...
 		}
 
-		if !reflect.DeepEqual(fluentdStatus, cluster.Status.Collection.LogCollection.FluentdStatus) {
-			logrus.Infof("Updating status of Fluentd")
-			cluster.Status.Collection.LogCollection.FluentdStatus = fluentdStatus
-
-			if err = sdk.Update(cluster); err != nil {
-				return fmt.Errorf("Failed to update Cluster Logging Fluentd status: %v", err)
+		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
```
richm: if cluster logging does not exist, is that an error condition?

pkg/k8shandler/collection.go
```diff
@@ -47,18 +49,22 @@ func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
...
+				if !reflect.DeepEqual(fluentdStatus, cluster.Status.Collection.LogCollection.FluentdStatus) {
+					logrus.Infof("Updating status of Fluentd")
```
richm: you can just use `logrus.Info` since you aren't doing any formatting here

pkg/k8shandler/collection.go
```diff
@@ -56,7 +56,7 @@ func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
 				if !reflect.DeepEqual(fluentdStatus, cluster.Status.Collection.LogCollection.FluentdStatus) {
-					logrus.Infof("Updating status of Fluentd")
+					logrus.Info("Updating status of Fluentd")
```
richm: I think it is useful to print this, but not more than once if looping in the `RetryOnConflict`

pkg/k8shandler/collection.go
```diff
@@ -92,7 +92,7 @@ func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
 				if !reflect.DeepEqual(rsyslogStatus, cluster.Status.Collection.LogCollection.RsyslogStatus) {
-					logrus.Infof("Updating status of Rsyslog")
+					logrus.Info("Updating status of Rsyslog")
```
richm: I think it is useful to print this, but not more than once if looping in the `RetryOnConflict`

pkg/k8shandler/collection.go
```diff
@@ -505,7 +505,8 @@ func updateFluentdDaemonsetIfRequired(desired *apps.DaemonSet) (err error) {
 		if flushBuffer {
 			current.Spec.Template.Spec.Containers[0].Env = append(current.Spec.Template.Spec.Containers[0].Env, v1.EnvVar{Name: "FLUSH_AT_SHUTDOWN", Value: "True"})
 			if err = sdk.Update(current); err != nil {
-				return fmt.Errorf("Failed to prepare Fluentd daemonset to flush its buffers: %v", err)
+				logrus.Errorf("Failed to prepare Fluentd daemonset to flush its buffers: %v", err)
```
richm: I don't think you need to print the error here - instead, print the error as the error return from the `RetryOnConflict` - if you really need to print something here, make it `Debugf` instead.

pkg/k8shandler/curation.go
```diff
@@ -49,7 +49,7 @@ func CreateOrUpdateCuration(cluster *logging.ClusterLogging) (err error) {
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
 				if !reflect.DeepEqual(curatorStatus, cluster.Status.Curation.CuratorStatus) {
-					logrus.Infof("Updating status of Curator")
+					logrus.Info("Updating status of Curator")
```
richm: I think it is useful to print this, but not more than once if looping in the RetryOnConflict

pkg/k8shandler/logstore.go
```diff
@@ -39,7 +39,7 @@ func CreateOrUpdateLogStore(cluster *logging.ClusterLogging) (err error) {
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
 				if !reflect.DeepEqual(elasticsearchStatus, cluster.Status.LogStore.ElasticsearchStatus) {
-					logrus.Infof("Updating status of Elasticsearch")
+					logrus.Info("Updating status of Elasticsearch")
```
richm: I think it is useful to print this, but not more than once if looping in the RetryOnConflict

pkg/k8shandler/visualization.go
```diff
@@ -58,7 +58,7 @@ func CreateOrUpdateVisualization(cluster *logging.ClusterLogging) (err error) {
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
 				if !reflect.DeepEqual(kibanaStatus, cluster.Status.Visualization.KibanaStatus) {
-					logrus.Infof("Updating status of Kibana")
+					logrus.Info("Updating status of Kibana")
```
richm: I think it is useful to print this, but not more than once if looping in the RetryOnConflict

pkg/k8shandler/collection.go
ewolinetz: Would doing something like the following make sense then? Its kind of ugly but prevents us from updating when not necessary...
```
if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
	if !reflect.DeepEqual(fluentdStatus, cluster.Status.Collection.LogCollection.FluentdStatus) {
		logrus.Info("Updating status of Fluentd")
			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
				if !reflect.DeepEqual(fluentdStatus, cluster.Status.Collection.LogCollection.FluentdStatus) {
					cluster.Status.Collection.LogCollection.FluentdStatus = fluentdStatus
					return sdk.Update(cluster)
				}
			}
			return nil
		})
		if retryErr != nil {
			return fmt.Errorf("Failed to update Cluster Logging Fluentd status: %v", retryErr)
		}
	}
}
```

richm:
lgtm - squash and we can commit

ewolinetz:
commits squashed

PR #49, opened 2018-12-04
make deploy-example will now deploy full cluster

richm:
This allows you to do a single `make deploy-example` and get
a full cluster.  It assumes you have the latest
`elasticsearch-operator` cloned to `../elasticsearch-operator`


richm:
with this commit and https://github.com/openshift/origin-aggregated-logging/pull/1442 we are able to run the full CI tests against a logging cluster deployed with operators

Makefile
```diff
@@ -77,6 +78,9 @@ deploy-image:
 deploy: deploy-setup image deploy-image
 	hack/deploy.sh
 
+deploy-example: deploy
+	oc create -n $(NAMESPACE) -f hack/cr.yaml
```
ewolinetz: `hack/cr.yaml` is missing the managed spec field, `hack/examples/cr.yaml` has it though.
I'm find with whichever one we use but we should ensure it has that spec and lets remove the dupe

richm:
working now - please re-lgtm

PR #50, opened 2018-12-05
Dep bump for EO change and updating elasticsearch spec to expose ReplicationPolicy

ewolinetz:
Follow up from https://github.com/openshift/elasticsearch-operator/pull/33


josefkarasek:
Can you default to `FullReplication`? Two reasons for it:
* it's easier to hack on the Elasticsearch Operator when all the indices are not in red state after a cluster restart
* if you deploy a single node, there's no real difference between the respective replication strategies

ewolinetz:
@josefkarasek this is just the test CR, this isn't how users would normally deploy. My thought was this would save resources on the test nodes since there are less shards (e.g. for our CI tests). 

The indices would be red as long as any primary shards aren't placed which is independent of any shard replication.

PR #55, opened 2018-12-11
Updating to generate own signing CA for Cluster Logging

ewolinetz:
Updates the CLO to no longer require a secret to be created and mounted by customers. Will create a signing CA if one does not exist and store it in a secret.

Addresses https://jira.coreos.com/browse/LOG-255

### DIFF
--- a/pkg/k8shandler/certificates.go
+++ b/pkg/k8shandler/certificates.go
@@ -10,11 +10,63 @@ import (
 	"path"
 
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/util/retry"
 
 	sdk "github.com/operator-framework/operator-sdk/pkg/sdk"
 	k8sutil "github.com/operator-framework/operator-sdk/pkg/util/k8sutil"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
+
+// golang doesn't allow for const maps
+var secretCertificates = map[string]map[string]string{
+	"master-certs": map[string]string{
+		"masterca":  "ca.crt",
+		"masterkey": "ca.key",
+	},
+	"elasticsearch": map[string]string{
+		"elasticsearch.key": "elasticsearch.key",
+		"elasticsearch.crt": "elasticsearch.crt",
+		"logging-es.key":    "logging-es.key",
+		"logging-es.crt":    "logging-es.crt",
+		"admin-key":         "system.admin.key",
+		"admin-cert":        "system.admin.crt",
+		"admin-ca":          "ca.crt",
+	},
+	"kibana": map[string]string{
+		"ca":   "ca.crt",
+		"key":  "system.logging.kibana.key",
+		"cert": "system.logging.kibana.crt",
+	},
+	"kibana-proxy": map[string]string{
+		"server-key":  "kibana-internal.key",
+		"server-cert": "kibana-internal.crt",
+	},
+	"curator": map[string]string{
+		"ca":       "ca.crt",
+		"key":      "system.logging.curator.key",
+		"cert":     "system.logging.curator.crt",
+		"ops-ca":   "ca.crt",
+		"ops-key":  "system.logging.curator.key",
+		"ops-cert": "system.logging.curator.crt",
+	},
+	"fluentd": map[string]string{
+		"app-ca":     "ca.crt",
+		"app-key":    "system.logging.fluentd.key",
+		"app-cert":   "system.logging.fluentd.crt",
+		"infra-ca":   "ca.crt",
+		"infra-key":  "system.logging.fluentd.key",
+		"infra-cert": "system.logging.fluentd.crt",
+	},
+	"rsyslog": map[string]string{
+		"app-ca":     "ca.crt",
+		"app-key":    "system.logging.rsyslog.key",
+		"app-cert":   "system.logging.rsyslog.crt",
+		"infra-ca":   "ca.crt",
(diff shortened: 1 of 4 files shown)


pkg/k8shandler/curation.go
```diff
@@ -255,7 +254,7 @@ func updateCuratorIfRequired(desired *batch.CronJob) (err error) {
 	current := desired.DeepCopy()
 
 	if err = sdk.Get(current); err != nil {
-		if apierrors.IsNotFound(err) {
+		if errors.IsNotFound(err) {
```
richm: where does `errors` come from now if not `"k8s.io/apimachinery/pkg/api/errors"`?

scripts/cert_generation.sh
```diff
@@ -197,13 +214,28 @@ function generate_certs() {
...
+    extension_index=2
+    use_comma=1
+  fi
+
   for name in ${cert_names//,/}; do
-    extension_names="${extension_names},DNS.${extension_index}:${name}"
+    if [ $use_comma = 1 ]; then
+      extension_names="${extension_names},DNS.${extension_index}:${name}"
+    else
+      extension_names="${extension_names}DNS.${extension_index}:${name}"
```
richm: ok - but since `$use_comma` is `0` here, you could just say `extension_names="DNS.${extension_index}:${name}"`

pkg/k8shandler/certificates.go
```diff
@@ -45,6 +50,72 @@ func extractSecretToFile(namespace string, secretName string, key string, toFile
...
+			"masterkey":  utils.GetFileContents("/tmp/_working_dir/ca.key"),
+			"kibanacert": utils.GetFileContents("/tmp/_working_dir/kibana-internal.crt"),
+			"kibanakey":  utils.GetFileContents("/tmp/_working_dir/kibana-internal.key"),
+		})
+
+	utils.AddOwnerRefToObject(secret, utils.AsOwner(logging))
+
+	err = sdk.Create(secret)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		return fmt.Errorf("Failure constructing Curator secret: %v", err)
```
jcantrill: Is this error message correct? `Curator`

pkg/k8shandler/certificates.go
```diff
@@ -45,6 +50,72 @@ func extractSecretToFile(namespace string, secretName string, key string, toFile
...
+			"masterca":   utils.GetFileContents("/tmp/_working_dir/ca.crt"),
+			"masterkey":  utils.GetFileContents("/tmp/_working_dir/ca.key"),
+			"kibanacert": utils.GetFileContents("/tmp/_working_dir/kibana-internal.crt"),
+			"kibanakey":  utils.GetFileContents("/tmp/_working_dir/kibana-internal.key"),
+		})
+
+	utils.AddOwnerRefToObject(secret, utils.AsOwner(logging))
+
+	err = sdk.Create(secret)
+	if err != nil && !errors.IsAlreadyExists(err) {
```
jcantrill: Isn't it better to return early if the secret exists already?

pkg/k8shandler/certificates.go
```diff
@@ -45,6 +50,72 @@ func extractSecretToFile(namespace string, secretName string, key string, toFile
...
+		return fmt.Errorf("Failure constructing Curator secret: %v", err)
+	}
+
+	if err != nil && errors.IsAlreadyExists(err) {
```
jcantrill: Style... since both blocks depend on `err != nil` why not return early if it is nil which removes indenting, conditions, improves readability

pkg/k8shandler/certificates.go
```diff
@@ -45,6 +50,72 @@ func extractSecretToFile(namespace string, secretName string, key string, toFile
 	return nil
 }
 
+func writeSecret(logging *v1alpha1.ClusterLogging) (err error) {
+
+	secret := utils.Secret(
+		"logging-master-ca",
```
jcantrill: Do we need to still call this 'logging-master-ca'? Is there a way to be more descriptive and drop the 'logging'?  Is it really named correctly if we are including ca, cert, keys?

scripts/cert_generation.sh
```diff
@@ -228,5 +261,6 @@ generate_certs 'system.logging.curator'
 generate_certs 'system.admin'
 
 # TODO: get es SAN DNS, IP values from es service names
```
jcantrill: Why do we want to do this?  Isnt this an issue when the pods go up and down?

pkg/k8shandler/certificates.go
ewolinetz: if it exists we still want to make sure the certs aren't expired, if they were then we need to update the secret because we generated new ones.

pkg/k8shandler/certificates.go
ewolinetz: it isn't, good catch

pkg/k8shandler/curation.go
ewolinetz: we had that import twice... once named as `apierrors` and once as `k8s.io/apimachinery/pkg/api/errors`

pkg/k8shandler/certificates.go
ewolinetz: Updating it to be:
```
if err != nil {
		if !errors.IsAlreadyExists(err) {
			return fmt.Errorf("Failure constructing logging-master-ca secret: %v", err)
		}

		current := secret.DeepCopy()
		...
```

pkg/k8shandler/certificates.go
ewolinetz: renamed to "master-certs"

PR #70, opened 2019-01-02
Changing annotation default behavior for all-in-one, renaming spec fi…

ewolinetz:
…elds, updating ES spec types

This requires https://github.com/openshift/elasticsearch-operator/pull/55


hack/cr.yaml
```diff
@@ -2,17 +2,15 @@ apiVersion: "logging.openshift.io/v1alpha1"
 kind: "ClusterLogging"
 metadata:
   name: "example"
```
jcantrill: can we change this to 'logging' or 'cluster-logging' since there is no reason the 'example' cant be what we recommend

pkg/utils/utils.go
```diff
@@ -20,12 +20,12 @@ import (
 )
 
 const WORKING_DIR = "/tmp/_working_dir"
-const ALLINONE_ANNOTATION = "io.openshift.clusterlogging.alpha/allinone"
+const SPLIT_ANNOTATION = "io.openshift.clusterlogging.alpha/splitinstallation"
```
jcantrill: I still think that if we intend to not support 'split' log flows in the current incarnation we should drop the concept here all together and look to utilize our pipeline concept.  This is our opportunity to clean up 4.0:
*  It is going to be green deployments, 
* we still need to figure out how to adopt existing clusters which we can address split scenarios then,
* ops doesnt utilize this arrangement now
* Split is the 2% use case and not the 98%

hack/cr.yaml
ewolinetz: given that this is under `hack`, I personally would like to leave it as "example" to discourage customers using this file since customers should be using OLM to create this normally.

pkg/utils/utils.go
jcantrill: Conversation with @tkatarki, he agreed we will remove the notion of 'split' deployment in 4.0. This is  the time for us to clean the slate.

ewolinetz: This is something we have customers today using though and we agreed in our meeting that we would flip the functionality of not having the annotation from being that we split to being it is all-in-one so that if a customer still wanted to split their installation it would be far less painful than if we didn't provide a way to do so.

Also, when we go to adopt an existing split install cluster if we want to move away from a split install we will need to have a way to move ES data into a consolidated cluster.

hack/cr.yaml
ewolinetz: I think we should make sure that the CR created by OLM doesn't have `example` as its name... i would agree that `cluster-logging` is the most appropriate name there

jcantrill: We'll address on next PR to OLM to update with the lastest CR bits

ewolinetz:
waiting for changes on https://github.com/openshift/elasticsearch-operator/pull/54 to be merged in

ewolinetz:
@jcantrill status spec change made

pkg/k8shandler/status.go
```diff
@@ -149,7 +149,7 @@ func getElasticsearchStatus(namespace string) ([]v1alpha1.ElasticsearchStatus, e
 
 			nodeStatus := v1alpha1.ElasticsearchStatus{
 				ClusterName:   node.Name,
-				Replicas:      node.Spec.Nodes[0].Replicas,
+				Replicas:      node.Spec.Nodes[0].NodeCount,
```
jcantrill: nit.  this should probably be 'NodeCount' or a more clear status for when we actually support alternte deployment types would be some form of `NodeCount: {master: 1, data: 1, client:1}'.

PR #71, opened 2019-01-03
Create role and cluster role binding for reading sharing-config cm

ewolinetz:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1662397

Will require OLM role update (cc @jcantrill)


richm:
lgtm - but can we merge this without the OLM role change?

ewolinetz:
@richm we can, but the operator won't have permissions to manage its role and role binding (when deployed by means other than our makefile)

PR #72, opened 2019-01-03
fixes formating code to format repo code only and run on build

jcantrill:
This PR:

* runs formatting code on build
* only formats repo code


Makefile
```diff
@@ -61,7 +62,7 @@ image: imagebuilder
 	$(IMAGE_BUILDER) -t $(IMAGE_TAG) . $(IMAGE_BUILDER_OPTS)
 
 fmt:
-	@gofmt -l -w $(SRC)
```
ewolinetz: why is this not just `gofmt -l -w pkg` and `gofmt -l -w test` ?

jcantrill: ...and cmd?  Are these 2 separate lines?  Is that a viable syntax?

richm: no - the syntax would be 
`gofmt -l -w pkg && gofmt -l -w test`

jcantrill: updated.

PR #77, opened 2019-01-08
added more prerequisites details

jatanmalde:
-  Added the requirement of docker package and service up and running.

signoff by Jatan Malde (jmalde@redhat.com)


README.md
```diff
@@ -29,6 +29,7 @@ deploying the operator assumes:
 * `make` targets are executed from the `openshift/cluster-logging-operator` root directory
 * various other commands such as `imagebuilder` and `operator-sdk` - it is suggested
   that you use the logging `sdk_setup.sh` script at https://raw.githubusercontent.com/openshift/origin-aggregated-logging/master/hack/sdk_setup.sh
+* Ensure you have `docker` package installed and `docker` daemon up and running on the workstation you are running these commands.
```
jcantrill: This should probably just mention container runtime as CRIO replaces docker in 4.0 as far as I know.  @richm thoughts on how to properly word this?

jatanmalde: I think container runtime would be appropriate. I will make that change and would readd it.

richm: well - we currently use docker `docker push` to push images to the remote repo running in the cluster - is there another way to do that without docker?  `crictl` or `skopeo`?

richm: I'm working on a way to do that in another way with buildconfigs - but I think many devs will want to build images locally and push to the remote registry in the cluster

jatanmalde: @richm yes docker push is one which I could see being used for the image so maybe, for now, we could mention these updates on the doc so that users testing the repo would not find it difficult setting up the cluster-logging-operator.

richm: @jatanmalde for the typical QE use case, you won't need to build and push the images, so perhaps just a note that if you use `REMOTE_REGISTRY=true` you will need to have `docker` available

jatanmalde: @richm I added the changes now, could check if that would be okay?

richm:
looks good - please squash into a single commit

README.md
```diff
@@ -29,6 +29,8 @@ deploying the operator assumes:
 * `make` targets are executed from the `openshift/cluster-logging-operator` root directory
 * various other commands such as `imagebuilder` and `operator-sdk` - it is suggested
   that you use the logging `sdk_setup.sh` script at https://raw.githubusercontent.com/openshift/origin-aggregated-logging/master/hack/sdk_setup.sh
+  
+**Note:** If you are using `REMOTE_REGISTRY=true`, ensure you have `docker` package installed and `docker` daemon up and running on the workstation you are running these commands.
```
richm: was there a change other than this?

jatanmalde: No I think a note and the requirements of docker package and service if remote registry is in picture was all in it.

jatanmalde:
@richm Thanks.

PR #80, opened 2019-01-09
Adding check if redundancy policy has changed

ewolinetz:
Noticed this gap while trying to recreate https://bugzilla.redhat.com/show_bug.cgi?id=1664497


ewolinetz:
/hold
while testing

PR #81, opened 2019-01-10
fix 1662273. Add roles and bindings to allow prometheus to access log…

jcantrill:
…ging

Initially fixes https://bugzilla.redhat.com/show_bug.cgi?id=1662273


jcantrill:
@ewolinetz I will revert the names.  I only renamed them because its many roles and many rolebindings

PR #84, opened 2019-01-14
Log collector switch check

ewolinetz:
If the log collector is switched from `fluentd` to `rsyslog` or visa-versa we will clean up/remove the old configuration (when we are managing the cluster).

Addresses https://jira.coreos.com/browse/LOG-319

### DIFF
--- a/pkg/k8shandler/collection.go
+++ b/pkg/k8shandler/collection.go
@@ -32,6 +32,8 @@ func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
 	}
 
 	if cluster.Spec.Collection.Logs.Type == logging.LogCollectionTypeFluentd {
+		removeRsyslog(cluster)
+
 		if err = createOrUpdateFluentdServiceAccount(cluster); err != nil {
 			return
 		}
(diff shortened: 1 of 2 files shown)


ewolinetz:
/hold

I would like to have a discussion regarding the creation of new objects when the CR is `Unmanaged`. As it stands we will create a new cluster, however we will not update any of the objects in such a state.

I believe we should only create objects if we are `Managed`.

jcantrill:
> /hold
> 
> I would like to have a discussion regarding the creation of new objects when the CR is `Unmanaged`. As it stands we will create a new cluster, however we will not update any of the objects in such a state.
> 
> I believe we should only create objects if we are `Managed`.

I think it is reasonable to expect the operator to spin up the cluster and then take no additional actions.  Without having a deployment what would be the point of creating the CR initially as Unmanaged?  There are no deployed resources upon which you, as the administrator, can modify to your liking.

ewolinetz:
> I think it is reasonable to expect the operator to spin up the cluster and then take no additional actions. Without having a deployment what would be the point of creating the CR initially as Unmanaged? There are no deployed resources upon which you, as the administrator, can modify to your liking.

This make sense to me and I think that was the initial reasoning behind why we implemented it as such... and as I was typing out a counter point to play devil's advocate it reaffirmed our current approach. I think I was mixing up `Unmanaged` and `Paused` (which we don't use).

ewolinetz:
/hold cancel
tested and verified on my cluster

PR #85, opened 2019-01-15
refactor to introduce local struct

jcantrill:
This change is to propose address https://jira.coreos.com/browse/LOG-357:

* introduction of a local struct to allow use of attaching methods
* Add code style based on golang conventions
* Improve readability and promote code reuse

### DIFF
--- a/pkg/k8shandler/collection.go
+++ b/pkg/k8shandler/collection.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	"github.com/sirupsen/logrus"
 	v1 "k8s.io/api/core/v1"
@@ -15,10 +16,13 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/util/retry"
 
-	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
 	sdk "github.com/operator-framework/operator-sdk/pkg/sdk"
 	apps "k8s.io/api/apps/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+)
+
+const (
+	clusterLoggingPriorityClassName = "cluster-logging"
 )
 
 var (
@@ -26,7 +30,7 @@ var (
 	timeout       = time.Second * 1800
 )
 
-func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
+func (cluster *ClusterLogging) CreateOrUpdateCollection() (err error) {
 
 	// there is no easier way to check this in golang without writing a helper function
 	// TODO: write a helper function to validate Type is a valid option for common setup or tear down
@@ -35,15 +39,15 @@ func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
 			return
 		}
 
-		if err = createOrUpdateCollectorServiceAccount(cluster); err != nil {
+		if err = cluster.createOrUpdateCollectorServiceAccount(); err != nil {
 			return
 		}
 	} else {
-		if err = utils.RemoveServiceAccount(cluster, "logcollector"); err != nil {
+		if err = utils.RemoveServiceAccount(cluster.Namespace, "logcollector"); err != nil {
 			return
 		}
 
-		if err = utils.RemovePriorityClass(cluster, "cluster-logging"); err != nil {
+		if err = utils.RemovePriorityClass(clusterLoggingPriorityClassName); err != nil {
 			return
 		}
 	}
@@ -68,7 +72,7 @@ func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
 
 		printUpdateMessage := true
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
+			if exists := cluster.Exists(); exists {
 				if !reflect.DeepEqual(fluentdStatus, cluster.Status.Collection.Logs.FluentdStatus) {
 					if printUpdateMessage {
 						logrus.Info("Updating status of Fluentd")
@@ -107,7 +111,7 @@ func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
 
 		printUpdateMessage := true
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
+			if exists := cluster.Exists(); exists {
 				if !reflect.DeepEqual(rsyslogStatus, cluster.Status.Collection.Logs.RsyslogStatus) {
 					if printUpdateMessage {
 						logrus.Info("Updating status of Rsyslog")
@@ -129,17 +133,18 @@ func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
 	return nil
 }
 
-func removeFluentd(cluster *logging.ClusterLogging) (err error) {
+func removeFluentd(cluster *ClusterLogging) (err error) {
 	if cluster.Spec.ManagementState == logging.ManagementStateManaged {
-		if err = utils.RemoveConfigMap(cluster, "fluentd"); err != nil {
+
+		if err = utils.RemoveConfigMap(cluster.Namespace, "fluentd"); err != nil {
 			return
 		}
 
-		if err = utils.RemoveSecret(cluster, "fluentd"); err != nil {
+		if err = utils.RemoveSecret(cluster.Namespace, "fluentd"); err != nil {
 			return
 		}
 
-		if err = utils.RemoveDaemonset(cluster, "fluentd"); err != nil {
+		if err = utils.RemoveDaemonset(cluster.Namespace, "fluentd"); err != nil {
 			return
 		}
 	}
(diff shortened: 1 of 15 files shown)


pkg/k8shandler/collection.go
```diff
@@ -5,36 +5,42 @@ import (
 	"io/ioutil"
 	"reflect"
 
+	"strings"
+	"time"
+
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
ewolinetz: fyi this is unnecessary, the defaulted name it gets is the last part of the package which is already "v1"

pkg/k8shandler/collection.go
```diff
@@ -5,36 +5,42 @@ import (
...
 var (
 	retryInterval = time.Second * 30
 	timeout       = time.Second * 1800
 )
 
-func CreateOrUpdateCollection(cluster *logging.ClusterLogging) (err error) {
+func (localLogging *ClusterLogging) CreateOrUpdateCollection() (err error) {
 
-	if err = createOrUpdateCollectionPriorityClass(cluster); err != nil {
+	cluster := localLogging.ClusterLogging
```
ewolinetz: if you update the function to be for `(cluster *logging.ClusterLogging)` does this line become unnecessary?

pkg/k8shandler/types.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+//DoesClusterLoggingExist determines if the specified cluster exists
+func (clusterLogging *ClusterLogging) DoesClusterLoggingExist() bool {
+	exists, _ := utils.DoesClusterLoggingExist(clusterLogging.ClusterLogging)
+	return exists
+}
+
+//createOrUpdateServiceAccount creates or updates a ServiceAccount for logging with the given name
+func (logging *ClusterLogging) createOrUpdateServiceAccount(name string) error {
```
ewolinetz: I would argue this belongs in `utils` and the functions that return an object definition would belong in here.

pkg/k8shandler/collection.go
jcantrill: ack.  this is copy/past from some other file

pkg/k8shandler/collection.go
jcantrill: This is really a 'stop gap' to make refactoring more digestible.  I would expect this line to be unnecessary after refactoring all the places where its needed

pkg/k8shandler/collection.go
jcantrill: interestingly, this is being reverted for me.  I wonder if there is some intelligence to 'go fmt' that is adding it back in

jcantrill:
@jcantrill   Here is my 2 cents about the PR as implemented by adding methods to the `pkg/api` folder:
* It basically requires you to more or less move all `cluster.foo()` logic which was in k8shandler to the `pkg/api` folder.  That seems less then desirable but probably a symmantic/style difference.  
* The meat of the logic which is in k8shandler also now becomes versioned because `pkg/api` is versioned.  This may make future updates non-intuative and or more difficult.  One solution would be to introduce interfaces but I don't see how we would need multiple versions from version to version given the logic probably wont drift much
* A better approach might be to introduce a simple wrapper object simply for the reason you can create it with the internal as an anonymous type and all fields just 'bubble' up.  It may also make it cleaner for us to migrate versions in future.

jcantrill:
@richm @ewolinetz   I have added a second commit based on my previous comment and recent experience with the code base.  It's my opinion we should work towards the newest comment in lieu of the first commit; it uses a 'wrapper' struct and allows us to keep the pks/api package clean.  This seems to fit into the pattern of other projects and operators as best as I can determine.  I personally don't find it any more challenging and think it will allow us to better consume version changes as well as evolve the 'private' api without needing to modify the 'public' api

richm:
lgtm but I don't really understand the motivation of this refactoring

jcantrill:
> lgtm but I don't really understand the motivation of this refactoring

The main motivations are:
* Improve readability
* Encapsulate certain logic, OO style in places where it makes sense in lieu of relying on functional aspects
* Collapse duplicate logic into reusable functions

My belief is by doing so the code will be cleaner, easier to understand, and maintain

ewolinetz:
/hold
I would argue that since this doesn't fix anything directly we should
refrain from merging it yet during a freeze.

pkg/stub/handler.go
```diff
@@ -34,49 +33,49 @@ func (h *Handler) Handle(ctx context.Context, event sdk.Event) error {
...
 }
 
-func Reconcile(cluster *logging.ClusterLogging) (err error) {
+func Reconcile(spec *logging.ClusterLogging) (err error) {
 	exists := true
-
+	cluster := k8shandler.NewClusterLogging(spec)
 	// Reconcile certs
-	if exists, cluster = utils.DoesClusterLoggingExist(cluster); exists {
+	if exists = cluster.Exists(); exists {
```
ewolinetz: we can just shorten this to `if cluster.Exists() {`

ewolinetz: That previous function also served to get us the most recent iteration of the cluster object so that we could update it without conflicts... does this still cover that case?

jcantrill: `cluster` is  a pointer meaning the data associated with the struct to which it points is mutable.  I followed the `sdk.Get` function and the way I interpreted it was that it would 'fill' the struct identified by the pointer. If read correctly then yes its covered.  I would prefer an additional set of eyes to ensure it behaves as I expect.

pkg/utils/factory.go
```diff
@@ -44,6 +47,9 @@ func Secret(secretName string, namespace string, data map[string][]byte) *core.S
 	}
 }
 
+func NewServiceAccount(accountName string, namespace string) *core.ServiceAccount {
+	return ServiceAccount(accountName, namespace)
```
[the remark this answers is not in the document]
jcantrill: It is only a stop gap until we move ALL the code to use `New....`.   The golang style convention is prefix functions with `New` if they represent constructors, which this is.  We should move to golang conventions as much as possible.

pkg/utils/utils.go
```diff
@@ -39,12 +32,26 @@ var COMPONENT_IMAGES = map[string]string{
...
+	Meta() metav1.ObjectMeta
+}
+
+func AllInOne(o KubernetesResource) bool {
+
+	//_, ok := logging.ObjectMeta.Annotations[SPLIT_ANNOTATION]
+	return true
+}
+
+func AsOwner(o KubernetesResource) metav1.OwnerReference {
```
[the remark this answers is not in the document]
jcantrill: I think this is a standin until KubernetesResource goes away to minimize impact of the other places change would need to be made

jcantrill:
@ewolinetz rebased to include latest and refactored 'Exists()' to check based on a copy

jcantrill:
block PR on https://github.com/openshift/cluster-logging-operator/pull/105

richm:
registry issue fixed - now running into the usual "no more aws resources - cannot install" issue
/test e2e-aws

pkg/apis/logging/v1alpha1/temp.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+func DoesClusterLoggingExist(cluster *ClusterLogging) bool {
+
+	// this is to avoid an invalid memory address panic
+	if cluster == nil {
+		return false
+	}
+
+	if err := sdk.Get(cluster); err != nil {
```
ewolinetz: didn't we want to avoid updating the ClusterLogging object? this will do that since we pass it as a pointer.
I think we want to do a deepcopy of it and then `get` on that object

pkg/k8shandler/curation.go
```diff
@@ -48,7 +48,7 @@ func CreateOrUpdateCuration(cluster *logging.ClusterLogging) (err error) {
 
 		printUpdateMessage := true
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
+			if exists := logging.DoesClusterLoggingExist(cluster); exists {
```
ewolinetz: shouldn't this be `cluster.Exists()` ?

pkg/apis/logging/v1alpha1/temp.go
jcantrill: We do that here: https://github.com/openshift/cluster-logging-operator/pull/85/files#diff-156742c24dc12665a82a7fed5cf7c8abR57   I'll update in this method too

pkg/k8shandler/logstore.go
```diff
@@ -39,7 +39,7 @@ func CreateOrUpdateLogStore(cluster *logging.ClusterLogging) (err error) {
 
 		printUpdateMessage := true
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
+			if exists := logging.DoesClusterLoggingExist(cluster); exists {
```
ewolinetz: `cluster.Exists()`?

pkg/k8shandler/visualization.go
```diff
@@ -59,7 +59,7 @@ func CreateOrUpdateVisualization(cluster *logging.ClusterLogging) (err error) {
 
 		printUpdateMessage := true
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			if exists, cluster := utils.DoesClusterLoggingExist(cluster); exists {
+			if exists := logging.DoesClusterLoggingExist(cluster); exists {
```
ewolinetz: `cluster.Exists()` ?

pkg/k8shandler/visualization.go
```diff
@@ -83,39 +83,40 @@ func CreateOrUpdateVisualization(cluster *logging.ClusterLogging) (err error) {
 
 func removeKibana(cluster *logging.ClusterLogging) (err error) {
 	if cluster.Spec.ManagementState == logging.ManagementStateManaged {
-		if err = utils.RemoveServiceAccount(cluster, "kibana"); err != nil {
+		namespace := cluster.Namespace
```
ewolinetz: if we do this in the visualization file, can we do this in the other ones too so they look consistent?

pkg/k8shandler/curation.go
jcantrill: This PR only does a partial refactoring to minimize impact across the code base.  it handles changes related to the collector

pkg/k8shandler/visualization.go
jcantrill: yes

jcantrill:
test e2e-aws

pkg/k8shandler/collection.go
```diff
@@ -350,7 +355,7 @@ func createOrUpdateRsyslogConfigMap(logging *logging.ClusterLogging) error {
 	)
 	rsyslogConfigMaps["rsyslog"] = rsyslogConfigMap
 	for name, cm := range rsyslogConfigMaps {
-		utils.AddOwnerRefToObject(cm, utils.AsOwner(logging))
+		logging.AddOwnerRefTo(cm)
```
ewolinetz: +1

ewolinetz:
/lgtm
I would like to also eventually condense all api objects to their own files e.g. anything `serviceaccount` related would be in a `serviceaccount.go` etc. That can be a future pr.

Also thanks for removing all the `AllInOne` logic that wasn't being used anymore.

PR #86, opened 2019-01-15
Removing components when not specified in the CR

ewolinetz:
Resolves https://jira.coreos.com/browse/LOG-321


ewolinetz:
ES logic dependent on commit in https://github.com/openshift/elasticsearch-operator/pull/62

PR #88, opened 2019-01-18
fix e2e-test for ci

jcantrill:
Fix e2e tests for use with ci

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 CURPATH=$(PWD)
 TARGET_DIR=$(CURPATH)/_output
+KUBECONFIG?=$(HOME)/.kube/config
 
 GOBUILD=go build
 BUILD_GOPATH=$(TARGET_DIR):$(TARGET_DIR)/vendor:$(CURPATH)/cmd
@@ -83,7 +84,7 @@ deploy: deploy-setup deploy-image
 deploy-example: deploy
 	oc create -n $(NAMESPACE) -f hack/cr.yaml
 
-test-e2e: deploy-image
+test-e2e:
 	hack/test-e2e.sh
 
 undeploy:
--- a/README.md
+++ b/README.md
@@ -28,82 +28,67 @@ deploying the operator assumes:
 * Logged into a cluster with an admin user who has the cluster role of `cluster-admin`
 * Access to a clone of the `openshift/elasticsearch-operator` repository
 * `make` targets are executed from the `openshift/cluster-logging-operator` root directory
-* various other commands such as `imagebuilder` and `operator-sdk` - it is suggested
-  that you use the logging `sdk_setup.sh` script at https://raw.githubusercontent.com/openshift/origin-aggregated-logging/master/hack/sdk_setup.sh
-  
-**Note:** If you are using `REMOTE_REGISTRY=true`, ensure you have `docker` package installed and `docker` daemon up and running on the workstation you are running these commands.
+* various other commands such as `imagebuilder` and `operator-sdk`. The
+   [sdk_setup.sh](https://raw.githubusercontent.com/openshift/origin-aggregated-logging/master/hack/sdk_setup.sh) will setup your environment.
 
 The deployment can be optionally modified using any of the following:
 
-*  `IMAGE_BUILDER` is the command to build the container image (default: `docker build`)
-*  `EXCLUSIONS` is list of manifest files that should be ignored (default: '')
-*  `OC` is the openshift binary to use to deploy resources (default: `oc` in path)
-*  `REMOTE_REGISTRY` Set to `true` if you are running the cluster on a different machine
-    than the one you are developing on. For example, if you are running a cluster in a
-    local libvirt or minishift environment, you may want to build the image on the host
-    and push them to the cluster running in the VM.
-    You will need a username with a password (i.e. not the default `system:admin` user).
-    If your cluster was deployed with the `allow_all` identity provider, you can create
-    a user like this: `oc login --username=admin --password=admin`, then assign it rights:
-    `oc login --username=system:admin`
-    `oc adm policy add-cluster-role-to-user cluster-admin admin`
-    If you used the new `openshift-installer`, it created a user named `kubeadmin`
+| Env Var | Default | Description|
+|---------|---------|------------|
+|`IMAGE_BUILDER`|`imagebuilder`| The command to build the container image|
+|`EXCLUSIONS`|none|The list of manifest files that should will be ignored|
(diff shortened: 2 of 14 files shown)


Makefile
```diff
@@ -83,7 +84,7 @@ deploy: deploy-setup deploy-image
 deploy-example: deploy
 	oc create -n $(NAMESPACE) -f hack/cr.yaml
 
-test-e2e: deploy-image
+test-e2e:
```
ewolinetz: why is this removed from here? the e2e tests currently rely on an image to be produced

test/e2e/clusterlogging_test.go
```diff
@@ -49,75 +47,13 @@ func TestClusterLogging(t *testing.T) {
...
 	// run subtests
 	t.Run("clusterlogging-group", func(t *testing.T) {
+
 		t.Run("Cluster with fluentd", ClusterLoggingClusterFluentd)
 		time.Sleep(time.Minute * 1) // wait for objects to be deleted/cleaned up
 		t.Run("Cluster with rsyslog", ClusterLoggingClusterRsyslog)
 	})
 }
 
-func createRequiredClusterRoleAndBinding(f *framework.Framework, ctx *framework.TestCtx) error {
```
ewolinetz: this didn't break e2e tests? I would think that this would cause errors in the CLO since it wouldn't have the cluster permissions to do anything with oauthclients and priorityclasses...

Makefile
jcantrill: It does rely on the image to be produced but CI has a mechanism for us to use the one built by CI and not one we have to both build and deploy.  This is the reason for the introduction of [1].  I think it's reasonable to require people who are manually running this test to explicitly build and deploy their own image.  IMO we need to cater to CI which is supposed to make our lives easier.
[1] https://github.com/openshift/cluster-logging-operator/pull/88/files#diff-aa5c0c7202e9c241e0c57a1d903f9ec2R21

test/e2e/clusterlogging_test.go
jcantrill: The tests were failing and I couldn't understand why.  I thought these were just duplicates for resources in the manifest but I see there are some differences.  Reverting.

hack/buildconfig-template.yaml
```diff
@@ -0,0 +1,31 @@
+---
```
jcantrill: Removing in favor of https://github.com/openshift/cluster-logging-operator/pull/89/files#diff-4c7729766fb0e3362bb4ecdd4880efbeR1

Makefile
ewolinetz: Ack, can we update our README to describe this then?

Makefile
```diff
@@ -1,5 +1,6 @@
 CURPATH=$(PWD)
 TARGET_DIR=$(CURPATH)/_output
+KUBECONFIG?=$(HOME)/.kube/config
```
richm: Is this necessary?  by default, if KUBECONFIG is not set, the oc command will look in $HOME/.kube/config

test/e2e/clusterlogging_test.go
```diff
@@ -13,13 +13,13 @@ import (
...
 	framework "github.com/operator-framework/operator-sdk/pkg/test"
-	rbac "k8s.io/api/rbac/v1beta1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	rbac "k8s.io/api/rbac/v1beta1"
 )
 
 var (
 	retryInterval        = time.Second * 5
-	timeout              = time.Second * 60
+	timeout              = time.Minute * 2
```
ewolinetz: just to keep it consistent can you make this 120 sec instead?

hack/test-e2e.sh
```diff
@@ -39,9 +45,9 @@ oc adm policy add-scc-to-user privileged -z rsyslog -n openshift-logging
 # allows rsyslog to query k8s api for all namespace/pod info
 oc adm policy add-cluster-role-to-user cluster-reader -z rsyslog -n openshift-logging
 
-TEST_NAMESPACE=openshift-logging go test ./test/e2e/... \
+TEST_NAMESPACE=${NAMESPACE} go test ./test/e2e/... \
   -root=$(pwd) \
-  -kubeconfig=$HOME/.kube/config \
+  -kubeconfig=${KUBECONFIG} \
```
richm: I see (comment from above) - is it necessary to pass in `-kubeconfig`?  It will not look at the env. var. $KUBECONFIG first, then $HOME/.kube/config?

test/e2e/clusterlogging_test.go
jcantrill: ack

hack/test-e2e.sh
jcantrill: I have no other answer then the e2e test failed unless I provided a path where it could be something other then $HOME/.kube/config.  My assumptions is maybe the CI templates setup the environment such that it writes the config to someplace other then $HOME.  Wouldn't this be valid for a node where you do not intend to support users

richm: ok - weird

Makefile
ewolinetz: That's really vague... by that I don't understand what I would set that variable to. I was meaning something to state that you will need to `make deploy-image` manually...

README.md
```diff
@@ -28,81 +28,68 @@ deploying the operator assumes:
...
-```
-$ REMOTE_REGISTRY=true make deploy-image
-```
+|Target|Description|
+|------|-----------|
+|`deploy-example`|Deploys all resources and creates an instance of the `cluster-logging-operator`, creates a ClusterLogging custom resource named `example`, starts the Elasticsearch, Kibana, and Fluentd pods running.|
+|`deploy`|Deploys all resources and creates an instance of the `cluster-logging-operator`, but does not create the CR nor start the components running.|
+|`undeploy`|Removes all `cluster-logging` resources and `openshift-logging` project|
+|`deploy-setup`|Setup the OKD cluster to support cluster-logging without deploying an instance of the `cluster-logging-operator`|
+|`deploy-image`|If you already have a remote cluster, and you just want to rebuild and redeploy the image to the remote cluster.  Use `make image` instead if you only want to build the imagebut not push it.|
```
ewolinetz: "image but"

Makefile
jcantrill: @ewolinetz better? https://github.com/openshift/cluster-logging-operator/pull/88/files#diff-04c6e90faac2675aa89e2176d2eec7d8R93

ewolinetz: works for me!

PR #89, opened 2019-01-18
support buildconfigs and no image build deployments

richm:
If USE_IMAGE_STREAM=true, setup a buildconfig and an
image stream in the remote repo, and build the operator image
there from the github source.  This may save a lot of time
as opposed to building the image locally and pushing the image
into the remote registry.
Also allow deploying without first building the image.  This
assumes something like a CI environment where the previous
CI stage builds the images needed to deploy and test.
depends on https://github.com/openshift/elasticsearch-operator/pull/65

### DIFF
--- a/hack/deploy-image.sh
+++ b/hack/deploy-image.sh
@@ -13,6 +13,29 @@ LOCAL_PORT=${LOCAL_PORT:-5000}
 
 tag=${tag:-"127.0.0.1:${registry_port}/$IMAGE_TAG"}
 
+if [ "${USE_IMAGE_STREAM:-false}" = true ] ; then
+    oc process \
+        -p CL_OP_GITHUB_URL=https://github.com/${CL_OP_GITHUB_REPO:-openshift}/cluster-logging-operator \
+        -p CL_OP_GITHUB_BRANCH=${CL_OP_GITHUB_BRANCH:-master} \
+        -f hack/image-stream-build-config-template.yaml | \
+      oc -n openshift create -f -
+    # wait for is and bc
+    for ii in $(seq 1 10) ; do
+        if oc -n openshift get bc cluster-logging-operator > /dev/null 2>&1 && \
+           oc -n openshift get is cluster-logging-operator > /dev/null 2>&1 ; then
+            break
+        fi
+        sleep 1
+    done
+    if [ $ii = 10 ] ; then
+        echo ERROR: timeout waiting for cluster-logging-operator buildconfig and imagestream to be available
+        exit 1
+    fi
+    # wait
+    oc -n openshift logs -f bc/cluster-logging-operator
+    exit 0
+fi
+
 ${IMAGE_TAG_CMD} $IMAGE_TAG ${tag}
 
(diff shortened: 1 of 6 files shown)


hack/deploy-image.sh
```diff
@@ -13,6 +13,29 @@ LOCAL_PORT=${LOCAL_PORT:-5000}
 
 tag=${tag:-"127.0.0.1:${registry_port}/$IMAGE_TAG"}
 
+if [ "${USE_IMAGE_STREAM:-false}" = true ] ; then
+    oc process \
+        -p CL_OP_GITHUB_URL=https://github.com/${CL_OP_GITHUB_REPO:-openshift}/cluster-logging-operator \
+        -p CL_OP_GITHUB_BRANCH=${CL_OP_GITHUB_BRANCH:-master} \
+        -f hack/image-stream-build-config.yaml | \
```
jcantrill: This file appears to be missing.  Is it a template like? https://github.com/openshift/cluster-logging-operator/pull/88/files#diff-571fb67fc88fd2f824bfc43d0c3ecc79R1

richm: fixed

hack/image-stream-build-config.yaml
```diff
@@ -0,0 +1,49 @@
...
+        ref: ${CL_OP_GITHUB_BRANCH}
+      type: Git
+    strategy:
+      dockerStrategy:
+        dockerfilePath: Dockerfile
+      type: Docker
+parameters:
+-
+  description: 'URL for cluster-logging-operator fork'
+  name: CL_OP_GITHUB_URL
```
jcantrill: Any reason not to simplify this `GITHUB_URL` and `GITHUB_BRANCH` or

richm: I wanted them to be consistent with the environment variable names - we have to use a unique prefix `CL_OP` or `ES_OP` for the environment variables because we will typically set both sets of env. vars. at the same time.

hack/deploy-image.sh
```diff
@@ -13,6 +13,29 @@ LOCAL_PORT=${LOCAL_PORT:-5000}
...
+            break
+        fi
+        sleep 1
+    done
+    if [ $ii = 10 ] ; then
+        echo ERROR: timeout waiting for cluster-logging-operator buildconfig and imagestream to be available
+        exit 1
+    fi
+    # build and wait
+    oc -n openshift start-build bc/cluster-logging-operator --wait
```
ewolinetz: same question here as https://github.com/openshift/elasticsearch-operator/pull/65/files#r249199081

hack/image-stream-build-config.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: v1
```
jcantrill: Can we rename this to something that ends with '-template.yaml'  to better identify its a resource that will require additional processing inorder to properly use.

richm:
I had a rebase conflict - please re-/lgtm

PR #95, opened 2019-01-29
Updating deps to pull in latest ES cr change

ewolinetz:
/hold

PR #98, opened 2019-02-05
Fixing setting of node resources in ES CR

ewolinetz:
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1671276

### DIFF
--- a/pkg/k8shandler/logstore.go
+++ b/pkg/k8shandler/logstore.go
@@ -108,8 +108,7 @@ func getElasticsearchCR(logging *logging.ClusterLogging, elasticsearchName strin
 		Roles:        []v1alpha1.ElasticsearchNodeRole{"client", "data", "master"},
 		NodeCount:    logging.Spec.LogStore.NodeCount,
 		NodeSelector: logging.Spec.LogStore.NodeSelector,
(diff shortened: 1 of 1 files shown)


ewolinetz:
/hold
I may need to retest this after #95

pkg/k8shandler/logstore.go
```diff
@@ -108,9 +108,6 @@ func getElasticsearchCR(logging *logging.ClusterLogging, elasticsearchName strin
 		Roles:        []v1alpha1.ElasticsearchNodeRole{"client", "data", "master"},
```
josefkarasek: For the purpose of Cluster logging, we only allow CDM nodes?
If yes, then this PR is good.

josefkarasek:
> We still have the ability to define different Resource requirements on different nodes

I think I wrote this days ago and it got published now, when I clicked at *submit review* :-1:

ewolinetz:
> For the purpose of Cluster logging, we only allow CDM nodes?

Correct, for now this is the only scenario we support when it comes from CLO. 

> We still have the ability to define different Resource requirements on different nodes

With CLO we are only going to expose the ability to configure the resources for all nodes at this time...

PR #100, opened 2019-02-06
update config docs to include nodeSelector

jcantrill:
cc @thoraxe

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -48,6 +48,19 @@ block with valid memory (e.g. 16Gi) and CPU values (e.g 1):
             cpu:
             memory:
 ```
+### Node Selectors
+Each component specification allows the component to target a specific node.  This is defined
+by modifying the `nodeSelector` block to include key/value pairs that correspond
+to specifically labeled nodes:
+```
+  spec:
+    logStore:
+      type: "elasticsearch"
+      elasticsearch:
+        nodeSelector:
+          node-type: infra
+          cluster-logging-component: es
+```
 
 ## Data Aggregation and Storage
 An Elasticsearch cluster is responsible for log aggregation.  Following is a sample


jcantrill:
cc @mburke5678

mburke5678:
@jcantrill What are the components that a user can specify a node selector? Are they the same as the components listed in your doc for image?
https://github.com/openshift/cluster-logging-operator/blob/4e31f478bd5760f9573dd25e27fb0311cf08c505/docs/configuration.md#component-images

Also, what are the values for the `cluster-logging-component` parameter for each?
Do we have recommendations for a *better* node type for each?

thoraxe:
LGTM

thoraxe:
@mburke5678 node labeling schemes are 100% up to the end user with the exception of `node-role.kubernetes.io/worker=""` which should (must?) be applied to any schedulable non-master node.

To that end, the example is... an example. `node-type` is not a "standard" label that I know of (it's `node-role.kubernetes.io/infra=""` and neither is `cluster-logging-component`.

If we are making recommendations for node _type_ what I think you are suggesting is that we make recommendations for node _features_ (eg: memory, #cores, etc). There isn't really such a thing as a node _type_ that I know of.

thoraxe:
@pweil- i don't think you need to call out anything other than a link to `MachineSet` documentation (for the case where you want to create nodes specifically to host logging components). Otherwise you run the risk of each operator re-duplicating some minimal documentation about the use of `MachineSets` and providing conflicting information.

In other words, the cloud team should document `MachineSets` and how to use them and apply node labeling and etc. "our" (logging) docs should refer to those docs with something to the effect of "if you want to create special nodes for logging components, check out this documentation on how to create nodes with labels".

Does that make any sense?

thoraxe:
@pweil- sorry, I completely misread your comment. You did say "that links to".

pweil-:
> if you want to create special nodes for logging components, check out this documentation on how to create nodes with labels

yes, this is enough to call out the use case in an obvious way :+1:

jcantrill:
@thoraxe 
> If we are making recommendations for node _type_ what I think you are suggesting is that we make recommendations for node _features_ (eg: memory, #cores, etc). There isn't really such a thing as a node _type_ that I know of.

Node labels are arbitrary as far as I know.  We should change it if there is a better example; there is nothing specific in this draft

jcantrill:
@mburke5678 
> @jcantrill What are the components that a user can specify a node selector? Are they the same as the components listed in your doc for image?
> https://github.com/openshift/cluster-logging-operator/blob/4e31f478bd5760f9573dd25e27fb0311cf08c505/docs/configuration.md#component-images
> 
Correct.
> Also, what are the values for the `cluster-logging-component` parameter for each?
> Do we have recommendations for a _better_ node type for each?
> Or, is `cluster-logging-component: es` a label on a node?
> 
as stated previously, AFAIK the key/value pairs are arbitrary.  Waiting on @thoraxe to identify if we should supply a better example.

thoraxe:
It depends on whether we care about product or upstream. Upstream doesn't care about subscriptions/infra nodes. So, if we use any example, perhaps something as simple as `node-role.kubernetes.io/logging=""` would work. Otherwise we can go the completely generic `foo=bar` route.

jcantrill:
> It depends on whether we care about product or upstream. Upstream doesn't care about subscriptions/infra nodes. So, if we use any example, perhaps something as simple as `node-role.kubernetes.io/logging=""` would work. Otherwise we can go the completely generic `foo=bar` route.

Openshift cluster-logging will not work on upstream because of our dependency on various openshift auth apis.

PR #101, opened 2019-02-07
Updating CLO to create role and binding for logcollector SA to use privileged scc

ewolinetz:


ewolinetz:
/hold
manual for testing

pkg/k8shandler/collection.go
```diff
@@ -241,6 +241,28 @@ func createOrUpdateCollectorServiceAccount(cluster *logging.ClusterLogging) erro
 		return fmt.Errorf("Failure creating Log collector privileged role binding: %v", err)
 	}
 
+	subject := utils.Subject(
```
richm:
```
collection.go:244:10: no new variables on left side of :=
```

hack/test-e2e.sh
```diff
@@ -5,7 +5,7 @@ if [ -n "${IMAGE_CLUSTER_LOGGING_OPERATOR:-}" ] ; then
   source "$(dirname $0)/common"
 fi
 
-IMAGE_CLUSTER_LOGGING_OPERATOR=${IMAGE_CLUSTER_LOGGING_OPERATOR:-quay.io/openshift/origin-cluster-logging-operator:latest}
+IMAGE_CLUSTER_LOGGING_OPERATOR=${IMAGE_CLUSTER_LOGGING_OPERATOR:-openshift/origin-cluster-logging-operator:latest}
```
richm: why?

ewolinetz: I'm seeing that the logcollector SA doesn't have the proper permissions even though I don't see this issue in my env. I'm testing to see if the issue is that we aren't actually testing the image being built. The correct fix would likely be in the release repo if that is the case.

hack/test-e2e.sh
```diff
@@ -5,7 +5,9 @@ if [ -n "${IMAGE_CLUSTER_LOGGING_OPERATOR:-}" ] ; then
   source "$(dirname $0)/common"
 fi
 
+IMAGE_STREAM_CLUSTER_LOGGING_OPERATOR=$(oc get is origin-cluster-logging-operator -n openshift-logging -o jsonpath='{.status.dockerImageRepository}')
```
richm: This doesn't work for me in a CI cluster e.g.
```
oc get project
NAME                 DISPLAY NAME                                                                         STATUS
ci-op-m57t3frq       ci-op-m57t3frq - pull-ci-openshift-origin-aggregated-logging-master-full-integ-aws   Active
```
the namespace name is unique for each ci run and is available in the env. var. `$OPENSHIFT_BUILD_NAMESPACE`.
```
oc -n ci-op-m57t3frq get is
NAME       DOCKER REPO                                             TAGS                                                                   UPDATED
pipeline   registry.svc.ci.openshift.org/ci-op-m57t3frq/pipeline   logging-kibana5,logging-fluentd,logging-elasticsearch5 + 6 more...     About an hour ago
release    registry.svc.ci.openshift.org/ci-op-m57t3frq/release    latest                                                                 About an hour ago
stable     registry.svc.ci.openshift.org/ci-op-m57t3frq/stable     logging-kibana5,logging-fluentd,logging-elasticsearch5 + 108 more...   About an hour ago
```
I think ci always uses `pipeline` as the name of the `is`.
```
reg=$( oc -n ci-op-m57t3frq get is pipeline -o jsonpath='{.status.dockerImageRepository}' )
echo $reg
docker-registry.default.svc:5000/ci-op-m57t3frq/pipeline
```
then I think the `image:` will be something like `docker-registry.default.svc:5000/ci-op-m57t3frq/pipeline:cluster-logging-operator` - note - without the `origin-`

hack/test-e2e.sh
```diff
@@ -5,7 +5,13 @@ if [ -n "${IMAGE_CLUSTER_LOGGING_OPERATOR:-}" ] ; then
   source "$(dirname $0)/common"
 fi
 
+IMAGE_STREAM_CLUSTER_LOGGING_OPERATOR=$(oc get is pipeline -n $OPENSHIFT_BUILD_NAMESPACE -o jsonpath='{.status.dockerImageRepository}')
```
jcantrill: The required image should be available with this var [1] as I disussed with @bbguimaraes.  If this is not working as advertised we should talk it over with the him, @stevekuznetsov and the rest of the test platform team

[1] https://github.com/openshift/ci-operator/blob/master/TEMPLATES.md#image_component

ewolinetz: Would we want the `IMAGE_` or `LOCAL_IMAGE_` in this case though? which is the one that was built as part of this CI run?

bbguimaraes: @jcantril: On second thought, that variable is available for the
(`ci-operator`) template, but it won't be visible to the test unless it is
listed as a parameter (it is not).

The `IMAGE_FORMAT` variable [is][0] available in the test, so it could be used.
The [format][1] is very similar to `IMAGE_<component>`.

I don't know if this is how we want to deploy and test an operator, however.
@smarterclayton may.

[0]: https://github.com/openshift/release/blob/d71c33b7b096c33da6d2aa8948160e9169242908/ci-operator/templates/openshift/installer/cluster-launch-installer-src.yaml#L102-L103
[1]: https://github.com/openshift/ci-operator/blob/master/TEMPLATES.md#image_format

richm:
this seems wrong:
```
hack/test-e2e.sh
+ set -e
+ '[' -n '' ']'
+ IMAGE_CLUSTER_LOGGING_OPERATOR=quay.io/openshift/origin-cluster-logging-operator:latest
+ IMAGE_MANIFEST_CLUSTER_LOGGING_OPERATOR=quay.io/openshift/origin-cluster-logging-operator:latest
```
The image built by the test pipeline and stored in the ci registry should be something like this: https://github.com/openshift/origin-aggregated-logging/pull/1503/files#diff-dd61b3f86178f248fd3b67e74e8d07efR31
```
    # we are running in the CI environment
    # OPENSHIFT_BUILD_NAMESPACE=ci-op-xxx
    # IMAGE_FORMAT=registry.svc.ci.openshift.org/ci-op-xxx/stable:${component}
```
So the image should be something like `registry.svc.ci.openshift.org/ci-op-xxx/pipeline:cluster-logging-operator`

hack/test-e2e.sh
smarterclayton: You are allowed to use IMAGE_FORMAT in your own testing, because you may want to access tags or images that aren't part of the release payload.  For instance, if you build your own tests image, or want to access official components directly like the CLI image or the tests image

smarterclayton: Generally use `IMAGE_X` or `LOCAL_IMAGE_X` over `IMAGE_FORMAT`, but we won't change the templates to add every possible one, so `IMAGE_FORMAT` is the tool we give the operator tests.

richm:
ok - squash into single commit and lgtm

pkg/utils/types.go
```diff
@@ -271,3 +272,79 @@ func PriorityClass(priorityclassName string, priorityValue int32, globalDefault
 		Description:   description,
 	}
 }
+
+func PolicyRule(apiGroups, resources, resourceNames, verbs []string) rbac.PolicyRule {
```
jcantrill: Please use accepted golang  style conventions to name these functions  `NewPolicyRule`, `NewPolicyRules`, `NewSubject`, etc

ewolinetz:
@jcantrill updated function names

PR #103, opened 2019-02-11
Default mem and cpu when not defined for CL

jcantrill:
fix https://bugzilla.redhat.com/show_bug.cgi?id=1675381

Default recommended memory and cpu when nothing is defined for ClusterLogging.  This defaults the values to be the same as OOTB for 3.11


pkg/apis/logging/v1alpha1/types.go
```diff
@@ -40,14 +40,14 @@ type VisualizationSpec struct {
 }
 
 type KibanaSpec struct {
-	Resources    v1.ResourceRequirements `json:"resources"`
-	NodeSelector map[string]string       `json:"nodeSelector,omitempty"`
-	Replicas     int32                   `json:"replicas"`
+	Resources    *v1.ResourceRequirements `json:"resources"`
```
ewolinetz: is there a reason we're moving to use a pointer here? We don't do this in the EO currently

jcantrill: Move to a pointer to know if someone specifies resource or not (via nil).  This is a clean way to know if we use what is provided or sane default

pkg/k8shandler/visualization.go
```diff
@@ -7,7 +7,8 @@ import (
 
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
ewolinetz: these are duplicates with different names... please remove the `corev1` one

pkg/k8shandler/curation.go
```diff
@@ -2,12 +2,14 @@ package k8shandler
 
 import (
 	"fmt"
+	"reflect"
+
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	"github.com/sirupsen/logrus"
-	"k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
ewolinetz: duplicates, please resolve

pkg/k8shandler/collection.go
```diff
@@ -10,6 +10,7 @@ import (
 
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
```
ewolinetz: duplicates

pkg/k8shandler/logstore.go
```diff
@@ -3,11 +3,12 @@ package k8shandler
 import (
 	"fmt"
 
+	"reflect"
+
 	"github.com/openshift/elasticsearch-operator/pkg/apis/elasticsearch/v1alpha1"
+	corev1 "k8s.io/api/core/v1"
```
ewolinetz: for naming consistency can we make this `v1` instead of `corev1` ?

Makefile
```diff
@@ -89,6 +90,8 @@ deploy-no-build: deploy-setup
 deploy-example: deploy
 	oc create -n $(NAMESPACE) -f hack/cr.yaml
 
+test-unit: 
+	@go test $(PKGS)
```
ewolinetz: +100 on adding unit tests, can we add them to another package directory other than `k8shander` though?

ewolinetz:
one nit about unit test directory structure, otherwise lgtm

Makefile
jcantrill: I grabbed the change from another operator so it should glob test files in any of the packages except those here: https://github.com/openshift/cluster-logging-operator/pull/103/files#diff-b67911656ef5d18c4ae36cb6741b7965R19

ewolinetz: right, im just asking that maybe we do something like `/pkg/unittests/` or something under `/pkg` so that it is easily found and distinguished from the e2e tests

jcantrill: @ewolinetz this is not how 'unittests' are organized in golang.  They live alongside the code, most likely because of visibility requirements.

PR #104, opened 2019-02-12
Exposing elasticsearch service with a route

nhosoi:
-- an alternative method of `openshift_logging_es_allow_external`: True
     if this is True, Elasticsearch will be exposed as a Route

https://jira.coreos.com/browse/LOG-283

### DIFF
--- a/docs/unmanaged_configuration.md
+++ b/docs/unmanaged_configuration.md
@@ -25,3 +25,39 @@ oc set env dc/fluentd MERGE_JSON_LOG=true
 * Possible log loss due to Elasticsearch rejecting documents due to inconsistent type mappings
 * Potential buffer storage leak caused by rejected message cycling
 * Overwrite of data for field with same names
+
+## ElasticSearch
+### Exposing elasticsearch service with a route
+```
+# Get ElasticSearch pod name
+ESPOD=$( oc get pods -l component=elasticsearch -o name | sed -e "s/pod\///" )
+
+# Extract CA cert.
+oc exec $ESPOD -- cat /etc/openshift/elasticsearch/secret/admin-ca > ./admin-ca
+
+# Create an yaml file for the ElasticSearch service
+cat << EOF > my_es_route.yaml
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: elasticsearch
+  namespace: openshift-logging
+spec:
+  host:
+  to:
+    kind: Service
+    name: elasticsearch
+  tls:
+    termination: reencrypt
+    caCertificate: |-
+EOF
+cat ./admin-ca | sed -e "s/^/      /" >> my_es_route.yaml
+echo "    destinationCACertificate: |-" >> my_es_route.yaml
+cat ./admin-ca | sed -e "s/^/      /" >> my_es_route.yaml
+
+# Create the service
+oc create -f my_es_route.yaml
+
+# Check the ElasticSearch service is exposed
+curl --silent --insecure -H "Authorization: Bearer $( oc whoami -t )" "https://$( oc get route elasticsearch -o jsonpath='{.spec.host}' ):443/.operations.*/_search" | jq
+```


docs/unmanaged_configuration.md
```diff
@@ -25,3 +25,50 @@ oc set env dc/fluentd MERGE_JSON_LOG=true
...
+metadata:
+  name: elasticsearch
+  namespace: openshift-logging
+spec:
+  host: elasticsearch.openshift-logging.svc.cluster.local
+  to:
+    kind: Service
+    name: elasticsearch
+  tls:
+    termination: passthrough
```
richm: This isn't exactly the same as https://github.com/openshift/openshift-ansible/blob/5032be34a3e7746a77e19c561b17198aaae3148e/roles/openshift_logging_elasticsearch/tasks/main.yaml#L508 which creates a reencrypt route.  https://github.com/openshift/openshift-ansible/blob/5032be34a3e7746a77e19c561b17198aaae3148e/roles/openshift_logging_elasticsearch/templates/route_reencrypt.j2

docs/unmanaged_configuration.md
```diff
@@ -25,3 +25,50 @@ oc set env dc/fluentd MERGE_JSON_LOG=true
 * Possible log loss due to Elasticsearch rejecting documents due to inconsistent type mappings
 * Potential buffer storage leak caused by rejected message cycling
 * Overwrite of data for field with same names
+
+## ElasticSearch
+### openshift_logging_es_allow_external
```
jcantrill: We should make no references to ansible style variables here.  The wording should be along the lines of "Exposing elasticsearch service with a route".   The goal is to describe how to enable features from 3.11 but avoid mentioning ansible since it isn't what we are using in 4.x.

docs/unmanaged_configuration.md
jcantrill: We should continue to use a re-encrypt route if possible.  Doesn't `passthrough` mean the termination ends at the service requiring the ES pod to be able to handle the certs?  @ewolinetz looking at the ansible code seems the destination ca is required for a re-encrypt route.  Is the one used by the operator accessible to an admin?

docs/unmanaged_configuration.md
```diff
@@ -25,3 +25,50 @@ oc set env dc/fluentd MERGE_JSON_LOG=true
...
+* Create a route for ElasticSearch.
+```
+$ oc create -f - << EOF
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: elasticsearch
+  namespace: openshift-logging
+spec:
+  host: elasticsearch.openshift-logging.svc.cluster.local
```
jcantrill: If we leave this blank, the platform will assign it a fully qualified name using its rules.  This means there will be no need for the steps to get the canonical name, or populate /etc/hosts

docs/unmanaged_configuration.md
```diff
@@ -25,3 +25,50 @@ oc set env dc/fluentd MERGE_JSON_LOG=true
...
+```
+* Add the ElasticSearch IP address and FQDN into /etc/hosts on the host the ElasticSearch is to be accessed.
+```
+# echo $ESIPADDR $ESFQDN >> /etc/hosts
+```
+* Get admin key, cert and CA cert from ElasticSearch secret
+```
+for i in admin-key admin-ca admin-cert; do oc exec $ESPOD -- cat /etc/openshift/elasticsearch/secret/$i > $i ; done
+```
+Assume the files are located at /path/to/{admin-ca,admin-cert,admin-key} on the host.
```
jcantrill: Assuming re-encrypt, these are the certs you, as an admin, provide? we should name them differently.  Do we need to provide a set of certs for re-encrypt or just the dest_ca_crt?

docs/unmanaged_configuration.md
nhosoi: Yes, I also noticed kibana route has termination: reencrypt/Redirect...
Let me try replacing passthrough with it.

docs/unmanaged_configuration.md
nhosoi: @jcantrill, @richm, Could you please help me?
I restarted with this yaml file.
http://pastebin.test.redhat.com/713292
It seems it's badly broken... :(
If I leave the host blank, then the spec.host is filled with this FQDN (note: ntest0 is a cluster name.)
```
  spec:
    host: elasticsearch-openshift-logging.apps.ntest0.devcluster.openshift.com
```
Then, my attempt to search in ElasticSearch with curl failed with this error.
curl: (51) SSL: certificate subject name 'system.admin' does not match target host name 'elasticsearch-openshift-logging.apps.ntest0.devcluster.openshift.com'

I'd guess my spec.tls.key, spec.tls.certificate, spec.tls.caCertificate, and spec.tls.destinationCACertificate should be wrong.  I put admin-key, admin-cert, admin-ca and admin-ca from elasticsearch secret there, individually. 

Looking at the ansible role, es_key is from system.logging.es.key,  es_cert is from system.logging.es.crt, and es_ca is from ca.crt.  Are they different from admin-key, -cert and -ca by any chance?

Also, according to the following task, it seems we need to generate destinationCACertificate.  But it's not clear to me how to do it manually...
https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/tasks/main.yaml#L470-L478

Thanks, in advance...

richm: > @jcantrill, @richm, Could you please help me?
> I restarted with this yaml file.
> http://pastebin.test.redhat.com/713292
> It seems it's badly broken... :(
> If I leave the host blank, then the spec.host is filled with this FQDN (note: ntest0 is a cluster name.)
> 
> ```
>   spec:
>     host: elasticsearch-openshift-logging.apps.ntest0.devcluster.openshift.com
> ```

Yes, you must have a host name.  In the ansible code, you must provide the hostname in openshift_logging_es_hostname (or use the default value): https://github.com/openshift/openshift-ansible/blob/release-3.11/roles/openshift_logging/defaults/main.yml#L118

This hostname is used by the router to figure out which service to use.  That is - all of the connections are coming into the router host + port.  The router looks at the requested hostname to figure out which service to send the request to.

I guess the route code auto-generates a hostname if you do not provide one.

> 
> Then, my attempt to search in ElasticSearch with curl failed with this error.
> curl: (51) SSL: certificate subject name 'system.admin' does not match target host name 'elasticsearch-openshift-logging.apps.ntest0.devcluster.openshift.com'
> 
> I'd guess my spec.tls.key, spec.tls.certificate, spec.tls.caCertificate, and spec.tls.destinationCACertificate should be wrong. I put admin-key, admin-cert, admin-ca and admin-ca from elasticsearch secret there, individually.

tls.key and tls.certificate must be the _server cert_, not admin-cert which is a _client_ cert.

> 
> Looking at the ansible role, es_key is from system.logging.es.key, es_cert is from system.logging.es.crt, and es_ca is from ca.crt. Are they different from admin-key, -cert and -ca by any chance?

Yes, they are server certs.

> 
> Also, according to the following task, it seems we need to generate destinationCACertificate. But it's not clear to me how to do it manually...
> https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/tasks/main.yaml#L470-L478
> 
> Thanks, in advance...

The destinationCACertificate is the CA cert for the CA that issued the elasticsearch server certs.

The way re-encrypt routes work is this:

client -> router -> elasticsearch

The router terminates the TLS connection, opens a new connection to elasticsearch, and forwards the request to elasticsearch.  Also note that you cannot use client cert auth with a router re-encrypt route - for client cert auth to work, the TLS termination must be done by Elasticsearch, which can be done with a passthrough route, or by using externalIPs on the service.  The original reason we added support for openshift_logging_es_allow_external was for CloudForms to use token based auth to access Elasticsearch from outside of the cluster, which is accomplished by a re-encrypt route.

Elasticsearch has a TLS server cert and key, issued by the logging CA.  The router also has a TLS server cert and key, issued by the logging CA (in the openshift-ansible code).  I think the router TLS server cert and key can be the same as the Elasticsearch TLS server cert key, not sure how the openshift-ansible code works.

ewolinetz: I have to look at the arguments for the command again to see if it can cover all gaps, but have we thought about providing an `oc expose` command for the elasticsearch service instead of a template file?

nhosoi: Thank you, @richm, @ewolinetz.

I think I extracted the server key and cert and put them in the spec.tls.key and cert.  But my attempt to search against the exposed elasticsearch fails with "Authentication finally failed"...  Does this mean the key and cert to "reencrypt" (router -> elasticsearch) are broken?  (Or is this curl command line not appropriate to test the exposed elasticsearch addr?)
```
$ curl --cacert ./admin-ca --cert ./logging-es.crt --key ./logging-es.key https://${ESFQDN}:443/.operations.*/_search
Authentication finally failed
$ curl --cacert ./admin-ca --cert ./admin-cert --key ./admin-key https://${ESFQDN}:443/.operations.*/_search
Authentication finally failed
```

richm: > Thank you, @richm, @ewolinetz.
> 
> I think I extracted the server key and cert and put them in the spec.tls.key and cert. But my attempt to search against the exposed elasticsearch fails with "Authentication finally failed"... Does this mean the key and cert to "reencrypt" (router -> elasticsearch) are broken? (Or is this curl command line not appropriate to test the exposed elasticsearch addr?)
> 
> ```
> $ curl --cacert ./admin-ca --cert ./logging-es.crt --key ./logging-es.key https://${ESFQDN}:443/.operations.*/_search
> Authentication finally failed
> $ curl --cacert ./admin-ca --cert ./admin-cert --key ./admin-key https://${ESFQDN}:443/.operations.*/_search
> Authentication finally failed
> ```

You cannot use client cert/key auth with a re-encrypt route.  You'll have to use token based auth e.g. see https://github.com/openshift/origin-aggregated-logging/blob/master/hack/testing/util.sh#L210

nhosoi: Thanks a lot, @richm !!  The following command line [1] returns the search results successfully!  One question remains...  The curl_es_with_token function hardcodes the port 9200 for the elasticsearch.  But my test hangs if I put 9200, while 443 works.  Is it something I should put in the elasticsearch route manifest?  Or is it ok if it works with 443?

[1]
`curl --silent --insecure -H "Authorization: Bearer $( oc whoami -t )" "https://elasticsearch.openshift-logging.svc.cluster.local:443/.operations.*/_search" | jq`

nhosoi: Another note: as @jcantrill suggested, without specifying spec.host in the original manifest nor adding IP FQDN to the local /etc/hosts, the command line to verify works!  It makes the steps simpler. :)
`curl --silent --insecure -H "Authorization: Bearer $( oc whoami -t )" "https://$( oc get route elasticsearch -o jsonpath='{.spec.host}' ):443/.operations.*/_search" | jq`

richm: When using a route, you have to use port `443` (or just leave it off - when using a url which begins with `https`, the port `443` is used by default).  The router listens on port `443`, and redirects your request based on the hostname to `eshost:9200`

docs/unmanaged_configuration.md
nhosoi: I tested with just the dest_ca_crt and verified it worked.  Thus, I'm going to drop the set of certs. :)

I got another question here...  I compared with the route for kibana which has both dest_ca_crt and ca_crt sharing the same cert value.  Are there any reason to do so?  Should the ElasticSearch route follow that way - having both dest_ca_crt and ca_crt?  Or having just the dest_ca_crt is good enough?

richm: We should do whatever openshift-ansible did when it created the reencrypt route for Elasticsearch.  If it is reusing the certs, that's fine.  If it is creating new certs, then we will have to do that too.

nhosoi: Ok.  This is the output from v3.11 elasticsearch route created with openshift_logging_es_allow_external=True.
http://pastebin.test.redhat.com/714785

I.e., there are both dest_ca_crt and ca_crt are placed.  I'd follow it.  Thanks!

docs/unmanaged_configuration.md
```diff
@@ -25,3 +25,39 @@ oc set env dc/fluentd MERGE_JSON_LOG=true
...
+## ElasticSearch
+### Exposing elasticsearch service with a route
+```
+# Get ElasticSearch pod name
+ESPOD=$( oc get pods -l component=elasticsearch -o name | sed -e "s/pod\///" )
+
+# Extract CA cert.
+oc exec $ESPOD -- cat /etc/openshift/elasticsearch/secret/admin-ca > ./admin-ca
+
+# Create an yaml file for the ElasticSearch service
```
jcantrill: Is it possible to essentially do this same step with `oc expose <servicename>`?

docs/unmanaged_configuration.md
```diff
@@ -25,3 +25,39 @@ oc set env dc/fluentd MERGE_JSON_LOG=true
...
+EOF
+cat ./admin-ca | sed -e "s/^/      /" >> my_es_route.yaml
+echo "    destinationCACertificate: |-" >> my_es_route.yaml
+cat ./admin-ca | sed -e "s/^/      /" >> my_es_route.yaml
+
+# Create the service
+oc create -f my_es_route.yaml
+
+# Check the ElasticSearch service is exposed
+curl --silent --insecure -H "Authorization: Bearer $( oc whoami -t )" "https://$( oc get route elasticsearch -o jsonpath='{.spec.host}' ):443/.operations.*/_search" | jq
```
jcantrill: Do we need to specify the port `443` since that is the default for `https`

docs/unmanaged_configuration.md
```diff
@@ -25,3 +25,39 @@ oc set env dc/fluentd MERGE_JSON_LOG=true
 * Possible log loss due to Elasticsearch rejecting documents due to inconsistent type mappings
 * Potential buffer storage leak caused by rejected message cycling
 * Overwrite of data for field with same names
+
+## ElasticSearch
```
jcantrill: Elasticsearch ... lower case `S`

docs/unmanaged_configuration.md
```diff
@@ -25,3 +25,39 @@ oc set env dc/fluentd MERGE_JSON_LOG=true
...
+### Exposing elasticsearch service with a route
+```
+# Get ElasticSearch pod name
```
jcantrill: Elasticsearch ... lower case `S`

docs/unmanaged_configuration.md
```diff
@@ -25,3 +25,39 @@ oc set env dc/fluentd MERGE_JSON_LOG=true
...
+spec:
+  host:
+  to:
+    kind: Service
+    name: elasticsearch
+  tls:
+    termination: reencrypt
+    caCertificate: |-
+EOF
+cat ./admin-ca | sed -e "s/^/      /" >> my_es_route.yaml
```
jcantrill: @richm Can this really only be the admin-ca we generate?  I thought the point was admin CA should only be relevant from the router to ES, and client to router is a well known CA (e.g. provided by Openshift, etc)

docs/unmanaged_configuration.md
richm: I don't see where `oc expose` allows you to specify the certs/keys

docs/unmanaged_configuration.md
richm: no

docs/unmanaged_configuration.md
richm: I think this is the CA cert for the CA that issued the Elasticsearch server cert.  `admin-ca` is slightly misleading - we are assuming that the same CA was used to issue the ES admin cert and the ES http(s) server cert.

richm: We could also get the `admin-ca` cert from the secret:
```bash
oc extract secret/elasticsearch --to=. --keys=admin-ca
```

jcantrill: I'm not sure that answers my question.  The caCertificate is browser->router.  That should be provided outside the context of logging should it not?  Should it be a BYO ca? Or am I misunderstanding here?

richm: caCertificate is optional - https://docs.okd.io/latest/architecture/networking/routes.html#secured-routes "Edge Termination" (which is the same as re-encrypt, in terms of the client/browser facing part)

The key and cert are required - these are the key and cert specifically for the route - not sure what the old openshift-ansible code does for these, if it reuses the Elasticsearch https server cert/key, or generates new server cert/key specifically for the route - https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_logging_elasticsearch/tasks/main.yaml#L508

The caCertificate is optional, depending on if the cert was issued by a well known CA (e.g. something like Verisign which is already in the browsers' CA list), or some sort of corporate CA which is by policy installed on all machines.  In our case, since we are creating the cert/key, we will have to put our CA cert in the caCertificate field, and browsers and other clients will need a copy of this CA cert (or allow insecure connections like `curl -k`).


as far as destinationCACertificate goes:

https://docs.okd.io/latest/architecture/networking/routes.html#secured-routes "Re-encryption Termination"

"destinationCACertificate is required for re-encryption. destinationCACertificate specifies a CA certificate to validate the endpoint certificate, securing the connection from the router to the destination pods. If the service is using a service signing certificate, or the administrator has specified a default CA certificate for the router and the service has a certificate signed by that CA, this field can be omitted.

If the destinationCACertificate field is left empty, the router automatically leverages the certificate authority that is generated for service serving certificates, and is injected into every pod as /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt. This allows new routes that leverage end-to-end encryption without having to generate a certificate for the route. This is useful for custom routers or the F5 router, which might not allow the destinationCACertificate unless the administrator has allowed it."

In our case, we cannot use the default service-ca.crt because our Elasticsearch server cert is custom issued by our custom CA.

docs/unmanaged_configuration.md
nhosoi: Thanks.  Updated.

docs/unmanaged_configuration.md
nhosoi: Thanks. Replaced the termination type with reencrypt

docs/unmanaged_configuration.md
```diff
@@ -25,3 +25,39 @@ oc set env dc/fluentd MERGE_JSON_LOG=true
 * Possible log loss due to Elasticsearch rejecting documents due to inconsistent type mappings
 * Potential buffer storage leak caused by rejected message cycling
 * Overwrite of data for field with same names
+
+## ElasticSearch
+### Exposing elasticsearch service with a route
+```
+# Get ElasticSearch pod name
+ESPOD=$( oc get pods -l component=elasticsearch -o name | sed -e "s/pod\///" )
```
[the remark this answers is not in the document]
nhosoi: Thanks!  Updated.

docs/unmanaged_configuration.md
nhosoi: Thanks. Removed ":443".

docs/unmanaged_configuration.md
nhosoi: :p  Replaced all "ElasticSearch" with "Elasticsearch".

docs/unmanaged_configuration.md
nhosoi: ditto...

docs/unmanaged_configuration.md
```diff
@@ -25,3 +25,39 @@ oc set env dc/fluentd MERGE_JSON_LOG=true
...
+
+# Extract CA cert.
+oc exec $ESPOD -- cat /etc/openshift/elasticsearch/secret/admin-ca > ./admin-ca
```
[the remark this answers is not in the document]
nhosoi: Shall I replace the line 36 with this command line as @richm suggested?
> oc extract secret/elasticsearch --to=. --keys=admin-ca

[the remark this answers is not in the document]
richm: If you need a cert/key from a secret, then yes, `oc extract` is the best way to do it.

nhosoi:
Closing this pr and moving it to EO doc with simplified instructions:
> not to document the same procedure used in 3.11 openshift-ansible elasticsearch role 
> for the reencrypt route, but just point to the existing openshift route docs, and tell them 
> how to get the CA cert to use for destinationCACertificate.

PR #108, opened 2019-03-04
use ZeroRedundancy for tests - add cr for aws deploys

richm:
tests use single es instance so use ZeroRedundancy
create a cr.yaml for use with OCP 4.0 installer on aws


ewolinetz:
overall lgtm, why the aws specific cr file though?

richm:
> overall lgtm, why the aws specific cr file though?

The only way to deploy on 4.0 currently is on aws, and the current aws machine defaults do not allow you to run elasticsearch.  So it is nice to have a file that you can just pass to `oc create -f` that will just work without any modifications.

PR #110, opened 2019-03-04
updating dns name generation for certs and updating hash alg

ewolinetz:
addresses https://bugzilla.redhat.com/show_bug.cgi?id=1684332 issue for 4.0


richm:
Is `sha256` the best current choice for signature algorithm?

ewolinetz:
@richm we could use sha512 ...

ewolinetz:
@jcantrill i think we would need to make a similar change for ansible where we change the md value in the signing confs

PR #116, opened 2019-03-12
Expanding and adding default cronjob schedule

ewolinetz:
Adding tests to verify as well


jcantrill:
@ewolinetz is there documentation we need to add?

ewolinetz:
@mburke5678 may need to update the docs to reflect the new default...

PR #118, opened 2019-03-15
Update OAuthClient to include GrantMethod

enj:
This field will be required in 4.x

@jcantrill


jcantrill:
/lgtm

cc @ewolinetz @josefkarasek  I think we also need this change for EO

PR #113, opened 2019-03-08
add manifest and update dockerfile for marketplace

jcantrill:
This adds manifest dir for use by Marketplace

PR #120, opened 2019-03-15
remove release operator label

jcantrill:
This PR removes the CLO from the CVO build pipeline

cc @bparees


ewolinetz:
based on slack discussions, lgtm. I think we need to switch this to another one though

jcantrill:
> based on slack discussions, lgtm. I think we need to switch this to another one though

What do you mean? This PR just removes us from the CVO pipeline and https://github.com/openshift/cluster-logging-operator/pull/113 will at some point add us to the RH operator hub pipeline

ewolinetz:
I was referring to `com.redhat.delivery.appregistry=true`

Issue #119, opened 2019-03-15
Errors using the deploy-example goal


jamezp:
I'm trying to get clustered logging work on a local instance of minishift (or cluster up) and am facing some issues.

## First issue
The first issue is it seems the `ELASTICSEARCH_OP_REPO` needs to be explicitly set otherwise I see this error:
```
+ popd
~/tmp/wildfly-efk/cluster-logging-operator
+ CREATE_ES_SECRET=false
+ NAMESPACE=openshift-logging
+ make -C hack/../../elasticsearch-operator deploy-setup
make[1]: *** hack/../../elasticsearch-operator: No such file or directory.  Stop.
make: *** [Makefile:91: deploy-setup] Error 2
```

## Second Issue
The second issue is the `vendor/github.com/openshift/elasticsearch-operator/hack/deploy-setup.sh` script has line where it point's to an invalid directory.
```
pushd vendor/github.com/coreos/prometheus-operator/example/prometheus-operator-crd
  for file in prometheusrule.crd.yaml servicemonitor.crd.yaml; do 
    oc create -n ${NAMESPACE} -f ${file} ||:
  done
popd
```

The directory `vendor/github.com/coreos/prometheus-operator/example/prometheus-operator-crd` does not exist.

## Third Issue
This one could actually just be the environment I'm attempting to use. It could be that I'm using OpenShift 3 and it looks like this targets OpenShift 4. Anyway the error is:

```
--> FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base as 1
--> RUN INSTALL_PKGS="       openssl       " &&     yum install -y $INSTALL_PKGS &&     rpm -V $INSTALL_PKGS &&     yum clean all &&     mkdir /tmp/_working_dir &&     chmod og+w /tmp/_working_dir
Loaded plugins: ovl, product-id, search-disabled-repos, subscription-manager
This system is not receiving updates. You can use subscription-manager on the host to register and assign subscriptions.
http://base-4-0.ocp.svc/rhel-fast-datapath/repodata/repomd.xml: [Errno 14] curl#6 - "Could not resolve host: base-4-0.ocp.svc; Unknown error"
Trying other mirror.


 One of the configured repositories failed (rhel-fast-datapath),
 and yum doesn't have enough cached data to continue. At this point the only
 safe thing yum can do is fail. There are a few ways to work "fix" this:

     1. Contact the upstream for the repository and get them to fix the problem.

[...]

PR #122, opened 2019-03-18
Script fixes

jamezp:
A couple fixes found in #119 for the scripts.

### DIFF
--- a/hack/common
+++ b/hack/common
@@ -6,7 +6,7 @@ fi
 alias oc=${OC:-oc}
 
 repo_dir="$(dirname $0)/.."
-ELASTICSEARCH_OP_REPO=${ELASTICSEARCH_OP_REPO:-${repo_dir}/../elasticsearch-operator}
+ELASTICSEARCH_OP_REPO=${ELASTICSEARCH_OP_REPO:-${repo_dir}/vendor/github.com/openshift/elasticsearch-operator}
 
 ADMIN_USER=${ADMIN_USER:-kubeadmin}
 ADMIN_PSWD=${ADMIN_USER:-admin123}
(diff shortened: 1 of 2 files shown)


ewolinetz:
I'm good with using the vendored in elasticsearch-operator. 
Can you explain why we're removing a script from the vendored directory though?

jamezp:
@ewolinetz It's just removing the processing of the `vendor/github.com/coreos/prometheus-operator/example/prometheus-operator-crd` directory as that directory doesn't exist. It's possible the correct solution is to add the directory, but I'm not sure what should go in there.

ewolinetz:
ack, I didn't expand that diff to see that. The only major concern I have with that is when we do a `dep ensure -update` we'll lose that change so it seems very fragile to me.

Also, without those files the Elasticsearch operator will have errors in it

richm:
The problem is that these scripts/Makefiles were written with developers in mind, and with those developers being both elasticsearch-operator and cluster-logging-operator developers at the same time, and who have both elasticsearch-operator and cluster-logging-operator checked out at the same level under ~/go/src/github.com/openshift (or whatever the GOPATH is) - not sure how to rewrite the scripts to give developers the ability to use either `../elasticsearch-operator` or `vendor/..../elasticsearch-operator`

jamezp:
Ah okay. This PR definitely could be wrong for sure. I was just attempting the [quick start example](https://github.com/openshift/cluster-logging-operator#quick-start). It's definitely possible I don't have the environment setup correct.

if this is wrong too feel free to close. I doesn't look like I'll be able to use this OpenShift 3.x anyway. I was just hoping to so I could test some WildFly/JBoss EAP logging changes that are coming in :)

jcantrill:
Closing Per @richm comment.  These are intended for use by dev

PR #123, opened 2019-03-18
Restrict to single instance of CL

jcantrill:
LOG-358. Restrict CL to single instance called "instance"

This PR additionally introduces an abstraction to allow mocking calls to runtime

### DIFF
--- a/hack/cr.yaml
+++ b/hack/cr.yaml
@@ -1,7 +1,7 @@
 apiVersion: "logging.openshift.io/v1alpha1"
 kind: "ClusterLogging"
 metadata:
-  name: "example"
+  name: "instance"
 spec:
   managementState: "Managed"
   logStore:
--- a/pkg/runtime/runtime.go
+++ b/pkg/runtime/runtime.go
@@ -1,0 +1,28 @@
+package runtime
+
+import (
+	"github.com/operator-framework/operator-sdk/pkg/sdk"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/util/retry"
+)
+
+//RetryOnConflict is a place for "k8s.io/client-go/util/retry#RetryOnConflict"
+type RetryOnConflict func(backoff wait.Backoff, fn func() error) error
+
+//SdkUpdate is a place holder for "github.com/operator-framework/operator-sdk/pkg/sdk#Update"
+type SdkUpdate func(object sdk.Object) error
+
+//OperatorRuntime is an adapter to the underlying runtime for the operator to
+//all mocking for testing
+type OperatorRuntime struct {
+	RetryOnConflict RetryOnConflict
+	Update          SdkUpdate
+}
+
+//New returns the default runtime
+func New() *OperatorRuntime {
+	return &OperatorRuntime{
+		RetryOnConflict: retry.RetryOnConflict,
+		Update:          sdk.Update,
+	}
+}
--- a/pkg/stub/handler_test.go
+++ b/pkg/stub/handler_test.go
@@ -1,0 +1,95 @@
+package stub
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
+	"github.com/openshift/cluster-logging-operator/pkg/runtime"
+	"github.com/operator-framework/operator-sdk/pkg/sdk"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+)
+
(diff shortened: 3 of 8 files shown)


jcantrill:
@richm we will need to make sure to update origin-aggregated-logging once this merges

pkg/stub/handler_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+func newClusterLogging(name string) *logging.ClusterLogging {
+	return &logging.ClusterLogging{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: name,
+		},
+	}
+
+}
+func TestUpdateStatusToIgnoreWhenNoError(t *testing.T) {
```
ewolinetz: is it really necessary to test this?

pkg/stub/handler_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+	spec := newClusterLogging("myclusterLogging")
+	if err := updateStatusToIgnore(spec, mockRuntime); err != nil {
+		t.Errorf("Exp. no error when updating the status for %q: %v", spec.Name, err)
+	}
+	if spec.Status.Message != singletonMessage {
+		t.Errorf("Exp. ClusterLogging status message to indicate it is not the singleton instance but was %q", spec.Status.Message)
+	}
+}
+func TestUpdateStatusToIgnoreUpdatingWhenMessageIsAlreadyCorrect(t *testing.T) {
```
ewolinetz: same question here

pkg/stub/handler_test.go
```diff
@@ -0,0 +1,95 @@
...
+	spec := newClusterLogging("myclusterLogging")
+	spec.Status.Message = singletonMessage
+	if err := updateStatusToIgnore(spec, mockRuntime); err != nil {
+		t.Errorf("Exp. no error when updating the status for %q: %v", spec.Name, err)
+	}
+	if spec.Status.Message != singletonMessage {
+		t.Errorf("Exp. ClusterLogging status message to indicate it is not the singleton instance but was %q", spec.Status.Message)
+	}
+}
+func TestUpdateStatusWhenRetryOnConflictError(t *testing.T) {
```
ewolinetz: same question here

pkg/runtime/runtime.go
```diff
@@ -0,0 +1,28 @@
+package runtime
```
ewolinetz: this feels like an unnecessary layer of abstraction

pkg/stub/handler_test.go
jcantrill: Only to improve code coverage

pkg/stub/handler_test.go
jcantrill: Only to improve code coverage

pkg/runtime/runtime.go
jcantrill: This is fundamentally what the SDK will provide to abstract the dependency on the API server.  It's simply a mechanism by which we can test the code blocks for which we have nothing now and control the outcome of the bits that don't allow us to test now.  It's no different then Mockito as a java library or any other mocking. This is simply an opportunity to add code which will allow us to expand our testing.

jcantrill:
> some of this seems unnecessary and overly complex for the intention of this fix

Its more then required, but adds the groundwork to allow us to test more paths which we can not at the moment.  Any logic that hits the API service is untestable via a unit test, which is bad

ewolinetz:
> Its more then required, but adds the groundwork to allow us to test more paths which we can not at the moment. Any logic that hits the API service is untestable via a unit test, which is bad

IMO that is why we have e2e tests. the unit test should be able to be done independent of the API. Things like making sure we correctly evaluate that we can process a CR, that we correctly generate specs, etc are what we should be unit testing.

jcantrill:
> IMO that is why we have e2e tests. the unit test should be able to be done independent of the API. Things like making sure we correctly evaluate that we can process a CR, that we correctly generate specs, etc are what we should be unit testing.

It is a fairly common practice in software development to mock external dependencies; ref [1] which uses the classic example of mocking a database (in golang no less).  I don't understand why more testing isn't desirable.  There are various layers in the test cycle and catching things early is always beneficial. The problem with deferring any of this testing to e2e is it now requires an actual OKD cluster on which to run, which introduces 20+ min to the feedback cycle

[1] https://rollout.io/blog/unit-testing-external-dependencies/
[2] https://medium.com/@harrygogonis/testing-go-mocking-third-party-dependancies-4ab4e1c9bd3f

hack/cr.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: "logging.openshift.io/v1alpha1"
 kind: "ClusterLogging"
 metadata:
-  name: "example"
+  name: "instance"
```
richm: please make the same change to hack/cr-aws.yaml

PR #127, opened 2019-03-20
1666674 for 4x

jcantrill:
fix https://bugzilla.redhat.com/show_bug.cgi?id=1666674 for 4.x

### DIFF
--- a/pkg/k8shandler/logging.go
+++ b/pkg/k8shandler/logging.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/pkg/runtime"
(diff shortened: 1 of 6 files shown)


richm:
e2e-operator is not a test flake - looks like a real error in fluentd:
```
--- FAIL: TestClusterLogging (333.14s)
    --- FAIL: TestClusterLogging/clusterlogging-group (60.00s)
        --- FAIL: TestClusterLogging/clusterlogging-group/Cluster_with_fluentd (135.62s)
```

pkg/k8shandler/logging.go
```diff
@@ -14,10 +15,15 @@ import (
 
 type ClusterLogging struct {
 	*logging.ClusterLogging
+	Runtime *runtime.OperatorRuntime
```
ewolinetz: we're using a runtime?

jcantrill: maybe bad naming choice but its just the interface to allow mocking the api calls

PR #128, opened 2019-03-21
1691444 node selectors

jcantrill:
This bug fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1691444

By adding node selectors for the operands

Last commit is the only one of interest


jcantrill:
tests wont pass until merge of https://github.com/openshift/cluster-logging-operator/pull/124 and https://github.com/openshift/cluster-logging-operator/pull/123

ewolinetz:
can we also add a test in here to verify that if we don't specify a node selector that it is still empty/nil? (verifying the current deployment use case)

jcantrill:
> can we also add a test in here to verify that if we don't specify a node selector that it is still empty/nil? (verifying the current deployment use case)

Added

PR #130, opened 2019-03-26
Bump CL to v1

jcantrill:
This PR bumps CL to v1

https://jira.coreos.com/browse/LOG-370


hack/cr-aws.yaml
```diff
@@ -1,6 +1,5 @@
-apiVersion: "logging.openshift.io/v1alpha1"
-kind: "ClusterLogging"
-metadata:
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogging"metadata:
```
richm: this doesn't look right

pkg/apis/logging/v1/types.go
```diff
@@ -1,4 +1,4 @@
-package v1alpha1
+package v1
 
 import (
 	"github.com/openshift/elasticsearch-operator/pkg/apis/elasticsearch/v1alpha1"
```
richm:
```
	v1 "k8s.io/api/core/v1"
```
do you need to change this import to `corev1` as in `zz_generated.deepcopy.go`?

jcantrill: only if there is a conflict and compile error.  That file is generated so they have taken the liberty of using a naming convention

hack/cr-aws.yaml
jcantrill: fixed

richm:
can this be merged independent of https://github.com/openshift/elasticsearch-operator/pull/106?

PR #133, opened 2019-03-27
rsyslog - support action retry, max wait

richm:
This makes rsyslog have the same behavior as fluentd
```
        retry_max_interval "#{ENV['ES_RETRY_WAIT'] || '300'}"
        retry_forever true
```
Ref https://jira.coreos.com/browse/LOG-399


files/rsyslog/55-elasticsearch.conf
```diff
@@ -61,6 +61,8 @@ ruleset(name="try_es") {
             queue.maxFileSize=`echo $ES_QUEUE_MAXFILESIZE`
             queue.checkpointInterval=`echo $ES_QUEUE_CHECKPOINTINTERVAL`
             queue.syncqueuefiles="on"
+            action.resumeRetryCount="-1"
+            action.resumeIntervalMax=`echo $ES_RETRY_WAIT`
```
nhosoi: I cannot find resumeIntervalMax but resumeInterval in the doc and the sources...  Could you point me to the reference?

richm: good catch - `action.resumeIntervalMax` is not supported in our version of rsyslog, even though it is listed on the page, and doesn't say `new in 8.xx version` - https://www.rsyslog.com/doc/v8-stable/configuration/actions.html#general-action-parameters

richm:
please review and lgtm

ewolinetz:
not sure why you put hold, but lgtm

richm:
the hold is because I thought we were not supposed to merge this for 4.1 - @bparees @jcantrill please confirm

bparees:
confirmed.  unless not merging this leaves us exposed to some GA functionality breakage due to the pieces of this that already landed.

but if it only relates to enabling what was going to be tech preview in 4.1, then it should be held for 4.2 instead.

jcantrill:
@richm i changed the base.  is there a jira card we can ref here

richm:
> @richm i changed the base. is there a jira card we can ref here

there is now - https://jira.coreos.com/browse/LOG-399

bparees:
why is this merging when the exception was not approved?

bparees:
oh sorry, i see it's a 4.2 branch
/hold cancel

PR #141, opened 2019-03-29
Generate manifest from CSV

jcantrill:
This PR uses a script to generate deployment manifest based on the CSV


hack/deploy.sh
```diff
@@ -5,7 +5,8 @@ set -euxo pipefail
 source "$(dirname $0)/common"
 
 if [ $REMOTE_REGISTRY = false ] ; then
-    oc create -n ${NAMESPACE} -f manifests/05-deployment.yaml
+    $repo_dir/hack/gen-olm-artifacts.sh ${CSV_FILE} ${NAMESPACE} 'dep' \
+    |  oc create -f -
```
richm: if you ran deploy-setup.sh it would create the `dep` already - will it cause an error if the `dep` already exists?

jcantrill: looks like it would.  For equivalence I will remove `dep` from `deploy-setup` which should make it equivalent

test/e2e/clusterlogging_test.go
```diff
@@ -215,78 +202,37 @@ func clusterLoggingUpgradeClusterTest(t *testing.T, f *framework.Framework, ctx
...
-		return fmt.Errorf("failed to get currentOperator: %v", err)
-	}
-
-	currentOperator.Spec.Template.Spec.Containers[0].Env = currentEnv
-	err = f.Client.Update(goctx.TODO(), currentOperator)
-	if err != nil {
-		return fmt.Errorf("could not update cluster-logging-operator with prior image values %v", err)
-	}
-
-	err = CheckForElasticsearchImageName(t, f.Client, namespace, "elasticsearch", "quay.io/openshift/origin-logging-elasticsearch5:latest", retryInterval, timeout)
```
ewolinetz: shouldn't we keep this?

hack/deploy.sh
```diff
@@ -5,7 +5,8 @@ set -euxo pipefail
 source "$(dirname $0)/common"
 
 if [ $REMOTE_REGISTRY = false ] ; then
-    oc create -n ${NAMESPACE} -f manifests/05-deployment.yaml
+    $repo_dir/hack/gen-olm-artifacts.sh ${CSV_FILE} ${NAMESPACE} 'dep' \
```
ewolinetz: can you explain what this line is doing? why is there a `dep` call?

jcantrill: possibly bad naming... it short hand for generate the kind 'deployment'

test/e2e/clusterlogging_test.go
jcantrill: No.  This is what we talked about on slack.  This begins check where we modify the image back to the original which is re-testing the fact that the operator can react to image changes

ewolinetz: missed the check for it earlier...

PR #143, opened 2019-03-29
Bug 1694097 - Check if nodeselectors have changed

ewolinetz:
Updating components to check for nodeselector change

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1694097

This must go after https://github.com/openshift/elasticsearch-operator/pull/114 and requires a `dep` update

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -199,7 +199,7 @@
   branch = "master"
   name = "github.com/openshift/elasticsearch-operator"
   packages = ["pkg/apis/elasticsearch/v1"]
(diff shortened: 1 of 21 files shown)


richm:
do you need to rebase/vendor in the latest elasticsearch-operator?

ewolinetz:
@richm yes, I updated the EO API slightly so that we can specify the nodeSelector as a common setting similar to the resourceRequirements and Image. These tests will fail until I can update the vendored in code (https://github.com/openshift/elasticsearch-operator/pull/114)

PR #144, opened 2019-04-01
Bug 1694522 - Do not compare UUIDs when we dont provide them

ewolinetz:
Since CLO does not specify a UUID for the nodes for EO, we need to ignore that the GenUUID field has changed on the CR when doing a comparison to see if changes are required to be made.

Also required to update vendor to get GenUUID field in api.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1694522


pkg/k8shandler/logstore.go
```diff
@@ -120,7 +123,6 @@ func (cluster *ClusterLogging) newElasticsearchCR(elasticsearchName string) *ela
 		Roles:        []elasticsearch.ElasticsearchNodeRole{"client", "data", "master"},
 		NodeCount:    cluster.Spec.LogStore.NodeCount,
 		NodeSelector: cluster.Spec.LogStore.NodeSelector,
-		Resources:    *resources,
```
jcantrill: Did we lose this?  It looks like we update to include a default cpuRequest.  Is `esNode` the common defaults?

ewolinetz: we didn't, but I'd rather us specify it just at the common node level than at both the individual node and the common node.

josefkarasek: +1 Bigly

pkg/k8shandler/logstore.go
```diff
@@ -229,9 +231,98 @@ func isElasticsearchCRDifferent(current *elasticsearch.Elasticsearch, desired *e
...
+
+	// if we didn't find a role match, then that means changes were made
+	if !foundRoleMatch {
+		different = true
+	}
+
+	return desired, different
+}
+
+func areNodeRolesSame(lhs, rhs elasticsearch.ElasticsearchNode) bool {
```
josefkarasek: For the record, this is style which I prefer (like better)

```
    func getRole(roles []ElasticsearchNodeRole, roleType ElasticsearchNodeRole) bool {
        for _, role := range roles {
            if role == roleType {
                return true
            }
        }
        return false
    }
    // ...
    // not equal matters here
    if getRole(lhs.Roles, elasticsearch.ElasticsearchRoleClient) != getRole(rhs.Roles, elasticsearch.ElasticsearchRoleClient) {
        return false
    }
    // ...
    return true
```

wking:
/hold

I dunno why the restest bot is picking on this PR, but it is, despite the fact that this won't go green until at least the conflicts are resolved.  I'm dropping a `/hold` to call the bot off until that rebase has happened, feel free to `/hold cancel` afterwards.

richm:
Usually the needs-rebase label is added automatically - not sure why it wasn't added

ewolinetz:
still checking vendored changes /hold

PR #149, opened 2019-04-03
Bug 1695494 - Adding resource checks to all components

ewolinetz:
Adding checks for the different logging components to check if their resources have changed, warranting a rollout.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1695494

### DIFF
--- a/pkg/k8shandler/collection.go
+++ b/pkg/k8shandler/collection.go
@@ -250,14 +250,19 @@ func isDaemonsetDifferent(current *apps.DaemonSet, desired *apps.DaemonSet) (*ap
 	different := false
 
 	if !utils.AreSelectorsSame(current.Spec.Template.Spec.NodeSelector, desired.Spec.Template.Spec.NodeSelector) {
-		logrus.Infof("Daemonset nodeSelector change found, updating '%s'", current.Name)
+		logrus.Infof("Collector nodeSelector change found, updating '%s'", current.Name)
 		current.Spec.Template.Spec.NodeSelector = desired.Spec.Template.Spec.NodeSelector
 		different = true
 	}
 
 	if isDaemonsetImageDifference(current, desired) {
-		logrus.Infof("Daemonset image change found, updating %q", current.Name)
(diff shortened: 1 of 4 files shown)


pkg/k8shandler/collection.go
```diff
@@ -323,6 +329,25 @@ func updateCurrentDaemonsetImages(current *apps.DaemonSet, desired *apps.DaemonS
 	return current
 }
 
+func areResourcesDifferent(current *apps.Daemonset, desired *apps.Daemonset) (bool, *apps.Daemonset) {
```
richm: you could have a single `areResourcesDifferent` function - use introspection to determine if object is a Daemonset or a Deployment or a CronJob or whatever

pkg/k8shandler/curation.go
```diff
@@ -315,5 +315,11 @@ func isCuratorDifferent(current *batch.CronJob, desired *batch.CronJob) (*batch.
 		different = true
 	}
 
+	if changed, updated := utils.CompareResources(current.Spec.JobTemplate.Spec.Template.Spec.Containers[0], desired.Spec.JobTemplate.Spec.Template.Spec.Containers[0]); changed {
```
richm: use `areResourcesDifferent`

richm:
ok - if it passes, squash it and i'll lgtm

ewolinetz:
@richm should pass now, confirmed it works on my local env. lets make sure ci tests come back clean first

pkg/k8shandler/collection.go
```diff
@@ -261,6 +261,11 @@ func isDaemonsetDifferent(current *apps.DaemonSet, desired *apps.DaemonSet) (*ap
 		different = true
 	}
 
+	if utils.AreResourcesDifferent(current, desired) {
+		logrus.Infof("Daemonset resource(s) change found, updating %q", current.Name)
```
richm: Use `Collector` instead of `Daemonset` to be consistent with the other log messages below

ewolinetz: updated

PR #112, opened 2019-03-05
Bug fix 1683359 - fluentd metrics not found

josefkarasek:

PR #150, opened 2019-04-05
LOG-258 - Fluentd alerting rules

nhosoi:
Resurrecting the code for fluentd alerting rules from original pr/139


pkg/k8shandler/fluentd.go
```diff
@@ -22,6 +22,7 @@ const (
 	metricsPortName   = "metrics"
 	prometheusCAFile  = "/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt"
 	metricsVolumeName = "fluentd-metrics"
+	alertsFile        = "/etc/prometheus/fluentd_prometheus_alerts.yaml"
```
ewolinetz: why can't we point this to `/usr/bin/files/fluentd_prometheus_alerts.yaml` instead? it would keep our `.docker` files more simple...

nhosoi: I thought that is what we prefer according to the comment by @jcantrill...
https://github.com/openshift/cluster-logging-operator/pull/112#pullrequestreview-217739790

jcantrill: Per my conversation from stand up, we could leave this here but seems like the wrong place. We should follow conventions and as far as I know, `/usr/bin/files` is not a standard.

richm: `/usr/share/prometheus/fluentd_prometheus_alerts.yaml`

PR #162, opened 2019-04-28
Add e2e case: switch log collector type in clusterlogging CR

QiaolingTang:
@jcantrill PTAL, thanks.
cc @anpingli


test/e2e/clusterlogging_test.go
```diff
@@ -51,7 +52,7 @@ func TestClusterLogging(t *testing.T) {
 
 		for _, collector := range []string{"fluentd", "rsyslog"} {
 			t.Run(collector, func(t *testing.T) {
-				t.Parallel()
+				//				t.Parallel()
```
jcantrill: @ewolinetz can you speak to this? It parallelizes the tests?

test/e2e/clusterlogging_test.go
```diff
@@ -236,3 +242,49 @@ func clusterLoggingUpgradeTest(t *testing.T, f *framework.Framework, ctx *framew
...
+			Namespace: namespace,
+		},
+	}
+
+	exampleName := types.NamespacedName{Name: exampleClusterLogging.GetName(), Namespace: namespace}
+	if err = f.Client.Get(goctx.TODO(), exampleName, exampleClusterLogging); err != nil {
+		return fmt.Errorf("failed to get exampleClusterLogging: %v", err)
+	}
+
+	if exampleClusterLogging.Spec.Collection.Logs.Type == "fluentd" {
```
jcantrill: I think there are defined constants for these types

test/e2e/clusterlogging_test.go
ewolinetz: i believe so, though it did seem that they ran in series during a `make test-e2e`

test/e2e/clusterlogging_test.go
ewolinetz: https://github.com/openshift/cluster-logging-operator/blob/master/pkg/apis/logging/v1/types.go#L220-L221

QiaolingTang: Thanks, I'll change it to use const vars.

test/e2e/clusterlogging_test.go
QiaolingTang: I observed that when use `t.Parallel()`, the value of `collector` is always equal to `rsyslog` in `clusterLoggingInitialDeploymentTest` and `clusterLoggingUpgradeTest`.  Found a gist about this: https://gist.github.com/posener/92a55c4cd441fc5e5e85f27bca008721

test/e2e/clusterlogging_test.go
```diff
@@ -53,6 +54,7 @@ func TestClusterLogging(t *testing.T) {
 	t.Run("collectors", func(t *testing.T) {
 
 		for _, collector := range []string{"fluentd", "rsyslog"} {
+			collector := collector
```
richm: ?

ewolinetz:
i believe this has actually already been addresed by https://github.com/openshift/cluster-logging-operator/commit/93e721b6fd0fa470ccdd4fb45fc04fefb9b08848#diff-c3388c285fdfb9fce38392732d2c2f2b

jcantrill:
@QiaolingTang please rebase because of https://github.com/openshift/cluster-logging-operator/pull/227

PR #163, opened 2019-04-29
LOG-413. Correct manifest for rh-certified operator pipeline

jcantrill:
This PR fixes up the manifest for the rh certified pipeline to resolve https://jira.coreos.com/browse/LOG-413


manifests/4.1/image-references
```diff
@@ -0,0 +1,24 @@
...
+      kind: DockerImage
+      name: quay.io/openshift/origin-logging-curator5:latest
+  - name: logging-fluentd
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-logging-fluentd:latest
+  - name: rsyslog
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-logging-rsyslog:latest
```
nhosoi: No need to be `docker.io/viaq/rsyslog:latest` any more?

If we search, these 2 hit in the CLO code.  I'd think the first one is (or both of them could be) obsolete???
```
Makefile:	RSYSLOG_IMAGE=quay.io/viaq/rsyslog:latest \
controller-manifests/cluster-logging.v4.1.0.clusterserviceversion.yaml:                    value: "docker.io/viaq/rsyslog:latest"
```

jcantrill: @richm I'm removing this from the imageref file since rsyslog will not be in 4.1  We will have to add it in for 4.2

Dockerfile
```diff
@@ -4,6 +4,8 @@ COPY . .
 RUN make
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+WORKDIR /go/src/github.com/openshift/cluster-logging-operator
```
ewolinetz: we have two `WORKDIR` entries in this Dockerfile

Dockerfile
```diff
@@ -12,10 +14,11 @@ RUN INSTALL_PKGS=" \
     yum clean all && \
     mkdir /tmp/_working_dir && \
     chmod og+w /tmp/_working_dir
-COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/_output/bin/cluster-logging-operator /usr/bin/
-COPY scripts/* /usr/bin/scripts/
-COPY files/ /usr/bin/files/
-ADD controller-manifests /manifests
+COPY --from=builder _output/bin/cluster-logging-operator /usr/bin/
+COPY --from=builder scripts/* /usr/bin/scripts/
+COPY --from=builder files/ /usr/bin/files/
```
ewolinetz: this is going to break some of the components like rsyslog...
https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/rsyslog.go#L54

nhosoi: Also, the COPY target has been moved to /usr/share/logging/ in master-post-release...
https://github.com/openshift/cluster-logging-operator/blob/master-post-release/Dockerfile#L18

jcantrill: why would it break them?  the builder image has all the source already.  This just copies it from their instead

Dockerfile
jcantrill: I wasnt certain if WORKDIR applies to seperately

PR #167, opened 2019-05-03
add es image to manifest

jcantrill:


manifests/4.1/image-references
```diff
@@ -6,6 +6,10 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/openshift/origin-cluster-logging-operator:latest
+  - name: logging-elasticsearch5
+    from:
+      kind: DockerImage
+      name: quay.io/openshift/origin-logging-elasticsearch5:latest
```
bparees: how is this operator consuming the value?

bparees: ah i think i found it.

PR #168, opened 2019-05-03
Bug 1706134: Allowing scale up past 3 nodes for ES Cr

ewolinetz:
Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1706134

### DIFF
--- a/pkg/k8shandler/logstore.go
+++ b/pkg/k8shandler/logstore.go
@@ -119,19 +119,40 @@ func (cluster *ClusterLogging) newElasticsearchCR(elasticsearchName string) *ela
 		}
 	}
 
-	esNode := elasticsearch.ElasticsearchNode{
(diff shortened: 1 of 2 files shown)


ewolinetz:
/hold
i want to test the upgrade scenario

ewolinetz:
From a CLO perspective this works great, the issue comes when we go from `>= 3` nodes to `> 3` nodes because we do not remove the old `cdm` nodes in the EO code.
@jcantrill @richm @josefkarasek thoughts? 

If I manually remove the old deployments things look good, not sure about storage for migrations though...

ewolinetz:
another option is to update this PR such that instead of creating separate `client data` `client master` nodes when `> 3` we instead create only 3 `client data master` nodes and the excess are `client data` nodes.

PR #173, opened 2019-05-07
Ff 41 to master

jcantrill:
FF 4.1 to sync master


stevekuznetsov:
This should have blocked you from merging and I'm investigating as to why but please please lets not merge to these branches. They are frozen.

https://github.com/openshift/cluster-logging-operator/issues/155

jcantrill:
> This should have blocked you from merging and I'm investigating as to why but please please lets not merge to these branches. They are frozen.
> 
> #155

These FF PRs were created at behest of ART who believed they were building off the 4.1 branches instead of master.  There is communication break down someplace as to the branches which are live and those which are not.

PR #174, opened 2019-05-07
Bug 1707465: Adding disk-pressure toleration for fluentd

ewolinetz:
To address fluentd not being scheduled to nodes where it may have left buffer files causing disk-pressure, allow fluentd to be scheduled on nodes where that taint has been applied so it can attempt to clear its buffer files.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1707465

### DIFF
--- a/pkg/k8shandler/fluentd.go
+++ b/pkg/k8shandler/fluentd.go
@@ -256,6 +256,11 @@ func newFluentdPodSpec(logging *logging.ClusterLogging, elasticsearchAppName str
 			Operator: v1.TolerationOpExists,
 			Effect:   v1.TaintEffectNoSchedule,
 		},
+		v1.Toleration{
+			Key:      "node.kubernetes.io/disk-pressure",
+			Operator: v1.TolerationOpExists,
+			Effect:   v1.TaintEffectNoSchedule,
+		},
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/k8shandler/fluentd.go
```diff
@@ -256,6 +256,11 @@ func newFluentdPodSpec(logging *logging.ClusterLogging, elasticsearchAppName str
 			Operator: v1.TolerationOpExists,
 			Effect:   v1.TaintEffectNoSchedule,
 		},
+		v1.Toleration{
+			Key:			"node.kubernetes.io/disk-pressure",
```
richm: formatting
I'm not sure how the toleration works - by default fluentd would not be scheduled to nodes which have disk pressure, and the toleration would allow fluentd to be scheduled to those nodes?  That is, what does a user do to get fluentd scheduled to nodes where it needs to clear the buffer backlog?
We also need this for rsyslog

ewolinetz: Yes, this allows it to tolerate that taint so that it can be scheduled to those nodes too.

ewolinetz: Added rsyslog as well

PR #175, opened 2019-05-10
Bug 1709454: fixing missing RBAC for metadata-reader cluster role

ewolinetz:
making sure fluentd tests run and fixing so that fluentd is correctly installed

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1709454

### DIFF
--- a/manifests/4.1/cluster-logging.v4.1.0.clusterserviceversion.yaml
+++ b/manifests/4.1/cluster-logging.v4.1.0.clusterserviceversion.yaml
@@ -200,6 +200,15 @@ spec:
           - clusterrolebindings
           verbs:
           - "*"
+        - apiGroups:
+          - ""
+          resources:
+          - pods
+          - namespaces
+          verbs:
+          - get
+          - list
+          - watch
       deployments:
       - name: cluster-logging-operator
         spec:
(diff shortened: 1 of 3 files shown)


ewolinetz:
@jcantrill we were missing permissions in our csv... not sure impact on this for getting it out for 4.1.0

manifests/4.1/cluster-logging.v4.1.0.clusterserviceversion.yaml
```diff
@@ -176,6 +176,7 @@ spec:
           - monitoring.coreos.com
           resources:
           - servicemonitors
+          - prometheusrules
```
nhosoi: Fluentd PrometheusRules were merged to 4.2 (master-post-release branch).  Is it ok to put it in the master branch?

ewolinetz: good catch, i just assumed it needed to go in without seeing if it failed without it. i'll verify this works for master even without it

jcantrill:
@ewolinetz per our conversation, please open a bz for this to go into a 4.1.z for a 0 day.  I think you will have to open this against the 4.1 branch when it opens cc @bparees

PR #180, opened 2019-05-20
Bug 1710404: No longer set CPU Limit for default ES resource requirements

josefkarasek:
Don't set cpu limit on elasticsearch. Only request at least 1 CPU and 16Gi of RAM

https://bugzilla.redhat.com/show_bug.cgi?id=1710404


ewolinetz:
unit test is failing, please fix otherwise LGTM

PR #182, opened 2019-05-20
use make deploy with oal logging CI

richm:
Allows the use of `make deploy` with oal logging CI to
deploy logging.  Set `IMAGE_OVERRIDE` and `EO_IMAGE_OVERRIDE`
to specify the operator images to use.

This also gets rid of the need to create a `/usr/share/logging`
on the local machine - instead, for test purposes, the env var
`LOGGING_SHARE_DIR` sets the location to the files to load.
If this is not set, the default `/usr/share/logging` is used
(e.g. inside the container).

Depends on https://github.com/openshift/elasticsearch-operator/pull/146


richm:
please review

PR #185, opened 2019-05-22
Bug 1712721 - Fixing checking if nodes are different when scaling from 3 to 4

ewolinetz:
Also fixing ordering of nodes defined to prevent prior node recreation

Needs to be cherry picked to `release-4.1` as well for https://bugzilla.redhat.com/show_bug.cgi?id=1712955

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1712721


ewolinetz:
we need to hold off on merging the 4.1 cherrypick until after 4.1 is unlocked

PR #190, opened 2019-05-28
Force all CL components to be allocated to linux nodes

lukas-vlcek:
See [LOG-411](https://jira.coreos.com/browse/LOG-411)

### DIFF
--- a/pkg/k8shandler/curation.go
+++ b/pkg/k8shandler/curation.go
@@ -186,7 +186,7 @@ func newCuratorCronJob(cluster *logging.ClusterLogging, curatorName string, elas
 		{Name: "config", ReadOnly: true, MountPath: "/etc/curator/settings"},
 	}
 
-	curatorPodSpec := NewPodSpec(
+	curatorPodSpec := NewPodSpecForLinuxNode(
 		"curator",
 		[]v1.Container{curatorContainer},
 		[]v1.Volume{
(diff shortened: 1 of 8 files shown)


lukas-vlcek:
My first PR to CLO. I still need to figure out if tests are passing... bear with me.

jcantrill:
> Are we only going to support our components running on Linux for the time being?
> Is there a reason to create non-linux node selectors?

Right now we only support running on linux

lukas-vlcek:
@jcantrill If we are sure we are going to support only linux nodes going forward then we do not need to introduce a new function and just bake the linux selector into original `func NewPodSpec` instead (and properly document this aspect of it).
Is this what you would prefer more?

ewolinetz:
>If we are sure we are going to support only linux nodes going forward then we do not need to introduce a new function and just bake the linux selector into original func NewPodSpec instead (and properly document this aspect of it).
Is this what you would prefer more?

I would prefer that we just add the selector into the original function, that was why I was asking if we were going to support non-linux installations.

jcantrill:
> @jcantrill If we are sure we are going to support only linux nodes going forward then we do not need to introduce a new function and just bake the linux selector into original `func NewPodSpec` instead (and properly document this aspect of it).
> Is this what you would prefer more?

I concur with @ewolinetz to reusing the existing function.  The only component I anticipate living on a non-linux node is the collector which we need to make several other adjustments to support anyway.

lukas-vlcek:
Closing in favour of https://github.com/openshift/cluster-logging-operator/pull/191

lukas-vlcek:
@jcantrill @ewolinetz I did requested changes and created new PR to drive it from my fork instead of branch in upstream.

PR #191, opened 2019-05-29
Force all CL components to be allocated to linux nodes

lukas-vlcek:
See [LOG-411](https://jira.coreos.com/browse/LOG-411)

### DIFF
--- a/pkg/k8shandler/curation_test.go
+++ b/pkg/k8shandler/curation_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
@@ -29,8 +30,9 @@ func TestNewCuratorCronJobWhenFieldsAreUndefined(t *testing.T) {
 	if resources.Requests[v1.ResourceCPU] != defaultFluentdCpuRequest {
 		t.Errorf("Exp. the default CPU request to be %v", defaultCuratorCpuRequest)
 	}
-	if podSpec.NodeSelector != nil {
-		t.Errorf("Exp. the nodeSelector to be %T but was %T", map[string]string{}, podSpec.NodeSelector)
+	// check node selecor
+	if podSpec.NodeSelector == nil {
+		t.Errorf("Exp. the nodeSelector to contains the linux allocation selector but was %T", podSpec.NodeSelector)
 	}
 }
 
@@ -114,7 +116,8 @@ func TestNewCuratorCronJobWhenScheduleDefined(t *testing.T) {
 }
 func TestNewCuratorCronJobWhenNodeSelectorDefined(t *testing.T) {
 	expSelector := map[string]string{
-		"foo": "bar",
+		"foo":             "bar",
+		utils.OsNodeLabel: utils.LinuxValue,
 	}
 	cluster := &logging.ClusterLogging{
 		Spec: logging.ClusterLoggingSpec{
--- a/pkg/k8shandler/pod_test.go
+++ b/pkg/k8shandler/pod_test.go
@@ -1,0 +1,88 @@
+package k8shandler
+
+import (
+	"testing"
+
+	"github.com/openshift/cluster-logging-operator/pkg/utils"
+	v1 "k8s.io/api/core/v1"
+)
+
+func TestNodeAllocationLabelsForPod(t *testing.T) {
+
+	// Create pod with nil selectors, we expect a new selectors map will be created
+	// and it will contain only linux allocation selector.
+	podSpec := NewPodSpec(
+		"Foo",
+		[]v1.Container{},
+		[]v1.Volume{},
+		nil,
+	)
+
+	if podSpec.NodeSelector == nil {
+		t.Errorf("Exp. the nodeSelector to contains the linux allocation selector but was %T", podSpec.NodeSelector)
+	}
+	if len(podSpec.NodeSelector) != 1 {
+		t.Errorf("Exp. single nodeSelector but %d were found", len(podSpec.NodeSelector))
+	}
+	if podSpec.NodeSelector[utils.OsNodeLabel] != utils.LinuxValue {
+		t.Errorf("Exp. the nodeSelector to contains %s: %s pair", utils.OsNodeLabel, utils.LinuxValue)
+	}
+
+	// Create pod with some "foo" selector, we expect a new linux box selector will be added
+	// while existing selectors will be left intact.
+	podSpec = NewPodSpec(
+		"Foo",
+		[]v1.Container{},
+		[]v1.Volume{},
+		map[string]string{"foo": "bar"},
+	)
+
+	if podSpec.NodeSelector == nil {
+		t.Errorf("Exp. the nodeSelector to contains the linux allocation selector but was %T", podSpec.NodeSelector)
+	}
+	if len(podSpec.NodeSelector) != 2 {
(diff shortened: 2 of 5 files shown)


pkg/utils/utils.go
```diff
@@ -52,6 +56,32 @@ func AreSelectorsSame(lhs, rhs map[string]string) bool {
 	return true
 }
 
+// ForceLinuxNodeSelector takes given selector map and returns a selector map with linux node selector added into it.
+// If there is already a node type selector and is different from "linux" then it is overridden to "linux" type.
+// See https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#interlude-built-in-node-labels
+func ForceLinuxNodeSelector(selectors map[string]string) map[string]string {
```
ewolinetz: instead of `ForceLinuxNodeSelector` can we call it something like `AddLinuxNodeSelector` or `SetLinuxNodeSelector`?

pkg/utils/utils.go
```diff
@@ -52,6 +56,32 @@ func AreSelectorsSame(lhs, rhs map[string]string) bool {
...
+	if os, ok := selectors[OS_NODE_LABEL]; ok {
+		if os == LINUX_NODE_LABEL_VALUE {
+			return selectors
+		}
+		// Can this not be "linux"?
+		logrus.Warnf("Unexpected node selector value: %s=%s", OS_NODE_LABEL, os)
+		return selectors
+	}
+	newSelectors := make(map[string]string)
+	if selectors == nil {
```
ewolinetz: this is redundant with the logic above. did you mean to `selectors = map[string]string{}` after instead?

pkg/utils/utils.go
```diff
@@ -52,6 +56,32 @@ func AreSelectorsSame(lhs, rhs map[string]string) bool {
...
+		newSelectors = map[string]string{}
+	}
+
+	for k, v := range selectors {
```
ewolinetz: instead of doing this, could we not just `selectors[OS_NODE_LABEL] = LINUX_NODE_LABEL_VALUE` and return `selectors`?

pkg/utils/utils.go
ewolinetz: I also think `EnsureLinuxNodeSelector` could make sense here

pkg/utils/utils.go
lukas-vlcek: You are right the logic there was quite chaotic. Fixed.

pkg/utils/utils.go
lukas-vlcek: Fixed

lukas-vlcek:
@ewolinetz @jcantrill @josefkarasek PTAL.
Notice that I created a brand new test file and consolidated existing tests in it (these are tests related only to node selectors). There is also one interesting scenario when client provides node selector that is different from what we would like to see (i.e. not the `"linux"`). Right now it is keep and a warning log is written. There is also test for it. Let me know if you think we need to treat such situation differently.

pkg/k8shandler/curation_test.go
```diff
@@ -29,9 +30,6 @@ func TestNewCuratorCronJobWhenFieldsAreUndefined(t *testing.T) {
 	if resources.Requests[v1.ResourceCPU] != defaultFluentdCpuRequest {
 		t.Errorf("Exp. the default CPU request to be %v", defaultCuratorCpuRequest)
 	}
-	if podSpec.NodeSelector != nil {
```
ewolinetz: instead of removing this, we should expect that it would not be nil... can you make this report an error if `podSpec.NodeSelector == nil` ?

pkg/k8shandler/visualization_test.go
```diff
@@ -57,9 +58,6 @@ func TestNewKibanaPodSpecWhenFieldsAreUndefined(t *testing.T) {
 	if resources.Requests[v1.ResourceCPU] != defaultKibanaCpuRequest {
 		t.Errorf("Exp. the default CPU request to be %v", defaultKibanaCpuRequest)
 	}
-	if podSpec.NodeSelector != nil {
```
ewolinetz: same ask here as with curation_test.go

pkg/k8shandler/pod_test.go
```diff
@@ -0,0 +1,88 @@
...
+		t.Errorf("Exp. the nodeSelector to contains the linux allocation selector but was %T", podSpec.NodeSelector)
+	}
+	if len(podSpec.NodeSelector) != 1 {
+		t.Errorf("Exp. single nodeSelector but %d were found", len(podSpec.NodeSelector))
+	}
+	if podSpec.NodeSelector[utils.OsNodeLabel] != utils.LinuxValue {
+		t.Errorf("Exp. the nodeSelector to contains %s: %s pair", utils.OsNodeLabel, utils.LinuxValue)
+	}
+
+	// Create pod with some "non-linux" selector, we expect will stay unchanged.
```
ewolinetz: why do we expect this to be unchanged? shouldn't `EnsureLinuxNodeSelector` always set the OS node label?

pkg/utils/utils.go
```diff
@@ -52,6 +56,25 @@ func AreSelectorsSame(lhs, rhs map[string]string) bool {
...
+	if selectors == nil {
+		return map[string]string{OsNodeLabel: LinuxValue}
+	}
+	if os, ok := selectors[OsNodeLabel]; ok {
+		if os == LinuxValue {
+			return selectors
+		}
+		// Can this not be "linux"?
+		logrus.Warnf("Unexpected node selector value: %s=%s", OsNodeLabel, os)
+	} else {
```
ewolinetz: i would expect that we always make sure the os node label is `LinuxValue` based on the description of the card and function name... is this not the case?

pkg/k8shandler/curation_test.go
lukas-vlcek: Does it count that we have such test here? https://github.com/openshift/cluster-logging-operator/pull/191/files#diff-c61d3b9279a00c2a95db3179686043d7R21
Or you would like me to add such test for curation and visualization as well?

ewolinetz: I suppose it is redundant, I figured it would just be good to ensure that when the curator (and kibana) pods are created they still have the node selector (to prevent regression within those individual functions)

pkg/utils/utils.go
lukas-vlcek: This is probably the case when user explicitly set the node selector (which by default is not set). I can make it override the value anyway, just wanted to make sure this is what we really want to do (i.e. ignore and override users settings).

ewolinetz: @jcantrill do you know the answer to this?

lukas-vlcek: Fixed.

pkg/k8shandler/pod_test.go
lukas-vlcek: Fixed.

lukas-vlcek:
It now overrides the node selector if it is not `linux` and it dumps warn log.

pkg/utils/utils.go
```diff
@@ -52,6 +56,24 @@ func AreSelectorsSame(lhs, rhs map[string]string) bool {
...
+func EnsureLinuxNodeSelector(selectors map[string]string) map[string]string {
+	if selectors == nil {
+		return map[string]string{OsNodeLabel: LinuxValue}
+	}
+	if os, ok := selectors[OsNodeLabel]; ok {
+		if os == LinuxValue {
+			return selectors
+		}
+		// Selector is provided but is not "linux"
+		logrus.Warnf("Overridding node selector value: %s=%s to %s", OsNodeLabel, os, LinuxValue)
```
ewolinetz: s/Overridding/Overriding/

ewolinetz:
one nit otherwise LGTM

pkg/utils/utils.go
lukas-vlcek: Fixed

PR #194, opened 2019-05-31
LOG-372 - rsyslog - moving undefined objects to a specified property 

nhosoi:


nhosoi:
@richm, [1] is a sample debug output dumped at [2].  The test fields "undefined1, undefined2, etc." are moved under "undefined".  But the original fields are still in the top level json.  If we can have their property such as `$!undefined1`, we could do `unset $!undefined1`.  But what we have in the config is a local variable $.ii, by which `unset` does not work to remove the original objects...  Do you think there's a way to remove the original somehow?

[1] - `$!: { "level": "info",  "hostname": "localhost.localdomain", "@timestamp": "2019-05-31T00:38:01.449420+00:00", "undefined1": "undefined1", "undefined11": "1111", "undefined12": "True", "empty1": "", "undefined2": { "undefined2": "undefined2", "undefined23": "False" }, "undefined3": { "emptyvalue": "" }, "undefined4": "undefined4", "undefined5": "undefined5", "undefined": "{\"_STREAM_ID\":\"9f76e2d623774883a7c398912117357f\",\"undefined1\":\"undefined1\",\"undefined11\":\"1111\",\"undefined12\":\"True\",\"empty1\":\"\",\"undefined2\":\"{ \"undefined2\": \"undefined2\", \"undefined23\": \"False\" }\",\"undefined3\":\"{ \"emptyvalue\": \"\" }\",\"undefined4\":\"undefined4\",\"undefined5\":\"undefined5\",\"_SYSTEMD_INVOCATION_ID\":\"1e8e74590f3b4f9e8f6f85d20b30d8ed\"}", "message": "I0531 00:38:01.449231    1034 helpers.go:846] eviction manager: observations: signal=nodefs.inodesFree, available: 62285567, capacity: 62389184, time: 2019-05-31 00:38:01.443151228 +0000 UTC m=+82213.908760902",  ...  }`

[2] - https://github.com/nhosoi/cluster-logging-operator/blob/undefined/files/rsyslog/65-viaq-formatting.conf#L326

richm:
One thing to investigate is to change the code that parses json to add the fields to $. instead of $!.  Then, instead of generating a lookup table for the fields, generate config code to add only those desired fields from $. to $!.  For example if DEFAULT_KEEP_FIELDS=foo,bar,bazgenerate code like thisset $!foo = $.foo;set $!bar = $.bar;set $!baz = $.baz;All of the fields in $. will be discarded automatically. Sent from my Verizon, Samsung Galaxy smartphone
-------- Original message --------From: Noriko Hosoi <notifications@github.com> Date: 5/30/19  7:16 PM  (GMT-07:00) To: openshift/cluster-logging-operator <cluster-logging-operator@noreply.github.com> Cc: Richard Megginson <rmeggins@redhat.com>, Mention <mention@noreply.github.com> Subject: Re: [openshift/cluster-logging-operator]  [WIP] - LOG-372 - rsyslog - moving undefined objects to a specified property  (#194) @richm, [1] is a sample debug output dumped at [2].  The test fields "undefined1, undefined2, etc." are moved under "undefined".  But the original fields are still in the top level json.  If we can have their property such as $!undefined1, we could do unset $!undefined1.  But what we have in the config is a local variable $.ii, by which unset does not work to remove the original objects...  Do you think there's a way to remove the original somehow?
[1] - $!: { "level": "info",  "hostname": "localhost.localdomain", "@timestamp": "2019-05-31T00:38:01.449420+00:00", "undefined1": "undefined1", "undefined11": "1111", "undefined12": "True", "empty1": "", "undefined2": { "undefined2": "undefined2", "undefined23": "False" }, "undefined3": { "emptyvalue": "" }, "undefined4": "undefined4", "undefined5": "undefined5", "undefined": "{\"_STREAM_ID\":\"9f76e2d623774883a7c398912117357f\",\"undefined1\":\"undefined1\",\"undefined11\":\"1111\",\"undefined12\":\"True\",\"empty1\":\"\",\"undefined2\":\"{ \"undefined2\": \"undefined2\", \"undefined23\": \"False\" }\",\"undefined3\":\"{ \"emptyvalue\": \"\" }\",\"undefined4\":\"undefined4\",\"undefined5\":\"undefined5\",\"_SYSTEMD_INVOCATION_ID\":\"1e8e74590f3b4f9e8f6f85d20b30d8ed\"}", "message": "I0531 00:38:01.449231    1034 helpers.go:846] eviction manager: observations: signal=nodefs.inodesFree, available: 62285567, capacity: 62389184, time: 2019-05-31 00:38:01.443151228 +0000 UTC m=+82213.908760902",  ...  }
[2] - https://github.com/nhosoi/cluster-logging-operator/blob/undefined/files/rsyslog/65-viaq-formatting.conf#L326

—You are receiving this because you were mentioned.Reply to this email directly, view it on GitHub, or mute the thread.
[
{
"@context": "http://schema.org",
"@type": "EmailMessage",
"potentialAction": {
"@type": "ViewAction",
"target": "https://github.com/openshift/cluster-logging-operator/pull/194?email_source=notifications\u0026email_token=AAAFUGQNU5B7GGM2ZMHKPWTPYB36HA5CNFSM4HRPMWVKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGODWT5YSQ#issuecomment-497540170",
"url": "https://github.com/openshift/cluster-logging-operator/pull/194?email_source=notifications\u0026email_token=AAAFUGQNU5B7GGM2ZMHKPWTPYB36HA5CNFSM4HRPMWVKYY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGODWT5YSQ#issuecomment-497540170",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
"@type": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]

files/rsyslog/60-mmk8s.conf
```diff
@@ -17,31 +17,31 @@ if ((strlen($!CONTAINER_NAME) > 0) and (strlen($!CONTAINER_ID_FULL) > 0)) or
...
                 if strlen($!stream) == 0 then {
-                    set $!stream = $.ii!stream;
+                    set $.all!stream = $.ii!stream;
                 }
                 if strlen($!log) == 0 then {
-                    set $!log = $.ii!log;
+                    set $.all!log = $.ii!log;
                 } else {
-                    reset $!log = $!log & $.ii!log;
+                    reset $.all!log = $.log & $.ii!log;
```
richm: `reset $.all!log = $.all!log & $.ii!log;`
If it is a problem that `$.all!log` does not exist before calling reset on it, then maybe something like this:
`} else if strlen($.all!log) == 0 then {
  set $.all!log = $.ii!log;
} else {
  reset $.all!log =  $.all!log & $.ii!log;
}
```

File: files/rsyslog/60-mmk8s.conf [hunk 2]
```diff
@@ -17,31 +17,31 @@ if ((strlen($!CONTAINER_NAME) > 0) and (strlen($!CONTAINER_ID_FULL) > 0)) or
...
```
richm: see above

files/rsyslog/rsyslog.sh
```diff
@@ -228,6 +232,83 @@ fi
...
+for field in $( echo $EXTRA_KEEP_FIELDS | sed -e "s/,/ /g" )
+do
+    cat >> $keep_fields_conf << EOF
+if strlen(\$.all!$field) > 0 then {
+    set \$!$field = \$.all!$field;
+}
+EOF
+    echo "    {\"index\": \"$field\", \"value\": \"true\"}," >> $keep_fields_lookup
+done
+echo '# action(type="omfile" template="debugTemplate" file="/var/tmp/rsyslog_debug.log")' >> $keep_fields_conf
```
richm: There is a built-in debug template `RSYSLOG_DebugFormat` - it is quite verbose https://www.rsyslog.com/doc/v8-stable/configuration/templates.html#reserved-template-names

files/rsyslog/60-mmk8s.conf
nhosoi: Thank you, @richm.  Yes, it is a bug...  Fixed it.

files/rsyslog/60-mmk8s.conf
nhosoi: Fixed.

nhosoi:
Hi @richm, it turned out the json format built by this code [0] is broken. :(  If I pass the result string to mmnormalize, it fails.  I.e., the string is put into "unparsed-data" value ([2] is a sample result string).
[0] https://github.com/nhosoi/cluster-logging-operator/blob/undefined/files/rsyslog/rsyslog.sh#L256-L269

This template with the json format option [1] looked promising, but it seems it's designed to support "fixed" properties.  For instance, outname does not take a variable such as $.ii!key.  So, we cannot pass $.ii!key for the property name and $.ii!value for the value...  Also, it's not easy to call it in the foreach loop since it automatically adds an escape character in each loop.  Do you have any idea how to generate a valid json from $.ii!key and $.ii!value in the foreach loop?

[1] https://www.rsyslog.com/doc/v8-stable/configuration/templates.html

[2] "unparsed-data": "{\"level\":\"unknown\",\"pipeline_metadata\":\"{ \"collector\": { \"name\": \"rsyslog\", \"inputname\": \"rsyslogd\", \"received_at\": \"2019-06-03T21:04:58.621746+00:00\", \"ipaddr4\": \"127.0.0.1\", \"ipaddr6\": \"::1\", \"version\": \"8.37.0-9.el7 \" } }\",\"undefined1\":\"undefined1\",\"undefined11\":\"1111\",\"undefined12\":\"True\",\"empty1\":\"\",\"undefined2\":\"{ \"undefined2\": \"undefined2\", \"undefined23\": \"False\" }\",\"undefined3\":\"{ \"emptyvalue\": \"\" }\",\"undefined4\":\"undefined4\",\"undefined5\":\"undefined5\"}"

files/rsyslog/rsyslog.sh
```diff
@@ -228,6 +232,83 @@ fi
...
+    }
+    set \$!undefined = exec_template('makeJsonTemplate');
+}
+EOF
+
+for field in $( echo $DEFAULT_KEEP_FIELDS | sed -e "s/,/ /g" )
+do
+    cat >> $keep_fields_conf << EOF
+if strlen(\$.all!$field) > 0 then {
+    set \$!$field = \$.all!$field;
```
richm: Then after this line
`unset \$.all!$field;`
to "move" `$field` from `$.all` to `$!`

files/rsyslog/rsyslog.sh
```diff
@@ -228,6 +232,83 @@ fi
...
+}
+EOF
+    echo "    {\"index\": \"$field\", \"value\": \"true\"}," >> $keep_fields_lookup
+done
+
+for field in $( echo $EXTRA_KEEP_FIELDS | sed -e "s/,/ /g" )
+do
+    cat >> $keep_fields_conf << EOF
+if strlen(\$.all!$field) > 0 then {
+    set \$!$field = \$.all!$field;
```
richm: Then after this line
`unset \$.all!$field;`
to "move" `$field` from `$.all` to `$!`

files/rsyslog/rsyslog.sh
```diff
@@ -228,6 +232,83 @@ fi
...
+}
```
richm: After removing the `$field` from `$.all`, `$.all` will contain _only_ the "undefined" fields - the fields _not_ specified by `DEFAULT_KEEP_FIELDS` or `EXTRA_KEEP_FIELDS`.  Then, "move" `$.all` to `$!$undefined`:
`set \$!$undefined = \$.all;`
I don't know if this will work - https://www.rsyslog.com/doc/v8-stable/rainerscript/variable_property_types.html does not explicitly say what will happen if you assign an object from `$.` to what you want to be a new object in `$!`.

Also, I think we are reaching the limits of what we can do with generated config code + rscript.  If we need to do something much more complex than this, we will probably have to look into using https://www.rsyslog.com/doc/v8-stable/configuration/modules/mmexternal.html.  This is similar to what the new rsyslog_exporter is doing: https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/10-viaq-modules.conf#L34-L38 - it is using `omprog` to execute an external program to serve the prometheus metrics.  The external program is a go program: https://github.com/openshift/origin-aggregated-logging/tree/master/rsyslog/go/src/github.com/soundcloud/rsyslog_exporter

We may have to do something similar for complex record manipulation - write a go program like the `rsyslog_exporter` to parse/process/format a new JSON record.  I think go would be a good balance between ease of coding and performance.  If the performance is bad we could always write our own C rsyslog plugin but I'd rather avoid that.

nhosoi: Thanks, @richm!  It looks much better (and simpler)!  Let me go this way a little more.

nhosoi:
@richm, currently the following environment variables [1] are supported with this limitation [2] on KEEP_EMPTY_FIELDS.  

Two parameters undefined_dot_replace_char and undefined_max_num_fields are left.  Interms of undefined_dot_replace_char, so far I have not found how to rename the property name dynamically.  I think this is needed to solve the type mismatch in Elasticsearch...  We have UNDEFINED_TO_STRING to convert the undefined json to json string.  Do you think it's acceptable to workaround the problem?  And I think there is no benefit to have undefined_max_num_fields for rsyslog.

[1]
```
CDM_USE_UNDEFINED=false
DEFAULT_KEEP_FIELDS=CEE,time,@timestamp,aushape,ci_job,collectd,docker,fedora-ci,file,foreman,geoip,hostname,ipaddr4,ipaddr6,kubernetes,level,message,namespace_name,namespace_uuid,offset,openstack,ovirt,pid,pipeline_metadata,rsyslog,service,systemd,tags,testcase,tlog,viaq_msg_id
EXTRA_KEEP_FIELDS=""
CDM_UNDEFINED_NAME="undefined"
KEEP_EMPTY_FIELDS=""
UNDEFINED_TO_STRING=false
```
[2] 
`KEEP_EMPTY_FIELDS currently only works if CDM_USE_UNDEFINED is enabled and the target fields are defined.  That is, 1) undefined empty fields are cleaned up even if the field name is listed in KEEP_EMPTY_FIELDS.  2) If CDM_USE_UNDEFINED is disabled, empty fields are always cleaned up.`

Notes: the configuration file 10-keep_fields.conf is generated and placed in ${RSYSLOG_WORKDIRECTORY} since the ConfigMap directory cannot be read-write mode.

richm:
> @richm, currently the following environment variables [1] are supported with this limitation [2] on KEEP_EMPTY_FIELDS.
> 
> Two parameters undefined_dot_replace_char and undefined_max_num_fields are left. Interms of undefined_dot_replace_char, so far I have not found how to rename the property name dynamically. I think this is needed to solve the type mismatch in Elasticsearch...

Yes, it is needed to avoid the situation like https://github.com/ViaQ/fluent-plugin-viaq_data_model#undefined_dot_replace_char

Hmm - we may have reached the limits of what we can do in rscript.  The really ugly way to do it would be to convert the record to a string, do a string/regexp search&replace, convert back to a record . . . 

> We have UNDEFINED_TO_STRING to convert the undefined json to json string. Do you think it's acceptable to workaround the problem?

Not sure what you mean.

> And I think there is no benefit to have undefined_max_num_fields for rsyslog.

Not sure what you mean.  The benefit is for Elasticsearch, which has performance problems if there are too many fields per record.  Can you use a `foreach` to check for this?
```
if cnum(`echo $UNDEFINED_MAX_NUM_FIELDS`) > -1 then {
  set $.nfields = 0;
  foreach ($.ii in $.undefinedfields) do {
    reset $.nfields = $.nfields + 1;
  }
  if $.nfields > cnum(`echo $UNDEFINED_MAX_NUM_FIELDS`) then {
    $!undefinedfieldname = exec_template("convert_undefinedfields_to_JSON_string");
    not sure what goes here but we then need to remove all of those fields from the record . ..
```
hmm - maybe not - again, we may have reached the limits of what we can do in rscript.  I wouldn't spend too much more time on this.  If you can't figure out how to support dot_replace_char and max_num_fields in rscript, then let's proceed with what you do have, and open another PR to implement support for dot_replace_char and max_num_fields (probably by writing a mmexternal module in golang for origin-aggregated-logging/rsyslog).

> 
> [1]
> 
> ```
> CDM_USE_UNDEFINED=false
> DEFAULT_KEEP_FIELDS=CEE,time,@timestamp,aushape,ci_job,collectd,docker,fedora-ci,file,foreman,geoip,hostname,ipaddr4,ipaddr6,kubernetes,level,message,namespace_name,namespace_uuid,offset,openstack,ovirt,pid,pipeline_metadata,rsyslog,service,systemd,tags,testcase,tlog,viaq_msg_id
> EXTRA_KEEP_FIELDS=""
> CDM_UNDEFINED_NAME="undefined"
> KEEP_EMPTY_FIELDS=""
> UNDEFINED_TO_STRING=false
> ```
> 
> [2]
> `KEEP_EMPTY_FIELDS currently only works if CDM_USE_UNDEFINED is enabled and the target fields are defined. That is, 1) undefined empty fields are cleaned up even if the field name is listed in KEEP_EMPTY_FIELDS. 2) If CDM_USE_UNDEFINED is disabled, empty fields are always cleaned up.`
> 
> Notes: the configuration file 10-keep_fields.conf is generated and placed in ${RSYSLOG_WORKDIRECTORY} since the ConfigMap directory cannot be read-write mode.

nhosoi:
> hmm - maybe not - again, we may have reached the limits of what we can do in rscript. I wouldn't spend too much more time on this. If you can't figure out how to support dot_replace_char and max_num_fields in rscript, then let's proceed with what you do have, and open another PR to implement support for dot_replace_char and max_num_fields (probably by writing a mmexternal module in golang for origin-aggregated-logging/rsyslog).

All right.  I will open another PR for them.  Do you think there are any golang mmexternal examples I can learn from?

How about this one?  Is it ok to remove [WIP]?  (I thought I put /hold in the beginning, but it was removed somehow... :p)  My concern is if we write mmexternal module, it might be simpler for the module to let it do all...  But for now, I'm going to work on dot_replace_char and max_num_fields first.

nhosoi:
> Do you think there are any golang mmexternal examples I can learn from?

Golang example may not exist yet, but there are some in the other languages, so it'd help me...

https://github.com/rsyslog/rsyslog/tree/master/plugins/external

richm:
> > hmm - maybe not - again, we may have reached the limits of what we can do in rscript. I wouldn't spend too much more time on this. If you can't figure out how to support dot_replace_char and max_num_fields in rscript, then let's proceed with what you do have, and open another PR to implement support for dot_replace_char and max_num_fields (probably by writing a mmexternal module in golang for origin-aggregated-logging/rsyslog).
> 
> All right. I will open another PR for them. Do you think there are any golang mmexternal examples I can learn from?

Not golang and mmexternal specifically.  It is basically a program (any program) that reads from stdin, parses JSON (or raw, not sure what that is), does some modification to the data, then writes a JSON string (or raw?) to stdout.
We use `omprog` (which is similar to mmexternal, except that it doesn't write the output to stdout) for the rsyslog prometheus exporter - https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/10-viaq-modules.conf#L34 - rsyslog writes the impstats in JSON format to the external program `/usr/local/bin/rsyslog_exporter`, which then parses the JSON read from stdin - the program source is here - https://github.com/openshift/origin-aggregated-logging/tree/master/rsyslog/go/src/github.com/soundcloud/rsyslog_exporter - and is built into our rsyslog image here - https://github.com/openshift/origin-aggregated-logging/blob/master/rsyslog/Dockerfile#L24

The only reason I suggested golang is 1) it is very fast 2) we are already using it in the rsyslog image.

> 
> How about this one? Is it ok to remove [WIP]? (I thought I put /hold in the beginning, but it was removed somehow... :p) My concern is if we write mmexternal module, it might be simpler for the module to let it do all... But for now, I'm going to work on dot_replace_char and max_num_fields first.

Yes, please remove the `[WIP]`, and add back the `/hold` if you are not ready for this PR to be merged.
And yes - it may be that using mmexternal with a golang program is fast enough to do all of this, and probably much simpler.

files/rsyslog/60-mmk8s.conf
```diff
@@ -17,31 +17,31 @@ if ((strlen($!CONTAINER_NAME) > 0) and (strlen($!CONTAINER_ID_FULL) > 0)) or
             action(type="mmnormalize" rulebase="/etc/rsyslog.d/multiline-json.rulebase")
             foreach ($.ii in $!multilinejson) do {
                 if strlen($!@timestamp) == 0 then {
```
richm: should this be `$.all!@timestamp` now? Same with `$.all!stream` and `$.all!log` below.  and in the crio section

nhosoi: Ah, you are right.  I misunderstood the logic... :p  Fixing them.  Thanks!

nhosoi:
Closing this PR in favor of the implementation using mmexternal plugin,

PR #198, opened 2019-06-10
LOG-411 Ensure linux node selector in ElasticsearchCR

lukas-vlcek:
Ensure linux node selector is added to storage/Elasticsearch CR.
The node selector should propagate to elasticsearch-operator transparently,
meaning all pods started by elasticsearch-operator will be allocated
to linux nodes.

### DIFF
--- a/pkg/k8shandler/logstore.go
+++ b/pkg/k8shandler/logstore.go
@@ -162,9 +162,10 @@ func newElasticsearchCR(cluster *logging.ClusterLogging, elasticsearchName strin
 		},
 		Spec: elasticsearch.ElasticsearchSpec{
 			Spec: elasticsearch.ElasticsearchNodeSpec{
-				Image:        utils.GetComponentImage("elasticsearch"),
-				Resources:    *resources,
-				NodeSelector: cluster.Spec.LogStore.NodeSelector,
+				Image:     utils.GetComponentImage("elasticsearch"),
+				Resources: *resources,
+				// We always want to ensure the "linux" node selector is used, see LOG-411
+				NodeSelector: utils.EnsureLinuxNodeSelector(cluster.Spec.LogStore.NodeSelector),
 			},
 			Nodes:            esNodes,
 			ManagementState:  elasticsearch.ManagementStateManaged,
--- a/pkg/k8shandler/logstore_test.go
+++ b/pkg/k8shandler/logstore_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	elasticsearch "github.com/openshift/elasticsearch-operator/pkg/apis/logging/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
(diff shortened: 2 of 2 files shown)


ewolinetz:
My concern here is that since the Elasticsearch Operator is responsible ultimately for the node selector used for Elasticsearch that we have that gap. Someone could set the ClusterLogging CR to be `Unmanaged` and then update the node selector on the Elasticsearch CR to not be specific to `linux`. This is different than how the other components we install are handled.

lukas-vlcek:
@ewolinetz yes, it is different, agree. On one hand I like the fact that I do not have to copy&paste parts of the code from CLO to EO and that everything is controlled from a single operator. On the other hand, as you point out, when CLO is in `Unmanaged` state then it is possible to remove the `linux` node selector for Elasticsearch pods. BTW, does it mean that EO could re-allocate existing pods to different pods?

lukas-vlcek:
For the record, we decided that I will replicate or "copy&paste" the relevant code directly from CLO to EO for now and then, later, when the code is merged into EO we can reuse it in CLO because CLO vendors EO (though this plan may miss some low-level details).
Basically, in the end we would like to move the code from CLO to EO.

richm:
Did something recently merge related to this?  oal ci is broken e.g. https://github.com/openshift/origin-aggregated-logging/pull/1658#issuecomment-500596831
```
FAILURE after 61.934s: test/cluster/rollout.sh:68: executing 'oc rollout status deployment/kibana' expecting any result and text 'successfully rolled out'; re-trying every 0.2s until completion or 60.000s: the command timed out
```
One of the problems is that there are no nodes which match the node selector: https://storage.googleapis.com/origin-ci-test/pr-logs/pull/openshift_origin-aggregated-logging/1658/pull-ci-openshift-origin-aggregated-logging-master-full-integ-aws/908/artifacts/full-integ-aws/scripts/check-EFK-running/artifacts/kibana-7fd98ccfb-4lkj6.yaml
```
    resources:
      limits:
        memory: 736Mi
      requests:
        cpu: 100m
        memory: 736Mi
...
  nodeSelector:
    kubernetes.io/os: linux
    logging-ci-test: "true"
...
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: 2019-06-10T20:27:22Z
    message: '0/6 nodes are available: 2 Insufficient cpu, 5 node(s) didn''t match
      node selector.'
    reason: Unschedulable
    status: "False"
    type: PodScheduled
  phase: Pending
```
@lukas-vlcek @ewolinetz @josefkarasek @jcantrill what do we need to do to fix this?

ewolinetz:
@richm can we add in a `oc label nodes --all kubernetes.io/os=linux` command for out CI? that would sort it out

richm:
> @richm can we add in a `oc label nodes --all kubernetes.io/os=linux` command for out CI? that would sort it out

Sure, I'll do that.  But how will this be handled in production?  Will the node tuning operator apply this label?

richm:
https://github.com/openshift/origin-aggregated-logging/pull/1661

ewolinetz:
> Sure, I'll do that. But how will this be handled in production? Will the node tuning operator apply this label?

I would guess so. @bparees do we know for sure that this will be the case?

richm:
> I would guess so. @bparees do we know for sure that this will be the case?

afaict, this is already being done in CI - so the cause of the oal logging ci problem is unknown - still investigating

bparees:
> @richm can we add in a oc label nodes --all kubernetes.io/os=linux command for out CI? that would sort it out

i would have expected the nodes to already have this label.

lukas-vlcek:
Replaced by https://github.com/openshift/elasticsearch-operator/pull/164

PR #202, opened 2019-06-20
LOG-372 - rsyslog - support improved undefined field handling

nhosoi:
Exacuting undefined_field in the mmexternal plugin.

Depending on https://github.com/openshift/origin-aggregated-logging/pull/1671

### DIFF
--- a/files/rsyslog/65-viaq-formatting.conf
+++ b/files/rsyslog/65-viaq-formatting.conf
@@ -96,6 +96,18 @@ if strlen($!_MACHINE_ID) > 0 then {
         set $!systemd!t!AUDIT_SESSION = $!_AUDIT_SESSION;
     }
     unset $!_AUDIT_SESSION;
+    if strlen($!_STREAM_ID) > 0 then {
+        set $!systemd!t!_STREAM_ID = $!_STREAM_ID;
+    }
+    unset $!_STREAM_ID;
+    if strlen($!_SYSTEMD_INVOCATION_ID) > 0 then {
+        set $!systemd!t!_SYSTEMD_INVOCATION_ID = $!_SYSTEMD_INVOCATION_ID;
+    }
+    unset $!_SYSTEMD_INVOCATION_ID;
+    if strlen($!_LINE_BREAK) > 0 then {
+        set $!systemd!t!LINE_BREAK = $!_LINE_BREAK;
+    }
+    unset $!_LINE_BREAK;
     if strlen($!_BOOT_ID) > 0 then {
         set $!systemd!t!BOOT_ID = $!_BOOT_ID;
     }
--- a/files/rsyslog/67-mmexternal.conf
+++ b/files/rsyslog/67-mmexternal.conf
@@ -1,0 +1,15 @@
+if (strlen(`echo $USE_MMEXTERNAL`) > 0) and (`echo $USE_MMEXTERNAL` == "true") then {
+    module(load="mmexternal")
+    action(type="mmexternal" binary="/usr/local/bin/undefined_field" interface.input="fulljson")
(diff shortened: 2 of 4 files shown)


files/rsyslog/65-viaq-formatting.conf
```diff
@@ -96,6 +96,14 @@ if strlen($!_MACHINE_ID) > 0 then {
...
     }
     unset $!_AUDIT_SESSION;
+    if strlen($!_STREAM_ID) > 0 then {
+        set $!systemd!t!_STREAM_ID = $!_STREAM_ID;
+    }
+    unset $!_STREAM_ID;
+    if strlen($!_SYSTEMD_INVOCATION_ID) > 0 then {
+        set $!systemd!t!_SYSTEMD_INVOCATION_ID = $!_SYSTEMD_INVOCATION_ID;
+    }
+    unset $!_SYSTEMD_INVOCATION_ID;
```
richm: Please also add support for the new `systemd.t.LINE_BREAK` field - https://github.com/ViaQ/elasticsearch-templates/commit/aae0fcb56b282b2c8f2d49f7c1893dacd5a20d6f

files/rsyslog/65-viaq-formatting.conf
```diff
@@ -246,14 +254,12 @@ if strlen($!level) > 0 then {
     unset $.lclevel;
     unset $.normlevel;
 } else {
-    if $!stream == "stdout" then {
+    if $!message startswith "I" then {
```
richm: No, please don't add this yet.  I think we need a separate commit to fix https://bugzilla.redhat.com/show_bug.cgi?id=1726639 `fluentd misbehaving and marking an info log wrong level to 'err'` - we will need to add more parsing, and we will need to coordinate the fix with fluentd.

files/rsyslog/67-mmexternal.conf
```diff
@@ -0,0 +1,15 @@
+if (strlen(`echo $USE_MMEXTERNAL`) > 0) and (`echo $USE_MMEXTERNAL` == "true") then {
+    module(load="mmexternal")
+    action(type="mmexternal" binary="/usr/local/bin/undefined_field" interface.input="fulljson")
```
richm: so `/usr/local/bin/undefined_field` also does the skip empty field handling?

nhosoi: Yes, it does.  And undefined_field skip empty supports CDM_KEEP_EMPTY_FIELDS, while the else part mmnormalization with parse_json_skip_empty.rulebase does not.

The reason why I added USE_MMEXTERNAL and SKIP_EMPTY is for the performance measurement.  Just in case mmexternal shows degradation, we could provide Plan B to the people who don't care the undefined and CDM_KEEP_EMPTY_FIELDS.  (Or we could completely turn it off.)

files/rsyslog/65-viaq-formatting.conf
```diff
@@ -96,6 +96,18 @@ if strlen($!_MACHINE_ID) > 0 then {
...
     unset $!_AUDIT_SESSION;
+    if strlen($!_STREAM_ID) > 0 then {
+        set $!systemd!t!_STREAM_ID = $!_STREAM_ID;
+    }
+    unset $!_STREAM_ID;
+    if strlen($!_SYSTEMD_INVOCATION_ID) > 0 then {
+        set $!systemd!t!_SYSTEMD_INVOCATION_ID = $!_SYSTEMD_INVOCATION_ID;
+    }
+    unset $!_SYSTEMD_INVOCATION_ID;
+    if strlen($!_SYSTEMD_INVOCATION_ID) > 0 then {
```
richm: `if strlen($!_LINE_BREAK) > 0 then {`

PR #204, opened 2019-06-25
add art csv metadata

bparees:


manifests/4.2/cluster-logging.v4.2.0.clusterserviceversion.yaml
```diff
@@ -61,8 +61,6 @@ metadata:
 spec:
   displayName: Cluster Logging
 
-  replaces: clusterlogging.v4.1.0
```
bparees: @jwforres can you confirm we don't need this anymore?

manifests/art.yaml
```diff
@@ -0,0 +1,16 @@
+updates:
+  - file: "{MAJOR}.{MINOR}/cluster-logging.v{MAJOR}.{MINOR}.0.clusterserviceversion.yaml" # relative to this file
+    update_list:
+    # replace metadata.name value
+    - search: "clusterlogging.v{MAJOR}.{MINOR}.2"
```
ewolinetz: the reason we are replacing a ".2" is because we have a `4.1.2` csv? i don't see that version in master to be replaced...

bparees: i think you're right, this needs to be `clusterlogging.v{MAJOR}.{MINOR}.0` i think, to match https://github.com/openshift/cluster-logging-operator/pull/204/files#diff-5766fe294b4d426421892b30ee5365eaR257

will fix.

manifests/art.yaml
```diff
@@ -0,0 +1,16 @@
+updates:
+  - file: "{MAJOR}.{MINOR}/cluster-logging.v{MAJOR}.{MINOR}.0.clusterserviceversion.yaml" # relative to this file
+    update_list:
+    # replace metadata.name value
+    - search: "clusterlogging.v{MAJOR}.{MINOR}.0"
```
bparees: @ewolinetz fixed to match https://github.com/openshift/cluster-logging-operator/pull/204/files#diff-5766fe294b4d426421892b30ee5365eaR6

thanks

manifests/4.2/cluster-logging.v4.2.0.clusterserviceversion.yaml
bparees: @richm @ewolinetz unless you guys still need/want it for your manual/CI upgrade testing?

bparees:
@adammhaile @sosiouxme @jwforres @shawn-hurley ok i think this gets the master branch to where it should be in terms of all the "replaces", "skipranges", "version", etc, as well as adding a little doc to help us maintain it in the future.

I also updated the 4.1 CSV though my assumption is that the 4.1 CSV is ignored in the master branch.

ptal, if this is correct i'll ensure the ELO files are also correct as well as make sure all our 4.1 branches are right.

manifests/art.yaml
```diff
@@ -0,0 +1,16 @@
...
+      replace: "clusterlogging.{FULL_VER}"
+    # replace entire version line, otherwise would replace 4.2.0 anywhere
+    - search: "version: 4.2.0"
+      replace: "version: {FULL_VER}"
+    - search: 'olm.skipRange: ">=4.1.0 <4.2.0"'
+      replace: 'olm.skipRange: ">=4.1.0 <{FULL_VER}"'
+  - file: "cluster-logging.package.yaml"
+    update_list:
+    - search: "currentCSV: clusterlogging.v{MAJOR}.{MINOR}.2"
```
shawn-hurley: I think this needs to be `.0` instead of `.2`?

bparees: yup, thanks, updated.

manifests/4.2/cluster-logging.v4.2.0.clusterserviceversion.yaml
ewolinetz: i feel like `replaces` is just causing issues currently and it seems that we can just use `skipRange` instead so i'm good with this being removed

PR #205, opened 2019-06-26
Bug 1712721 - Ensure we preserve GenUUID for nodes when doing diffs

ewolinetz:
When scaling up from 3 to 4 nodes the CLO was not preserving UUIDs that were generated and populated in the ElasticsearchCR by the EO resulting in an error message that the previously known UUID was removed.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1712721

### DIFF
--- a/pkg/k8shandler/logstore_test.go
+++ b/pkg/k8shandler/logstore_test.go
@@ -7,6 +7,7 @@ import (
 	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	elasticsearch "github.com/openshift/elasticsearch-operator/pkg/apis/logging/v1"
+	esutils "github.com/openshift/elasticsearch-operator/test/utils"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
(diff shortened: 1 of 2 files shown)


pkg/k8shandler/logstore_test.go
```diff
@@ -347,3 +348,77 @@ func TestDifferenceFoundWhenNodeCountExceeds4(t *testing.T) {
...
+					NodeCount: 3,
+				},
+			},
+		},
+	}
+	elasticsearchCR := newElasticsearchCR(cluster, "test-app-name")
+	dataUUID := esutils.GenerateUUID()
+	elasticsearchCR.Spec.Nodes[0].GenUUID = &dataUUID
+
+	cluster = &logging.ClusterLogging{
```
jcantrill: I don't recall if this is saved via pointer or value but can you reduce the lines, reusing the `cluster` and simply bump the node count so you don't need to redefine everything?

pkg/k8shandler/logstore_test.go
```diff
@@ -347,3 +348,77 @@ func TestDifferenceFoundWhenNodeCountExceeds4(t *testing.T) {
...
+	if !different {
+		t.Errorf("Expected that difference would be found due to node count change")
+	}
+
+	if diffCR.Spec.Nodes[0].GenUUID == nil || *diffCR.Spec.Nodes[0].GenUUID != dataUUID {
+		t.Errorf("Expected that original GenUUID would be preserved as %v but was %v", dataUUID, diffCR.Spec.Nodes[0].GenUUID)
+	}
+}
+
+func TestGenUUIDPreservedWhenNodeCountChanges(t *testing.T) {
```
jcantrill: How is this different from the previous test aside from 1->3 and 3->4 ?  What about the code makes this significant?

ewolinetz: This was the nature of the bz, when we scale between 1 and 3 the UUID was preserved, when we scaled up past 3 to 4 the UUID would disappear. This was due to the fact we were creating a new `node` and not pulling the existing UUID from the CR after doing a comparison to see if the nodes changed.

pkg/k8shandler/logstore_test.go
ewolinetz: I probably could, this is the same as our other functions though... tech debt?

PR #206, opened 2019-06-27
create 4.2 channel and remove unnecessary 4.1 content from 4.2 branch

bparees:


bparees:
@jcantrill @ewolinetz @richm this is the new directory structure/channel definitions that @shawn-hurley and @jwforres and I came up with.  ptal.

our 4.1 channel will have to be named "preview" since that's how it went out, but going forward I expect us to to have a new channel for each openshift z-stream.

We do not need to maintain the content for other channels, in a given branch....only the content for the channel that branch represents, hence me deleting the 4.1 content here.

Dockerfile
```diff
@@ -4,7 +4,7 @@ COPY . .
 RUN make
 
 FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
-ARG CSV=4.1
+ARG CSV=4.2
```
bparees: @ewolinetz @jcantrill 
it's a good thing I deleted the 4.1 dir or this might have been overlooked.......

PR #207, opened 2019-06-27
Bug 1724833: Cleanup 4.1 manifest content

bparees:
aligning w/ similar changes in master: https://github.com/openshift/cluster-logging-operator/pull/206

patch manager: there is no master BZ associated w/ this change because this is cleaning up issues that were specific to the 4.1 resource files.

### DIFF
--- a/manifests/4.1/cluster-logging.v4.1.0.clusterserviceversion.yaml
+++ b/manifests/4.1/cluster-logging.v4.1.0.clusterserviceversion.yaml
@@ -3,7 +3,7 @@
 apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
-  name: clusterlogging.v4.1.2
+  name: clusterlogging.v4.1.0
   namespace: placeholder
   annotations:
     capabilities: Seamless Upgrades
@@ -15,7 +15,7 @@ metadata:
     containerImage: quay.io/openshift/origin-cluster-logging-operator:latest
     createdAt: 2018-08-01T08:00:00Z
     support: AOS Logging
-    olm.skipRange: ">=4.1.0 <4.1.2"
+    olm.skipRange: ">=4.1.0 <4.1.0"
     alm-examples: |-
         [
             {
@@ -60,6 +60,7 @@ metadata:
             }
         ]
 spec:
+  version: 4.1.0
   displayName: Cluster Logging
 
   description: |
@@ -248,7 +249,6 @@ spec:
                     value: "quay.io/openshift/origin-oauth-proxy:latest"
                   - name: RSYSLOG_IMAGE
                     value: "quay.io/openshift/origin-logging-rsyslog:latest"
-  version: 4.1.2
   customresourcedefinitions:
     owned:
     - name: clusterloggings.logging.openshift.io
--- a/manifests/art.yaml
+++ b/manifests/art.yaml
@@ -2,15 +2,15 @@ updates:
   - file: "{MAJOR}.{MINOR}/cluster-logging.v{MAJOR}.{MINOR}.0.clusterserviceversion.yaml" # relative to this file
     update_list:
     # replace metadata.name value
-    - search: "clusterlogging.v{MAJOR}.{MINOR}.2"
+    - search: "clusterlogging.v{MAJOR}.{MINOR}.0"
       replace: "clusterlogging.{FULL_VER}"
-    # replace entire version line, otherwise would replace 4.1.0 anywhere
-    - search: "version: 4.1.2"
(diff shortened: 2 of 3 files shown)


bparees:
Now that the ART pipeline is doing proper substitution there should be no need to leave the "4.1.2" values in place.  Hoping to avoid some future confusion by setting these to 4.1.0.

@jcantrill @ewolinetz ptal.  i guess i'll need to get a BZ created to get this cherry-pick approved.

@sosiouxme @adammhaile can you confirm this is safe to do in the 4.1 stream at this point?

manifests/4.1/cluster-logging.v4.1.0.clusterserviceversion.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
-  name: clusterlogging.v4.1.2
+  name: clusterlogging.v4.1.0
```
richm: Is it going to be a problem if there is already a CSV with `name: clusterlogging.v4.1.2` out there?  That is, will OLM ignore the CSV with `name: clusterlogging.v4.1.0` even though it is supposed to be "newer"?

bparees: @richm this value is going to be substituted by the ART pipeline, it will not be published as "name: clusterlogging.v4.1.0", it will be published as "name: clusterlogging.v4.1.4" (and then 4.1.5, and then 4.1.6, etc).

ewolinetz:
this is looks like its basically a cherrypick
/lgtm

bparees:
/hold
need ART to confirm these changes make sense for 4.1 at this point (going back to using 4.1.0 as the search text, not 4.1.2)  @sosiouxme @adammhaile

bparees:
testing something....
/label cherry-pick-approved

sosiouxme:
LGTM

bparees:
/hold cancel
@sosiouxme reviewed in slack (thanks!)

manifests/4.1/cluster-logging.v4.1.0.clusterserviceversion.yaml
```diff
@@ -15,7 +15,7 @@ metadata:
     containerImage: quay.io/openshift/origin-cluster-logging-operator:latest
     createdAt: 2018-08-01T08:00:00Z
     support: AOS Logging
-    olm.skipRange: ">=4.1.0 <4.1.2"
+    olm.skipRange: ">=4.1.0 <4.1.0"
```
jwforres: no version could ever match this range, i know this is just the placeholder, but is it going to cause the semver library that OLM is using to be upset, or is it not going to care. I just don't want to break the logging CI if OLM will wind up choking on this.

bparees: @ecordell can you weigh in?

ecordell:
```go
import (
	"github.com/blang/semver"
	"github.com/stretchr/testify/require"
)

func TestBlang(t *testing.T) {
	r, err := semver.ParseRange(">=4.1.0 <4.1.0")
	require.NoError(t, err)
	v, err := semver.ParseTolerant("4.1.0")
	require.NoError(t, err)
	require.False(t, r(v))
}
```

This test passes, so:

- the range will be parsed with no issues
- 4.1.0 will not match the range

bparees: so it sounds like this will work as is, however if someone wanted to use it to test upgrading from a true 4.1.0 operator, to a new operator they just built (using this CSV) they could not, they'd have to modify the CSV.

so i guess i'll just change this to "4.1.1" to make everyone's lives easier.

bparees: well, i take it back...i'd rather not do that unless it becomes necessary.  Since it's not necessary currently, i'm going to suggest we proceed w/ what we've got.  If this skipRange becomes problematic in the future for upgrade testing, we can revisit at that point.

jwforres:
Ok all concerns were addressed on this, patch manager is @crawford this week. I'll let him make final call to approve it for merge.

crawford:
@bparees I'm not familiar enough to understand what effect this is going to have (especially the strange semver range). Can you provide details in the commit message for me and future readers?

bparees:
@crawford done.

crawford:
/retitle Bug 1724833: Cleanup 4.1 manifest content

crawford:
No content changes.

/lgtm

PR #209, opened 2019-07-11
LOG-372 - rsyslog - support improved undefined field handling

nhosoi:
fix problems with mmexternal

Opening a PR based upon the fix by @richm.  Thanks!
> I think I've figured out why rsyslog is broken - https://github.com/richm/cluster-logging-operator/tree/rsyslog-cleanup


nhosoi:
> You also need to set `USE_MMEXTERNAL=false` by default until the origin-aggregated-logging change that adds `/usr/local/bin/undefined_field` is added.

It should be already added by OAL pr/#1671...
Merge pull request #1671 from nhosoi/undefined_external

You don't see it in your test env?

richm:
> You don't see it in your test env?

Now I do.

PR #214, opened 2019-07-18
rsyslog configuration

nhosoi:
- Prioritizing mmnormalize/parse_json_skip_empty over mmexternal
- Removing stream field.

### DIFF
--- a/files/rsyslog/65-viaq-formatting.conf
+++ b/files/rsyslog/65-viaq-formatting.conf
@@ -258,16 +258,9 @@ if strlen($!level) > 0 then {
     unset $.lclevel;
     unset $.normlevel;
 } else {
-    if $!stream == "stdout" then {
-        set $!level = "info";
-    } else {
-        if $!stream == "stderr" then {
-            set $!level = "err";
-        } else {
-            set $!level = "unknown";
-        }
-    }
+    set $!level = "unknown";
 }
+unset $!stream;
 
 # add eventrouter
 
--- a/files/rsyslog/rsyslog.sh
+++ b/files/rsyslog/rsyslog.sh
@@ -6,26 +6,115 @@ export LOGGING_FILE_PATH=${LOGGING_FILE_PATH:-"/var/log/rsyslog/rsyslog.log"}
 export RSYSLOG_WORKDIRECTORY=${RSYSLOG_WORKDIRECTORY:-/var/lib/rsyslog.pod}
 export MERGE_JSON_LOG=${MERGE_JSON_LOG:-false}
 export UNDEFINED_DEBUG=${UNDEFINED_DEBUG:-false}
-# if false, mmexternal undefined_field won't be called.
-export USE_MMEXTERNAL=${USE_MMEXTERNAL:-true}
-# if true, mmnormalize skipempty is executed.
-# effective when USE_MMEXTERNAL is false.
-export SKIP_EMPTY=${SKIP_EMPTY:-false}
 export RSYSLOG_FILE_READ_FROM_TAIL=${RSYSLOG_FILE_READ_FROM_TAIL:-off}
 export RSYSLOG_JOURNAL_READ_FROM_TAIL=${RSYSLOG_JOURNAL_READ_FROM_TAIL:-off}
 
-# Create a config file for undefined fields from env vars
+# Undefined field configuraion
+CDM_USE_UNDEFINED=${CDM_USE_UNDEFINED:-false}
+CDM_KEEP_EMPTY_FIELDS=${CDM_KEEP_EMPTY_FIELDS:-""}
+CDM_UNDEFINED_TO_STRING=${CDM_UNDEFINED_TO_STRING:-false}
+CDM_UNDEFINED_DOT_REPLACE_CHAR=${CDM_UNDEFINED_DOT_REPLACE_CHAR:-"UNUSED"}
+if [ "${CDM_USE_UNDEFINED,,}" = "false" -a "${CDM_KEEP_EMPTY_FIELDS}" = "" -a "${CDM_UNDEFINED_TO_STRING,,}" = "false" -a "${CDM_UNDEFINED_DOT_REPLACE_CHAR,,}" = "UNUSED" ] ; then
+    if [ "${SKIP_EMPTY:-}" = "" ] ; then
+        if [ "${USE_MMEXTERNAL:-}" = "" ] ; then
+            # default - mmnormalize skip-empty
+            export SKIP_EMPTY=true
+            export USE_MMEXTERNAL=false
+        elif [ "${USE_MMEXTERNAL,,}" = "true" ] ; then
+            export SKIP_EMPTY=false
+            export USE_MMEXTERNAL=true
+            echo "INFO: USE_MMEXTERNAL was set to true but none of the UNDEFINED fields were set. Consider using SKIP_EMPTY=true and unsetting USE_MMEXTERNAL instead."
+        else
+            export SKIP_EMPTY=true
+            export USE_MMEXTERNAL=false
+            echo "INFO: SKIP_EMPTY was not set but USE_MMEXTERNAL was set to false. Enabling SKIP_EMPTY. Consider unsetting both to avoid this message in the future."
+        fi
+    elif [ "${SKIP_EMPTY,,}" = "true" ] ; then
+        if [ "${USE_MMEXTERNAL:-}" = "" ] ; then
+            export SKIP_EMPTY=true
+            export USE_MMEXTERNAL=false
+        elif [ "${USE_MMEXTERNAL,,}" = "true" ] ; then
+            export SKIP_EMPTY=true
+            export USE_MMEXTERNAL=false
+            echo "INFO: Both USE_MMEXTERNAL and SKIP_EMPTY were set to true. Respecting SKIP_EMPTY. Consider unsetting USE_MMEXTERNAL to avoid this message in the future."
+        else
+            export SKIP_EMPTY=true
+            export USE_MMEXTERNAL=false
+        fi
+    else
+        if [ "${USE_MMEXTERNAL:-}" = "" ] ; then
+            export SKIP_EMPTY=false
+            export USE_MMEXTERNAL=true
+            echo "INFO: SKIP_EMPTY was set to false, but USE_MMEXTERNAL was not set and none of the UNDEFINED fields were configured. Enabling USE_MMEXTERNAL. Consider using SKIP_EMPTY=true instead."
+        elif [ "${USE_MMEXTERNAL,,}" = "true" ] ; then
+            export SKIP_EMPTY=false
+            export USE_MMEXTERNAL=true
+            echo "INFO: SKIP_EMPTY was set to false and USE_MMEXTERNAL was set to true, but none of the UNDEFINED fields were configured. Consider using SKIP_EMPTY=true and unsetting USE_MMEXTERNAL instead."
+        else
+            export SKIP_EMPTY=true
+            export USE_MMEXTERNAL=false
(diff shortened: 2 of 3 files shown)


files/rsyslog/65-viaq-formatting.conf
```diff
@@ -268,6 +268,7 @@ if strlen($!level) > 0 then {
         }
     }
 }
+unset $!stream
```
richm: need a semicolon `;` here.

files/rsyslog/rsyslog.sh
```diff
@@ -7,25 +7,35 @@ export RSYSLOG_WORKDIRECTORY=${RSYSLOG_WORKDIRECTORY:-/var/lib/rsyslog.pod}
...
+\"CDM_KEEP_EMPTY_FIELDS\" : \"${CDM_KEEP_EMPTY_FIELDS}\", \
+\"CDM_UNDEFINED_TO_STRING\" : ${CDM_UNDEFINED_TO_STRING}, \
+\"CDM_UNDEFINED_DOT_REPLACE_CHAR\" : \"${CDM_UNDEFINED_DOT_REPLACE_CHAR}\", \
 \"CDM_UNDEFINED_MAX_NUM_FIELDS\" : ${CDM_UNDEFINED_MAX_NUM_FIELDS:--1}, \
 \"MERGE_JSON_LOG\" : ${MERGE_JSON_LOG}, \
 \"UNDEFINED_DEBUG\" : ${UNDEFINED_DEBUG}}" > $undefined_conf
+else
+    export SKIP_EMPTY=false
+    export USE_MMEXTERNAL=true
+    rm -f $undefined_conf
```
richm: you need to write `$undefined_conf` here.

files/rsyslog/rsyslog.sh
```diff
@@ -7,25 +7,35 @@ export RSYSLOG_WORKDIRECTORY=${RSYSLOG_WORKDIRECTORY:-/var/lib/rsyslog.pod}
...
 \"CDM_UNDEFINED_NAME\" : \"${CDM_UNDEFINED_NAME:-"undefined"}\", \
-\"CDM_KEEP_EMPTY_FIELDS\" : \"${CDM_KEEP_EMPTY_FIELDS:-""}\", \
-\"CDM_UNDEFINED_TO_STRING\" : ${CDM_UNDEFINED_TO_STRING:-false}, \
-\"CDM_UNDEFINED_DOT_REPLACE_CHAR\" : \"${CDM_UNDEFINED_DOT_REPLACE_CHAR:-"UNUSED"}\", \
+\"CDM_KEEP_EMPTY_FIELDS\" : \"${CDM_KEEP_EMPTY_FIELDS}\", \
+\"CDM_UNDEFINED_TO_STRING\" : ${CDM_UNDEFINED_TO_STRING}, \
+\"CDM_UNDEFINED_DOT_REPLACE_CHAR\" : \"${CDM_UNDEFINED_DOT_REPLACE_CHAR}\", \
 \"CDM_UNDEFINED_MAX_NUM_FIELDS\" : ${CDM_UNDEFINED_MAX_NUM_FIELDS:--1}, \
 \"MERGE_JSON_LOG\" : ${MERGE_JSON_LOG}, \
 \"UNDEFINED_DEBUG\" : ${UNDEFINED_DEBUG}}" > $undefined_conf
```
richm: you need to `rm -f $undefined_conf` here

files/rsyslog/65-viaq-formatting.conf
richm: Also replace this
```
    if $!stream == "stdout" then {
        set $!level = "info";
    } else {
        if $!stream == "stderr" then {
            set $!level = "err";
        } else {
            set $!level = "unknown";
        }
    }
```
with this
```
    set $!level = "unknown";
```

nhosoi:
Thank you, Rich...

I'm shocked to learn all the codes were broken... (;_;)

files/rsyslog/rsyslog.sh
```diff
@@ -7,25 +7,35 @@ export RSYSLOG_WORKDIRECTORY=${RSYSLOG_WORKDIRECTORY:-/var/lib/rsyslog.pod}
...
 export RSYSLOG_JOURNAL_READ_FROM_TAIL=${RSYSLOG_JOURNAL_READ_FROM_TAIL:-off}
 
-# Create a config file for undefined fields from env vars
+CDM_USE_UNDEFINED=${CDM_USE_UNDEFINED:-false}
+CDM_KEEP_EMPTY_FIELDS=${CDM_KEEP_EMPTY_FIELDS:-""}
+CDM_UNDEFINED_TO_STRING=${CDM_UNDEFINED_TO_STRING:-false}
+CDM_UNDEFINED_DOT_REPLACE_CHAR=${CDM_UNDEFINED_DOT_REPLACE_CHAR:-"UNUSED"}
 undefined_conf="/var/lib/rsyslog.pod/undefined.json"
-echo "{\"CDM_USE_UNDEFINED\" : ${CDM_USE_UNDEFINED:-false}, \
+if [ "${CDM_USE_UNDEFINED}" = "false" -a "${CDM_KEEP_EMPTY_FIELDS}" = "" -a "${CDM_UNDEFINED_TO_STRING}" = "false" -a "${CDM_UNDEFINED_DOT_REPLACE_CHAR}" = "UNUSED" ] ; then
```
richm: Thinking about this some more - I think the intention here is to "do the right thing" if the user *has not* set either `SKIP_EMPTY` or `USE_MMEXTERNAL`.  That is, if I do `oc set env ds/rsyslog USE_MMEXTERNAL=true` that means I want to use mmexternal even if I did not set CDM_USE_UNDEFINED, CDM_UNDEFINED_TO_STRING, etc.  You could print a message like "INFO: USE_MMEXTERNAL was set but none of the UNDEFINED fields were set.  Consider using SKIP_EMPTY=true instead." to let the user know that they may have done something unintentional.
Likewise if the user sets `SKIP_EMPTY=true` but one of `CDM_USE_UNDEFINED`, etc. undefined parameters are set.  You could print a message like "INFO: SKIP_EMPTY was set but one or more of CDM_USE_UNDEFINED, etc. were set.  Consider unsetting SKIP_EMPTY and using USE_MMEXTERNAL=true instead."

files/rsyslog/rsyslog.sh
```diff
@@ -7,25 +7,35 @@ export RSYSLOG_WORKDIRECTORY=${RSYSLOG_WORKDIRECTORY:-/var/lib/rsyslog.pod}
...
+    export SKIP_EMPTY=true
+    export USE_MMEXTERNAL=false
+    rm -f $undefined_conf
+else
+    export SKIP_EMPTY=false
+    export USE_MMEXTERNAL=true
+    # Create a config file for undefined fields from env vars
```
richm:
```bash
if [ "$SKIP_EMPTY" = false -a "$USE_MMEXTERNAL" = true ] ; then
  create $undefined_conf
else
  rm -f $undefined_conf
fi
```

files/rsyslog/rsyslog.sh
nhosoi: Ok.  So, we don't set SKIP_EMPTY and USE_MMEXTERNAL automatically in rsyslog.sh, but we respect the users' settings.  If their configuration is not enough, like missing CDM_USE_UNDEFINED with USE_MMEXTERNAL, we give an appropriate warning.

Now, when nothing is set, the default behaviour should be SKIP_EMPTY=true && USE_MMEXTERNAL=false, right?

files/rsyslog/rsyslog.sh
nhosoi: Thanks!  I'm going to modify this PR as suggested...

richm: Yes - buildcop status is Severe - because we're all trying to merge 14,000,012 PRs by tomorrow :-(

files/rsyslog/rsyslog.sh
```diff
@@ -6,26 +6,115 @@ export LOGGING_FILE_PATH=${LOGGING_FILE_PATH:-"/var/log/rsyslog/rsyslog.log"}
...
+            export SKIP_EMPTY=true
+            export USE_MMEXTERNAL=false
+        elif [ "${USE_MMEXTERNAL,,}" = "true" ]; then
+            export SKIP_EMPTY=false
+            export USE_MMEXTERNAL=true
+            echo "INFO: USE_MMEXTERNAL was set but none of the UNDEFINED fields were set. Consider using SKIP_EMPTY=true instead."
+        else
+            export SKIP_EMPTY=true
+            export USE_MMEXTERNAL=false
+            echo "INFO: SKIP_EMPTY was set but USE_MMEXTERNAL was set to false. Enabling SKIP_EMPTY."
```
richm: `INFO: SKIP_EMPTY was not set but USE_MMEXTERNAL was set to false. Enabling SKIP_EMPTY.  Consider unsetting USE_MMEXTERNAL to avoid this message in the future.`

nhosoi: I guess I'd better append "to avoid this message in the future" to each INFO message?

files/rsyslog/rsyslog.sh
```diff
@@ -6,26 +6,115 @@ export LOGGING_FILE_PATH=${LOGGING_FILE_PATH:-"/var/log/rsyslog/rsyslog.log"}
...
+    elif [ "${SKIP_EMPTY,,}" = "true" ]; then
+        if [ "${USE_MMEXTERNAL:-}" = "" ]; then
+            # default - mmnormalize skip-empty
+            export SKIP_EMPTY=true
+            export USE_MMEXTERNAL=false
+            echo "INFO: SKIP_EMPTY was set to true and USE_MMEXTERNAL was not set but the UNDEFINED fields were configured. Consider using USE_MMEXTERNAL=true instead."
+        elif [ "${USE_MMEXTERNAL,,}" = "true" ]; then
+            export SKIP_EMPTY=false
+            export USE_MMEXTERNAL=true
+            echo "INFO: Both USE_MMEXTERNAL and SKIP_EMPTY were set to true. Respecting USE_MMEXTERNAL."
```
richm: I think if the user sets `SKIP_EMPTY=true` that should take precedence, even if the user has set both `USE_MMEXTERNAL=true` and the UNDEFINED field configuration.

files/rsyslog/rsyslog.sh
```diff
@@ -6,26 +6,115 @@ export LOGGING_FILE_PATH=${LOGGING_FILE_PATH:-"/var/log/rsyslog/rsyslog.log"}
...
+        if [ "${USE_MMEXTERNAL:-}" = "" ]; then
+            export SKIP_EMPTY=false
+            export USE_MMEXTERNAL=true
+        elif [ "${USE_MMEXTERNAL,,}" = "true" ]; then
+            export SKIP_EMPTY=false
+            export USE_MMEXTERNAL=true
+        else
+            export SKIP_EMPTY=false
+            export USE_MMEXTERNAL=true
+            echo "INFO: Both USE_MMEXTERNAL and SKIP_EMPTY were set to false. Enabling USE_MMMEXTERNAL."
```
richm: `INFO: Both USE_MMEXTERNAL and SKIP_EMPTY were set to false, but the UNDEFINED field configuration is enabled. Enabling USE_MMEXTERNAL.`

files/rsyslog/rsyslog.sh
nhosoi: Hmmm, but doesn't s/he expect to process logs with UNDEFINED configurations?

files/rsyslog/rsyslog.sh
richm: Yes, good idea, in case the user is wondering how to get rid of this message at startup . . .

files/rsyslog/rsyslog.sh
richm: But `SKIP_EMPTY=true` and the precedent is to honor the `SKIP_EMPTY=true` setting over everything else.

PR #215, opened 2019-07-18
rsyslog - fix impstats file format

richm:
the msg field is already in JSON format - just print it


nhosoi:
/lgtm

Yes, I also noticed it and was wondering in my testing... :)  Sooooo glad you fixed it!

PR #218, opened 2019-07-29

nhosoi:
There was a small window where the daemonset is up and its service account is
missing for the collectors.  Based on the rule "we should try to delete objects
in the reverse order that we created them", reversing the order of removing the
daemonset and the service account.

Applying the rule to curator and kibana, as well.


richm:
note that I don't think this will completely fix the Bug 1715370: Some orphaned logs sent to .operations.* index by mistake.  It may help, there are other problems.

richm:
> note that I don't think this will completely fix the Bug 1715370: Some orphaned logs sent to .operations.* index by mistake. It may help, there are other problems.

Never mind.  I was thinking of another bz.

PR #220, opened 2019-07-31
Bug 1722944: rsyslog log collector dropping logs

richm:
Increase the elasticsearch queue size - the goal is that the limit
should come from the available disk space, not the number of
records allowed in the queue.
Improve the stats queue so that it does not interfere with the
other outputs.
Label all actions - in conjunction with https://github.com/openshift/origin-aggregated-logging/pull/1713
will greatly improve stats legibility


richm:
ok - ready to review

files/rsyslog/55-elasticsearch.conf
```diff
@@ -62,9 +62,10 @@ ruleset(name="try_es") {
             queue.maxFileSize=`echo $ES_QUEUE_MAXFILESIZE`
             queue.checkpointInterval=`echo $ES_QUEUE_CHECKPOINTINTERVAL`
             queue.syncqueuefiles="on"
+            queue.size=`echo $ES_QUEUE_SIZE`
             action.resumeRetryCount="-1"
             )
-            # not yet supported - action.resumeIntervalMax=`echo $ES_RETRY_WAIT`
+            # not yet supported - RHEL 8.2 rebase - action.resumeIntervalMax=`echo $ES_RETRY_WAIT`
```
nhosoi: Out of curiosity... OAL vendored_src does not contain the latest rsyslog source codes?

richm: it contains the latest rsyslog source code from rhel8

nhosoi: Then, we could uncomment???

richm: rhel 8.2 has not been released yet . . .

PR #221, opened 2019-07-31
Bug 1734592: Empty fields are skipped when set SKIP_EMPTY=false and …

nhosoi:
…USE_MMEXTERNAL=false in rsyslog daemonset.

When both SKIP_EMPTY and USE_MMEXTERNAL are set to false, respect the configuration with the note:
  INFO: Both USE_MMEXTERNAL and SKIP_EMPTY were set to false and none of the environment variables
        to manage undefined and empty fields were configured. No undefined field as well as skip
        empty field handling will be executed.

### DIFF
--- a/files/rsyslog/rsyslog.sh
+++ b/files/rsyslog/rsyslog.sh
@@ -23,11 +23,11 @@ if [ "${CDM_USE_UNDEFINED,,}" = "false" -a "${CDM_KEEP_EMPTY_FIELDS}" = "" -a "$
         elif [ "${USE_MMEXTERNAL,,}" = "true" ] ; then
             export SKIP_EMPTY=false
             export USE_MMEXTERNAL=true
-            echo "INFO: USE_MMEXTERNAL was set to true but none of the UNDEFINED fields were set. Consider using SKIP_EMPTY=true and unsetting USE_MMEXTERNAL instead."
+            echo "INFO: USE_MMEXTERNAL was set to true but none of the environment variables to manage undefined and empty fields were set. Consider using SKIP_EMPTY=true and unsetting USE_MMEXTERNAL instead."
         else
             export SKIP_EMPTY=true
             export USE_MMEXTERNAL=false
-            echo "INFO: SKIP_EMPTY was not set but USE_MMEXTERNAL was set to false. Enabling SKIP_EMPTY. Consider unsetting both to avoid this message in the future."
+            echo "INFO: SKIP_EMPTY was not set but USE_MMEXTERNAL was set to false. Enabling SKIP_EMPTY. Consider unsetting both environment variables."
         fi
     elif [ "${SKIP_EMPTY,,}" = "true" ] ; then
         if [ "${USE_MMEXTERNAL:-}" = "" ] ; then
@@ -36,7 +36,7 @@ if [ "${CDM_USE_UNDEFINED,,}" = "false" -a "${CDM_KEEP_EMPTY_FIELDS}" = "" -a "$
         elif [ "${USE_MMEXTERNAL,,}" = "true" ] ; then
             export SKIP_EMPTY=true
             export USE_MMEXTERNAL=false
-            echo "INFO: Both USE_MMEXTERNAL and SKIP_EMPTY were set to true. Respecting SKIP_EMPTY. Consider unsetting USE_MMEXTERNAL to avoid this message in the future."
+            echo "INFO: Both USE_MMEXTERNAL and SKIP_EMPTY were set to true. Respecting SKIP_EMPTY. Consider unsetting USE_MMEXTERNAL."
         else
             export SKIP_EMPTY=true
             export USE_MMEXTERNAL=false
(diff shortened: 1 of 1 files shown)


nhosoi:
@mburke5678, I'm trying to improve the info message from the rsyslog container at the startup time.  Could you please take a look and give me your advice?  Thanks!

files/rsyslog/rsyslog.sh
```diff
@@ -45,15 +45,15 @@ if [ "${CDM_USE_UNDEFINED,,}" = "false" -a "${CDM_KEEP_EMPTY_FIELDS}" = "" -a "$
         if [ "${USE_MMEXTERNAL:-}" = "" ] ; then
             export SKIP_EMPTY=false
             export USE_MMEXTERNAL=true
-            echo "INFO: SKIP_EMPTY was set to false, but USE_MMEXTERNAL was not set and none of the UNDEFINED fields were configured. Enabling USE_MMEXTERNAL. Consider using SKIP_EMPTY=true instead."
+            echo "INFO: SKIP_EMPTY was set to false, but USE_MMEXTERNAL was not set and none of the environment variables to manage undefined and empty fields were configured. Enabling USE_MMEXTERNAL. Consider using SKIP_EMPTY=true instead."
```
mburke5678: _USE_MMEXTERNAL was not set _
This means that U`USE_MMEXTERAL` is `true` by default, right? 
_INFO: SKIP_EMPTY was set to false, USE_MMEXTERNAL is true by default but none of the environment variables to manage undefined and empty fields were configured. Consider configuring the CDM_* environment variables _ or consider using SKIP_EMPTY=true instead._

Something to tell the user what to do in this case?

files/rsyslog/rsyslog.sh
```diff
@@ -45,15 +45,15 @@ if [ "${CDM_USE_UNDEFINED,,}" = "false" -a "${CDM_KEEP_EMPTY_FIELDS}" = "" -a "$
...
         elif [ "${USE_MMEXTERNAL,,}" = "true" ] ; then
             export SKIP_EMPTY=false
             export USE_MMEXTERNAL=true
-            echo "INFO: SKIP_EMPTY was set to false and USE_MMEXTERNAL was set to true, but none of the UNDEFINED fields were configured. Consider using SKIP_EMPTY=true and unsetting USE_MMEXTERNAL instead."
+            echo "INFO: SKIP_EMPTY was set to false and USE_MMEXTERNAL was set to true, but none of the environment variables to manage undefined and empty fields were configured. Consider using SKIP_EMPTY=true and unsetting USE_MMEXTERNAL instead."
```
mburke5678: _Consider using SKIP_EMPTY=true and unsetting USE_MMEXTERNAL instead_
Should you change this to suggest setting the CDM_* environment variables? If the user set `USE_MMEXTERNAL` wouldn't that suggest the user wants to use MMEXTERNAL, but neglected to set the variables?

_Consider setting the CDM_* environment variables or using SKIP_EMPTY=true and unsetting USE_MMEXTERNAL._

files/rsyslog/rsyslog.sh
```diff
@@ -94,7 +94,7 @@ else
         else
             export SKIP_EMPTY=false
             export USE_MMEXTERNAL=true
-            echo "INFO: Both USE_MMEXTERNAL and SKIP_EMPTY were set to false, but the UNDEFINED fields were configured. Enabling USE_MMEXTERNAL. Consider using USE_MMEXTERNAL=true to avoid this message in the future."
+            echo "INFO: Both USE_MMEXTERNAL and SKIP_EMPTY were set to false, but the environment variables to manage undefined and empty fields were configured. Enabling USE_MMEXTERNAL. Consider using USE_MMEXTERNAL=true to avoid this message in the future."
```
mburke5678: Do we need _ to avoid this message in the future_? The other messages do not have this statement. I think they make sense without that statement.

mburke5678:
@nhosoi I left a couple of small suggestions. The messages look good to me.

files/rsyslog/rsyslog.sh
nhosoi: Under these settings, no undefined field handling is not executed, but just empty fields are skipped.  When this is what a user wants to do, setting SKIP_EMPTY=true and use mmnormalize would be preferable for the performance reason.  (that's the reason why SKIP_EMPTY was introduced.)  So, if only empty fields are to be skipped, we'd like to recommend SKIP_EMPTY=true.

files/rsyslog/rsyslog.sh
nhosoi: All right.  Let me remove that part _ to avoid this message in the future_.

nhosoi:
> @nhosoi I left a couple of small suggestions. The messages look good to me.

Thanks a lot, @mburke5678!!  I removed  _to avoid this message in the future_ and pushed the new patch.

richm:
lgtm - @mburke5678 ?

mburke5678:
LGTM

PR #223, opened 2019-08-01
Bug 1735511: Rsyslog couldn't parse json log which has prefix space before the json string when set MERGE_JSON_LOG=true

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1735511
Change the json parser to look for leading whitespace

Note: The multiline_json parser doesn't need this because its json
is computer generated by docker, not human generated as in the bz


richm:
ready - please review

PR #225, opened 2019-08-02
Bug 1735548:  pipeline_metadata.collector.ipaddr6 is incorrect when using rsyslog as log collector.

VladMasarik:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1735548 .
Part of that Bugzilla is also Fluentd fix for the same incorrect field. However, the fix requires opening a PR in [Origin Aggregate Logging](https://github.com/openshift/origin-aggregated-logging) repository.

This is the Fluentd PR https://github.com/openshift/origin-aggregated-logging/pull/1718

### DIFF
--- a/files/rsyslog/rsyslog.sh
+++ b/files/rsyslog/rsyslog.sh
@@ -174,6 +174,9 @@ else
 fi
 RSYSLOG_VERSION=`/usr/sbin/rsyslogd -v | awk -F'[ ,]+' '/^rsyslogd / {print $2}'`
 PIPELINE_VERSION="${RSYSLOG_VERSION} ${DATA_VERSION}"
+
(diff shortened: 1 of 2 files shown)


files/rsyslog/rsyslog.sh
```diff
@@ -174,6 +174,9 @@ else
 fi
 RSYSLOG_VERSION=`/usr/sbin/rsyslogd -v | awk -F'[ ,]+' '/^rsyslogd / {print $2}'`
 PIPELINE_VERSION="${RSYSLOG_VERSION} ${DATA_VERSION}"
+
+IPADDR6="" # So as to omit this field from logs. See this PR for more info.
```
VladMasarik: Not sure how to add label, but this is still WIP. I did not know to which PR to refer.

richm: I added the do-not-merge/work-in-progress label to the PR

PR #227, opened 2019-08-05
LOG-459. Add OLM like workflow to e2e

jcantrill:
Fixes: https://jira.coreos.com/browse/LOG-459


hack/deploy-image.sh
```diff
@@ -60,6 +60,7 @@ rc=0
 for ii in $( seq 1 5 ) ; do
     if push_image ${IMAGE_TAG} ${tag} ; then
         rc=0
+        oc -n openshift get imagestreams | grep cluster-logging-operator
```
richm: Does this mean that `push_image` can somehow succeed, but the image isn't actually pushed?

hack/test-e2e.sh
```diff
@@ -32,21 +38,30 @@ https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prom
...
+  fi
+}
+
+deploy_operator 0 "${repo_dir}/manifests" "true" "${IMAGE_CLUSTER_LOGGING_OPERATOR}"
 
-$repo_dir/hack/gen-olm-artifacts.sh ${CSV_FILE} ${NAMESPACE} 'crd' >> ${global_manifest}
-$repo_dir/hack/gen-olm-artifacts.sh ${EO_CSV_FILE} ${NAMESPACE} 'crd' >> ${global_manifest}
-$repo_dir/hack/gen-olm-artifacts.sh ${EO_CSV_FILE} ${NAMESPACE} >> ${manifest}
+oc -n openshift-cluster-version scale deployment cluster-version-operator --replicas=0
+oc -n openshift-operator-lifecycle-manager scale deployment olm-operator --replicas=0
```
richm: You probably need some sort of `wait` here - e.g. https://github.com/openshift/origin-aggregated-logging/blob/master/hack/deploy-logging.sh#L208

test/e2e/clusterlogging_test.go
```diff
@@ -121,8 +121,9 @@ func clusterLoggingInitialDeploymentTest(t *testing.T, f *framework.Framework, c
 			LogStore: logging.LogStoreSpec{
 				Type: logging.LogStoreTypeElasticsearch,
 				ElasticsearchSpec: logging.ElasticsearchSpec{
-					NodeCount: 1,
-					Resources: esResources,
+					NodeCount:        1,
+					Resources:        esResources,
+					RedundancyPolicy: elasticsearch.ZeroRedundancy,
```
richm: this change looks unrelated?

test/e2e/clusterlogging_test.go
```diff
@@ -160,18 +162,17 @@ func clusterLoggingInitialDeploymentTest(t *testing.T, f *framework.Framework, c
 	if err != nil {
 		return err
 	}
+	err = WaitForElasticsearchResouce(t, f, namespace, retryInterval, timeout)
```
richm: Resource - and unrelated?

test/e2e/utils.go
```diff
@@ -10,11 +10,33 @@ import (
 	core "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
 	dynclient "sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+func WaitForElasticsearchResouce(t *testing.T, f *framework.Framework, namespace string, retryInterval, timeout time.Duration) error {
```
richm: Resource - and unrelated?

hack/deploy-image.sh
jcantrill: No.  I added this as a convenience when manually pushing your images so you could see the full pull spec as an output in the execution of the target.  Suggestions how to make it more obvious?  Maybe an echo or even capturing the output and spliting/combining it into an actual pull spec?

hack/test-e2e.sh
```diff
@@ -32,21 +38,30 @@ https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prom
 oc create -n ${NAMESPACE} -f \
 https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/servicemonitor.crd.yaml || :
 
-repo_dir="$(dirname $0)/.."
-$repo_dir/hack/gen-olm-artifacts.sh ${CSV_FILE} ${NAMESPACE} 'ns' | oc create -f - || :
+repo_dir=$(echo "$(cd "$(dirname "$1")"; pwd -P)")
```
jcantrill: @richm not sure if i should revert this.  I was seeing strange output trying to deploy the EO and then reverted that logic when I realized the e2e test doesn't need to deploy the EO to do a test on the CLO and what it does.

hack/deploy-image.sh
richm: No, I think this is fine.

hack/test-e2e.sh
jcantrill: I think i may be able to revert this.  This was added to handle the upgrade portion of the test but I think we should fix the upgrade part by going through a typical OLM workflow which would make this unnecessary.

test/e2e/clusterlogging_test.go
jcantrill: It is not.  There actually is an OLM validation failure that doesn't manifest itself now in our e2e test because we don't go through OLM.

test/e2e/clusterlogging_test.go
jcantrill: will correct... and it is related because we only need to validate the CLO creates an ES resource.  We will defer additional e2e testing for the EO to that repo.

hack/testing/test-010-deploy-via-olm-minimal.sh
```diff
@@ -0,0 +1,98 @@
...
+
+if [ "$?" != "0" ] ; then
+	os::log::error "Error deploying operator via OLM using manifest: $manifest"
+	exit 1
+fi
+
+TIMEOUT_MIN=$((2 * $minute))
+
+# wait for operator to be ready
+os::cmd::try_until_text "oc -n $NAMESPACE get deployment cluster-logging-operator -o jsonpath={.status.availableReplicas} --ignore-not-found" "1" ${TIMEOUT_MIN}
```
richm: I get the cluster-logging-operator - but I don't get an elasticsearch-operator - should I?

richm: hmm - I guess this test doesn't need EO - for a real test, will need to use olm-test-script to deploy EO

jcantrill: No.  This test is simply to verify the CLO deployment and that it creates the resources we expect.  It intentionally is simple like the legacy test.  I would expect we add a broader set of tests like OAL test suite or something that actually verifies pods spin up, etc.

hack/testing/test-020-olm-upgrade.sh
```diff
@@ -0,0 +1,141 @@
...
+
+# deploy EO
+cat<<EOL | oc create -f -
+apiVersion: v1
+kind: List
+items:
+- apiVersion: v1
+  kind: Namespace
+  metadata:
+    name: openshift-operators-redhat
```
richm: Are we still using `openshift-operators-redhat` for 4.2?  I suggest using a variable like `EO_NS` instead of having `openshift-operators-redhat` hardcoded everywhere.

hack/testing/test-020-olm-upgrade.sh
```diff
@@ -0,0 +1,141 @@
...
+else
+  IMAGE_CLUSTER_LOGGING_OPERATOR=${IMAGE_CLUSTER_LOGGING_OPERATOR:-quay.io/openshift/origin-cluster-logging-operator:latest}
+fi
+
+KUBECONFIG=${KUBECONFIG:-$HOME/.kube/config}
+
+oc create -n ${NAMESPACE} -f \
+https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/prometheusrule.crd.yaml || :
+oc create -n ${NAMESPACE} -f \
+https://raw.githubusercontent.com/coreos/prometheus-operator/master/example/prometheus-operator-crd/servicemonitor.crd.yaml || :
```
richm: Are these still necessary?  Are they for prometheus scraping only?

hack/testing/test-020-olm-upgrade.sh
```diff
@@ -0,0 +1,141 @@
...
+
+repo_dir="$(dirname $0)/../.."
+
+# deploy EO
+cat<<EOL | oc create -f -
```
richm: Isn't this what the olm-test-script does?

hack/testing/utils
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+
+indent() {
+  INDENT="      "
+  sed "s/^/$INDENT/" | sed "s/^${INDENT}\($1\)/${INDENT:0:-2}- \1/"
+}
+
+deployConfigMapForCatalogSource() {
```
richm: rather than vendoring in olm-test-script, you are rewriting it?

richm:
@jcantrill you expressed reluctance at having to git clone origin-aggregated-logging to get the test files - I agree - for oal I changed it so that it will no longer git clone clo and eo - instead I wrote a recursive github file grabber - https://github.com/openshift/origin-aggregated-logging/pull/1734/files#diff-0144f31f058e341bb59f3d9838899a29R102 and https://github.com/openshift/origin-aggregated-logging/pull/1734/files#diff-0144f31f058e341bb59f3d9838899a29R75

You could do the same - grab `hack/deploy-logging.sh`, `hack/test-logging.sh`, and `hack/testing/template/logging-ci-test-runner-template.yaml`

hack/testing/test-020-olm-upgrade.sh
jcantrill: They are for scraping but I believe our operator still pragmatically creates subscriptionmonitors and fails without the CRDs.  Would be good to test on a fresh cluster to see if these are already present

hack/testing/test-020-olm-upgrade.sh
jcantrill: Not exactly.  This is 'deploy from marketplace' which I stole from your code.  The script is 'deploy from my source'

hack/testing/test-020-olm-upgrade.sh
jcantrill: Changes comming

hack/testing/test-020-olm-upgrade.sh
richm: They appear to be - in the oal PR I did not need to do this

hack/testing/test-020-olm-upgrade.sh
```diff
@@ -0,0 +1,94 @@
...
+  end_seconds=$(date +%s)
+  runtime="$(($end_seconds - $start_seconds))s"
+  local return_code="$?"
+  set +e
+  oc -n openshift-operators-redhat -o yaml get subscription elasticsearch-operator > $ARTIFACT_DIR/subscription-eo.yml 2>&1 ||:
+  oc logs -n ${NAMESPACE} deployment/cluster-logging-operator > $ARTIFACT_DIR/cluster-logging-operator.log 2>&1 ||:
+  oc -n ${NAMESPACE} -o yaml get subscription cluster-logging-operator > $ARTIFACT_DIR/subscription-clo.yml 2>&1 ||:
+  oc -n ${NAMESPACE} -o yaml get configmap cluster-logging > $ARTIFACT_DIR/configmap-for-catalogsource-clo.yml 2>&1 ||:
+  oc -n ${NAMESPACE} -o yaml get catalogsource cluster-logging > $ARTIFACT_DIR/catalogsource-clo.yml 2>&1 ||:
+  oc  -n openshift-operator-lifecycle-manager logs deployment/catalog-operator > $ARTIFACT_DIR/catalog-operator.logs --since=$runtime 2>&1 ||:
```
richm: `--since` should come right after `logs`

hack/testing/test-020-olm-upgrade.sh
```diff
@@ -0,0 +1,94 @@
...
+}
+trap cleanup exit
+
+if [ -n "${IMAGE_CLUSTER_LOGGING_OPERATOR:-}" ] ; then
+  source "$repo_dir/hack/common"
+fi
+if [ -n "${IMAGE_FORMAT:-}" ] ; then
+  IMAGE_CLUSTER_LOGGING_OPERATOR=$(sed -e "s,\${component},cluster-logging-operator," <(echo $IMAGE_FORMAT))
+else
+  IMAGE_CLUSTER_LOGGING_OPERATOR=${IMAGE_CLUSTER_LOGGING_OPERATOR:-quay.io/openshift/origin-cluster-logging-operator:latest}
```
richm: I think we are not using quay.io?  If not, use `registry.svc.ci.openshift.org/origin/4.2:cluster-logging-operator`

manifests/cluster-logging.package.yaml
```diff
@@ -1,5 +1,5 @@
 #! package-manifest: ./deploy/chart/catalog_resources/rh-operators/clusterlogging.v4.1.0.clusterserviceversion.yaml
 packageName: cluster-logging
 channels:
-- name: 4.2
+- name: "4.2"
```
richm: We definitely need this fixed asap

hack/lib/OWNERS
```diff
@@ -0,0 +1,13 @@
+reviewers:
+  - stevekuznetsov
+  - smarterclayton
+  - kargakis
+  - soltysh
+  - liggitt
+  - jhadvig
+  - fabianofranz
```
bparees: this owners file looks wrong/out of date.

hack/testing/utils
```diff
@@ -0,0 +1,85 @@
...
+}
+
+deploy_config_map_catalog_source() {
+  local namespace=$1
+  local manifest_dir=$2
+  local image=${3:-}
+
+  local version=$(basename $(find $manifest_dir -type d | sort -r | head -n 1))
+
+	local CRD=$(sed '/^#!.*$/d' $manifest_dir/$version/*crd.yaml | grep -v -- "---" | indent apiVersion)
```
bparees: indentation

hack/vendor/olm-test-script/README.md
```diff
@@ -0,0 +1,29 @@
+# OLM Test Script
+
+## Description
+This script can be used to test install your operator's manifest file in an OCP cluster
+
+### Variables
```
bparees: operator image env var?

hack/testing/test-010-deploy-via-olm-minimal.sh
```diff
@@ -0,0 +1,87 @@
...
+oc create ns ${NAMESPACE} || :
+
+
+os::cmd::expect_success "oc create -f ${repo_dir}/vendor/github.com/openshift/elasticsearch-operator/manifests/${version}/elasticsearches.crd.yaml"
+
+
+os::log::info "Deploying operator from ${manifest}"
+NAMESPACE=${NAMESPACE} \
+VERSION=${version} \
+OPERATOR_IMAGE=${IMAGE_CLUSTER_LOGGING_OPERATOR} \
```
bparees: what about the operand image name?    ideally that should also come from the IMAGE_FORMAT so we can get the one built by the PR and reuse this in the aggregated logging repo

hack/vendor/olm-test-script/README.md
```diff
@@ -0,0 +1,29 @@
...
+| *DEBUG* | default false | Set this to true to enable debug output of the e2e-olm script |
+
+| *TEST_NAMESPACE* | default olm-test | This is the namespace where the test subscription, operator group, and csv will be installed to. Note: this will not be created by this script. |
+
+| *MANIFEST_DIR* | default ./deploy/manifests | This is the path to your operator's manifest directory. |
+
+| *VERSION* | default 4.1 | The version directory under $MANIFEST_DIR where your operator's csv file exists. |
+
+### Execution
+MANIFEST_DIR=/data/src/github.com/openshift/ansible-service-broker ./e2e-olm.sh
```
bparees: ansible service broker?

hack/vendor/olm-test-script/e2e-olm.sh
```diff
@@ -0,0 +1,104 @@
...
+CRD=$(sed '/^#!.*$/d' $MANIFEST_DIR/$VERSION/*crd.yaml | grep -v -- "---" | indent apiVersion)
+PKG=$(sed '/^#!.*$/d' $MANIFEST_DIR/*package.yaml | indent packageName)
+CSV=$(sed '/^#!.*$/d' $MANIFEST_DIR/$VERSION/*version.yaml | sed 's/namespace: placeholder/namespace: '$TEST_NAMESPACE'/' |grep -v -- "---" |  indent apiVersion)
+
+if [ -n "${OPERATOR_IMAGE:-}" ] ; then
+  CSV=$(echo "$CSV" | sed -e "s~containerImage:.*+~containerImage: ${OPERATOR_IMAGE}~" | indent apiVersion)
+  CSV=$(echo "$CSV" | sed -e "s~image:.*~image: ${OPERATOR_IMAGE}\n~" | indent ApiVersion)
+fi
+
+cat > /tmp/configmap.yaml <<EOF | sed 's/^  *$//'
```
bparees: why isn't this sharing code with `deploy_config_map_catalog_source` ?

hack/testing/test-010-deploy-via-olm-minimal.sh
jcantrill: This particular test does no verification specific to the deployed operand; it is basically a revised version of the original test which only validated the operator produced the desired resources.  I would envison that we either add more tests or expand this one to possibly validate the image used by the operand

hack/testing/test-020-olm-upgrade.sh
jcantrill: I'll change it but seems we could simply drop this line as it should default to doing and using whats in the manifest

hack/vendor/olm-test-script/README.md
jcantrill: All items related to `vendor` are meant to be straight copies from the original repo in order foster the original repo being maintained and the vedor here being periodically updated.  Seems worthy to try and maintain that

hack/vendor/olm-test-script/README.md
jcantrill: same

hack/vendor/olm-test-script/e2e-olm.sh
jcantrill: its came before the function.  I think it would be worth contributing the function back to shawn's repo or figuring out where a 'olm test tools' repo should live

manifests/cluster-logging.package.yaml
jcantrill: What action are you asking me to take here?

hack/testing/test-020-olm-upgrade.sh
jcantrill: removed

hack/lib/OWNERS
jcantrill: removed.

hack/testing/test-010-deploy-via-olm-minimal.sh
bparees: but i assume someday you'll want to invoke this test from your aggregated-logging repo also.  So is there any reason not to make it fully parameterized to consume "the images built by the PR job" in preparation for that?

hack/vendor/olm-test-script/README.md
bparees: ok.

hack/vendor/olm-test-script/README.md
bparees: ok

hack/test-e2e.sh
```diff
@@ -1,58 +1,30 @@
...
-  -singleNamespace
+#!/bin/bash
+
+current_dir=$(dirname "${BASH_SOURCE[0]}" )
+source "${current_dir}/lib/init.sh"
+
+for test in $( find "${current_dir}/testing" -type f -name 'test-*.sh' | sort); do
+	os::log::info "running e2e $test "
+	for i in $(seq 1 60); do
+		set +e
```
richm: why not just 
```bash
if ! oc get project openshift-logging > /dev/null 2>&1 ; then
    break
fi
sleep 1
```

hack/testing/assertions
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+assert_resources_exist(){
+	# verify deployments -- kibana, curator
+	os::cmd::try_until_success "oc -n $NAMESPACE get deployment kibana" ${TIMEOUT_MIN}
```
richm: Does this spew a lot of output?

hack/testing/test-001-e2e-legacy.sh
```diff
@@ -0,0 +1,79 @@
...
+
+
+$repo_dir/hack/gen-olm-artifacts.sh ${CSV_FILE} ${NAMESPACE} 'ns' | oc create -f - || :
+
+$repo_dir/hack/gen-olm-artifacts.sh ${CSV_FILE} ${NAMESPACE} >> ${manifest}
+sed -i "s,quay.io/openshift/origin-cluster-logging-operator:latest,${IMAGE_CLUSTER_LOGGING_OPERATOR}," ${manifest}
+if [ -n "${IMAGE_FORMAT:-}" ] ; then
+  for comp in logging-curator5 logging-elasticsearch5 logging-fluentd logging-kibana5 logging-oauth-proxy logging-rsyslog ; do
+    img=$(sed -e "s,\${component},$comp," <(echo $IMAGE_FORMAT))
+    sed -i "s,quay.io/openshift/origin-${comp}:latest,${img}," ${manifest}
```
richm: ok for now - but this will break if we ever change `quay.io` to something else - this is how I handled it in origin-aggregated-logging, which does not depend on looking for the string `quay.io/something` - https://github.com/openshift/origin-aggregated-logging/pull/1734/files#diff-0144f31f058e341bb59f3d9838899a29R177

hack/testing/assertions
jcantrill: not unless debug is true

hack/testing/test-001-e2e-legacy.sh
```diff
@@ -0,0 +1,79 @@
...
+}
+trap cleanup exit
+
+if [ -n "${IMAGE_CLUSTER_LOGGING_OPERATOR:-}" ] ; then
+  source "$(dirname $0)/../common"
+fi
+if [ -n "${IMAGE_FORMAT:-}" ] ; then
+  IMAGE_CLUSTER_LOGGING_OPERATOR=$(sed -e "s,\${component},cluster-logging-operator," <(echo $IMAGE_FORMAT))
+else
+  IMAGE_CLUSTER_LOGGING_OPERATOR=${IMAGE_CLUSTER_LOGGING_OPERATOR:-quay.io/openshift/origin-cluster-logging-operator:latest}
```
richm:
```
   IMAGE_CLUSTER_LOGGING_OPERATOR=${IMAGE_CLUSTER_LOGGING_OPERATOR:-registry.svc.ci.openshift.org/origin/4.2:cluster-logging-operator}
```

hack/testing/test-010-deploy-via-olm-minimal.sh
```diff
@@ -0,0 +1,87 @@
...
+}
+trap cleanup exit
+
+if [ -n "${IMAGE_CLUSTER_LOGGING_OPERATOR:-}" ] ; then
+  source "$(dirname $0)/../common"
+fi
+if [ -n "${IMAGE_FORMAT:-}" ] ; then
+  IMAGE_CLUSTER_LOGGING_OPERATOR=$(sed -e "s,\${component},cluster-logging-operator," <(echo $IMAGE_FORMAT))
+else
+  IMAGE_CLUSTER_LOGGING_OPERATOR=${IMAGE_CLUSTER_LOGGING_OPERATOR:-quay.io/openshift/origin-cluster-logging-operator:latest}
```
richm: see above

hack/testing/test-010-deploy-via-olm-minimal.sh
```diff
@@ -0,0 +1,87 @@
...
+
+# deploy cluster logging
+os::cmd::expect_success "oc -n $NAMESPACE create -f ${repo_dir}/hack/cr.yaml"
+
+# assert deployment
+assert_resources_exist
+
+# deploy alt collector
+os::log::info Patching the collector to deploy an alternate
+os::cmd::try_until_success 'oc -n $NAMESPACE get clusterlogging instance -o yaml | sed -e "s/type: fluentd/type: rsyslog/" | oc replace -f -' ${TIMEOUT_MIN}
```
richm: Use `os::cmd::expect_success` instead of `os::cmd::try_until_success`

I would also suggest that this isn't really the code under test, so you don't have to use `os::cmd::expect_success` here either - just run the command - if it returns an error the script will exit and go through `cleanup` since you are using `set -e`

hack/test-e2e.sh
jcantrill: if the project doesn't terminate in a timely manor it means you potentially run into a condition where you are starting the new test before you have a "clean slate"

hack/testing/test-010-deploy-via-olm-minimal.sh
```diff
@@ -0,0 +1,87 @@
...
+export NAMESPACE="openshift-logging"
+repo_dir="$(dirname $0)/../.."
+manifest=${repo_dir}/manifests
+version=$(basename $(find $manifest -type d | sort -r | head -n 1))
+
+cleanup(){
+  local return_code="$?"
+  set +e
+  oc logs -n ${NAMESPACE} deployment/cluster-logging-operator > $ARTIFACT_DIR/cluster-logging-operator.log 2>&1 ||:
+  oc delete ns ${NAMESPACE} --wait=false --ignore-not-found
```
richm: If you do `--wait=true` here, when the `oc delete` command returns, are you guaranteed that you can immediately do an `oc adm new-project $NAMESPACE` that will succeed?

hack/testing/test-020-olm-upgrade.sh
```diff
@@ -0,0 +1,94 @@
...
+  oc -n openshift-operators-redhat -o yaml get subscription elasticsearch-operator > $ARTIFACT_DIR/subscription-eo.yml 2>&1 ||:
+  oc logs -n ${NAMESPACE} deployment/cluster-logging-operator > $ARTIFACT_DIR/cluster-logging-operator.log 2>&1 ||:
+  oc -n ${NAMESPACE} -o yaml get subscription cluster-logging-operator > $ARTIFACT_DIR/subscription-clo.yml 2>&1 ||:
+  oc -n ${NAMESPACE} -o yaml get configmap cluster-logging > $ARTIFACT_DIR/configmap-for-catalogsource-clo.yml 2>&1 ||:
+  oc -n ${NAMESPACE} -o yaml get catalogsource cluster-logging > $ARTIFACT_DIR/catalogsource-clo.yml 2>&1 ||:
+  oc  -n openshift-operator-lifecycle-manager logs --since=$runtime deployment/catalog-operator > $ARTIFACT_DIR/catalog-operator.logs 2>&1 ||:
+  oc  -n openshift-operator-lifecycle-manager logs --since=$runtime deployment/olm-operator > $ARTIFACT_DIR/olm-operator.logs 2>&1 ||:
+  
+  for item in "crd/elasticsearches.logging.openshift.io" "crd/clusterloggings.logging.openshift.io" "ns/openshift-logging" "ns/openshift-operators-redhat"; do
+    oc delete $item --wait=false --ignore-not-found --force --grace-period=0
```
richm: `--wait=true`?

hack/testing/test-001-e2e-legacy.sh
```diff
@@ -0,0 +1,79 @@
...
+cleanup(){
+  local return_code="$?"
+  set +e
+
+  oc logs deployment/cluster-logging-operator > $ARTIFACT_DIR/operator.logs
+  cat $manifest > $ARTIFACT_DIR/manifest
+  cat $global_manifest > $ARTIFACT_DIR/global_manifest
+
+  oc delete -f $manifest --ignore-not-found
+  oc delete -f $global_manifest --ignore-not-found
```
richm: `--wait=true`?

hack/testing/test-010-deploy-via-olm-minimal.sh
richm: not sure if you missed this comment - but it looks like elsewhere you are using `--wait=true` - use it here too

hack/testing/test-010-deploy-via-olm-minimal.sh
```diff
@@ -0,0 +1,88 @@
...
+
+if [ "$?" != "0" ] ; then
+	os::log::error "Error deploying operator via OLM using manifest: $manifest"
+	exit 1
+fi
+
+TIMEOUT_MIN=$((2 * $minute))
+
+# wait for operator to be ready
+os::cmd::try_until_text "oc -n $NAMESPACE get deployment cluster-logging-operator -o jsonpath={.status.availableReplicas} --ignore-not-found" "1" ${TIMEOUT_MIN}
```
richm: what does `--ignore-not-found` do here?  Does this mean if the deployment does not exist, then `oc get deployment` can return success, and we can attempt to create the CR without the operator deployment existing?

hack/testing/test-020-olm-upgrade.sh
```diff
@@ -0,0 +1,98 @@
...
+else
+  IMAGE_CLUSTER_LOGGING_OPERATOR=${IMAGE_CLUSTER_LOGGING_OPERATOR:-registry.svc.ci.openshift.org/origin/4.2:cluster-logging-operator}
+fi
+
+
+# deploy EO
+os::cmd::expect_success 'deploy_marketplace_operator "openshift-operators-redhat" "elasticsearch-operator"'
+
+# verify operator is ready
+os::cmd::try_until_text "oc -n openshift-operators-redhat get deployment elasticsearch-operator -o jsonpath={.status.availableReplicas} --ignore-not-found" "1" ${TIMEOUT_MIN}
```
richm: what does `--ignore-not-found` do here?

hack/testing/test-020-olm-upgrade.sh
```diff
@@ -0,0 +1,98 @@
...
+
+# deploy CLO
+os::cmd::expect_success 'deploy_marketplace_operator "openshift-logging" "cluster-logging-operator" "cluster-logging"'
+
+# verify operator is ready
+os::cmd::try_until_text "oc -n openshift-logging get deployment cluster-logging-operator -o jsonpath={.status.updatedReplicas} --ignore-not-found" "1" ${TIMEOUT_MIN}
```
richm: what does `--ignore-not-found` do here?

hack/testing/test-020-olm-upgrade.sh
```diff
@@ -0,0 +1,98 @@
...
+payload="{\"op\":\"replace\",\"path\":\"/spec/source\",\"value\":\"cluster-logging\"}"
+payload="$payload,{\"op\":\"replace\",\"path\":\"/spec/sourceNamespace\",\"value\":\"openshift-logging\"}"
+payload="$payload,{\"op\":\"replace\",\"path\":\"/spec/channel\",\"value\":\"$version\"}"
+os::cmd::expect_success "oc -n $NAMESPACE patch subscription cluster-logging-operator --type json -p '[$payload]'"
+
+#verify deployment is rolled out
+os::cmd::try_until_text "oc -n openshift-logging get deployment cluster-logging-operator -o jsonpath={.spec.template.spec.containers[0].image}" "${IMAGE_CLUSTER_LOGGING_OPERATOR}" ${TIMEOUT_MIN}
+
+# verify operator is ready
+os::cmd::try_until_text "oc -n openshift-logging get deployment cluster-logging-operator -o jsonpath={.status.updatedReplicas} --ignore-not-found" "1" ${TIMEOUT_MIN}
```
richm: what does `--ignore-not-found` do here?

hack/testing/test-010-deploy-via-olm-minimal.sh
richm: ok - I see - I think the way `os::cmd::try_until_text` works is that it is really implemented like `try_until_success_and_text` - so it should already ignore the return value if it is a failure code - so the `--ignore-not-found` is unused but harmless

PR #228, opened 2019-08-06

nhosoi:
fluentd, rsyslog - Shorten the TerminationGracePeriod from 30 sec to 10 sec.
    Sleep for the duration to wait for the collector pod is deleted,
    then delete secret and service account.

### DIFF
--- a/pkg/k8shandler/fluentd.go
+++ b/pkg/k8shandler/fluentd.go
@@ -2,6 +2,7 @@ package k8shandler
 
 import (
 	"fmt"
+	"time"
 
 	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
 	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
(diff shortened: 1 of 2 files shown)


nhosoi:
Note: this is not the final solution - shortening the TerminationGracePeriodSeconds may break the left over logs in the cache.  Also, we still don't know the optimal number for the option.

nhosoi:
> /lgtm
> 
> Please add a test to origin-aggregated-logging zzz-rsyslog.sh - check for .orphaned at the beginning of the test (to see if some other test caused `.orphaned`, after stopping fluentd and starting rsyslog (to see if it is fluentd shutdown or rsyslog startup that creates `.orphaned`), and at the end of the test - we currently put back fluentd in the `cleanup` of that test, but move that into the actual test, then we can see if deleting rsyslog and adding back fluentd causes `.orphaned`

I thought we don't want to merge this, but look for a better solution...

Regarding the test case, it requires "oc delete clusterlogging instance" then re-create it.  Is it possible to do that from zzz-rsyslog.sh?

richm:
> I thought we don't want to merge this, but look for a better solution...

sorry, I thought you did want to merge it . . .

> Regarding the test case, it requires "oc delete clusterlogging instance" then re-create it.

I guess that would be another test case . . .

Then, I guess never mind - do you want to revert this PR?

PR #229, opened 2019-08-06
Bug 1737286: The logs collected by rsyslog are not parsed

richm:
https://bugzilla.redhat.com/show_bug.cgi?id=1737286
Get rid of the mmnormalize based SKIP_EMPTY processing.  In order
for it to be useful for our case we need to be able to specify
CDM_KEEP_EMPTY_FIELDS since we preserve the empty `message`
field by default.
Add more fluentd functionality - unconditionally delete the
MESSAGE and log fields, ignoring PRESERVE_JSON_LOG.  Keep the
empty message field in the record.
If we have performance problems with the mmexternal based
undefined_field.go, we can revisit adding skip_empty_fields
support to the mmnormalize json parser, or write
undefined_field.go as an rsyslog C module.

### DIFF
--- a/files/rsyslog/65-viaq-formatting.conf
+++ b/files/rsyslog/65-viaq-formatting.conf
@@ -16,6 +16,10 @@ if strlen($!MESSAGE) > 0 then {
             unset $!originalmsg;
             unset $!unparsed-data;
         }
+        # NOTE: We don't currently use PRESERVE_JSON_LOG - we unconditionally delete
+        # the original field - this is in accordance with what fluentd does - it
+        # may preserve the original field in the parse_json plugin, but it will
+        # unconditionally remove the field in the viaq filter
     }
     # ensure that $!message is set and $!MESSAGE is unset
     # if rsyslog is case sensitive, then $!MESSAGE == $!message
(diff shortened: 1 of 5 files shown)


files/rsyslog/65-viaq-formatting.conf
```diff
@@ -16,36 +16,47 @@ if strlen($!MESSAGE) > 0 then {
...
+        # may preserve the original field in the parse_json plugin, but it will
+        # unconditionally remove the field in the viaq filter
     }
     # ensure that $!message is set and $!MESSAGE is unset
     # if rsyslog is case sensitive, then $!MESSAGE == $!message
     if $!MESSAGE == $!message then {
         set $.save_message = $!message;
         unset $!MESSAGE; # which also unsets $!message
         set $!message = $.save_message;
-        unset $.save_message;
```
nhosoi: Why is this line removed? $.save_message is consumed in the line before?  (I thought "set" is a copy.)

files/rsyslog/rsyslog.sh
```diff
@@ -42,109 +43,22 @@ issue_deprecation_warnings() {
...
     # Create a config file for undefined fields from env vars
-    echo "{\"CDM_USE_UNDEFINED\" : ${CDM_USE_UNDEFINED}, \
+    echo "{\"CDM_USE_UNDEFINED\" : ${CDM_USE_UNDEFINED:-false}, \
 \"CDM_DEFAULT_KEEP_FIELDS\" : \"${CDM_DEFAULT_KEEP_FIELDS:-"CEE,time,@timestamp,aushape,ci_job,collectd,docker,fedora-ci,file,foreman,geoip,hostname,ipaddr4,ipaddr6,kubernetes,level,message,namespace_name,namespace_uuid,offset,openstack,ovirt,pid,pipeline_metadata,rsyslog,service,systemd,tags,testcase,tlog,viaq_msg_id"}\", \
 \"CDM_EXTRA_KEEP_FIELDS\" : \"${CDM_EXTRA_KEEP_FIELDS:-""}\", \
 \"CDM_UNDEFINED_NAME\" : \"${CDM_UNDEFINED_NAME:-"undefined"}\", \
-\"CDM_KEEP_EMPTY_FIELDS\" : \"${CDM_KEEP_EMPTY_FIELDS}\", \
-\"CDM_UNDEFINED_TO_STRING\" : ${CDM_UNDEFINED_TO_STRING}, \
-\"CDM_UNDEFINED_DOT_REPLACE_CHAR\" : \"${CDM_UNDEFINED_DOT_REPLACE_CHAR}\", \
+\"CDM_KEEP_EMPTY_FIELDS\" : \"${CDM_KEEP_EMPTY_FIELDS:-}\", \
```
nhosoi: Do we want to set "message" to CDM_KEEP_EMPTY_FIELDS, by default? How about "kubernetes"?

files/rsyslog/65-viaq-formatting.conf
richm: It is an optimization - we don't use `$.` variables in the final message - it might be "cheaper" for rsyslog to just remove all of them at the end of record processing, rather than having to delete each one of them separately

files/rsyslog/rsyslog.sh
richm: I don't think we want to add `kubernetes` by default - yet - let's see if people actually want all of the labels and annotations

PR #230, opened 2019-08-16

nhosoi:
In the Cluster Logging instance deletion, resources under Cluster Logging
were deleted independently, which could leave a short time period when the
fluentd or rsyslog daemonset was still running but its logcollector service
account was removed.  That made the logs processed in the time miss kubernetes
information including the namespace name.

In this patch, the following changes are made:
- Setting foregroundDeletion finalizer to collector service account
- Making the ownerReference of fluentd and rsyslog daemonset the collector service account
- Setting blockOwnerDeletion to true in the ownerReference
- The foregroundDeletion informs whether the collector service account is being deleted
  or not via reconciler. If it is being deleted, letting the collector service account
  deletion wait until the other objects are all removed, then, delete the finalizer and
  delete the service account.

### DIFF
--- a/manifests/4.2/cluster-logging.v4.2.0.clusterserviceversion.yaml
+++ b/manifests/4.2/cluster-logging.v4.2.0.clusterserviceversion.yaml
@@ -139,6 +139,7 @@ spec:
           - configmaps
           - secrets
           - serviceaccounts
+          - serviceaccounts/finalizers
           verbs:
           - "*"
         - apiGroups:
--- a/pkg/k8shandler/collection.go
+++ b/pkg/k8shandler/collection.go
@@ -10,10 +10,13 @@ import (
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/util/retry"
 
 	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
 	apps "k8s.io/api/apps/v1"
+	core "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
 const (
@@ -29,10 +32,14 @@ var (
 	timeout       = time.Second * 1800
 )
 
+var serviceAccountLogCollectorUID types.UID
+
 //CreateOrUpdateCollection component of the cluster
 func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection() (err error) {
 
 	cluster := clusterRequest.cluster
+
+	var collectorServiceAccount *core.ServiceAccount
 
 	// there is no easier way to check this in golang without writing a helper function
 	// TODO: write a helper function to validate Type is a valid option for common setup or tear down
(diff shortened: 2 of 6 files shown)


pkg/k8shandler/collection.go
```diff
@@ -180,17 +196,31 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateCollectionPriorityCla
...
+		// This object is not being deleted.
+		if !utils.ContainsString(collectorServiceAccount.ObjectMeta.Finalizers, metav1.FinalizerDeleteDependents) {
+			collectorServiceAccount.ObjectMeta.Finalizers = append(collectorServiceAccount.ObjectMeta.Finalizers, metav1.FinalizerDeleteDependents)
+		}
+		err := clusterRequest.Create(collectorServiceAccount)
+		if err != nil && !errors.IsAlreadyExists(err) {
+			return nil, fmt.Errorf("Failure creating Log Collector service account: %v", err)
+		}
+		if len(collectorServiceAccount.ObjectMeta.UID) != 0 {
+			serviceAccountLogCollectorUID = collectorServiceAccount.ObjectMeta.UID
```
nhosoi: The UID in the logcollector service account is set when it is created.  So, I store it to use it in ownerReference later here.  If there is an api to retrieve it dynamically, we don't have to store the UID this way...  But so far I could not find it.

sample logcollector service account
```
apiVersion: v1
imagePullSecrets:
- name: logcollector-dockercfg-rgwsp
kind: ServiceAccount
metadata:
  creationTimestamp: "2019-08-15T22:12:46Z"
  finalizers:
  - foregroundDeletion
  name: logcollector
  namespace: openshift-logging
  ownerReferences:
  - apiVersion: logging.openshift.io/v1
    controller: true
    kind: ClusterLogging
    name: instance
    uid: cec9ff31-bfa9-11e9-b67d-029422b5c914
  resourceVersion: "468364"
  selfLink: /api/v1/namespaces/openshift-logging/serviceaccounts/logcollector
  uid: cfa30bf4-bfa9-11e9-a7b2-0a1781cc6f5c
```

pkg/k8shandler/collection.go
```diff
@@ -268,10 +298,14 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateCollectorServiceAccou
 
 	err = clusterRequest.Create(collectorReaderClusterRoleBinding)
 	if err != nil && !errors.IsAlreadyExists(err) {
-		return fmt.Errorf("Failure creating Log collector %q cluster role binding: %v", collectorReaderClusterRoleBinding.Name, err)
+		return nil, fmt.Errorf("Failure creating Log collector %q cluster role binding: %v", collectorReaderClusterRoleBinding.Name, err)
 	}
 
-	return nil
+	if delfinalizer == true {
```
richm:
```go
     if delfinalizer {
```
it's already a bool

nhosoi: Thanks!  Fixed.

nhosoi:
> have you tested this with the origin-aggregated-logging CI tests?

Yes!
```
+ echo finished hack/test-logging.sh at 2019-08-18 19:22:53-07:00
finished hack/test-logging.sh at 2019-08-18 19:22:53-07:00
+ '[' PASS = PASS ']'
```

manifests/4.2/cluster-logging.v4.2.0.clusterserviceversion.yaml
```diff
@@ -139,6 +139,7 @@ spec:
           - configmaps
           - secrets
           - serviceaccounts
+          - serviceaccounts/finalizers
```
richm: @jcantrill @ewolinetz do we need to add this rbac anywhere else?  Is this CSV the one used by customers?

PR #234, opened 2019-08-23
Bug 1732585: Modify kibana-proxy to rely on ServiceAccount as oauthcl…

jcantrill:
…ient

This bug resolves https://bugzilla.redhat.com/show_bug.cgi?id=1732585 by:
* Modifying kibana-proxy to rely on ServiceAccount as an oauthclient so that we no longer need to keep a secret and and oauthclient in sync

### DIFF
--- a/pkg/k8shandler/visualization.go
+++ b/pkg/k8shandler/visualization.go
@@ -4,9 +4,7 @@ import (
 	"bytes"
 	"fmt"
 	"reflect"
-	"time"
 
-	oauth "github.com/openshift/api/oauth/v1"
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	"github.com/sirupsen/logrus"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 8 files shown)


pkg/k8shandler/visualization.go
```diff
@@ -473,11 +380,11 @@ func newKibanaPodSpec(cluster *logging.ClusterLogging, kibanaName string, elasti
...
 		"--https-address=:3000",
 		"-provider=openshift",
-		"-client-id=kibana-proxy",
-		"-client-secret-file=/secret/oauth-secret",
+		"-client-id=system:serviceaccount:openshift-logging:kibana",
+		"-client-secret-file=/var/run/secrets/kubernetes.io/serviceaccount/token",
 		"-cookie-secret-file=/secret/session-secret",
 		"-upstream=http://localhost:5601",
-		"-scope=user:info user:check-access user:list-projects",
+		"-scope=user:info user:check-access",
```
richm: we don't need list-projects here?

jcantrill: I'm uncertain why this is here as the Kibana proxy never needed to list projects.  All of that particular logic is in the openshift-elasticsearch-plugin.

jcantrill: backtracking here as https://bugzilla.redhat.com/show_bug.cgi?id=1746695 was introduced.. not understood why our access script doesnt expose it

jcantrill:
Any cherrypick also needs https://github.com/openshift/cluster-logging-operator/pull/241

PR #237, opened 2019-08-30
Ignore GoLand folder

lukas-vlcek:


.gitignore
```diff
@@ -1,5 +1,7 @@
 pf.log
 _output
+# GoLand
```
richm: ok - but have you tried Visual Studio Code with the golang module?

lukas-vlcek: Yes I did. Worked fine.
But I was lucky to get subscription provided by Red Hat for GoLand and I do not think I am going to switch back to VS anytime soon :-)

PR #239, opened 2019-09-05
Bug 1746968: Remove hostpath mount /etc/hostname and pass via env var

jcantrill:
Resolves https://bugzilla.redhat.com/show_bug.cgi?id=1746968 by moving the value to a downward API

### DIFF
--- a/pkg/k8shandler/collection.go
+++ b/pkg/k8shandler/collection.go
@@ -334,7 +334,11 @@ func isDaemonsetDifferent(current *apps.DaemonSet, desired *apps.DaemonSet) (*ap
 		logrus.Infof("Collector resource(s) change found, updating %q", current.Name)
 		different = true
 	}
-
+	if !reflect.DeepEqual(current.Spec.Template.Spec.Containers[0].Env, desired.Spec.Template.Spec.Containers[0].Env) {
+		logrus.Infof("Collector container EnvVar change found, updating %q", current.Name)
+		current.Spec.Template.Spec.Containers[0].Env = desired.Spec.Template.Spec.Containers[0].Env
+		different = true
+	}
 	return current, different
 }
 
--- a/pkg/k8shandler/fluentd.go
+++ b/pkg/k8shandler/fluentd.go
@@ -215,6 +215,7 @@ func newFluentdPodSpec(logging *logging.ClusterLogging, elasticsearchAppName str
 	}
 
 	fluentdContainer.Env = []v1.EnvVar{
+		{Name: "NODE_NAME", ValueFrom: &v1.EnvVarSource{FieldRef: &v1.ObjectFieldSelector{FieldPath: "spec.nodeName"}}},
 		{Name: "MERGE_JSON_LOG", Value: "false"},
 		{Name: "PRESERVE_JSON_LOG", Value: "true"},
 		{Name: "K8S_HOST_URL", Value: "https://kubernetes.default.svc"},
@@ -245,7 +246,6 @@ func newFluentdPodSpec(logging *logging.ClusterLogging, elasticsearchAppName str
 		{Name: "varlibdockercontainers", ReadOnly: true, MountPath: "/var/lib/docker"},
 		{Name: "config", ReadOnly: true, MountPath: "/etc/fluent/configs.d/user"},
 		{Name: "certs", ReadOnly: true, MountPath: "/etc/fluent/keys"},
-		{Name: "dockerhostname", ReadOnly: true, MountPath: "/etc/docker-hostname"},
 		{Name: "localtime", ReadOnly: true, MountPath: "/etc/localtime"},
 		{Name: "dockercfg", ReadOnly: true, MountPath: "/etc/sysconfig/docker"},
 		{Name: "dockerdaemoncfg", ReadOnly: true, MountPath: "/etc/docker"},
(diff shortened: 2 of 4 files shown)


pkg/k8shandler/collection.go
```diff
@@ -334,7 +334,11 @@ func isDaemonsetDifferent(current *apps.DaemonSet, desired *apps.DaemonSet) (*ap
 		logrus.Infof("Collector resource(s) change found, updating %q", current.Name)
 		different = true
 	}
-
+	if !reflect.DeepEqual(current.Spec.Template.Spec.Containers[0].Env, desired.Spec.Template.Spec.Containers[0].Env) {
+		logrus.Infof("Collector container EnvVar change found, updating %q", current.Name)
+		current.Spec.Template.Spec.Containers[0].Env = desired.Spec.Template.Spec.Containers[0].Env
+		different = true
```
richm: Does this mean we will no longer be able to set environment variables in the fluentd/rsyslog daemonsets to control the behavior?

pkg/k8shandler/fluentd.go
```diff
@@ -215,6 +215,7 @@ func newFluentdPodSpec(logging *logging.ClusterLogging, elasticsearchAppName str
 	}
 
 	fluentdContainer.Env = []v1.EnvVar{
+		{Name: "NODE_NAME", ValueFrom: &v1.EnvVarSource{FieldRef: &v1.ObjectFieldSelector{FieldPath: "spec.nodeName"}}},
```
richm: we'll need to change fluentd and rsyslog to consume this
do we need to add this? If I `oc rsh` into a pod and do `env`, I see there are environment variables related to the hostname - are there some we can already use?

pkg/k8shandler/collection.go
jcantrill: this means if your operator is in 'managed' it will revert any changes to what is expected.  We will need to scale down our operator or make it unmanaged in order to do any testing unrelated to what the operator does

pkg/k8shandler/fluentd.go
jcantrill: I have not looked at the list but typically 'hostname' is the name of the pod.  The link you posted to me previously seemed to make it obvious that we don't use the node hostname anywhere so if we don't need the value then we could simply remove adding it all together

pkg/k8shandler/collection.go
richm: Right - I get that.
This is a big change in behavior.  This means, in 4.1, you could set env. vars. directly in the fluentd daemonset in Managed mode and they would apply.  However, big change in 4.2, if you set env. vars. directly in the fluentd daemonset in Managed mode, they will _not_ apply - they will be reverted.
If this is an intentional behavioral change in our product, we need to document this in the release notes and in the product documentation.

pkg/k8shandler/fluentd.go
richm: I'm looking at a recent 4.2 cluster.  It seems that all of our pods have a `HOSTNAME` env. var. added by Kubernetes?  OpenShift?
```
HOSTNAME=fluentd-xds7j
HOSTNAME=kibana-544698d578-qrltc
```
etc.
So I think we should just rely on the presence of the HOSTNAME env. var. in both fluentd and rsyslog, and get rid of the use of /etc/docker-hostname and /etc/hostname.  This means changes in origin-aggregated-logging/fluent-plugin-viaq_data_model and cluster-logging-operator.

pkg/k8shandler/collection.go
jcantrill: It is my opinion after operators were introduced that being able to modify env vars is not the correct behavior.  The whole point of the operator pattern is that we are supposed to know what is best in the operational behavior and we set the things that are necessary and expose knobs in the CR as needed.  The operator owns the resources it creates and has the right to ensure they are to the spec it defines.  The fact that you could tweak the env is really a bug in behavior and not by design.  If you need to tweak it then either you should: go into unmanaged or log an RFE for us to promote something into the CR.

As a side note, the forwarding feature will explicitly rely on an env var that contains a hash of its config and the fact that it changes as the mechanism to ensure pipeline changes are causing the DS to roll out

pkg/k8shandler/fluentd.go
jcantrill: This hostname is the actual pod itself.  Wouldn't the nodename on which the fluent pod ran be useful in logging to understand on which node a given pod was running?  I don't have a strong opinion about keeping in the env var so I defer to if you believe we should simply remove it.

pkg/k8shandler/collection.go
richm: > It is my opinion after operators were introduced that being able to modify env vars is not the correct behavior

I agree, but this is still a big change that may break customer deployments when they upgrade to 4.2.  When you take away features, expect to hear about it . . .

pkg/k8shandler/fluentd.go
richm: ok, I see.  Yes, this is very useful, we should keep this - we'll have to change origin-aggregated-logging/fluent-plugin-viaq_data_model/cluster-logging-operator to use it

ewolinetz:
do we have a PR to adjust fluentd and rsyslog to handle this?
otherwise lgtm

richm:
> do we have a PR to adjust fluentd and rsyslog to handle this?
> otherwise lgtm

No, not yet.  It's also not clear if it will break logging CI if we remove this mount . . .

richm:
afaict, this won't break anything - here are the places where /etc/docker-hostname is referenced:
https://github.com/ViaQ/fluent-plugin-viaq_data_model/blob/master/lib/fluent/plugin/filter_viaq_data_model.rb#L201 - use of /etc/docker-hostname and `@docker_hostname` are protected

https://github.com/openshift/cluster-logging-operator/blob/master/files/rsyslog/rsyslog.sh#L72
also protected

/lgtm

richm:
putting on hold temporarily
@jcantrill please add a change to `rsyslog/files/rsyslog.sh` so that it looks like this:
```bash
if [ -f /etc/docker-hostname ] ; then
    HOSTNAME=`cat /etc/docker-hostname`
    IPADDR4=`getent ahostsv4 $HOSTNAME | awk '/ STREAM / {print $1}'`
    IPADDR6=`getent ahostsv6 $HOSTNAME | awk '/ STREAM / {print $1}'`
elif [ -n "${NODE_NAME:-}" ] ; then
    HOSTNAME=$NODE_NAME
    IPADDR4=${NODE_IPV4:-`getent ahostsv4 $HOSTNAME | awk '/ STREAM / {print $1}'`}
    IPADDR6=`getent ahostsv6 $HOSTNAME | awk '/ STREAM / {print $1}'`
elif [ -n "${NODE_IPV4:-}" ] ; then
...
```

jcantrill:
> ```shell
> elif [ -n "${NODE_NAME:-}" ] ; then
>     HOSTNAME=$NODE_NAME
>     IPADDR4=${NODE_IPV4:-`getent ahostsv4 $HOSTNAME | awk '/ STREAM / {print $1}'`}
>     IPADDR6=`getent ahostsv6 $HOSTNAME | awk '/ STREAM / {print $1}'`
> ```
done

PR #240, opened 2019-09-06
Bug 1745394: Got many `[WARN ][c.f.s.a.BackendRegistry ] Authenticat…

nhosoi:
…ion finally failed for null` in ES pod log after changing log collector from fluentd to rsyslog.

Adding a check in rsyslog.sh for the certs/keys used by omelasticsearch,
for the cert/token used by mmkubernetes, and for certs/keys used to access
prometheus.

### DIFF
--- a/files/rsyslog/rsyslog.sh
+++ b/files/rsyslog/rsyslog.sh
@@ -107,8 +107,10 @@ if [ -z $ES_PORT ]; then
 fi
 
 # Check bearer_token_file for fluent-plugin-kubernetes_metadata_filter.
-if [ ! -s /var/run/secrets/kubernetes.io/serviceaccount/token ] ; then
-    echo "ERROR: Bearer_token_file (/var/run/secrets/kubernetes.io/serviceaccount/token) to access the Kubernetes API server is missing or empty."
+k8s_token="/var/run/secrets/kubernetes.io/serviceaccount/token"
+k8s_cacert="/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+if [ ! -s "${k8s_token}" -o ! -s "${k8s_cacert}" ] ; then
+    echo "ERROR: Bearer_token_file (${k8s_token}) or ca cert (${k8s_cacert}) to access the Kubernetes API server is missing or empty."
     exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


nhosoi:
Do we want to add the similar checks to the fluentd run.sh script although the similar errors have not been observed with the fluentd pod AFAIK...

files/rsyslog/rsyslog.sh
```diff
@@ -119,6 +121,28 @@ OPS_CLIENT_CERT=${OPS_CLIENT_CERT:-$ES_CLIENT_CERT}
...
+if [ ! -s "${ES_CLIENT_CERT}" -o ! -s "${OPS_CLIENT_CERT}" ] ; then
+    echo "ERROR: Client cert (${ES_CLIENT_CERT} or ${OPS_CLIENT_CERT}) to access Elasticsearch is missing or empty."
+    exit 1
+fi
+if [ ! -s "${ES_CLIENT_KEY}" -o ! -s "${OPS_CLIENT_KEY}" ] ; then
+    echo "ERROR: Client key (${ES_CLIENT_KEY} or ${OPS_CLIENT_KEY}) to access Elasticsearch is missing or empty."
+    exit 1
+fi
+
+if [[ ${ENABLE_PROMETHEUS_ENDPOINT} ]] ; then
```
richm: `if [ $ENABLE_PROMETHEUS_ENDPOINT = true ] ; then`

files/rsyslog/rsyslog.sh
```diff
@@ -107,8 +107,10 @@ if [ -z $ES_PORT ]; then
 fi
 
 # Check bearer_token_file for fluent-plugin-kubernetes_metadata_filter.
-if [ ! -s /var/run/secrets/kubernetes.io/serviceaccount/token ] ; then
-    echo "ERROR: Bearer_token_file (/var/run/secrets/kubernetes.io/serviceaccount/token) to access the Kubernetes API server is missing or empty."
+k8s_token="/var/run/secrets/kubernetes.io/serviceaccount/token"
+k8s_cacert="/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
+if [ ! -s "${k8s_token}" -o ! -s "${k8s_cacert}" ] ; then
+    echo "ERROR: Bearer_token_file (${k8s_token}) or ca cert (${k8s_cacert}) to access the Kubernetes API server is missing or empty."
```
richm: one consequence of this is that QE is going to report that rsyslog doesn't start at all and gives this error message - we'll need to catch that and close them as duplicates of 1745394 - I suppose rsyslog not starting at all (fail fast) is better than rsyslog starting but not actually working (fail late).

richm:
We need a way to reproduce this problem so that we reduce the cycle time.  Please try to find a way to reproduce this.  It doesn't have to be a CI test, maybe just a script that restarts rsyslog over and over until this error occurs.

files/rsyslog/rsyslog.sh
nhosoi: Thanks!  Fixed.

PR #250, opened 2019-10-08
link to openshift product docs for install

bparees:


bparees:
@jcantrill @ewolinetz we should update this to 4.2 once the 4.2 docs are available/linkable, since the instruction vary slightly.

@derekwaynecarr fyi.

richm:
> @jcantrill @ewolinetz we should update this to 4.2 once the 4.2 docs are available/linkable, since the instruction vary slightly.

too bad we can't just link to `https://docs.openshift.com/container-platform/latest/logging/efk-logging-deploying.html` where `latest` is a symlink that gets updated by docs when a new release is available

bparees:
> too bad we can't just link to https://docs.openshift.com/container-platform/latest/logging/efk-logging-deploying.html where latest is a symlink that gets updated by docs when a new release is available

yeah, although that would also mean users on old versions of the operator would get pointed to newer docs that might not be accurate for them either.

hopefully the docs stabilize to the point where it's not critical.

ewolinetz:
@bparees would it make sense to have a table so we could list different versions then?

bparees:
> @bparees would it make sense to have a table so we could list different versions then?

a table in the CSV itself?  seems awkward.  "here's a set of links, some of them may not work currently, only one of them relates to the operator you are actually in the process of installing....choose wisely!"

PR #252, opened 2019-10-10
convert component specs to pointers

jcantrill:
Convert component specs to pointers so that when reconciling CL we can properly skip components which are not intended to be deployed


jcantrill:
/hold cancel

Unable to write e2e test as api server rejects CL instance with empty component types

jcantrill:
cc @ewolinetz

PR #253, opened 2019-10-11
Fixing skopeo auth issue in deploy-image.sh

nhosoi:
[ERROR]
+++ dirname hack/deploy-image.sh
+ skopeo copy --dest-tls-verify=false docker-daemon:quay.io/openshift/origin-cluster-logging-operator:latest docker://127.0.0.1:5000/openshift/origin-cluster-logging-operator:latest
time="..." level=fatal msg="Error trying to reuse blob sha256:... at destination: unable to retrieve auth token: invalid username/password"

### DIFF
--- a/hack/common
+++ b/hack/common
@@ -164,7 +164,7 @@ function login_to_registry() {
             username=kubeadmin
         fi
     fi
-    podman login --tls-verify=false -u "$username" -p "$token" "$1" > /dev/null
+    docker login -u "$username" -p "$token" "$1" > /dev/null
 }
 
 function push_image() {


hack/common
```diff
@@ -165,10 +165,16 @@ function login_to_registry() {
         fi
     fi
     podman login --tls-verify=false -u "$username" -p "$token" "$1" > /dev/null
```
richm: it works if you use docker instead:
```bash
docker login -u "$username" -p "$token" "$1" > /dev/null
```
I guess we just can't use podman yet :-(

richm: So please revise this PR to only change the single line above

nhosoi: Is you suggestion replacing
`podman login --tls-verify=false -u "$username" -p "$token" "$1" > /dev/null`
with
`docker login -u "$username" -p "$token" "$1" > /dev/null`
?  And reverting the rest?  If yes, let me try the changes in my next deploy.  I've been using this patch for my .local CLO debugging... :)

richm: > Is you suggestion replacing

yes - see https://github.com/openshift/origin-aggregated-logging/pull/1771

nhosoi: Cool!  It worked.  Let me update this PR...

PR #256, opened 2019-10-17
CLO logs messages "Collector container EnvVar change fo…

nhosoi:
…und, updating \"fluentd\"" although there's no obvious fluentd EnvVar change.

Replacing reflect.DeepEqual with envValueEqual to skip the value provided
by the downward api or otherwise provided by kubernetes.

### DIFF
--- a/pkg/k8shandler/collection.go
+++ b/pkg/k8shandler/collection.go
@@ -271,8 +271,10 @@ func isDaemonsetDifferent(current *apps.DaemonSet, desired *apps.DaemonSet) (*ap
 		logrus.Infof("Collector resource(s) change found, updating %q", current.Name)
 		different = true
 	}
-	if !reflect.DeepEqual(current.Spec.Template.Spec.Containers[0].Env, desired.Spec.Template.Spec.Containers[0].Env) {
+
+	if !utils.EnvValueEqual(current.Spec.Template.Spec.Containers[0].Env, desired.Spec.Template.Spec.Containers[0].Env) {
 		logrus.Infof("Collector container EnvVar change found, updating %q", current.Name)
+
 		current.Spec.Template.Spec.Containers[0].Env = desired.Spec.Template.Spec.Containers[0].Env
 		different = true
 	}
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -240,3 +240,67 @@ func RemoveString(slice []string, s string) (result []string) {
 	}
 	return
 }
+
+
+/**
+EnvValueEqual - check if 2 EnvValues are equal or not
+Notes:
+- reflect.DeepEqual does not return expected results if the to-be-compared value is a pointer.
+- needs to adjust with k8s.io/api/core/v#/types.go when the types are updated.
+**/
+func EnvValueEqual(env1, env2 []v1.EnvVar) bool {
+	var found bool
+	if len(env1) != len(env2) {
+		return false
+	}
+	for _, elem1 := range env1 {
+		found = false
+		for _, elem2 := range env2 {
+			if elem1.Name == elem2.Name {
+				if elem1.Value != elem2.Value {
+					return false
+				}
+				if (elem1.ValueFrom != nil && elem2.ValueFrom == nil) ||
+				   (elem1.ValueFrom == nil && elem2.ValueFrom != nil) {
+					return false
+				}
+				if elem1.ValueFrom != nil {
+					found = EnvVarSourceEqual(*elem1.ValueFrom, *elem2.ValueFrom)
+				} else {
+					found = true
+				}
+				break
+			}
+		}
+		if !found {
+			return false
+		}
+	}
+	return true
+}
+
+func EnvVarSourceEqual(esource1, esource2 v1.EnvVarSource) bool {
+	if (esource1.FieldRef != nil && esource2.FieldRef == nil) ||
+	   (esource1.FieldRef == nil && esource2.FieldRef != nil) ||
+	   (esource1.ResourceFieldRef != nil && esource2.ResourceFieldRef == nil) ||
+	   (esource1.ResourceFieldRef == nil && esource2.ResourceFieldRef != nil) ||
+	   (esource1.ConfigMapKeyRef != nil && esource2.ConfigMapKeyRef == nil) ||
+	   (esource1.ConfigMapKeyRef == nil && esource2.ConfigMapKeyRef != nil) ||
+	   (esource1.SecretKeyRef != nil && esource2.SecretKeyRef == nil) ||
+	   (esource1.SecretKeyRef == nil && esource2.SecretKeyRef != nil) {
+		return false
+	}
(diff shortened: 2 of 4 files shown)


pkg/k8shandler/collection.go
```diff
@@ -366,3 +368,23 @@ func isBufferFlushRequired(current *apps.DaemonSet, desired *apps.DaemonSet) boo
 func getServiceAccountLogCollectorUID() types.UID {
 	return serviceAccountLogCollectorUID
 }
+
+func envValueEqual(cenv, denv []core.EnvVar) bool {
```
richm: I don't think this handles the case where the user may have added or deleted env vars to the daemonset.  I think you'll have to add logic to say "if celem not in denv then return false" and "if delem not in cenv then return false"

nhosoi: :p indeed...  thanks!

jcantrill:
@bparees do we really need a 4.3 bz to resolve this?  My impression is no

pkg/k8shandler/collection.go
jcantrill: There should be a unit test for this.  Wouldn't we be better to implement a whitelist of env vars that we simply skip checking?  Maybe rebuild the actual list as needed.

bparees:
> @bparees do we really need a 4.3 bz to resolve this? My impression is no

you don't need a 4.3 BZ to merge into master.  But if you are going to backport this to 4.2 or 4.1 you'll need a 4.3 BZ in verified state first.

pkg/k8shandler/collection.go
nhosoi: I've added a unit test.   Checking the EnvVar array as is, though...  Do you think it's ok?  Or better to rebuild a list?

pkg/utils/utils_test.go
```diff
@@ -46,3 +47,77 @@ func TestAreMapsSameWhenDifferent(t *testing.T) {
...
+		{Name: "PRESERVE_JSON_LOG", Value: "true"},
+		{Name: "K8S_HOST_URL", Value: "https://kubernetes.default.svc"},
+	}
+
+	if EnvValueEqual(currentenv, desiredenv) {
+		t.Errorf("EnvVarEqual returned true for the not equal inputs")
+	}
+}
+
+func TestEnvVarEqualShorter(t *testing.T) {
```
richm: Please also provide a test for when denv is longer than cenv

ewolinetz:
Given that Env vars are stored as `map[string]string` does this mean that https://github.com/openshift/cluster-logging-operator/blob/master/pkg/utils/utils.go#L45-L47 will have issues too?

Can we kill to birds with one stone and fix that function and use that for env var comparison?

pkg/utils/utils.go
```diff
@@ -240,3 +240,40 @@ func RemoveString(slice []string, s string) (result []string) {
...
+	if len(cenv) != len(denv) {
+		return false
+	}
+	delemcheck := make([]bool, len(denv))
+	for _, celem := range cenv {
+		found = false
+		if celem.ValueFrom != nil {
+			continue
+		}
+		for d, delem := range denv {
```
ewolinetz: should this be nested?
```
if delem.ValueFrom != nil {
  delemcheck[d] = true
```

pkg/utils/utils.go
```diff
@@ -240,3 +240,40 @@ func RemoveString(slice []string, s string) (result []string) {
...
+				}
+				found = true
+				break
+			}
+		}
+		if !found {
+			return false
+		}
+	}
+	for _, dc := range delemcheck {
```
ewolinetz: is this necessary since you're checking the array lengths and return `false` if `!found`

nhosoi: Actually, the all delemcheck codes were _not_ necessary...  It was meant for delem was superset of celem, but it's handled by len(celem) != len(delem).  So, I'm deleting them.  Thanks!

nhosoi:
Verified with this timeout increase, the e2e-operator test passes.  Reverting the change, let me run the test again...
```
diff --git a/test/e2e/clusterlogging_test.go b/test/e2e/clusterlogging_test.go
index b470ebdd..73e2d836 100644
--- a/test/e2e/clusterlogging_test.go
+++ b/test/e2e/clusterlogging_test.go
@@ -21,7 +21,7 @@ import (
 
 var (
        retryInterval        = time.Second * 5
-       timeout              = time.Second * 120
+       timeout              = time.Second * 300
        cleanupRetryInterval = time.Second * 1
        cleanupTimeout       = time.Second * 5
 )

```

pkg/utils/utils.go
```diff
@@ -240,3 +240,32 @@ func RemoveString(slice []string, s string) (result []string) {
...
 }
+
+func EnvValueEqual(cenv, denv []v1.EnvVar) bool {
+	var found bool
+	if len(cenv) != len(denv) {
+		return false
+	}
+	for _, celem := range cenv {
+		found = false
+		if celem.ValueFrom != nil {
```
ewolinetz: we want to also compare `ValueFrom` to `ValueFrom` if they're both available

nhosoi: My original plan was to skip comparing ValueFrom following the advice from Rich.  I'm afraid if we compare ValueFrom, we are to use reflect.DeepEqual?  That'd return false even if they look identical...

> Rich Megginson wrote: 
> I think the operator should ignore env vars where the value is provided by the downward api or otherwise provided by kubernetes.  That is, it is meaningless to compare the NODE_NAME value in the operator with the one from the currently running daemonset.

ewolinetz: this was something we brought up during scrum today... it'd be possible for users to have a `Value` compared to a `ValueFrom` otherwise and we would need to revert it to the appropriate one otherwise. Same with a `ValueFrom` vs another `ValueFrom`... if someone changes the variable to be based on the namespace rather than the pod name, it could cause issues in the fluentd image so we'd want to just keep that as we set it.

richm: Ok - if we expect the `ValueFrom` value to be identical in the current and the desired, then we should compare them too - I had thought perhaps that the value in the current one would have been altered by Kubernetes, that is, that Kubernetes would replace it with the actual value, but I guess that only happens when the actual pod is created.
So yes - we should compare `ValueFrom` also - but I guess we can't use `DeepEqual` because it doesn't work, and we don't know why . . .

pkg/utils/utils.go
```diff
@@ -240,3 +240,67 @@ func RemoveString(slice []string, s string) (result []string) {
 	}
 	return
 }
+
+
+/**
+EnvValueEqual - check if 2 EnvValues are equal or not
+Notes:
+- reflect.DeepEqual does not return expected results if the to-be-comared value is a pointer.
```
jcantrill: nit: spelling `compared`

nhosoi: *shame shame* :p Thanks!

PR #260, opened 2019-10-24
Log367 log forwarding resource

jcantrill:
Implement https://jira.coreos.com/browse/LOG-462 for 
https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/cluster-logging-log-forwarding.md


jcantrill:
cc @bparees @sichvoge

pkg/k8shandler/forwarding.go
```diff
@@ -0,0 +1,223 @@
...
+const (
+	internalOutputName       = "clo-default-output-es"
+	collectorSecretName      = "fluentd"
+	logStoreService          = "elasticsearch.openshift-logging.svc:9200"
+	defaultAppPipelineName   = "clo-default-app-pipeline"
+	defaultInfraPipelineName = "clo-default-infra-pipeline"
+	secureForwardConfHash    = "8163d9a59a20ada8ab58c2535a3a4924"
+
+	//ForwardingAnnotation  Annotate CL instance with a value of "enabled"
+	ForwardingAnnotation = "clusterlogging.openshift.io/logforwarding"
```
bparees: this is the critical annotation to enable the TP feature, right?  Can we put TP in the name?

e.g. `clusterlogging.openshift.io/logforwardingtechpreview`

pkg/k8shandler/forwarding.go
```diff
@@ -46,15 +46,19 @@ func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig() (config s
...
 
 func (clusterRequest *ClusterLoggingRequest) normalizeLogForwarding(namespace string, cluster *logging.ClusterLogging) logforward.ForwardingSpec {
+	logger.Debugf("Normalizing logforwarding from request: %v", clusterRequest)
+	logger.Debugf("ForwardingRequest: %v", clusterRequest.ForwardingRequest)
+	logger.Debugf("ForwardingSpec: %v", clusterRequest.ForwardingSpec)
 	if cluster.Spec.LogStore != nil && cluster.Spec.LogStore.Type == logging.LogStoreTypeElasticsearch {
-		if cluster.Spec.Forwarding == nil || (!cluster.Spec.Forwarding.DisableDefaultForwarding && len(cluster.Spec.Forwarding.Pipelines) == 0) {
-			cluster.Status.Forwarding = &logforward.ForwardingStatus{
+		if !clusterRequest.ForwardingSpec.DisableDefaultForwarding && len(clusterRequest.ForwardingSpec.Pipelines) == 0 {
+			logger.Debug("Configuring logforwarding to utilize the operator managed logsore")
```
richm: logstore

manifests/4.3/logforwardings.crd.yaml
```diff
@@ -0,0 +1,76 @@
...
+kind: CustomResourceDefinition
+metadata:
+  name: clusterloggings.logging.openshift.io
+spec:
+  group: logging.openshift.io
+  names:
+    kind: LogForwarding
+    listKind: LogForwardingList
+    plural: logforwardings
+    singular: logforwarding
```
jcantrill: @bparees In the interest of reserving the ability to have namespaced logforwarding in the future, should the kind maybe be `clusterlogforwarding`

bparees: that seems reasonable.  though in that case you could also make it a clusterscoped resource (is clusterlogging a cluster scoped resource today?)

jcantrill: no.

bparees: I guess hedge it by naming it clusterlogforwarding for now and we can reevaluate before we create/name the GA api.

hack/testing/utils
```diff
@@ -114,3 +116,58 @@ spec:
...
+    TARGET_NAMESPACE=${namespace} \
+    ${repo_dir}/hack/vendor/olm-test-script/e2e-olm.sh
+
+    if [ "$?" != "0" ] ; then
+	    os::log::error "Error deploying operator via OLM using manifest: $manifest"
+	    exit 1
+    fi
+
+    # wait for operator to be ready
+    os::cmd::try_until_text "oc -n $namespace get deployment $operatorName -o jsonpath={.status.availableReplicas} --ignore-not-found" "1" ${timeout}
```
ewolinetz: can we move this into a function of its own since we're using it multiple times?

hack/vendor/olm-test-script/e2e-olm.sh
```diff
@@ -59,9 +59,9 @@ test::object_assert() {
 }
 
 # Name of the configmap that we will create
-CONFIGMAP_NAME=${CONFIGMAP_NAME:-openshift-olm-test}
+CONFIGMAP_NAME=${CONFIGMAP_NAME:-openshift-olm-test$SUFFIX}
 
-CRD=$(sed '/^#!.*$/d' $MANIFEST_DIR/$VERSION/*crd.yaml | grep -v -- "---" | indent apiVersion)
+CRD=$(cat $MANIFEST_DIR/$VERSION/*crd.yaml | sed '/^#!.*$/d' | grep -v -- "---" | indent apiVersion)
```
ewolinetz: what was wrong with the prior implementation?

manifests/4.3/cluster-logging.v4.3.0.clusterserviceversion.yaml
```diff
@@ -20,46 +19,78 @@ metadata:
...
-                "visualization": {
-                  "type": "kibana",
-                  "kibana": {
-                    "replicas": 1
+          {
+            "apiVersion": "logging.openshift.io/v1",
+            "kind": "ClusterLogging",
+            "metadata": {
+              "name": "instance",
+              "namespace": "openshift-logging"
```
ewolinetz: don't we want to have the annotation if we're going to have the `logforwarding` definition?

pkg/apis/logging/v1alpha1/forwarding_types.go
```diff
@@ -0,0 +1,196 @@
...
+
+	//Pipelines is the status of the outputs
+	Pipelines []PipelineStatus `json:"pipelines,omitempty"`
+
+	//Outputs is the status of the outputs
+	Outputs []OutputStatus `json:"outputs,omitempty"`
+}
+
+//PipelineStatus is the status of a give pipeline
+type PipelineStatus struct {
```
ewolinetz: can we use `conditions` like other k8s components use? (including our operators)

test/e2e/logforwarding/forward_to_fluent_test.go
```diff
@@ -0,0 +1,155 @@
...
+			BeforeEach(func() {
+				if fluentDeployment, err = e2e.DeployFluendReceiver(rootDir, false); err != nil {
+					Fail(fmt.Sprintf("Unable to deploy fluent receiver: %v", err))
+				}
+
+				cr := helpers.NewClusterLogging(helpers.ComponentTypeCollector)
+				if err := e2e.CreateClusterLogging(cr); err != nil {
+					Fail(fmt.Sprintf("Unable to create an instance of cluster logging: %v", err))
+				}
+				forwarding := &logforward.LogForwarding{
```
JAORMX: This is missing TypeMeta: metav1.TypeMeta ; which defines the Kind of resource and APIVersion

pkg/apis/logging/v1alpha1/forwarding_types.go
```diff
@@ -0,0 +1,196 @@
...
+	PipelineStateReasonUnrecognizedSource   PipelineStateReason = "The source type is unrecognized"
+	PipelineStateReasonNonUniqueName        PipelineStateReason = "The pipeline is missing a unique name"
+	PipelineStateReasonMissingName          PipelineStateReason = "The pipeline is missing a name"
+	PipelineStateReasonMissingOutputs       PipelineStateReason = "The pipeline does not reference any outputs or referes more than exist"
+	PipelineStateReasonMissingSource        PipelineStateReason = "The pipeline is missing a source type"
+	PipelineStateReasonReservedNameConflict PipelineStateReason = "Name conflicts with an internally reserved name"
+)
+
+//OutputStatus of a given output
+type OutputStatus struct {
```
ewolinetz: same question about being `conditions`

pkg/controller/forwarding/forwarding_controller.go
```diff
@@ -0,0 +1,127 @@
...
+	}
+	if err = r.client.Status().Update(context.TODO(), instance); err != nil {
+		return reconcile.Result{}, err
+	}
+	if instance.Name != singletonName {
+		return reconcile.Result{}, nil
+	}
+
+	clInstance := &loggingv1.ClusterLogging{}
+	clName := types.NamespacedName{Name: singletonName, Namespace: openshiftNS}
```
ewolinetz: i think if we don't find the `clInstance` we shouldn't reconcile the same... 
should we split up the logic in `k8shandler.Reconcile` and move it into the controllers?

hack/vendor/olm-test-script/e2e-olm.sh
jcantrill: > what was wrong with the prior implementation?

It didnt properly indent and format multiple crd files for the configmap.

manifests/4.3/cluster-logging.v4.3.0.clusterserviceversion.yaml
jcantrill: > don't we want to have the annotation if we're going to have the `logforwarding` definition?

yes.  our guidelines require explicit opt into the feature.  @bparees should the example enable LF via annoation by default

manifests/4.3/logforwardings.crd.yaml
jcantrill: deferring to a GA discussion

pkg/controller/forwarding/forwarding_controller.go
jcantrill: > i think if we don't find the `clInstance` we shouldn't reconcile the same...
Why not? LF depends on CL to evaluate whether it should default routing to the operator managed logstore.  CL relies on LF to understand how to create the collector config.  Presumably there should be no issues processing the same since we aren't saving CL since its an empty struct.  If it is then we should fix that.
  
> should we split up the logic in `k8shandler.Reconcile` and move it into the controllers?
Not sure I understand the benefit

test/e2e/logforwarding/forward_to_fluent_test.go
jcantrill: > This is missing TypeMeta: metav1.TypeMeta ; which defines the Kind of resource and APIVersion

While this is true, the framework code actually serializes the struct and makes a straight REST call to the api server which must rely on the URI to determine the type instead of the struct.

pkg/generators/forwarding/fluentd/generators.go
```diff
@@ -0,0 +1,210 @@
...
+	"text/template"
+
+	logforward "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
+	"github.com/openshift/cluster-logging-operator/pkg/generators"
+	"github.com/openshift/cluster-logging-operator/pkg/logger"
+	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+var (
+	defaultInfraProjectPatterns = []string{"default", "openshift", "openshift-", "kube-"}
```
ewolinetz: we're calling them `default` here but we have no means to change it... can we just drop 'default'?

hack/testing/test-367-logforwarding.sh
```diff
@@ -0,0 +1,54 @@
...
+start_seconds=$(date +%s)
+os::test::junit::declare_suite_start "${BASH_SOURCE[0]}"
+
+ARTIFACT_DIR=${ARTIFACT_DIR:-"$repo_dir/_output"}
+if [ ! -d $ARTIFACT_DIR ] ; then
+  mkdir -p $ARTIFACT_DIR
+fi
+
+cleanup(){
+  os::log::info "Running cleanup"
```
JAORMX: I think `local return_code=$?` has to go above the os::log::info call

Else it'll get the value of the call to the info command and return success instead of catching the error.

pkg/generators/forwarding/fluentd/generators.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	data := struct {
+		CollectInfraLogs        bool
+		CollectAppLogs          bool
+		SourceInputLabels       []string
+		SourceToPipelineLabels  []string
+		PipelinesToOutputLabels []string
+		OutputLabels            []string
+	}{
```
ewolinetz: this doesn't require a `,` ?

jcantrill: > can we define an actual object type for this instead?

Is there a particular reason we need an actual object type instead of anonymous one?

pkg/generators/forwarding/fluentd/generators.go
```diff
@@ -0,0 +1,210 @@
...
+	out.WriteString(trimmed)
+	out.WriteString("\n")
+	return levelOut
+}
+
+//generateSourceToPipelineLabels generates fluentd label stanzas to fan source types to multiple pipelines
+func (engine *ConfigGenerator) generateSourceToPipelineLabels(sourcesToPipelines map[logforward.LogSourceType][]string) ([]string, error) {
+	configs := []string{}
+	for sourceType, pipelineNames := range sourcesToPipelines {
+		data := struct {
```
ewolinetz: create a type for this?

pkg/generators/forwarding/fluentd/generators.go
```diff
@@ -0,0 +1,210 @@
...
+		}
+		configs = append(configs, result)
+	}
+	return configs, nil
+}
+
+func (engine *ConfigGenerator) generatePipelineToOutputLabels(pipelines []logforward.PipelineSpec) ([]string, error) {
+	configs := []string{}
+	for _, pipeline := range pipelines {
+		data := struct {
```
ewolinetz: create a type for this?

hack/testing/utils
```diff
@@ -70,6 +70,8 @@ items:
 EOL
 
 fi
+  # wait for operator to be ready
+  os::cmd::try_until_text "oc -n $namespace get deployment $name -o jsonpath={.status.availableReplicas} --ignore-not-found" "1" $((2 * $minute))
```
JAORMX: I think this might be in the wrong spot.

pkg/generators/forwarding/fluentd/generators.go
jcantrill: removed as not need

pkg/k8shandler/clusterloggingrequest.go
```diff
@@ -24,15 +31,19 @@ func (clusterRequest *ClusterLoggingRequest) isManaged() bool {
...
-func (clusterRequest *ClusterLoggingRequest) Update(object runtime.Object) error {
-	logrus.Debugf("Updating: %v", object)
-	return clusterRequest.client.Update(context.TODO(), object)
+//Update the runtime Object or return error
+func (clusterRequest *ClusterLoggingRequest) Update(object runtime.Object) (err error) {
+	logrus.Tracef("Updating: %v", object)
+	if err = clusterRequest.client.Update(context.TODO(), object); err != nil {
+		logrus.Errorf("Error updating %v: %v", object.GetObjectKind(), err)
+	}
+	return err
```
ewolinetz: I don't think this won't contain the value above... when written as you have it the scope is changed. that's why you can't do something like: 
```
if err, val := doTheThing(); err != nil {
  printTheError(err)
}

// val won't be initialized here
doSomethingWithTheVal(val)
```

pkg/k8shandler/collection.go
```diff
@@ -50,6 +54,23 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection() (err err
...
+		if err = clusterRequest.RemoveServiceAccount("logcollector"); err != nil {
+			return
+		}
+
+		if err = clusterRequest.RemovePriorityClass(clusterLoggingPriorityClassName); err != nil {
+			return
+		}
+	}
+
+	if cluster.Spec.Collection.Logs.Type == logging.LogCollectionTypeFluentd {
```
ewolinetz: why is this broken out from the above if block?

pkg/k8shandler/fluentd.go
```diff
@@ -21,48 +25,51 @@ import (
 )
 
 const (
-	fluentdAlertsFile = "fluentd/fluentd_prometheus_alerts.yaml"
+	fluentdAlertsFile      = "fluentd/fluentd_prometheus_alerts.yaml"
+	fluentdCertsSecretName = "fluentd"
+	fluentdName            = "fluentd"
```
ewolinetz: is there a reason to have two const vars that represent the same data?

pkg/k8shandler/fluentd.go
```diff
@@ -387,9 +411,9 @@ func (clusterRequest *ClusterLoggingRequest) updateFluentdDaemonsetIfRequired(de
 	desired, different := isDaemonsetDifferent(current, desired)
 
 	if different {
-
+		current.Spec = desired.Spec
```
ewolinetz: i think we lose generated k8s metadata by doing this (since `current` is populated from a `Get`)

pkg/k8shandler/logstore.go
```diff
@@ -16,6 +16,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	ElasticsearchResourceName = "elasticsearch"
+	elasticsearchSecretName   = "elasticsearch"
```
ewolinetz: same question regarding two consts with same value

pkg/k8shandler/logstore.go
```diff
@@ -16,6 +16,11 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const (
+	ElasticsearchResourceName = "elasticsearch"
```
ewolinetz: `elasticsearchResourceName` ?

pkg/controller/forwarding/forwarding_controller.go
ewolinetz: > Why not? LF depends on CL to evaluate whether it should default routing to the operator managed logstore. CL relies on LF to understand how to create the collector config.

I'm referring to there being no CL... meaning our operator shouldn't be processing the CL at all (the current functionality). This controller is only going to fire from LF events. we would expect that our CL be stood up by the other controller

> Not sure I understand the benefit

Code readibility

pkg/apis/logging/v1alpha1/forwarding_types.go
ewolinetz: use in status:
https://github.com/openshift/cluster-logging-operator/blob/master/pkg/apis/logging/v1/clusterlogging_types.go#L30

definition:
https://github.com/openshift/cluster-logging-operator/blob/master/pkg/apis/logging/v1/clusterlogging_types.go#L241-L247

pkg/controller/forwarding/forwarding_controller.go
jcantrill: > > Why not? LF depends on CL to evaluate whether it should default routing to the operator managed logstore. CL relies on LF to understand how to create the collector config.
> 
> I'm referring to there being no CL... meaning our operator shouldn't be processing the CL at all (the current functionality). This controller is only going to fire from LF events. we would expect that our CL be stood up by the other controller

The state of LF directly affects the state of CL by way of the collector config.  You change pipelines and outputs it will be reflected in new fluent.conf and a bump in the fluent DS.  Assuming we have coded the rest of the CL logic properly there should be no reason for anything else about the stack to change; if it does then that is a bug.  I'm not sure I fully understand your concerns given the resources are intertwined and need to be reconciled as a pair.  The intention overall is to drive our operands to a desired state which can and will occur starting with either controller

pkg/k8shandler/clusterloggingrequest.go
jcantrill: To what `val` are you referring in this context? `Update` takes an object and returns an error

pkg/k8shandler/collection.go
jcantrill: hold over from rsyslog impl.  I'll move it

test/e2e/logforwarding/forward_to_fluent_test.go
jcantrill: Retract this statement and added a change

pkg/apis/logging/v1alpha1/sets.go
```diff
@@ -0,0 +1,19 @@
+package v1alpha1
+
+type LogSouceTypeSet map[LogSourceType]interface{}
```
richm: s/Souce/Source/g

jcantrill:
@ewolinetz @richm All the e2e tests passed.  I'm deferring releasing the hold pending your review/approval.  cc @JAORMX

richm:
lgtm

PR #264, opened 2019-10-28
Catch errors and fix e2e olm upgrade test

JAORMX:
The errors were being ignored because of catching the status of the exit
command too late (having a successful command before it)


JAORMX:
superseded by https://github.com/openshift/cluster-logging-operator/pull/268

PR #268, opened 2019-10-29
fix exit code of e2e tests to expose false positive

jcantrill:


hack/testing/test-020-olm-upgrade.sh
```diff
@@ -26,13 +26,14 @@ TIMEOUT_MIN=$((2 * $minute))
 NAMESPACE="openshift-logging"
 manifest_dir=${repo_dir}/manifests
 version=$(basename $(find $manifest_dir -type d | sort -r | head -n 1))
+previous_version=$(echo $version | awk '{print ($1*10-1)/10.0}')
```
richm: not sure why you need to have the extra math in there - assuming `$version` is `N.M` where `N` and `M` are digits:
```
echo "" | awk -v ver=4.2 '{print ver + 0.1}'
4.3
```

hack/testing/test-020-olm-upgrade.sh
```diff
@@ -65,13 +66,13 @@ fi
 
 
 # deploy EO
-os::cmd::expect_success 'deploy_marketplace_operator "openshift-operators-redhat" "elasticsearch-operator" "elasticsearch-operator" "true" '
+os::cmd::expect_success 'deploy_marketplace_operator "openshift-operators-redhat" "$previous_version" "elasticsearch-operator" "elasticsearch-operator" "true" '
```
richm:
```
deploy_marketplace_operator(){
  local ns=$1
  local name=$2
  local channel=$3
  local package=${4:-$name}
  local global=${5:-false}
```
so you want `$name` to be `$previous_version`?

hack/testing/test-020-olm-upgrade.sh
jcantrill: right

hack/testing/test-020-olm-upgrade.sh
jcantrill: wrong order

hack/testing/test-020-olm-upgrade.sh
```diff
@@ -26,13 +26,14 @@ TIMEOUT_MIN=$((2 * $minute))
 NAMESPACE="openshift-logging"
 manifest_dir=${repo_dir}/manifests
 version=$(basename $(find $manifest_dir -type d | sort -r | head -n 1))
+previous_version=$(echo $version | | awk '{print $1 - 0.1}')
```
richm: `| |` - remove one of the `|`

PR #270, opened 2019-10-31
enable upgrade test

jcantrill:
Attempt to reenable upgrade test

cc @shawn-hurley

### DIFF
--- a/hack/test-e2e.sh
+++ b/hack/test-e2e.sh
@@ -8,14 +8,13 @@ for test in $( find "${current_dir}/testing" -type f -name 'test-*.sh' | sort); 
 	os::log::info "======================================"
 	os::log::info "running e2e $test "
 	os::log::info "======================================"
-	if [ "$(basename $test)" == "test-020-olm-upgrade.sh" ] ; then
-		os::log::warning "Intentionally skipping $test"
-		continue
-	fi
 	if "${test}" ; then
 		os::log::info "e2e $test succeeded at $( date )"
 	else
-		os::log::warning "e2e $test failed at $( date )"
+
+		os::log::error "============= FAILED FAILED ============= "
+		os::log::error "e2e $test failed at $( date )"
+		os::log::error "============= FAILED FAILED ============= "
 		failed="true"
 	fi
 done
--- a/hack/testing/test-020-olm-upgrade.sh
+++ b/hack/testing/test-020-olm-upgrade.sh
@@ -9,12 +9,10 @@ if [ "${DEBUG:-}" = "true" ]; then
 fi
 
 repo_dir="$( cd "$(dirname "$0")/../.." ; pwd -P )"
-source "$repo_dir/hack/lib/init.sh"
-source "$repo_dir/hack/testing/assertions"
 source "$repo_dir/hack/testing/utils"
+source "$repo_dir/hack/testing/assertions"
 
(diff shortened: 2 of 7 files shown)


hack/testing/test-020-olm-upgrade.sh
```diff
@@ -83,6 +83,10 @@ os::cmd::expect_success "oc -n $NAMESPACE create -f ${repo_dir}/hack/cr.yaml"
 assert_resources_exist
 
 os::cmd::expect_success "deploy_config_map_catalog_source $NAMESPACE ${repo_dir}/manifests '${IMAGE_CLUSTER_LOGGING_OPERATOR}'"
+#patch catalog source
+#olm.skipRange: ">=4.2.0 <4.3.0"
+range=">=$previous_version <$(echo $version | awk '{print $1 + 0.1}')"
+os::cmd::expect_success "oc -n $NAMESPACE get configmap cluster-logging -o yaml | sed -e 's~olm.skipRange:.*~olm.skipRange: ${range}~' | oc replace -n $NAMESPACE -f -"
```
richm:
```
version=$(basename $(find $manifest_dir -type d | sort -r | head -n 1))
previous_version=$(echo $version | awk '{print $1 - 0.1}')
```
This means previous_version will be `4.2` - This means `range` and `olm.skipRange` will be `>=4.2 <4.3` - is this ok?  Do you have to have a 3 digit version here?  Also note that you could just use `$version` again:
```
range=">=$previous_version <$version"
```

Also - the `os::cmd::expect_success` and others are only really needed for code under test - that is, the actual code you are testing.  For things like deploying the config map source, editing the configmap, etc. - we aren't testing those things, and since we are running the script with `set -e`, the script will report and exit if that command fails.  Finally, using `os::cmd::xx` adds a _lot_ of noise to the output.

richm: re: using `$version` in range - never mind - it is good as it is
`range=">=$previous_version <$(echo $version | awk '{print $1 + 0.1}')"`
at least as far as the way you are using version - still don't know if it is ok to use 2 digit versions instead of 3 digit

jcantrill: updated to add third digit

jcantrill: @richm maybe the missing third digit is the reason this failed

richm: maybe - I don't know how it is supposed to work

hack/test-e2e.sh
```diff
@@ -8,14 +8,13 @@ for test in $( find "${current_dir}/testing" -type f -name 'test-*.sh' | sort);
...
-	if [ "$(basename $test)" == "test-020-olm-upgrade.sh" ] ; then
-		os::log::warning "Intentionally skipping $test"
-		continue
-	fi
 	if "${test}" ; then
 		os::log::info "e2e $test succeeded at $( date )"
 	else
-		os::log::warning "e2e $test failed at $( date )"
+
+		os::log::error "FAILED FAILED FAILED FAILED FAILED FAILED FAILED"
```
ewolinetz: is this just for debugging or is this being kept like this?

jcantrill: I was intending to leave permanently to make it make it more obvious where it failed

PR #271, opened 2019-10-31
Gather fluentd logs in e2e runs

JAORMX:
This adds a small script that gathers the fluentd logs from the nodes.

### DIFF
--- a/hack/lib/util/logs.sh
+++ b/hack/lib/util/logs.sh
@@ -1,0 +1,10 @@
+#!/bin/bash
+
+get_logging_pod_logs() {
+  local node
+  for node in $(oc get nodes -o jsonpath='{.items[*].metadata.name}') ; do
+      for logfile in $(oc adm node-logs "$node" --path=fluentd/); do
+          oc adm node-logs "$node" --path="fluentd/$logfile" > $ARTIFACT_DIR/$node-$logfile
+      done
+  done
+}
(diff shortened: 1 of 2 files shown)


hack/lib/util/logs.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+get_logging_pod_logs() {
+  local node
+  for node in $(oc get nodes --no-headers | awk '{print $1}') ; do
+      oc adm node-logs $node --path=fluentd/fluentd.log > $ARTIFACT_DIR/$node-fluentd.log
```
richm: hmm - is there some way to use `oc adm node-logs` to grab `fluentd/fluent.*.log` too?  If there is a lot of output, fluentd will do log rotation, and you will want to grab all of them, rotated logs too.

hack/lib/util/logs.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+get_logging_pod_logs() {
+  local node
+  for node in $(oc get nodes --no-headers | awk '{print $1}') ; do
```
richm: the usual idiom for getting only the name of some object is to use `-o jsonpath='{.items[*].metadata.name}'` - so
```bash
for node in $(oc get nodes -o jsonpath='{.items[*].metadata.name}')
```

hack/lib/util/logs.sh
JAORMX: Let me give it a try; I think it's possible

hack/lib/util/logs.sh
JAORMX: thanks!

PR #274, opened 2019-11-01
Enable audit log forwarding through pipelines

JAORMX:
This adds the "log.audit" sourceType to the logforwardings resource and reacts accordingly:

* auditd logs from the linux hosts are forwarded
* audit logs from the openshift-apiserver logs are forwarded
* audit logs from the kubernetes-apisrever logs are forwarded

This is not enabled by default and has to be explicitly enabled in the resource to take effect.


JAORMX:
@richm @jcantrill could you check this out?

PR #280, opened 2019-11-05
Bug 1768685: Improve missing secret message

jcantrill:
Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1768685

### DIFF
--- a/pkg/k8shandler/certificates.go
+++ b/pkg/k8shandler/certificates.go
@@ -7,7 +7,11 @@ import (
 	"github.com/openshift/cluster-logging-operator/pkg/logger"
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	"k8s.io/apimachinery/pkg/api/errors"
+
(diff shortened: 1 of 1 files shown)


pkg/k8shandler/certificates.go
```diff
@@ -68,7 +68,7 @@ func (clusterRequest *ClusterLoggingRequest) extractSecretToFile(secretName stri
 
 	// check to see if the map value exists
 	if !ok {
-		logger.Infof("No secret data %q found", key)
+		logger.Warnf("No secret data %q found. Please be aware but likely not an issue for deprecated keys: app-ca,app-key,app-cert,infra-ca,infra-key,infra-cert", key)
```
richm: then why not make the message `Info` level for deprecated keys, but `Warn` level for everything else?

jcantrill: better?

PR #284, opened 2019-11-06
Bug 1619479: Remove Kibana oauth-proxy param to allow re-login

jcantrill:
Remove parameter so Kibana re-login does not redirect more then once.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1619479


ewolinetz:
@jcantrill after your discussion with @bparees is this still the correct fix for the linked bz?

bparees:
@ewolinetz this PR was the outcome of our discussion/investigation of that BZ

PR #288, opened 2019-11-11
Remove legacy test that is covered by others and does not use OLM dep…

jcantrill:
…loyment


ewolinetz:
as long as we are still maintaining full coverage elsewhere lgtm

PR #289, opened 2019-11-12
 Enable logforwarding e2e test and fix fluentd forwarding test 

JAORMX:
This whole test suite was being skipped by accident. This re-enables
that.

Currently it seems that the admin managed-elasticsearch logstore e2e
tests are broken. So lets skip that and fix it in a subsequent PR.
This way the other tests stay covered.

The fluentd forwarding test was broken due to a race condition. When one
test ended, the other one started immediately, without waiting for the
resources to finish deleting. This fixes that.


jcantrill:
we can not skip the clo managed test given this is our primary use case for cluster logging

jcantrill:
closing in favor of #290

PR #290, opened 2019-11-13
Enable logforwarding e2e test and fix fluentd forwarding test

jcantrill:
This whole test suite was being skipped by accident. This re-enables
that.

Currently it seems that the admin managed-elasticsearch logstore e2e
tests are broken. So lets skip that and fix it in a subsequent PR.
This way the other tests stay covered.

The fluentd forwarding test was broken due to a race condition. When one
test ended, the other one started immediately, without waiting for the
resources to finish deleting. This fixes that.

PR #291, opened 2019-11-13
Clean up fluent-receiver secret in fluent forwarding e2e test

JAORMX:
This secret was being left out. That's fine in CI, but it leaves
leftovers for non-CI environments. So this PR cleans it up.


JAORMX:
@richm it doesn't wait. But from what I've seen, the deletion of secrets is fairly fast usually. The main problem we encountered was with pods, which is why #290 adds the grace period of zero seconds to the deletion.

PR #294, opened 2019-11-14
move the temp workdir as discussed for 1768042

jcantrill:
Asked to make the dir more obvious via chat in slack and discussed in https://bugzilla.redhat.com/show_bug.cgi?id=1768042#c8

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,8 +11,8 @@ RUN INSTALL_PKGS=" \
     yum install -y $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
     yum clean all && \
-    mkdir /tmp/_working_dir && \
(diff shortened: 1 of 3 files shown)


ewolinetz:
per the BZ, would it make more sense to instead make that error message more meaningful?
`"Operator unable to read local file to get contents: %v", err`

jcantrill:
> per the BZ, would it make more sense to instead make that error message more meaningful?
> `"Operator unable to read local file to get contents: %v", err`

That would probably make it more obvious that its a local message.  Change is made in `util.go`?

ewolinetz:
https://github.com/openshift/cluster-logging-operator/blob/master/pkg/utils/utils.go#L173

jcantrill:
@ewolinetz updated to reflect the error message you suggested

PR #296, opened 2019-11-14
fix up deploy to not depend on bash test lib

jcantrill:
Remove the deploy dependency on test framework.  cc @richm

### DIFF
--- a/hack/testing/utils
+++ b/hack/testing/utils
@@ -5,11 +5,28 @@ indent() {
   sed "s/^/$INDENT/" | sed "s/^${INDENT}\($1\)/${INDENT:0:-2}- \1/"
 }
 
+try_until_text() {
+  local cmd=$1
+  local expected=$2
+  local timeout=$3
+  local interval=${4:-0.2}
+  local now=$(date +%s)
+  local expire=$(($now + $timeout))
+  while [ $now -lt $expire ]; do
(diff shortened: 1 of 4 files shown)


hack/testing/utils
```diff
@@ -5,11 +5,26 @@ indent() {
...
+try_until_text() {
+  local cmd=$1
+  local expected=$2
+  local timeout=$3
+  local now=$(date +%s)
+  local expire=$(($now + $timeout))
+  while [ $now -lt $expire ]; do
+      if [[ "$($cmd)" == "${expected}" ]] ; then
+      exit 0
+    fi
```
richm: indentation
you should also have a `sleep $interval` here

use `return` instead of `exit`


This works a little bit differently than the hack/lib `try_until_text` but I think that's ok as long as people are aware that
- your version will not pass through shell variables in or out of the command under test since it uses a subshell, as opposed to the other version using `eval`
- your version uses an exact match for expected, whereas the other version expects a `grep` pattern

hack/testing/utils
```diff
@@ -175,4 +192,4 @@ function deploy_operator() {
     fi
 
     wait_for_deployment_to_be_ready $namespace $operatorName $timeout
-}
+}}
```
ewolinetz: accidental extra `}` ? I don't see an extra opening bracket...

PR #297, opened 2019-11-14
Use latest CI images from origin-aggregated-logging

JAORMX:
Lets ues the latest CI images from origin-aggregated-logging in the e2e
tests.

### DIFF
--- a/hack/vendor/olm-test-script/e2e-olm.sh
+++ b/hack/vendor/olm-test-script/e2e-olm.sh
@@ -7,6 +7,7 @@ TEST_NAMESPACE=${TEST_NAMESPACE:-olm-test}
 TARGET_NAMESPACE=${TARGET_NAMESPACE:-olm-test}
 CREATE_OPERATORGROUP=${CREATE_OPERATORGROUP:-"true"}
 OPERATOR_IMAGE=${OPERATOR_IMAGE:-""}
+IMAGE_FORMAT=${IMAGE_FORMAT:-""}
 
 # Get the manifests dir where the package and version dir are.
 MANIFEST_DIR=${MANIFEST_DIR:-$(realpath deploy/manifests)}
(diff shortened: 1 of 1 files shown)


JAORMX:
@richm This is the first try... not sure about the labels though. Didn't know where to find them.

hack/vendor/olm-test-script/e2e-olm.sh
```diff
@@ -70,6 +70,13 @@ if [ -n "${OPERATOR_IMAGE:-}" ] ; then
   CSV=$(echo "$CSV" | sed -e "s~image:.*~image: ${OPERATOR_IMAGE}\n~" | indent ApiVersion)
 fi
 
+ci_image_prefix="registry.svc.ci.openshift.org/origin/${VERSION}"
```
richm: while this should work, what the script really should do is to use `${IMAGE_FORMAT}`

you'll also need to update the elasticsearch image and the oauth-proxy image

https://github.com/openshift/origin-aggregated-logging/blob/master/hack/deploy-logging.sh#L171
and
https://github.com/openshift/origin-aggregated-logging/blob/master/hack/deploy-logging.sh#L195

JAORMX:
@richm the elasticsearch operator image is handled elsewhere. Or did you meanthe elasticsearch image in general? cause that seems to be handled different from everything else for some reason: https://github.com/openshift/cluster-logging-operator/blob/master/hack/testing/test-367-logforwarding.sh#L59

richm:
> @richm the elasticsearch operator image is handled elsewhere.

That will also need to use the one that `${IMAGE_FORMAT}` points at.

> Or did you meanthe elasticsearch image in general?

Yes - the elasticsearch image and the oauth-proxy image will have to come from the location that `${IMAGE_FORMAT}` points at.

> cause that seems to be handled different from everything else for some reason: https://github.com/openshift/cluster-logging-operator/blob/master/hack/testing/test-367-logforwarding.sh#L59

@jcantrill ?

JAORMX:
@richm @jcantrill it passed CI. Is there anywhere in the logs where we can see if the correct images were used?

hack/vendor/olm-test-script/e2e-olm.sh
```diff
@@ -70,6 +71,26 @@ if [ -n "${OPERATOR_IMAGE:-}" ] ; then
...
+
+# Use latest origin-aggregated-logging
+if [ -n "${IMAGE_FORMAT:-}" ] ; then
```
jcantrill: We need to figure out a way to pass this in. Ideally this script can be contributed back to be generic for operators and this breaks that: https://github.com/shawn-hurley/olm-test-script

richm:
> @richm @jcantrill it passed CI. Is there anywhere in the logs where we can see if the correct images were used?

If the test dumps the deployment objects or the actual pod objects, you can see from the `image:` field there which image was used.

hack/vendor/olm-test-script/e2e-olm.sh
JAORMX: Isn't this already passed though? it came from /hack/utils

JAORMX:
> > @richm @jcantrill it passed CI. Is there anywhere in the logs where we can see if the correct images were used?
> 
> If the test dumps the deployment objects or the actual pod objects, you can see from the `image:` field there which image was used.

It dumps the pod objects, but they're not picked up since they get deleted before the artifact gathering. So we don't know :confused:

richm:
> > > @richm @jcantrill it passed CI. Is there anywhere in the logs where we can see if the correct images were used?
> > 
> > 
> > If the test dumps the deployment objects or the actual pod objects, you can see from the `image:` field there which image was used.
> 
> It dumps the pod objects, but they're not picked up since they get deleted before the artifact gathering. So we don't know confused

. . . . which is why we need that test script enhancement that starts a background job that watches for new objects/pods, dumps the object contents, does something like `oc logs -f $pod`, etc. so we don't have to add a golang cleanup routine that does this.

hack/vendor/olm-test-script/e2e-olm.sh
richm: `${IMAGE_FORMAT:-}` comes from the CI environment - so if you are trying to test outside of the CI environment, you'll have to use the alternate methods as in https://github.com/openshift/origin-aggregated-logging/blob/master/hack/deploy-logging.sh#L187-L192

hack/vendor/olm-test-script/e2e-olm.sh
```diff
@@ -70,6 +71,26 @@ if [ -n "${OPERATOR_IMAGE:-}" ] ; then
...
+    FLUENTD_IMAGE=$(sed -e "s,\${component},logging-fluentd," <(echo $IMAGE_FORMAT))
+    KIBANA_IMAGE=$(sed -e "s,\${component},logging-kibana5," <(echo $IMAGE_FORMAT))
+    CURATOR_IMAGE=$(sed -e "s,\${component},logging-curator5," <(echo $IMAGE_FORMAT))
+    OAUTH_PROXY_IMAGE=$(sed -e "s,\${component},oauth-proxy," <(echo $IMAGE_FORMAT))
```
richm: Add `ELASTICSEARCH_IMAGE` also

hack/vendor/olm-test-script/e2e-olm.sh
```diff
@@ -70,6 +71,26 @@ if [ -n "${OPERATOR_IMAGE:-}" ] ; then
...
+else
+    FLUENTD_IMAGE="${ci_image_prefix}:logging-fluentd"
+    KIBANA_IMAGE="${ci_image_prefix}:logging-kibana5"
+    CURATOR_IMAGE="${ci_image_prefix}:logging-curator5"
+    OAUTH_PROXY_IMAGE="${ci_image_prefix}:oauth-proxy"
```
richm: Add `ELASTICSEARCH_IMAGE` also

hack/vendor/olm-test-script/e2e-olm.sh
```diff
@@ -70,6 +71,26 @@ if [ -n "${OPERATOR_IMAGE:-}" ] ; then
...
+fi
+
+CSV=$(echo "$CSV" | sed -e "s~quay.io/openshift/origin-logging-fluentd:latest~${FLUENTD_IMAGE}~" | indent apiVersion)
+CSV=$(echo "$CSV" | sed -e "s~quay.io/openshift/origin-logging-kibana5:latest~${KIBANA_IMAGE}~" | indent apiVersion)
+CSV=$(echo "$CSV" | sed -e "s~quay.io/openshift/origin-logging-curator5:latest~${CURATOR_IMAGE}~" | indent apiVersion)
+CSV=$(echo "$CSV" | sed -e "s~quay.io/openshift/origin-oauth-proxy:latest~${OAUTH_PROXY_IMAGE}~" | indent apiVersion)
```
richm: Add `ELASTICSEARCH_IMAGE` also

JAORMX:
This is replaced by https://github.com/openshift/cluster-logging-operator/pull/302

PR #300, opened 2019-11-18
Cleanup status messages as commented in 1770644

jcantrill:
Fix up status messages from:
```status:
  LastUpdated: null
  outputs:
  - lastUpdated: null
    name: fluentd-created-by-user
    state: Accepted
  pipelines:
  - conditions:
    - Message: ""
      Reason: ReservedNameConflict
      Status: "False"
      Type: Name
    lastUpdated: null
    name: pipeline[0]
    state: Dropped
  - conditions:
    - Message: ""
      Reason: ReservedNameConflict
      Status: "False"
      Type: Name
    lastUpdated: null
    name: pipeline[1]
    state: Dropped
  - lastUpdated: null
    name: clo-default-audit-pipeline
    state: Accepted
  sources:
  - logs.audit```

### DIFF
--- a/pkg/controller/forwarding/forwarding_controller.go
+++ b/pkg/controller/forwarding/forwarding_controller.go
@@ -110,7 +110,7 @@ func (r *ReconcileForwarding) Reconcile(request reconcile.Request) (reconcile.Re
 			// Request object not found, could have been deleted after reconcile request.
 			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
 			// Return and don't requeue
-			logger.Debugf("logforwarding-controller error fetching ClusterLogging instance: %v")
+			logger.Debugf("logforwarding-controller error fetching ClusterLogging instance: %v", err)
 			return reconcile.Result{}, err
 		}
 
--- a/pkg/apis/logging/v1alpha1/forwarding_types.go
+++ b/pkg/apis/logging/v1alpha1/forwarding_types.go
@@ -107,7 +107,7 @@ type ForwardingStatus struct {
 	Message string `json:"message,omitempty"`
 
 	// LastUpdated represents the last time that the status was updated.
-	LastUpdated metav1.Time
+	LastUpdated metav1.Time `json:"lastUpdated,omitempty"`
 
 	//LogSources lists the configured log sources
 	LogSources []LogSourceType `json:"sources,omitempty"`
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by ewolinetz** [comment]
@jcantrill can you give an example what the status was changed to in the pr description? Or describe what in the original was changed?

File: pkg/controller/forwarding/forwarding_controller.go [hunk 1]
```diff
@@ -98,9 +98,12 @@ func (r *ReconcileForwarding) Reconcile(request reconcile.Request) (reconcile.Re
 	//check for instancename and then update status
 	var reconcileErr error = nil
 	if instance.Name != singletonName {
-		instance.Status = logforwarding.NewForwardingStatus(logforwarding.LogForwardingStateRejected, logforwarding.LogForwardingReasonName, singletonMessage)
+		instance.Status = logforwarding.NewForwardingStatus()
```
ewolinetz: can we revert this? it makes more sense to me that we would return a status based on something we return and not have to set the fields again after.

pkg/controller/forwarding/forwarding_controller.go
```diff
@@ -98,9 +98,12 @@ func (r *ReconcileForwarding) Reconcile(request reconcile.Request) (reconcile.Re
...
+		instance.Status.State = logforwarding.LogForwardingStateRejected
+		instance.Status.Reason = logforwarding.LogForwardingReasonName
+		instance.Status.Message = singletonMessage
 	} else {
-		instance.Status = logforwarding.NewForwardingStatus(logforwarding.LogForwardingStateAccepted, logforwarding.LogForwardingReasonName, "")
+		instance.Status = logforwarding.NewForwardingStatus()
```
ewolinetz: since it looks like this pattern is used again, maybe we can create a different function that just returns a default StateAccepted status

jcantrill:
> @jcantrill can you give an example what the status was changed to in the pr description? Or describe what in the original was changed?

Ref the original comment.  It has things like `LastUpdated: null` or `Message: ""` which should not exist if they are not set.  This change sets the values for the dates and defines how they should properly serialized

pkg/controller/forwarding/forwarding_controller.go
jcantrill: reverted to be more like a proper constructor

pkg/controller/forwarding/forwarding_controller.go
jcantrill: reverted to be more like a proper constructor

ewolinetz:
>Ref the original comment. It has things like LastUpdated: null or Message: "" which should not exist if they are not set. This change sets the values for the dates and defines how they should properly serialized

Ack, I was asking that the initial commit message be updated to contain that info :)

PR #301, opened 2019-11-18
Bug 1771813: Add shared_key from secret

jcantrill:
Use the shared_key defined in the secret for fluent forward. Ref:

https://bugzilla.redhat.com/show_bug.cgi?id=1771813

### DIFF
--- a/pkg/generators/forwarding/fluentd/output_conf_forward_test.go
+++ b/pkg/generators/forwarding/fluentd/output_conf_forward_test.go
@@ -3,6 +3,7 @@ package fluentd
 import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
 	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
 	test "github.com/openshift/cluster-logging-operator/test"
 )
@@ -10,11 +11,11 @@ import (
 var _ = Describe("Generating fluentd secure forward output store config blocks", func() {
 
 	var (
+		err       error
 		outputs   []logging.OutputSpec
 		generator *ConfigGenerator
 	)
 	BeforeEach(func() {
-		var err error
 		generator, err = NewConfigGenerator()
 		Expect(err).To(BeNil())
 	})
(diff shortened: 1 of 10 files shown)


pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -462,7 +462,7 @@ const forwardTemplate = `{{- define "forward" }}
 	{{ if .Target.Secret }}
 	<security>
 		self_hostname "#{ENV['NODE_NAME']}"
-		shared_key {{ .Name }}
+		shared_key {{ .SecretValue "shared_key" }}
```
richm: This means the cleartext shared_key will be in the fluentd configmap, as opposed to doing something like https://github.com/openshift/origin-aggregated-logging/blob/release-3.11/fluentd/configs.d/input-post-forward-mux.conf#L9
```
  shared_key    "#{File.open('/etc/fluent/muxkeys/shared_key') do |f| f.readline end.rstrip}"
```
so that the key is either in the mounted secret, or in memory
@alanconway thoughts?

jcantrill: Relying on fluent to load the key would actually simply the code and this change I believe as there would be no need to actually load and unencode the secret.

pkg/generators/forwarding/fluentd/output_conf_forward_test.go
```diff
@@ -43,7 +44,7 @@ var _ = Describe("Generating fluentd secure forward output store config blocks",
 	   @type forward
 	   <security>
 	     self_hostname "#{ENV['NODE_NAME']}" 
-	     shared_key secureforward-receiver
+	     shared_key "#{File.open('/var/run/ocp-collector/secrets/my-infra-secret/shared_key') do |f| f.readline end.rstrip}"
```
richm: is it going to be a problem if you want to run the test and you have to use `sudo` to `mkdir -p /var/run/ocp-collector/secrets/my-infra-secret/`?

jcantrill: @richm Who has to `sudo mkdir`? The secret is mounted into the pod via the deployment so the container process should already have permissions to read the content of the file

richm: If this test will always be run inside the container, then correct.

I know we've had issues in the past with tests that required you to do things like `sudo mkdir /usr/share/logging` when running the test on your local machine outside of a container.  Maybe that doesn't apply to this test.

jcantrill: @richm to follow up on our slack convo here this code is only evaluated when the fluent pod starts.  There are no tests outside the container that would eval this.

richm: The answer is - this test assumes it will always be running in a fluentd deployed by cluster logging

PR #302, opened 2019-11-18
deploy logging with latest ci images

richm:


richm:
@bparees is this what you meant about using `ocp` instead of `origin` for the images (but this will only apply when running these tests outside of the CI environment)

bparees:
> @bparees is this what you meant about using ocp instead of origin for the images

yes

PR #304, opened 2019-11-25
Bug 1774864: Enable metrics for collector when LF enabled

jcantrill:
https://bugzilla.redhat.com/show_bug.cgi?id=1774864

### DIFF
--- a/hack/test-e2e.sh
+++ b/hack/test-e2e.sh
@@ -5,11 +5,13 @@ source "${current_dir}/lib/init.sh"
 source "${current_dir}/lib/util/logs.sh"
 
 for test in $( find "${current_dir}/testing" -type f -name 'test-*.sh' | sort); do
-	os::log::info "======================================"
+	os::log::info "==============================================================="
 	os::log::info "running e2e $test "
-	os::log::info "======================================"
+	os::log::info "==============================================================="
 	if "${test}" ; then
+		os::log::info "==========================================================="
 		os::log::info "e2e $test succeeded at $( date )"
+		os::log::info "==========================================================="
 	else
 
 		os::log::error "============= FAILED FAILED ============= "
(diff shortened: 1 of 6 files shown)


pkg/k8shandler/collection.go
```diff
@@ -149,6 +153,51 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateCollectionPriorityCla
 	return nil
 }
 
+func (clusterRequest *ClusterLoggingRequest) createCollectorMetricsRBAC() error {
+	clusterRole, err := clusterRequest.CreateClusterRole(
+		"log-collector-metrics",
```
JAORMX: shouldn't this be created as part of the manifests that deploy the operator?
Giving the operator permission to create clusterroles doesn't seem like a good idea :confused:

jcantrill: learned that OLM supports creation of resources from manifest files. modified accordingly

JAORMX: Thanks! that's way better!

JAORMX:
uhm.... I wonder if the new cluster role is  being picked up at all by the e2e job.

JAORMX:
@jcantrill you also need to add it to the clusterserviceversion https://github.com/openshift/cluster-logging-operator/blob/master/manifests/4.3/cluster-logging.v4.3.0.clusterserviceversion.yaml

JAORMX:
Here specifically https://github.com/openshift/cluster-logging-operator/blob/master/manifests/4.3/cluster-logging.v4.3.0.clusterserviceversion.yaml#L224

jcantrill:
> Here specifically https://github.com/openshift/cluster-logging-operator/blob/master/manifests/4.3/cluster-logging.v4.3.0.clusterserviceversion.yaml#L224

I was told adding to the CSV will not allow creation off the binding to the SA in a different namespace

PR #305, opened 2019-11-26
Bug 1768762: Persisting proxy config and trustbundle CM info

ewolinetz:
Addresses: 
  * https://bugzilla.redhat.com/show_bug.cgi?id=1768762
  * https://bugzilla.redhat.com/show_bug.cgi?id=1766187

TODO: this PR currently just contains clean up for status triggers and unnecessary configmap updates with trustbundle information. Will also add logic to keep the proxy changes in our components.

When we have an event caught by the `proxyconfig_controller` we see that the proxy config info is pushed into the respective components. However, when we do not reconcile an event through there we are using `nil` for the proxy config and trustbundle CA so we then remove it. This is complicated by the fact that status updates can trigger an event being picked up so we were likely removing any proxy changes before the scheduler could process the deployment/daemonset changes.

### DIFF
--- a/pkg/k8shandler/collection.go
+++ b/pkg/k8shandler/collection.go
@@ -6,7 +6,6 @@ import (
 	"strings"
 	"time"
 
-	"github.com/openshift/cluster-logging-operator/pkg/constants"
 	"github.com/openshift/cluster-logging-operator/pkg/logger"
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	"github.com/sirupsen/logrus"
@@ -38,7 +37,7 @@ var (
 var serviceAccountLogCollectorUID types.UID
 
 //CreateOrUpdateCollection component of the cluster
-func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection(proxyConfig *configv1.Proxy, trustedCABundleCM *core.ConfigMap) (err error) {
+func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection(proxyConfig *configv1.Proxy) (err error) {
 	cluster := clusterRequest.cluster
 	collectorConfig := ""
 	collectorConfHash := ""
@@ -81,13 +80,7 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection(proxyConfi
 			return
 		}
 
-		// Create or update cluster proxy trusted CA bundle.
-		err = clusterRequest.createOrUpdateTrustedCABundleConfigMap(constants.FluentdTrustedCAName)
-		if err != nil {
-			return
-		}
-
-		if err = clusterRequest.createOrUpdateFluentdDaemonset(collectorConfHash, proxyConfig, trustedCABundleCM); err != nil {
+		if err = clusterRequest.createOrUpdateFluentdDaemonset(collectorConfHash, proxyConfig); err != nil {
 			return
 		}
 
@@ -98,7 +91,7 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection(proxyConfi
 
 		printUpdateMessage := true
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			if !reflect.DeepEqual(fluentdStatus, cluster.Status.Collection.Logs.FluentdStatus) {
+			if !compareFluentdCollectorStatus(fluentdStatus, cluster.Status.Collection.Logs.FluentdStatus) {
 				if printUpdateMessage {
 					logrus.Info("Updating status of Fluentd")
 					printUpdateMessage = false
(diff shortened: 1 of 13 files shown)


jcantrill:
> When we have an event caught by the `proxyconfig_controller` we see that the proxy config info is pushed into the respective components. However, when we do not reconcile an event through there we are using `nil` for the proxy config and trustbundle CA so we then remove it. This is complicated by the fact that status updates can trigger an event being picked up so we were likely removing any proxy changes before the scheduler could process the deployment/daemonset changes.

Are we properly updating status with the statusupdate client?  Does the statusupdate client trigger events?  I would expect not.

pkg/k8shandler/collection.go
```diff
@@ -98,7 +98,7 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection(proxyConfi
 
 		printUpdateMessage := true
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			if !reflect.DeepEqual(fluentdStatus, cluster.Status.Collection.Logs.FluentdStatus) {
+			if !compareFluentdCollectorStatus(fluentdStatus, cluster.Status.Collection.Logs.FluentdStatus) {
```
jcantrill: Is it necessary to compare status? Why is status just not the current state of the collector?

pkg/k8shandler/curation.go
```diff
@@ -55,7 +55,8 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCuration() (err error
 
 		printUpdateMessage := true
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			if !reflect.DeepEqual(curatorStatus, cluster.Status.Curation.CuratorStatus) {
+
+			if !compareCuratorStatus(curatorStatus, cluster.Status.Curation.CuratorStatus) {
```
jcantrill: Same question here: why do we need to compare status?  Who else is updating it that we should get a conflict?  @bparees should i expect a conflict?

pkg/k8shandler/collection.go
ewolinetz: if conditions have changed from the last time we processed we need to make sure that we are updating it... if nothing has changed we don't want to bother making the Status update call

pkg/k8shandler/curation.go
ewolinetz: its not about preventing conflicts, its about not making unnecessary calls. If the status is exactly the same as what is out there, there is nothing to actually update so we want to spare k8s needing to process our update

ewolinetz:
> Are we properly updating status with the statusupdate client? Does the statusupdate client trigger events? I would expect not.

I didn't think it was supposed to, but through my testing I've noticed that it is... I believe this is as expected and how some teams get around not requeuing their requests. We still want to requeue because if the status doesn't change then we would never process anything again (it could not change if the cluster is all in a good state)

ewolinetz:
> Overall I like the cleanup and something that likely needs to be cherry-picked back to 4.3 but seems like it should probably be a separate PR unless you believe this is the only way to fix the referenced issue

I think some things may be separate from this fix... just things that made it easier to troubleshoot what was actually going on because of all the excess noise in the operator logs. I can split that out to a separate commit on this PR so that its easier to cherry-pick...

ewolinetz:
So far with these changes (and a hacked proxy choice) I see the env vars, and volume mounts added and persisted... I need to update our unit tests I believe and I'm going to also attempt to install my changes on a cluster with a configured proxy.

pkg/k8shandler/visualization.go
```diff
@@ -158,9 +205,27 @@ func (clusterRequest *ClusterLoggingRequest) removeKibana() (err error) {
...
 }
 
-func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaDeployment(proxyConfig *configv1.Proxy, trustedCABundleCM *v1.ConfigMap) (err error) {
+func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaDeployment(proxyConfig *configv1.Proxy) (err error) {
+
+	kibanaTrustBundle := &v1.ConfigMap{}
+
+	// Create cluster proxy trusted CA bundle.
+	if proxyConfig != nil {
+		err = clusterRequest.createOrUpdateTrustedCABundleConfigMap(constants.KibanaTrustedCAName)
```
bparees: why is this being done repeatedly?  we should be able to just define this CM as a resource in the CSV and be done with it. Same for the one for fluentd.

bparees: sorry i take that back... CVO allows create-only resources but i dunno if OLM does.  So if you can't tell OLM to just create the CM but not continue to reconcile/stomp it, then you need to create it yourself.  Still not sure you need to actively reconcile it (there should be no changes to apply)

ewolinetz: so to have OLM create the CMs, we would need to define them in the manifest, correct?
It may also make sense to move the creation of these configmaps into the proxy controller... I was keeping the call in this file because that's where it was before and the call itself should be idempotent.

bparees: yeah, as i said i don't think you can move them into the manifest after all because OLM is going to keep stomping the content of the configmap after the CAs get injected, which would be bad.  CVO has the ability to create a resource for you but then never touch it again, I don't think OLM can do that.

i think you can leave this here for now.

ewolinetz:
> I'm going to also attempt to install my changes on a cluster with a configured proxy.

Verified that we see logs flow to ES from fluentd -- I'm currently missing the oauth-proxy image on my cluster so Kibana is in an `ImagePullBackOff` status, I see the configurations on the proxy container but I need to update this PR further for unit test correction (possibly) so i'll verify the kibana bug as well first before removing the hold on this pr.

PR #309, opened 2019-11-27
Bug 1773847: Fixed fluentd sed error

igor-karpukhin:
Fixed the error when `run.sh` script for Fluentd tried to perform `sed` operations on a read-only FS:
```
sed: couldn't open temporary file /etc/fluent/configs.d/user/sedv62wWH: Read-only file system
```

### DIFF
--- a/files/fluentd/run.sh
+++ b/files/fluentd/run.sh
@@ -5,17 +5,6 @@ CFG_DIR=/etc/fluent/configs.d
 ENABLE_PROMETHEUS_ENDPOINT=${ENABLE_PROMETHEUS_ENDPOINT:-"true"}
 OCP_OPERATIONS_PROJECTS=${OCP_OPERATIONS_PROJECTS:-"default openshift openshift- kube-"}
 LOGGING_FILE_PATH=${LOGGING_FILE_PATH:-"/var/log/fluentd/fluentd.log"}
-OCP_FLUENTD_TAGS=""
-for p in ${OCP_OPERATIONS_PROJECTS}; do
-    if [[ "${p}" == *- ]] ; then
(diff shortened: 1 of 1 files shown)


pkg/utils/utils.go
```diff
@@ -357,30 +357,30 @@ func SetProxyEnvVars(proxyConfig *configv1.Proxy) []v1.EnvVar {
 			Name:  "HTTPS_PROXY",
 			Value: proxyConfig.Status.HTTPSProxy,
 		},
-		v1.EnvVar{
-			Name:  "https_proxy",
-			Value: proxyConfig.Status.HTTPSProxy,
-		})
+			v1.EnvVar{
```
ewolinetz: did `gofmt` do this?

ewolinetz:
@jcantrill @richm what are the repercussions from removing the snippet that this PR proposes? does the log forwarding cover this now?

pkg/utils/utils.go
igor-karpukhin: @ewolinetz yes, this is done by `gofmt`, `Makefile:94`

richm: The formatting does not look correct

richm:
> @jcantrill @richm what are the repercussions from removing the snippet that this PR proposes? does the log forwarding cover this now?

Not sure what will be the CI test implications, but the general approach is correct - when using log forwarding, log forwarding should be generating all of the config, correctly, and we should definitely not be modifying the config files in the image at runtime.

igor-karpukhin:
@jcantrill @ewolinetz @richm could you please review? E2E tests passed

PR #310, opened 2019-12-02
Bug 1776636: Provide external log link for pod

lukas-vlcek:
<s>Instruct Kibana UI to use the .all index-pattern when displaying
logs from the pod.</s>

### DIFF
--- a/pkg/k8shandler/visualization.go
+++ b/pkg/k8shandler/visualization.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/client-go/util/retry"
 
+	consolev1 "github.com/openshift/api/console/v1"
 	configv1 "github.com/openshift/api/config/v1"
 	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
 	apps "k8s.io/api/apps/v1"
@@ -23,8 +24,11 @@ import (
 )
 
 const (
-	kibanaServiceAccountName     = "kibana"
-	kibanaOAuthRedirectReference = "{\"kind\":\"OAuthRedirectReference\",\"apiVersion\":\"v1\",\"reference\":{\"kind\":\"Route\",\"name\":\"kibana\"}}"
+	kibanaServiceAccountName       = "kibana"
+	kibanaOAuthRedirectReference   = "{\"kind\":\"OAuthRedirectReference\",\"apiVersion\":\"v1\",\"reference\":{\"kind\":\"Route\",\"name\":\"kibana\"}}"
+	// The following strings are turned into JavaScript RegExps. Online tool to test them: https://regex101.com/
+	nodesAndContainersNamespaceFilter = "^(openshift-.*|kube-.*|openshift$|kube$|default$)"
+	appsNamespaceFilter               = "^((?!" + nodesAndContainersNamespaceFilter + ").)*$" // ^((?!^(openshift-.*|kube-.*|openshift$|kube$|default$)).)*$
 )
 
 var (
@@ -436,12 +440,14 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaConsoleExternal
 		return err
 	}
 
-	consoleExternalLogLink := NewConsoleExternalLogLink(
-		"kibana",
+	consoleExternalLogLinkForNodesAndContainers := NewConsoleExternalLogLink(
+		"kibana-admin",
 		cluster.Namespace,
 		"Show in Kibana",
 		strings.Join([]string{kibanaURL,
-			"/app/kibana#/discover?_g=(time:(from:now-1w,mode:relative,to:now))&_a=(columns:!(kubernetes.container_name,message),query:(query_string:(analyze_wildcard:!t,query:'",
+			"/app/kibana#/discover?_g=(time:(from:now-1w,mode:relative,to:now))&_a=(columns:!(kubernetes.container_name,message),index:",
+			"'.operations.*'",
+			",query:(query_string:(analyze_wildcard:!t,query:'",
 			strings.Join([]string{
 				"kubernetes.pod_name:\"${resourceName}\"",
 				"kubernetes.namespace_name:\"${resourceNamespace}\"",
@@ -449,18 +455,39 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaConsoleExternal
 			}, " AND "),
 			"')),sort:!('@timestamp',desc))"},
 			""),
+		nodesAndContainersNamespaceFilter,
 	)
 
-	utils.AddOwnerRefToObject(consoleExternalLogLink, utils.AsOwner(cluster))
+	consoleExternalLogLinkForApps := NewConsoleExternalLogLink(
+		"kibana",
+		cluster.Namespace,
+		"Show in Kibana",
+		strings.Join([]string{kibanaURL,
+			"/app/kibana#/discover?_g=(time:(from:now-1w,mode:relative,to:now))&_a=(columns:!(kubernetes.container_name,message),index:",
+			"'project.${resourceNamespace}.${resourceNamespaceUID}.*'",
+			",query:(query_string:(analyze_wildcard:!t,query:'",
+			strings.Join([]string{
+				"kubernetes.pod_name:\"${resourceName}\"",
+				"kubernetes.namespace_name:\"${resourceNamespace}\"",
+				"kubernetes.container_name.raw:\"${containerName}\"",
+			}, " AND "),
+			"')),sort:!('@timestamp',desc))"},
+			""),
+		appsNamespaceFilter,
+	)
+	links := []*consolev1.ConsoleExternalLogLink{consoleExternalLogLinkForNodesAndContainers, consoleExternalLogLinkForApps}
 
-	// In case the object already exists we delete it first
-	if err = clusterRequest.RemoveConsoleExternalLogLink("kibana"); err != nil {
-		return
-	}
-
-	err = clusterRequest.Create(consoleExternalLogLink)
-	if err != nil && !errors.IsAlreadyExists(err) {
-		return fmt.Errorf("Failure creating Kibana console external log link for %q: %v", cluster.Name, err)
+	for _, externalLink := range links {
+		utils.AddOwnerRefToObject(externalLink, utils.AsOwner(cluster))
+		// In case the object already exists we delete it first
+		if err = clusterRequest.RemoveConsoleExternalLogLink(externalLink.ObjectMeta.Name); err != nil {
+			return
+		}
+		err = clusterRequest.Create(externalLink)
+		if err != nil && !errors.IsAlreadyExists(err) {
+			return fmt.Errorf("Failure creating Kibana console external %q log link for %q: %v",
+				externalLink.ObjectMeta.Name, cluster.Name, err)
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


jcantrill:
What purpose is is this change?  This alias is only available to users who are cluster admins

pkg/k8shandler/visualization.go
```diff
@@ -372,7 +372,7 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaConsoleExternal
 		cluster.Namespace,
 		"Show in Kibana",
 		strings.Join([]string{kibanaURL,
-			"/app/kibana#/discover?_g=(time:(from:now-1w,mode:relative,to:now))&_a=(columns:!(kubernetes.container_name,message),query:(query_string:(analyze_wildcard:!t,query:'",
+			"/app/kibana#/discover?_g=(time:(from:now-1w,mode:relative,to:now))&_a=(columns:!(kubernetes.container_name,message),index:.all,query:(query_string:(analyze_wildcard:!t,query:'",
```
richm: Does this actually work for non-admin users?

pkg/k8shandler/visualization.go
```diff
@@ -432,32 +436,75 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaConsoleExternal
 		return err
 	}
 
-	consoleExternalLogLink := NewConsoleExternalLogLink(
-		"kibana",
+	consoleExternalLogLinkForAdmin := NewConsoleExternalLogLink(
+		"kibana-admin",
```
ewolinetz: have we coordinated with the console team regarding the object name change yet?

pkg/k8shandler/visualization.go
```diff
@@ -432,32 +436,75 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaConsoleExternal
...
+		cluster.Namespace,
+		"Show in Kibana",
+		strings.Join([]string{kibanaURL,
+			"/app/kibana#/discover?_g=(time:(from:now-1w,mode:relative,to:now))&_a=(columns:!(kubernetes.container_name,message),index:",
+			".operations.*",
```
ewolinetz: not all records in the operations index will have a pod, namespace, and container name... it may be logs that came from the node. do we really want to filter those out immediately?

pkg/k8shandler/visualization.go
jcantrill: @spadgett does the name matter?  its applied using a regex for the namespace correct?

pkg/k8shandler/visualization.go
```diff
@@ -432,32 +436,75 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaConsoleExternal
...
+		}
 	}
+	//utils.AddOwnerRefToObject(consoleExternalLogLinkForAdmin, utils.AsOwner(cluster))
+	//utils.AddOwnerRefToObject(consoleExternalLogLinkForNonAdmin, utils.AsOwner(cluster))
 
-	err = clusterRequest.Create(consoleExternalLogLink)
-	if err != nil && !errors.IsAlreadyExists(err) {
-		return fmt.Errorf("Failure creating Kibana console external log link for %q: %v", cluster.Name, err)
-	}
+	// In case the object already exists we delete it first
```
jcantrill: can we remove this if it is not required

pkg/k8shandler/visualization.go
spadgett: Right, console doesn't look at name. We list all console log links and apply them based on the namespace regex.

pkg/k8shandler/visualization.go
lukas-vlcek: I assume this link is displayed (thus also used) only when you list pods in web console and click logs tab. So in this context all logs should have all these filters applied.
Shall we address other user cases as well?

pkg/k8shandler/visualization.go
lukas-vlcek: Yes, we can. And I will remove it. The only reason I introduced this logic is to allow CLO re-create the external link objects on restart (aka `make run`) so that all code changes for the external links are reflected and can be immediately tested.
The only other alternative would be to delete relevant objects in K8s manually before doing `make run`, or is there other way?

ewolinetz: it is still required for normal operator reconciling... when we initially created this logic it was noticed that we could not do an `update` because some fields that we would possibly change were immutable so we need to do a `delete` and then a `create` every time.

pkg/k8shandler/visualization.go
ewolinetz: `kibana-non-admin` may be confusing for users... what about `kibana` and `kibana-admin` ?

lukas-vlcek: @ewolinetz I changed it to `kibana` and `kibana-admin`

lukas-vlcek:
PTAL again
Depends on https://github.com/openshift/console/pull/3745

pkg/k8shandler/visualization.go
```diff
@@ -432,31 +436,55 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaConsoleExternal
...
-		"kibana",
+	consoleExternalLogLinkForAdmin := NewConsoleExternalLogLink(
+		"kibana-admin",
 		cluster.Namespace,
 		"Show in Kibana",
 		strings.Join([]string{kibanaURL,
-			"/app/kibana#/discover?_g=(time:(from:now-1w,mode:relative,to:now))&_a=(columns:!(kubernetes.container_name,message),query:(query_string:(analyze_wildcard:!t,query:'",
+			"/app/kibana#/discover?_g=(time:(from:now-1w,mode:relative,to:now))&_a=(columns:!(kubernetes.container_name,message),index:",
+			"'.operations.*'",
+			",query:(query_string:(analyze_wildcard:!t,query:'",
```
ewolinetz: I'm still unsure that we want to have this query for the `.operations.*` index pattern.
I guess i'd like to know what we seek to accomplish by having an external link for admins -- is it to see specific container logs? see all container logs? see all infra logs that include containers and node logs?

lukas-vlcek: My understanding is that when this external log link is clicked then all the context is available. This means all project, pod and container are selected. If we miss any of these then the list of logs seen in the console UI can be different from what will be filtered in Kibana (for example if there are more containers, like app and proxy side car).

ewolinetz: I understand that, I'm wondering what our intent is with the admin link specifically. I can see it providing one of a few things and wanted to have that clarified:
* see specific container logs
* see all container logs
* see all infra logs that include containers and node logs

lukas-vlcek: > I believe the console will present one link or the other based upon a namespace regex

That is correct, I tested this and I can confirm that.

But the important part is in which context (i.e. **WHEN**) the link is displayed. I am aware of only a single case and that is when you navigate to `Pods > Pod Details` section and select the `Logs` tab. I am not aware of any other situation - @spadgett Is there any other?

In this context user have selected all of `Namespace`, `Pod` and `Container`.
And more specifically, this context already presents user with log stream window. My understanding is that the link "Show in Kibana" should present user with exactly the same set of logs:

- **For non-admin** user this is clear. We need to pass to Kibana `namespace`, `pod` and `container` for filtering and use index pattern for given project (this is what the PR does now).

- **For admin** we can do the same except the index pattern will be the `project.*` (hardcoded).

Does it make sense?
WDYT?

lukas-vlcek: In other words, when logged in as **the cluster admin** we see the following log stream in console:
Screenshot 2020-02-05 at 15 41 32

And upon clicking the "Show in Kibana" link I would expect that user (still cluster admin) should see the same logs:
Screenshot 2020-02-05 at 15 41 47
(logs need to be reverse ordered by time ^^)

jcantrill: > 
> * **For non-admin** user this is clear. We need to pass to Kibana `namespace`, `pod` and `container` for filtering and use index pattern for given project (this is what the PR does now).

Correct

> * **For admin** we can do the same except the index pattern will be the `project.*` (hardcoded).
> 

This depends.  If I am coming from an 'infra' namespace then it would be '.operations'  If I am coming from a non-infra namespace it would be 'project.*'

lukas-vlcek: @jcantrill 
> If I am coming from an 'infra' namespace then it would be '.operations'

What would be the correct approach to learn if user comes from "infra" namespace?
How about the following:

```
$ oc get projects | grep infra
openshift-infra                                                        Active
openshift-kni-infra                                                    Active
openshift-openstack-infra                                              Active
openshift-ovirt-infra                                                  Active
```

jcantrill: > @jcantrill
> 
> > If I am coming from an 'infra' namespace then it would be '.operations'
> 
> What would be the correct approach to learn if user comes from "infra" namespace?
> How about the following:
> 
> ```
> $ oc get projects | grep infra
> openshift-infra                                                        Active
> openshift-kni-infra                                                    Active
> openshift-openstack-infra                                              Active
> openshift-ovirt-infra                                                  Active
> ```

Infra means: any namespace that is openshift*, kube*, default

pkg/k8shandler/visualization.go
```diff
@@ -23,8 +24,12 @@ import (
 )
 
 const (
-	kibanaServiceAccountName     = "kibana"
-	kibanaOAuthRedirectReference = "{\"kind\":\"OAuthRedirectReference\",\"apiVersion\":\"v1\",\"reference\":{\"kind\":\"Route\",\"name\":\"kibana\"}}"
+	annotationOauthSecretUpdatedAt = "logging.openshift.io/oauthSecretUpdatedAt"
+	kibanaServiceAccountName       = "kibana"
+	kibanaOAuthRedirectReference   = "{\"kind\":\"OAuthRedirectReference\",\"apiVersion\":\"v1\",\"reference\":{\"kind\":\"Route\",\"name\":\"kibana\"}}"
+	// The following strings are turned into JavaScript RegExps. Online tool to test them: https://regex101.com/
+	adminNamespaceFilter           = "^(openshift-.*|kube-.*|default$)"
```
jcantrill: need to remove the dash in order to also capture 'openshft' and 'kube'

lukas-vlcek:
@jcantrill Fixed the naming nit as well. I think this PR is ready to go now. PTAL

pkg/k8shandler/visualization.go
```diff
@@ -23,8 +24,12 @@ import (
 )
 
 const (
-	kibanaServiceAccountName     = "kibana"
-	kibanaOAuthRedirectReference = "{\"kind\":\"OAuthRedirectReference\",\"apiVersion\":\"v1\",\"reference\":{\"kind\":\"Route\",\"name\":\"kibana\"}}"
+	annotationOauthSecretUpdatedAt = "logging.openshift.io/oauthSecretUpdatedAt"
```
ewolinetz: what are we going to use this for?
does this have overlap with https://github.com/openshift/cluster-logging-operator/pull/351 ?

jcantrill: Why is this in this change?  We added "oauth" bits in a separate PR  so the oauthclient would be dynamically generated.  I wold not expect any "oauthclient" parts to change in order to satisfy this bz

lukas-vlcek: This looks like a mistake caused by resolving a merge conflict. Good catch! PR should be updated now.

PR #312, opened 2019-12-02
fix lf timing issue

jcantrill:
Fix LogForwarding timing issue and other minor test fixes

cc @ewolinetz @igor-karpukhin  I discovered an issue of sysctl setting not being applied by the node-tuning-operator but I'm hoping its not a larger issue and just the cluster I stood up.  Bumping the timing here as previously only gave ES 2minutes to come up

We should run this a couple of times if it works

### DIFF
--- a/hack/testing/assertions
+++ b/hack/testing/assertions
@@ -1,5 +1,5 @@
 #!/bin/bash
-source "$(dirname $0)/utils"
+source "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )/utils"
 assert_resources_exist(){
 	# verify deployments -- kibana, curator
 	try_until_success "oc -n $NAMESPACE get deployment kibana" ${TIMEOUT_MIN}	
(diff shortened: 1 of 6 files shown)


hack/testing/assertions
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-source "$(dirname $0)/utils"
+source "$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )/utils"
```
ewolinetz: is it really necessary to `cd` as part of this `source` statement?

richm: in general, yes, if you want the "real" absolute path

igor-karpukhin:
Log forward test still fails

jcantrill:
all tests blocked by https://github.com/openshift/cluster-node-tuning-operator/pull/96

jcantrill:
In master at least, there was  a change in the node-tuning-operator where the sysctl changes required by ES are not being applied to the node.  waiting on the referenced PR to merge in order to unblock

PR #313, opened 2019-12-03
fix lf timing issue

jcantrill:
4.3 backport of https://github.com/openshift/cluster-logging-operator/pull/312  which 4.4 is blocked by https://github.com/openshift/cluster-node-tuning-operator/pull/96

### DIFF
--- a/hack/testing/utils
+++ b/hack/testing/utils
@@ -6,9 +6,32 @@ minute=$(( 60 * second ))
 
 LOGGING_NS="openshift-logging"
 
+log::info(){
+  echo " [INFO] $@"
+}
+
+log::error(){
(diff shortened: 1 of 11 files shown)


hack/testing/utils
```diff
@@ -6,10 +6,33 @@ minute=$(( 60 * second ))
 
 LOGGING_NS="openshift-logging"
 
+log::info(){
+  echo " [INFO] $1"
```
richm: use `$@` instead of `$1` - also below with `log::error`

jcantrill:
@bparees please approve since this is to try and address deficiencies in the testing code

jcantrill:
merged to resolve some test code issues

PR #316, opened 2019-12-06
Allow logforwarding tests to cleanup from go invocation

jcantrill:
pick up changes from 4.3 branch

ref: #313


jcantrill:
> /lgtm, but this is some kind of logic spread between go and bash. Maybe in the future, we could introduce helpers in golang to get config and logs from a specific resource.

That would be prefered but that means recreating a ton of code that already exists in oc

jcantrill:
Merging manually because of https://github.com/openshift/cluster-logging-operator/pull/316#pullrequestreview-328348480 to unblock other PRs

PR #318, opened 2019-12-07
Allow logforwarding tests to cleanup from go invocation

jcantrill:
More updates to make LF testing more reliable

Ref 4.4: https://github.com/openshift/cluster-logging-operator/pull/316


bparees:
no BZ needed, improving test resilience.

/lgtm

PR #323, opened 2019-12-12
Bug 1782740: Addressing LF causing cert rotation

ewolinetz:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1782740

Able to recreate the master-certs secret and elasticsearch secret being recreated by deploying CLO, creating a log forwarding object, then deleting the CLO pod. No longer observe that with this PR.

### DIFF
--- a/pkg/controller/forwarding/forwarding_controller.go
+++ b/pkg/controller/forwarding/forwarding_controller.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"time"
 
-	loggingv1 "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
 	logforwarding "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
 	"github.com/openshift/cluster-logging-operator/pkg/constants"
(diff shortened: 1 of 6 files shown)


ewolinetz:
issue seems to still be present

pkg/controller/forwarding/forwarding_controller.go
```diff
@@ -89,40 +89,54 @@ func (r *ReconcileForwarding) Reconcile(request reconcile.Request) (reconcile.Re
 			// Return and don't requeue
 			return reconcile.Result{}, nil
 		}
-		// Error reading the object - requeue the request.
-		return reconcile.Result{}, err
+		// Error reading the object -- requeue
+		return reconcileResult, err
```
richm: So when you want the request to be requeued, you return `reconcileResult`, and when you do not want the result to be requeued, you return `reconcile.Result{}`?

ewolinetz: yes, `reconcileResult` has a field `RequeueAfter` that puts this back on the event queue

ewolinetz:
Still want to do more regression testing before removing the `[WIP]`

PR #332, opened 2020-01-15
replace FROM for local builds

jcantrill:
This PR modifies the base image:
* The ART process replaces FROM with what's required for a release
* The ci pipeline replaces FROM based on what's defined in config: https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-logging-operator/openshift-cluster-logging-operator-master.yaml#L11-L13

This change will allow us to continue to build locally

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ WORKDIR /go/src/github.com/openshift/cluster-logging-operator
 COPY . .
 RUN make
 
-FROM registry.svc.ci.openshift.org/openshift/origin-v4.0:base
+FROM centos:centos7
 ARG CSV=4.4
 RUN INSTALL_PKGS=" \
       openssl \


jcantrill:
cc @richm Do you know if this will break origin-aggregated-logging?

ewolinetz:
what will be the process then for building our released images? wouldn't we want to use the current image then?

jcantrill:
> what will be the process then for building our released images? wouldn't we want to use the current image then?

You would have to substitute the FROM statement with the image used by ART and ensure you have access to the dependent rpms by supplying the correct repo files.  I will provide you with the document that describes how to do this.  Per my original comment, this is already being down for us by ART and ci-operator.  Technically you could pull the actual image if you needed from the private registry. 

As is, the current workflow is broken for local development.  This PR resolves that and CI and ART solves downstream

richm:
> cc @richm Do you know if this will break origin-aggregated-logging?

No - origin-aggregated-logging CI tests will continue to use whatever is the latest cluster-logging-operator image available in the CI registry

richm:
> As is, the current workflow is broken for local development. This PR resolves that and CI and ART solves downstream

How is it broken?  developers should be able to pull registry.svc.ci.openshift.org/openshift/origin-v4.0:base and should be able to set up the yum repos

jcantrill:
> > As is, the current workflow is broken for local development. This PR resolves that and CI and ART solves downstream
> 
> How is it broken? developers should be able to pull registry.svc.ci.openshift.org/openshift/origin-v4.0:base and should be able to set up the yum repos

It's broken in that there are no clear instructions on making it work without jumping through slack conversations and mojo docs to build the image.  I still was unable to build even after being pointed to a document that advised me how to get the repos.  Do we really need the actual base image if:
* Our local build is repeatable as is and produces a usable image
* Downstream processes do substitution in two different ways regardless

richm:
tl;dr do what you like, and it's already merged, but since you asked:

> > > As is, the current workflow is broken for local development. This PR resolves that and CI and ART solves downstream
> > 
> > 
> > How is it broken? developers should be able to pull registry.svc.ci.openshift.org/openshift/origin-v4.0:base and should be able to set up the yum repos
> 
> It's broken in that there are no clear instructions on making it work without jumping through slack conversations and mojo docs to build the image. I still was unable to build even after being pointed to a document that advised me how to get the repos.

I would suggest to keep trying to figure it out.  I'm not quite sure why it doesn't work anymore, given that until recently oal hack/build-images.sh worked, and so did clo and eo `make image`.  AFAIK it is a matter of getting the right pull secrets/credentials for the ubi registry, getting the right keys for the yum repo, and setting up the mount points for `imagebuilder` to mount the yum repo config and keys.  Yes, it is a big PITA that downstream/CI was changed which makes upstream developers jump through flaming hoops just to do something as simple as building an image, but it was done with a specific purpose/goal in mind.

> Do we really need the actual base image if:
> 
>     * Our local build is repeatable as is and produces a usable image
> 
>     * Downstream processes do substitution in two different ways regardless

You will always run the risk that something in the UBI image or yum repo is different than what's in centos7, and you will find that problem later rather than sooner.  That is, you may merrily be able to develop upstream for weeks with no problem, only then to find out from QE that something is broken because there is some difference between upstream centos7 and downstream ubi/repos.

PR #333, opened 2020-01-15
Bug 1791336: Fix fluent metrics config to support ipv6

jcantrill:
This PR resolves: https://bugzilla.redhat.com/show_bug.cgi?id=1791336


pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -108,6 +108,7 @@ var _ = Describe("Generating fluentd config", func() {
 			## ordered so that syslog always runs last...
 			<source>
 				@type prometheus
+                bind '::'
```
ewolinetz: formatting

PR #340, opened 2020-01-22
Use centos instead of RHEL to avoid subscription errors.

alanconway:
Backport c43a731b from 4.4 "replace FROM for local builds".
Ref https://github.com/openshift/cluster-logging-operator/pull/332
Use centos instead of RHEL to avoid subscription errors in `make deploy`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ WORKDIR /go/src/github.com/openshift/cluster-logging-operator
 COPY . .
 RUN make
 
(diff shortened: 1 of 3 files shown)


jcantrill:
@bparees I dont really need a bz for this dev change do I?

bparees:
> @bparees I dont really need a bz for this dev change do I?

1) it's not exactly a dev-only change.  This is the dockerfile our release content is built from also, right? (but the FROM line is always replaced)

2) we can override the BZ requirement but this can't merge until after 4.3GAs and we reopen the branch for z-stream work.

PR #341, opened 2020-01-23
Forwarding to external syslog, based on current fluentd plugin

ahadas:
This PR adds the ability to define a pipeline that outputs data to an external syslog server and uses the existing remote-syslog plugin.

An example of specification of syslog output within the `LogForwarding` CR:
```yaml
spec:
  outputs:
  - name: syslog
    type: syslog
    endpoint: "10.46.9.101:515"
    insecure: true
  pipelines:
  - name: logsapp
    inputSource: logs.audit
    outputRefs:
    - syslog
```

This will result in the following configuration section for fluentd:

```xml
 <label @SYSLOG>
        <match **>
        @type copy
        
                <store>
                @type syslog_buffered
                @id syslog
                remote_syslog 10.46.9.101
                port 515
                hostname ${hostname}
                facility user
                severity debug
                remove_tag_prefix
        </store>
</match>
</label>
```

Alternatively, the transport protocol can be set to udp by specifying the protocol within the `endpoint`:

```yaml
spec:
  outputs:
  - name: syslog
    type: syslog
    endpoint: "udp://10.46.9.101:515"
    insecure: true
  pipelines:
  - name: logsapp
    inputSource: logs.audit
    outputRefs:
    - syslog
```

### DIFF
--- a/manifests/4.5/logforwardings.crd.yaml
+++ b/manifests/4.5/logforwardings.crd.yaml
@@ -29,6 +29,7 @@ spec:
                     enum:
                       - elasticsearch
                       - forward
+                      - syslog
                   name: 
                     description: The name of the output
                     type: string
--- a/pkg/apis/logging/v1alpha1/forwarding_types.go
+++ b/pkg/apis/logging/v1alpha1/forwarding_types.go
@@ -74,6 +74,9 @@ const (
 
 	//OutputTypeForward configures the pipeline to send messages via Fluent's secure forward
 	OutputTypeForward OutputType = "forward"
+
+	//OutputTypeSyslog configures pipeline to send messages to an external syslog server through docebo/fluent-plugin-remote-syslog
+	OutputTypeSyslog OutputType = "syslog"
 )
 
 //LogForwardingReason The reason for the current state
--- a/pkg/generators/forwarding/fluentd/output_conf_syslog_test.go
+++ b/pkg/generators/forwarding/fluentd/output_conf_syslog_test.go
@@ -1,0 +1,108 @@
+package fluentd
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
+	test "github.com/openshift/cluster-logging-operator/test"
+)
+
+var _ = Describe("Generating external syslog server output store config blocks", func() {
+
+	var (
+		err       error
+		outputs   []logging.OutputSpec
+		generator *ConfigGenerator
+	)
+	BeforeEach(func() {
+		generator, err = NewConfigGenerator(false, false)
+		Expect(err).To(BeNil())
+	})
+
+	Context("based on syslog plugin", func() {
+		tcpConf := `<label @SYSLOG_RECEIVER>
+		<match **>
+		@type copy
+		<store>
+			@type syslog_buffered
+			@id syslog_receiver
+			remote_syslog sl.svc.messaging.cluster.local
+			port 9654
+			hostname ${hostname}
+			facility user
+			severity debug
+		</store>
+		</match>
+	</label>`
+
+		udpConf := `<label @SYSLOG_RECEIVER>
+		<match **>
+		@type copy
+		<store>
+			@type syslog
+			@id syslog_receiver
+			remote_syslog sl.svc.messaging.cluster.local
+			port 9654
+			hostname ${hostname}
+			facility user
+			severity debug
+		</store>
+		</match>
+	</label>`
+
+		Context("for protocol-less endpoint", func() {
+			BeforeEach(func() {
+				outputs = []logging.OutputSpec{
+					{
+						Type:     logging.OutputTypeSyslog,
+						Name:     "syslog-receiver",
+						Endpoint: "sl.svc.messaging.cluster.local:9654",
+					},
+				}
+			})
+			It("should produce well formed output label config", func() {
+				results, err := generator.generateOutputLabelBlocks(outputs)
+				Expect(err).To(BeNil())
+				Expect(len(results)).To(Equal(1))
+				test.Expect(results[0]).ToEqual(tcpConf)
+			})
+		})
+
+		Context("for tcp endpoint", func() {
+			BeforeEach(func() {
+				outputs = []logging.OutputSpec{
+					{
+						Type:     logging.OutputTypeSyslog,
+						Name:     "syslog-receiver",
+						Endpoint: "tcp://sl.svc.messaging.cluster.local:9654",
+					},
+				}
+			})
+			It("should produce well formed output label config", func() {
+				results, err := generator.generateOutputLabelBlocks(outputs)
+				Expect(err).To(BeNil())
+				Expect(len(results)).To(Equal(1))
+				test.Expect(results[0]).ToEqual(tcpConf)
+			})
+		})
+
+		Context("for udp endpoint", func() {
+			BeforeEach(func() {
+				outputs = []logging.OutputSpec{
+					{
+						Type:     logging.OutputTypeSyslog,
+						Name:     "syslog-receiver",
+						Endpoint: "udp://sl.svc.messaging.cluster.local:9654",
+					},
+				}
+			})
+			It("should produce well formed output label config", func() {
+				results, err := generator.generateOutputLabelBlocks(outputs)
+				Expect(err).To(BeNil())
+				Expect(len(results)).To(Equal(1))
+				test.Expect(results[0]).ToEqual(udpConf)
+			})
+		})
+	})
+})
--- a/pkg/generators/forwarding/fluentd/templates.go
+++ b/pkg/generators/forwarding/fluentd/templates.go
@@ -12,8 +12,10 @@ var templateRegistry = []string{
 	sourceToPipelineCopyTemplate,
 	outputLabelConfTemplate,
 	outputLabelConfNocopyTemplate,
+	outputLabelConfNoretryTemplate,
 	storeElasticsearchTemplate,
 	forwardTemplate,
+	storeSyslogTemplate,
 }
 
 const fluentConfTemplate = `{{- define "fluentConf" }}
@@ -484,6 +486,15 @@ const outputLabelConfNocopyTemplate = `{{- define "outputLabelConfNoCopy" }}
 </label>
 {{- end}}`
 
+const outputLabelConfNoretryTemplate = `{{- define "outputLabelConfNoRetry" }}
+<label {{.LabelName}}>
+	<match **>
+		@type copy
+{{include .StoreTemplate . "" | indent 4}}
+	</match>
+</label>
+{{- end}}`
+
 const forwardTemplate = `{{- define "forward" }}
 	# https://docs.fluentd.org/v1.0/articles/in_forward
 	@type forward
@@ -571,5 +582,17 @@ const storeElasticsearchTemplate = `{{- define "storeElasticsearch" }}
 		chunk_limit_size "#{ENV['BUFFER_SIZE_LIMIT'] || '8m' }"
 		overflow_action "#{ENV['BUFFER_QUEUE_FULL_ACTION'] || 'block'}"
 	</buffer>
+</store>
+{{- end}}`
+
+const storeSyslogTemplate = `{{- define "storeSyslog" }}
+<store>
+	@type {{.SyslogPlugin}}
+	@id {{.StoreID}}
+	remote_syslog {{.Host}}
+	port {{.Port}}
+	hostname ${hostname}
+	facility user
+	severity debug
 </store>
 {{- end}}`
--- a/test/e2e/logforwarding/syslog/forward_to_syslog_test.go
+++ b/test/e2e/logforwarding/syslog/forward_to_syslog_test.go
@@ -1,0 +1,144 @@
+package fluent
+
+import (
+	"fmt"
+	"runtime"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	apps "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+
+	logforward "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
+	"github.com/openshift/cluster-logging-operator/pkg/logger"
+	"github.com/openshift/cluster-logging-operator/test/helpers"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("LogForwarding", func() {
+	_, filename, _, _ := runtime.Caller(0)
+	logger.Infof("Running %s", filename)
+	var (
+		err              error
+		syslogDeployment *apps.Deployment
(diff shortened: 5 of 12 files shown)


richm:
Thanks.  I have some comments:

I think the functionality should be similar to what is in 3.11 - please take a look at https://github.com/openshift/origin-aggregated-logging/blob/release-3.11/fluentd/generate_syslog_config.rb to see how the fluentd config is generated - the new code should be similar

re: retry - the retry functionality is specific to the elasticsearch output, when es returns a specific 4xx error code which means the operation should be retried.  AFAIK, the syslog protocol has no such functionality.

richm:
Also, please review the documentation for the syslog feature in 3.11: https://docs.openshift.com/container-platform/3.11/install_config/aggregate_logging.html#sending-logs-to-external-rsyslog - the new implementation should offer similar functionality (but obviously not via env. vars or ansible parameters).

ahadas:
> Thanks. I have some comments:

@richm Thanks for the quick feedback :)

> 
> I think the functionality should be similar to what is in 3.11 - please take a look at https://github.com/openshift/origin-aggregated-logging/blob/release-3.11/fluentd/generate_syslog_config.rb to see how the fluentd config is generated - the new code should be similar
> 

Actually that's what I started with - I enabled the processing of the `USE_REMOTE_SYSLOG`  and `REMOTE_SYSLOG_HOST` env variables and took the output of that ruby script as a reference. What's the difference between the current output and that one that you see?

> re: retry - the retry functionality is specific to the elasticsearch output, when es returns a specific 4xx error code which means the operation should be retried. AFAIK, the syslog protocol has no such functionality.

Ack

richm:
> > Thanks. I have some comments:
> 
> @richm Thanks for the quick feedback :)
> 
> > I think the functionality should be similar to what is in 3.11 - please take a look at https://github.com/openshift/origin-aggregated-logging/blob/release-3.11/fluentd/generate_syslog_config.rb to see how the fluentd config is generated - the new code should be similar
> 
> Actually that's what I started with - I've used the `USE_REMOTE_SYSLOG` and `REMOTE_SYSLOG_HOST` env variables and took the output of that ruby script as a reference.

ok - good

> 
> > re: retry - the retry functionality is specific to the elasticsearch output, when es returns a specific 4xx error code which means the operation should be retried. AFAIK, the syslog protocol has no such functionality.
> 
> Ack

ahadas:
> > re: retry - the retry functionality is specific to the elasticsearch output, when es returns a specific 4xx error code which means the operation should be retried. AFAIK, the syslog protocol has no such functionality.
> 
> Ack

Done.
I've added another outputLabelConf template to reduce the chance for regressions with other outputs, hope it's ok.

ahadas:
Added all the properties for remote syslog that appear on the [4.1 documentation](https://docs.openshift.com/container-platform/4.1/logging/config/efk-logging-external.html#efk-logging-external-syslog_efk-logging-external)

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -553,3 +564,22 @@ const storeElasticsearchTemplate = `{{- define "storeElasticsearch" }}
...
 {{- end}}`
+
+const storeSyslogTemplate = `{{- define "storeSyslog" }}
+<store>
+	@type {{.RSyslogType}}
+	@id {{.StoreID}}
+	remote_syslog {{.Host}}
+	port {{.Port}}
+	hostname ${hostname}
+	facility {{.RSyslogFacility}}
```
jcantrill: 'user' - explicitly set based on https://github.com/openshift/origin-aggregated-logging/blob/release-3.11/fluentd/out_syslog_buffered.rb#L13-L21

same with all below

pkg/generators/forwarding/fluentd/rsyslog_conf.go
```diff
@@ -0,0 +1,46 @@
+package fluentd
```
jcantrill: I believe this file is obsolete if the remaining bits are opinionated

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -553,3 +564,22 @@ const storeElasticsearchTemplate = `{{- define "storeElasticsearch" }}
 	</buffer>
 </store>
 {{- end}}`
+
+const storeSyslogTemplate = `{{- define "storeSyslog" }}
```
jcantrill: Please add a unit test like https://github.com/openshift/cluster-logging-operator/blob/master/pkg/generators/forwarding/fluentd/output_conf_forward_test.go

manifests/4.4/logforwardings.crd.yaml
```diff
@@ -29,6 +29,7 @@ spec:
                     enum:
                       - elasticsearch
                       - forward
+                      - rsyslog
```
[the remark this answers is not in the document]
ahadas: done

pkg/apis/logging/v1alpha1/forwarding_types.go
```diff
@@ -56,6 +56,18 @@ type OutputSpec struct {
 	Name     string            `json:"name,omitempty"`
 	Endpoint string            `json:"endpoint,omitempty"`
 	Secret   *OutputSecretSpec `json:"secret,omitempty"`
+	RSyslog  *RemoteSyslogSpec `json:"rsyslog,omitempty"`
```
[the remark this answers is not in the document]
ahadas: Fair enough.
Not sure I understand the term "opinionated values" though. I assumed it means fetching these values from environment variables, please let me know if you meant something else

pkg/apis/logging/v1alpha1/forwarding_types.go
```diff
@@ -74,6 +86,9 @@ const (
 
 	//OutputTypeForward configures the pipeline to send messages via Fluent's secure forward
 	OutputTypeForward OutputType = "forward"
+
+	//OutputTypeSyslog configures pipeline to send messages to a remote syslog through TCP
+	OutputTypeSyslog OutputType = "rsyslog"
```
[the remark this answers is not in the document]
ahadas: Done

pkg/generators/forwarding/fluentd/rsyslog_conf.go
ahadas: Right, removed

pkg/generators/forwarding/fluentd/templates.go
ahadas: Changed, but can you please clarify this for me - I though that [generate_syslog_config.rb ](https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/generate_syslog_config.rb#L16-L34) is executed before and sets the defaults so the plugin's defaults are not being applied, and therefore the defaults I see in the [4.1 documentation](https://docs.openshift.com/container-platform/4.1/logging/config/efk-logging-external.html#efk-logging-external-syslog_efk-logging-external) - how come that eventually the plugin's defaults are set?

pkg/generators/forwarding/fluentd/templates.go
ahadas: Done

pkg/apis/logging/v1alpha1/forwarding_types.go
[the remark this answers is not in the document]
jcantrill: > Fair enough.
> Not sure I understand the term "opinionated values" though. I assumed it means fetching these values from environment variables, please let me know if you meant something else

"opinionated" means we as the logging team and the operator know better then the admin.  We set values to a sane default.  The alternative is we make everything highly configurable but this expands the test matrix and requires us to account for more test scenarios.  Ideally, we wish the admin to say "collect my logs and send them to THIS endpoint which is of type XYZ"

pkg/generators/forwarding/fluentd/templates.go
jcantrill: > Changed, but can you please clarify this for me - I though that [generate_syslog_config.rb ](https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/generate_syslog_config.rb#L16-L34) is executed before and sets the defaults so the plugin's defaults are not being applied, and therefore the defaults I see in the [4.1 documentation](https://docs.openshift.com/container-platform/4.1/logging/config/efk-logging-external.html#efk-logging-external-syslog_efk-logging-external) - how come that eventually the plugin's defaults are set?

Any config generation based on scripts and ruby files that are in the fluent image are no longer relevant; any reference provided to you is exactly that, a references for modifying it in LF.  At the heart of it, LogForwarding is a config generator for the collector and does not allow for any other tweaks other then what is possible via the LF api

pkg/generators/forwarding/fluentd/output_conf_syslog_test.go
```diff
@@ -0,0 +1,56 @@
...
+			test.Expect(results[0]).ToEqual(`<label @SYSLOG_RECEIVER>
+	<match **>
+	   @type copy
+	   <store>
+  	     @type "#{ENV['SYSLOG_TYPE'] || 'syslog_buffered'}"
+	     @id syslog_receiver
+	     remote_syslog sl.svc.messaging.cluster.local
+	     port 9654
+	     hostname ${hostname}
+	     facility "#{ENV['SYSLOG_FACILITY'] || 'user'}"
```
jcantrill: @richm if we are using LogForwarding and requiring the only way to modify conf is to:

1. Set ClusterLogging(Logforwarding?) to Unmanaged
1. Manually edit conf

Do you see any reason to still rely on env vars in lieu of explicitly setting a value?  I realize we have this elsewhere in our config generation but seems pointless given the mechanism to modify it.

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
```diff
@@ -0,0 +1,103 @@
...
+				for _, component := range components {
+					if err := e2e.WaitFor(component); err != nil {
+						Fail(fmt.Sprintf("Failed waiting for component %s to be ready: %v", component, err))
+					}
+				}
+
+			})
+
+			It("should send logs to the forward.Output logstore", func() {
+				Expect(e2e.LogStore.HasInfraStructureLogs(helpers.DefaultWaitForLogsTimeout)).To(BeTrue(), "Expected to find stored infrastructure logs")
```
jcantrill: Do you want to additionally check app and audit logs since you have those sources configured too?

pkg/apis/logging/v1alpha1/forwarding_types.go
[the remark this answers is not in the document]
ahadas: Got it

pkg/generators/forwarding/fluentd/templates.go
ahadas: Oh, I understand this part - that it's no longer relevant since introducing LF. It's just that you pointed me to the defaults in 3.11 that are determined by out_syslog_buffered.rb so I wondered if in 4.1 and below those default were ever applied in light of the configuration generated by generate_syslog_config.rb

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
ahadas: Not sure, I think not and I'll remove the other sources. But this part is still in progress - I try to make the rsyslog server to dump any traffic to a log file...

ahadas:
> @richm has stated the plugin we ship is old. If we need to switch the plugin or do something else to support the secure case and the most relevant protocol, please advise.

Seems that [this](https://github.com/docebo/fluent-plugin-remote-syslog) is the source of the plugin we use for syslog, right? it seems quite abandoned, but maybe we can use [this](https://github.com/zarqman/fluent-plugin-syslog-tls) as a reference and contribute tls support for it?

pkg/generators/forwarding/fluentd/templates.go
jcantrill: I pointed you to the defaults so you had a value to use in place of the templated values.  If there are others which you believe are more appropriate please use your best judgement

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
ahadas: Done, removed

ahadas:
@jcantrill @richm I believe all comments were addressed and I think that support for secured connections that can be added separately. Please let me know if there's anything else on me that blocks this.

sichvoge:
I think you are right. I am fine to close this and add anything we think also needs to be done for Verizon like the updated plugin and also TLS.

Let's wait for @jcantrill ack.

pkg/generators/forwarding/fluentd/output_conf_syslog_test.go
jcantrill: @ahadas per my discussion with @ewolinetz please explicitly set these values in lieu of them optionally being set by envvar.  Given the the fact the config is generated and to change anything about the config or DaemonSet env vars require you to go unmanaged, it makes more sense to simply require someone to update config directly.  Aside from that it lgtm

ahadas: Done

ahadas:
> Awaiting to see where we need to merge this but lgtm

@jcantrill  ping :)

manifests/4.5/logforwardings.crd.yaml
```diff
@@ -29,6 +29,7 @@ spec:
                     enum:
                       - elasticsearch
                       - forward
+                      - syslog-legacy
```
jcantrill: Nit here, but I don't think we need to call the type 'syslog-legacy'.  I think 'syslog' is sufficient regardless off the actual plugin.  From what I recall, any plugin to replace what we have would be a superset of what the current one provides.  Is that correct?

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
```diff
@@ -0,0 +1,144 @@
...
+		e2e              = helpers.NewE2ETestFramework()
+	)
+	BeforeEach(func() {
+		if err := e2e.DeployLogGenerator(); err != nil {
+			logger.Errorf("unable to deploy log generator. E: %s", err.Error())
+		}
+	})
+	Describe("when ClusterLogging is configured with 'forwarding' to an external syslog server", func() {
+
+		Context("with the legacy syslog plugin", func() {
```
jcantrill: "legacy" may be appropriate if we need to test a new and an old but won't we simply be able to swap out the plugin, update conf, and the essence of the test should remain

manifests/4.5/logforwardings.crd.yaml
ahadas: Yes, that's correct - in terms of protocols, the new plugin should be a superset of what the current one provides.

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
ahadas: Yeah, that's what I had in mind - to have an hybrid mode where the legacy one points to the current plugin and "syslog" points to the new plugin, at least until we'll have confidence in the new plugin to provide all the functionality of the current one and at the same quality (performance-wise).
But we can do it as part of the next PR that introduces the new plugin, by then we'll have better feeling of the gaps between the two and whether or not this hybrid mode makes sense - so removed the "legacy" part everywhere

PR #348, opened 2020-01-29
Bug 1796133: Exclude fluent/elasticsearch pod from log collection and write logs to STDOUT

jcantrill:
Write fluent logs to stdout

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1796133

### DIFF
--- a/files/fluentd/run.sh
+++ b/files/fluentd/run.sh
@@ -4,26 +4,8 @@ export MERGE_JSON_LOG=${MERGE_JSON_LOG:-false}
 CFG_DIR=/etc/fluent/configs.d
 ENABLE_PROMETHEUS_ENDPOINT=${ENABLE_PROMETHEUS_ENDPOINT:-"true"}
 OCP_OPERATIONS_PROJECTS=${OCP_OPERATIONS_PROJECTS:-"default openshift openshift- kube-"}
-LOGGING_FILE_PATH=${LOGGING_FILE_PATH:-"/var/log/fluentd/fluentd.log"}
 
-loggingargs=""
-if [ ${LOGGING_FILE_PATH} != "console" ] ; then
-    echo "============================="
-    echo "Fluentd logs have been redirected to: $LOGGING_FILE_PATH"
-    echo "If you want to print out the logs, use command:"
-    echo "oc exec <pod_name> -- logs"
-    echo "============================="
-
-    dirname=$( dirname $LOGGING_FILE_PATH )
-    if [ ! -d $dirname ] ; then
-        mkdir -p $dirname
-    fi
-    touch $LOGGING_FILE_PATH; exec >> $LOGGING_FILE_PATH 2>&1
-
-    loggingargs="-o $LOGGING_FILE_PATH --log-rotate-age $LOGGING_FILE_AGE --log-rotate-size $LOGGING_FILE_SIZE"
-fi
-
-fluentdargs="--no-supervisor $loggingargs"
+fluentdargs="--no-supervisor"
 # find the sniffer class file
 sniffer=$( gem contents fluent-plugin-elasticsearch|grep elasticsearch_simple_sniffer.rb )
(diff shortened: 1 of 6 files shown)


jcantrill:
cc @ewolinetz

pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -140,6 +140,7 @@ var _ = Describe("Generating fluentd config", func() {
 				@type tail
 				@id container-input
 				path "/var/log/containers/*.log"
+				exclude_path ['fluentd-*_openshift-logging_fluentd-*.log']
```
ewolinetz: nit, but is it necessary to have the container name in there too?
e.g. `'fluentd-*_openshift-logging_*.log'`

jcantrill: Corrected.

jcantrill:
@ewolinetz updated to also exclude ES pods

pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -140,12 +140,12 @@ var _ = Describe("Generating fluentd config", func() {
 				@type tail
 				@id container-input
 				path "/var/log/containers/*.log"
+				exclude_path ['fluentd-*_openshift-logging_*.log', 'elasticsearch-*_openshift-logging_*.log']
```
ewolinetz: we should also exclude kibana, currently it writes to stdout and it can be noisy

jcantrill: done

pkg/generators/forwarding/fluentd/source.go
```diff
@@ -24,8 +25,19 @@ func (engine *ConfigGenerator) generateSource(sources sets.String) (results []st
...
 		return results, fmt.Errorf("Unable to generate source configs for supported source types: %v", sources.List())
 	}
+	data := struct {
+		LoggingNamespace           string
+		CollectorPodNamePrefix     string
+		LogStorePodNamePrefix      string
+		VisualizationPodNamePrefix string
+	}{
+		constants.OpenshiftNS,
+		"fluentd",
```
ewolinetz: nit, but should we make this a constant too for uniformity?

ewolinetz:
tiny nit, otherwise :shipit:

jcantrill:
@ewolinetz I also realized i can simply remove the file writing from the run script since it now lives here

files/fluentd/run.sh
```diff
@@ -4,26 +4,8 @@ export MERGE_JSON_LOG=${MERGE_JSON_LOG:-false}
...
-    if [ ! -d $dirname ] ; then
-        mkdir -p $dirname
-    fi
-    touch $LOGGING_FILE_PATH; exec >> $LOGGING_FILE_PATH 2>&1
-
-    loggingargs="-o $LOGGING_FILE_PATH --log-rotate-age $LOGGING_FILE_AGE --log-rotate-size $LOGGING_FILE_SIZE"
-fi
 
-fluentdargs="--no-supervisor $loggingargs"
+fluentdargs="--no-supervisor"
```
ewolinetz: so i was actually just thinking about this... and this is probably a larger discussion but if we're logging to stdout how long does the container hold this information? and what happens to our running pod size the longer it lives? do we expect to just roll them periodically? or can we configure it to age out older stdout entries?

jcantrill: > so i was actually just thinking about this... and this is probably a larger discussion and not a blocker but if we're logging to stdout how long does the container hold this information? and what happens to our running pod size the longer it lives? do we expect to just roll them periodically? or can we configure it to age out older stdout entries?

I believe this is the responsibility of the runtime to manage the logfile and its rotation

PR #351, opened 2020-01-30
Bug 1781492: Updating so Kibana properly handles cert redeploys

ewolinetz:
Addresses: https://bugzilla.redhat.com/show_bug.cgi?id=1781492

@igor-karpukhin fyi this impacts your kibana CLO -> EO migration work


ewolinetz:
for some reason i'm seeing this try to delete a kibana pod when clo restarts because it wants to reconcile for every secret in `openshift-logging`.

I'm going to try to debug this, but as is we shouldn't merge this in.

PR #355, opened 2020-02-04
WIP: testify version of fluent_conf_test.go for review

alanconway:
For comparison with Ginkgo tests, review & discussion only.  Do not commit.

Signed-off-by: Alan Conway <aconway@redhat.com>


jcantrill:
I realize this is possible given what you expressed in the email thread but I would propose you have reverted the all the advantages of what I see to Ginko: intent and expressiveness.  This version of the test requires a developer to mentally understand the test from top to bottom and interpret the intent of the test from the terse name.  :-1:

alanconway:
Pushed an update that has the same detailed test names and variable/function scopes as your example.

jcantrill:
Closing as I expect this is simply to prove out alternate test options

PR #356, opened 2020-02-05
Forwarding to external syslog, based on a new fluentd plugin

ahadas:
This PR continues #341 and introduces an alternative way to specify syslog output within the `LogForwarding` CR that is based on a new plugin for fluentd (see openshift/origin-aggregated-logging#1826).

This PR fixes openshift/origin-aggregated-logging#1547 by enabling to transmit the logs over tcp+tls.

The secure connection can be specified as follow:
```yaml
spec:
  outputs:
  - name: syslog
    type: syslog
    endpoint: "10.46.9.101:515"
    secret:
      name: my-secret
  pipelines:
  - name: logsapp
    inputSource: logs.audit
    outputRefs:
    - syslog
```

Alternatively, the transport protocol can be set to udp (note the `endpoint` property):
```yaml
spec:
  outputs:
  - name: syslog
    type: syslog
    endpoint: "udp://10.46.9.101:515"
  pipelines:
  - name: logsapp
    inputSource: logs.audit
    outputRefs:
    - syslog
```
Or to unsecure tcp (which is also the case when no protocol is specified):
```yaml
spec:
  outputs:
  - name: syslog
    type: syslog
    endpoint: "tcp://10.46.9.101:515"
  pipelines:
  - name: logsapp
    inputSource: logs.audit
    outputRefs:
    - syslog
```

This allows us to deprecate the output type `syslog-legacy` and remove it once we feel the new plugin provides all the necessary things that were provided by the previous plugin.

### DIFF
--- a/manifests/4.4/logforwardings.crd.yaml
+++ b/manifests/4.4/logforwardings.crd.yaml
@@ -29,6 +29,7 @@ spec:
                     enum:
                       - elasticsearch
                       - forward
+                      - syslog
                       - syslog-legacy
                   name: 
                     description: The name of the output
--- a/manifests/4.4/logforwardings.crd.yaml
+++ b/manifests/4.4/logforwardings.crd.yaml
@@ -29,6 +29,7 @@ spec:
                     enum:
                       - elasticsearch
                       - forward
+                      - syslog-legacy
                   name: 
                     description: The name of the output
                     type: string
--- a/pkg/apis/logging/v1alpha1/forwarding_types.go
+++ b/pkg/apis/logging/v1alpha1/forwarding_types.go
@@ -74,6 +74,9 @@ const (
 
 	//OutputTypeForward configures the pipeline to send messages via Fluent's secure forward
 	OutputTypeForward OutputType = "forward"
+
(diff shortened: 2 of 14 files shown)


jcantrill:
Does this PR supercede #341?

ahadas:
> Does this PR supercede #341?

Yes, the commit of #341 is included here - I didn't want to lose the approval on #341 :)

ahadas:
> > Does this PR supercede #341?
> 
> Yes, the commit in #341 is included here - I didn't want to lose the approval on #341 :)

I've made changes to both #341 and this PR- the former enables defining outputs based on the existing plugin (and thus I expect its e2e tests to pass) and this one enables defining outputs based on the new plugin (so the e2e tests would pass only after openshift/origin-aggregated-logging#1826 is merged. And updated the descriptions of the PRs accordingly.

ahadas:
> /retest

It now reaches the point when I expected it to fail until openshift/origin-aggregated-logging#1547 gets in (locally it passes for me), so this one is also ready for review.

jcantrill:
Closing this PR as LF has changed significantly since this was opened.  Please review the current state of logforwarding and reopen if needed

PR #358, opened 2020-02-06
Bug 1799051: Add OLM annotations for ui hints

jcantrill:
https://bugzilla.redhat.com/show_bug.cgi?id=1799051

### DIFF
--- a/manifests/4.4/cluster-logging.v4.4.0.clusterserviceversion.yaml
+++ b/manifests/4.4/cluster-logging.v4.4.0.clusterserviceversion.yaml
@@ -7,6 +7,8 @@ metadata:
   name: clusterlogging.v4.4.0
   namespace: placeholder
   annotations:
+    "operatorframework.io/suggested-namespace": openshift-logging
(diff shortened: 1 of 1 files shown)


manifests/4.4/cluster-logging.v4.4.0.clusterserviceversion.yaml
```diff
@@ -299,15 +301,15 @@ spec:
                   - name: OPERATOR_NAME
                     value: "cluster-logging-operator"
                   - name: ELASTICSEARCH_IMAGE
-                    value: "quay.io/openshift/origin-logging-elasticsearch5:latest"
+                    value: registry.svc.ci.openshift.org/ocp/4.4:logging-elasticsearch5
```
ewolinetz: isn't the old pattern what will get replaced for us by the CI process?

jcantrill: Reverting..  this is not supposed to be here

PR #361, opened 2020-02-13
Add Role and RoleBinding for console-operator to read the sharing-config

jhadvig:
Related [PR](https://github.com/openshift/console-operator/pull/379) with some background [background](https://github.com/openshift/console-operator/pull/379#discussion_r378811466). Basically since the `openshift-logging` namaspace is not created by default on the cluster we need to add the required role and rolebinding for the console-operator here.

@benjaminapetersen fyi

/assign @jcantrill


jhadvig:
Closing this one based on the conversation with @jcantrill and @benjaminapetersen

PR #363, opened 2020-02-14
Bug 1802934: Fix incorrect olm.skipRange in 4.3 CSV

vfreex:
Fix an issue where the cluster-logging-operator couldn't be upgraded because it doesn't match the update_list defined in art.yaml nor configured in correct range.

Similar issue https://github.com/openshift/elasticsearch-operator/pull/237


anpingli:
The bugs is https://bugzilla.redhat.com/show_bug.cgi?id=1802934

PR #367, opened 2020-02-17
Bug 1800308: Updating to require trustbundle configmap

ewolinetz:
To address: https://bugzilla.redhat.com/show_bug.cgi?id=1800308

### DIFF
--- a/pkg/k8shandler/fluentd.go
+++ b/pkg/k8shandler/fluentd.go
@@ -380,7 +380,6 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, elasticsearchAppName str
 	}
 
 	if addTrustedCAVolume {
-		optional := true
 		fluentdPodSpec.Volumes = append(fluentdPodSpec.Volumes,
 			v1.Volume{
 				Name: constants.FluentdTrustedCAName,
@@ -389,7 +388,6 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, elasticsearchAppName str
 						LocalObjectReference: v1.LocalObjectReference{
 							Name: constants.FluentdTrustedCAName,
 						},
-						Optional: &optional,
 						Items: []v1.KeyToPath{
 							{
 								Key:  constants.TrustedCABundleKey,
@@ -413,26 +411,23 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateFluentdDaemonset(pipe
 	cluster := clusterRequest.cluster
 
 	fluentdTrustBundle := &v1.ConfigMap{}
-	if proxyConfig != nil {
-		// Create or update cluster proxy trusted CA bundle.
-		err = clusterRequest.createOrUpdateTrustedCABundleConfigMap(constants.FluentdTrustedCAName)
-		if err != nil {
-			return
-		}
-
-		// fluentd-trusted-ca-bundle
-		fluentdTrustBundleName := types.NamespacedName{Name: constants.FluentdTrustedCAName, Namespace: constants.OpenshiftNS}
-		if err := clusterRequest.client.Get(context.TODO(), fluentdTrustBundleName, fluentdTrustBundle); err != nil {
-			if !errors.IsNotFound(err) {
-				return err
-			}
-		}
+	// Create or update cluster proxy trusted CA bundle.
+	err = clusterRequest.createOrUpdateTrustedCABundleConfigMap(constants.FluentdTrustedCAName)
+	if err != nil {
+		return
 	}
 
 	fluentdPodSpec := newFluentdPodSpec(cluster, "elasticsearch", "elasticsearch", proxyConfig, fluentdTrustBundle, clusterRequest.ForwardingSpec)
 
 	fluentdDaemonset := NewDaemonSet("fluentd", cluster.Namespace, "fluentd", "fluentd", fluentdPodSpec)
 	fluentdDaemonset.Spec.Template.Spec.Containers[0].Env = updateEnvVar(v1.EnvVar{Name: "FLUENT_CONF_HASH", Value: pipelineConfHash}, fluentdDaemonset.Spec.Template.Spec.Containers[0].Env)
+
+	annotations, err := clusterRequest.getFluentdAnnotations(fluentdDaemonset)
+	if err != nil {
+		return err
+	}
+
+	fluentdDaemonset.Spec.Template.Annotations = annotations
 
 	uid := getServiceAccountLogCollectorUID()
 	if len(uid) == 0 {
@@ -450,7 +445,7 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateFluentdDaemonset(pipe
 
 	if clusterRequest.isManaged() {
 		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			return clusterRequest.updateFluentdDaemonsetIfRequired(fluentdDaemonset, fluentdTrustBundle)
+			return clusterRequest.updateFluentdDaemonsetIfRequired(fluentdDaemonset)
 		})
 		if retryErr != nil {
 			return retryErr
@@ -460,7 +455,7 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateFluentdDaemonset(pipe
 	return nil
 }
 
-func (clusterRequest *ClusterLoggingRequest) updateFluentdDaemonsetIfRequired(desired *apps.DaemonSet, trustedCABundleCM *v1.ConfigMap) (err error) {
+func (clusterRequest *ClusterLoggingRequest) updateFluentdDaemonsetIfRequired(desired *apps.DaemonSet) (err error) {
 	logger.DebugObject("desired fluent update: %v", desired)
 	current := &apps.DaemonSet{}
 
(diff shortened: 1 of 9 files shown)


pkg/k8shandler/visualization.go
```diff
@@ -283,17 +280,40 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaDeployment(prox
...
 				different = true
 				if kibanaDeployment.Spec.Template.ObjectMeta.Annotations == nil {
 					kibanaDeployment.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
 				}
-				kibanaDeployment.Spec.Template.ObjectMeta.Annotations[constants.TrustedCABundleHashName] = newTrustedCAHashedValue
+				kibanaDeployment.Spec.Template.ObjectMeta.Annotations[constants.TrustedCABundleHashName] = newTrustedCAHashValue
+			}
+
+			// Check secret hash has been updated or not
+			for _, secretName := range []string{"kibana", "kibana-proxy"} {
```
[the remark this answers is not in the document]
ewolinetz: this is to fix the logic that was being used before where we deleted the kibana pods in the case of the `kibana` or `kibana-proxy` secrets being updated
ref: https://github.com/openshift/cluster-logging-operator/pull/351

bparees:
@ewolinetz can you elaborate on how this addresses the bug in question?  seems like the bug in question was caused by us (for some reason) redeploying the operands when we should not (where redeploying meant deleting the pods).

this PR seems to:
1) switch from deleting pods to poking the deployment to trigger a redeployment (good, but not the core issue in the BZ)

2) some changes to how proxyconfig + proxyCA updates are handled - was something related to this the suspected cause of the frequent redeployments in the bug?

pkg/k8shandler/visualization.go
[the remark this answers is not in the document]
bparees: ok, so fluentd was already handling this properly?

pkg/k8shandler/visualization.go
```diff
@@ -283,17 +280,40 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaDeployment(prox
...
+
+				secret, err := clusterRequest.GetSecret(secretName)
+				if err != nil {
+
+				}
+				newSecretHashValue, err := calcSecretHashValue(secret)
+				if err != nil {
+
+				}
+				currentSecretHashValue, _ := current.Spec.Template.ObjectMeta.Annotations[hashKey]
```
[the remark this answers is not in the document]
ewolinetz: this can't be `nil` since its from the already existing object here which would just be empty -
https://godoc.org/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta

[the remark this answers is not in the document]
bparees: sorry, golang weirdness.  technically it is nil (or at least ==nil returns true) but map["key"] does not get a nil ptr error, it just returns nothing.

pkg/k8shandler/secret.go
```diff
@@ -92,3 +94,33 @@ func (clusterRequest *ClusterLoggingRequest) RemoveSecret(secretName string) err
 
 	return nil
 }
+
+func calcSecretHashValue(secret *core.Secret) (string, error) {
```
jcantrill: this is something that we can confirm with a test

pkg/k8shandler/secret.go
```diff
@@ -92,3 +94,33 @@ func (clusterRequest *ClusterLoggingRequest) RemoveSecret(secretName string) err
...
+		hashKeys = append(hashKeys, key)
+	}
+
+	sort.Strings(hashKeys)
+
+	for _, key := range hashKeys {
+		rawbytes = append(rawbytes, secret.Data[key]...)
+	}
+
+	hashValue, err = utils.CalculateMD5Hash(string(rawbytes))
```
jcantrill: is there a need to calculate the hash of the keys and the content.  Dont we need to be concerned with the key names changing?

pkg/k8shandler/visualization.go
```diff
@@ -79,28 +79,35 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateVisualization(proxyCo
...
-		printUpdateMessage := true
-		retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-			if !compareKibanaStatus(kibanaStatus, cluster.Status.Visualization.KibanaStatus) {
-				if printUpdateMessage {
-					logrus.Infof("Updating status of Kibana")
-					printUpdateMessage = false
-				}
-				cluster.Status.Visualization.KibanaStatus = kibanaStatus
-				return clusterRequest.UpdateStatus(cluster)
+func (clusterRequest *ClusterLoggingRequest) SetKibanaStatus() (err error) {
```
jcantrill: the name "UpdateKibanaStatus" would be better given you are not providing any args

pkg/k8shandler/visualization.go
```diff
@@ -283,17 +280,40 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaDeployment(prox
...
-			trustedCAHashedValue, _ := current.Spec.Template.ObjectMeta.Annotations[constants.TrustedCABundleHashName]
-			if trustedCAHashedValue != newTrustedCAHashedValue {
+			trustedCAHashValue, _ := current.Spec.Template.ObjectMeta.Annotations[constants.TrustedCABundleHashName]
+			if trustedCAHashValue != newTrustedCAHashValue {
 				different = true
 				if kibanaDeployment.Spec.Template.ObjectMeta.Annotations == nil {
 					kibanaDeployment.Spec.Template.ObjectMeta.Annotations = make(map[string]string)
 				}
-				kibanaDeployment.Spec.Template.ObjectMeta.Annotations[constants.TrustedCABundleHashName] = newTrustedCAHashedValue
+				kibanaDeployment.Spec.Template.ObjectMeta.Annotations[constants.TrustedCABundleHashName] = newTrustedCAHashValue
```
jcantrill: Does annotation change cause a redeploy?  @bparees ?

bparees: i believe any change to the podtemplate within the deployment causes a redeploy, but we should verify.

pkg/k8shandler/visualization.go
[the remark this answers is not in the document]
ewolinetz: fluentd doesn't need to be restarted in the case of a cert (secret) redeploy because it will reload certificates on its own. only ES and Kibana need to be externally restarted.

pkg/k8shandler/secret.go
ewolinetz: we don't need to worry about the keys changing because the operator ensures that it will be the correct key name. we need to know if the content (certs/keys) changed to know if we need to restart kibana

pkg/k8shandler/secret.go
ewolinetz: what would we be confirming?

pkg/k8shandler/visualization.go
ewolinetz: yes:
* @igor-karpukhin and i confirmed that in the case where we saw more than one kibana pod the difference between the replica sets was the annotation value
* manually testing this change when I deleted the configmap and forced a new annotation value to be generated i saw a redeploy/update

igor-karpukhin:
@ewolinetz , PR looks good, but the E2E failed. Also, should we first merge this PR and then I will backport it to my branch?

ewolinetz:
>  can you elaborate on how this addresses the bug in question? seems like the bug in question was caused by us (for some reason) redeploying the operands when we should not (where redeploying meant deleting the pods).

The core of the bug is due to timing with updates, so ideally we want to limit the amount of triggers that can cause us to reconcile to restrict unnecessary churn with updates. Essentially what I saw on my cluster was we would do an initial deployment (for kibana) and then we would immediately update it to add in the hashes. If the configmap was not yet injected with the trusted CA it would continue to error out and then we would try to update again after with the hash value. All of this could leave the cluster in a state where our replicaset kept trying to scale up a pod and the pod would be terminated.

pkg/k8shandler/fluentd.go
```diff
@@ -514,6 +499,61 @@ func (clusterRequest *ClusterLoggingRequest) updateFluentdDaemonsetIfRequired(de
...
+		return annotations, fmt.Errorf("%v does not yet contain expected key %v", fluentdTrustBundle.Name, constants.TrustedCABundleKey)
+	}
+
+	trustedCAHashValue, err := calcTrustedCAHashValue(fluentdTrustBundle)
+	if err != nil {
+		return annotations, fmt.Errorf("unable to calculate trusted CA value. E: %s", err.Error())
+	}
+
+	if trustedCAHashValue == "" {
+		return annotations, fmt.Errorf("Did not receive hashvalue for trusted CA value")
```
jcantrill: Does this need to return an error?  Isn't an empty value valid?

pkg/k8shandler/secret.go
jcantrill: the value produced is the value expected.  What if someone comes along and removes the sort statement.  It's likely immaterial given you would drive the change to the new value of the algo of the function but asking the question

pkg/k8shandler/fluentd.go
ewolinetz: it isn't, we want require that the hash is populated with something so that we don't unnecessarily update and cause a redeploy.

bparees:
/hold

this seems to be spinning

ewolinetz:
I think we're blocked by https://github.com/openshift/cluster-logging-operator/pull/373 -- if we have es6 as the image used for this test, then we need the EO changes for es6 to be merged in...

PR #374, opened 2020-02-19
Bug 1803196: Move shared config map to openshift-config-managed NS

periklis:
To address: https://bugzilla.redhat.com/show_bug.cgi?id=1803196

### DIFF
--- a/pkg/k8shandler/visualization.go
+++ b/pkg/k8shandler/visualization.go
@@ -29,6 +29,12 @@ const (
 	// The following strings are turned into JavaScript RegExps. Online tool to test them: https://regex101.com/
 	nodesAndContainersNamespaceFilter = "^(openshift-.*|kube-.*|openshift$|kube$|default$)"
 	appsNamespaceFilter               = "^((?!" + nodesAndContainersNamespaceFilter + ").)*$" // ^((?!^(openshift-.*|kube-.*|openshift$|kube$|default$)).)*$
+
+	loggingSharedConfigMapNamePre44x     = "sharing-config"
+	loggingSharedConfigRolePre44x        = "sharing-config-reader"
+	loggingSharedConfigRoleBindingPre44x = "openshift-logging-sharing-config-reader-binding"
+	loggingSharedConfigMapName           = "logging-shared-config"
+	loggingSharedConfigNs                = "openshift-config-managed"
 )
 
 var (
@@ -201,7 +207,11 @@ func (clusterRequest *ClusterLoggingRequest) removeKibana() (err error) {
 			return
 		}
 
-		if err = clusterRequest.RemoveConfigMap("sharing-config"); err != nil {
+		if err = clusterRequest.RemoveConfigMap(loggingSharedConfigMapNamePre44x); err != nil {
+			return
+		}
+
+		if err = clusterRequest.RemoveSharedConfigMap(loggingSharedConfigMapName, loggingSharedConfigNs); err != nil {
 			return
 		}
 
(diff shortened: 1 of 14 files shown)


jcantrill:
@jhadvig would this be better to use static roles and bindings that are created by OLM instead of the operator like: https://github.com/openshift/cluster-logging-operator/blob/master/manifests/4.4/0100_clusterroles.yaml   I presume it would simplify the PR

manifests/4.4/cluster-logging.v4.4.0.clusterserviceversion.yaml
```diff
@@ -247,6 +247,8 @@ spec:
         - apiGroups:
           - rbac.authorization.k8s.io
           resources:
+          - roles
```
jcantrill: I think this can be replaced by the static roles created by OLM.  See the manifest directory

pkg/k8shandler/visualization.go
```diff
@@ -204,7 +209,7 @@ func (clusterRequest *ClusterLoggingRequest) removeKibana() (err error) {
 			return
 		}
 
-		if err = clusterRequest.RemoveConfigMap("sharing-config"); err != nil {
+		if err = clusterRequest.RemoveSharedConfigMap(loggingSharedConfigMapName, loggingSharedConfigNs); err != nil {
```
jcantrill: Since we may have a scenario where we are upgrading the operator.  I believe we will either have to move this code or have similar code which ALWAYS tries to remove the configmap that was deployed in earlier versions.  As it stands this will only remove the new configmap created by this PR

manifests/4.4/cluster-logging.v4.4.0.clusterserviceversion.yaml
periklis: :+1:

pkg/k8shandler/visualization.go
periklis: What about the role and rolebinding to give `system:authenticaed` access to the shared config map? Shouldn't we delete them in the old namespace and create them in the new one for the migration path?

jcantrill: > What about the role and rolebinding to give `system:authenticated` access to the shared config map? Shouldn't we delete them in the old namespace and create them in the new one for the migration path?

I'm not sure what the 'old' code did exactly but I thought it was to create the shared config in the logging namespace.  Did it also create roles and binding? Maybe specifically for a console service account?  If so, then yes we should remove them.  This PR should:
* rename the shared config
* create shared config in the shared namespace
* create roles/bindings so it can create the config in the appropriate NS
* cleanup all the bits associated with the 'old' location

periklis: The old code created the shared config in the logging namespace and in addition a Role and Rolebinding for all authenticated users to access this namespace in companion when the kibana route was created/updated. Since we reconciled this role and rolebinding in addition to the configmap, we need to remove it in the logging space and reconciled it in the config-managed namespace.

Ok hold on, I will amend this PR then.

manifests/4.4/cluster-logging.v4.4.0.clusterserviceversion.yaml
periklis: Moved to static roles

pkg/k8shandler/visualization.go
periklis: Amended reconcilation for shared config role and rolebinding as well as deleting old ones.

syedriko:
nit: it would be nice to separate purely formatting changes into their own commit so the bug fix is easier to follow.

pkg/k8shandler/visualization.go
```diff
@@ -397,58 +417,80 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaRoute() error {
 		}
 	}
 
+	if err := clusterRequest.createOrUpdateKibanaSharedConfigMap(); err != nil {
+		return err
+	}
+
+	if err := clusterRequest.createOrUpdateKibanaSharedConfigRBAC(); err != nil {
```
jcantrill: You shouldn't need any of this code since the OLM operator is consuming your static resources and creating RBAC for you.

pkg/k8shandler/rbac.go
```diff
@@ -94,6 +95,78 @@ func NewClusterRoleBinding(bindingName, roleName string, subjects []rbac.Subject
...
+	err := clusterRequest.Delete(role)
+	if err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("Failure deleting role for %q: %v", clusterRequest.cluster.Name, err)
+	}
+
+	return nil
+}
+
+// CreateOrUpdateRoleBinding creates and returns a new rolebinding resource or updates an existing or returns an error
+func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateRoleBinding(name, namespace, roleName string, subjects []rbac.Subject) (*rbac.RoleBinding, error) {
```
jcantrill: I dont think any of this is needed if we are relying on static RBAC genertion

pkg/k8shandler/visualization.go
periklis: The extra RBAC reconciliation is removed now. I kept the cleanup of the old pair of Role/Rolebinding in `openshift-logging`.

pkg/k8shandler/rbac.go
periklis: This methods are dropped because of zero usage in the code base.

pkg/k8shandler/visualization.go
```diff
@@ -769,7 +765,7 @@ func newKibanaPodSpec(cluster *logging.ClusterLogging, kibanaName string, elasti
 
 func createSharedConfig(namespace, kibanaAppURL, kibanaInfraURL string) *v1.ConfigMap {
 	return NewConfigMap(
-		"sharing-config",
+		loggingSharedConfigMapName,
```
benjaminapetersen: Can we update `kibanaAppURL` -> `kibanaPublicAppURL`?  Not sure about `kibanaInfraURL`.

jhadvig: I think it should be `kibanaAppPublicURL`

PR #378, opened 2020-02-21
Update operator-sdk to 0.15.2, improvements to code generation.

alanconway:
- use a known operator-sdk version
- improved auto-generation of code if $(SRC) changes

The first commit in the PR improves the makefile and uses operator-sdk 0.8.2.
The second commit upgrades to operator-sdk 0.15.2.

I didn't see any problems with the upgrade, generated code is identical except for the "generated by" comment. However I'm not clear on how the .yaml in manifests/ is generated so I may be missing something.


ewolinetz:
in the past with this operator we've controlled the sdk version using the `Gopkg.toml` file and then ran `dep ensure -update`

it seems like you added a sdk bin file for this?

alanconway:
> in the past with this operator we've controlled the sdk version using the `Gopkg.toml` file and then ran `dep ensure -update`

@ewolinetz that makes sense, but it's not what the Makefile was doing. It was doing:  
     git clone https://github.com/operator-framework/operator-sdk --branch master $(GOPATH)/src/github.com/operator-framework/operator-sdk 
followed by install stuff under sudo! So a bug in our makefile could clobber a developer's laptop or a CI box, which I wanted to avoid. 

> it seems like you added a sdk bin file for this?

Yep. The benefits of downloading a binary are: it's quick, it's reproducible, it can only get official releases (not unstable dev branches) and the binary is independent of everything else on the local machine.

If you prefer a locally-built approach I can do it. We need to avoid clobbering anything installed on the local GOPATH (it might be in use for work other than our build), and we definitely don't want to touch anything that requires sudo!

What do you think?

jcantrill:
> I didn't see any problems with the upgrade, generated code is identical except for the "generated by" comment. However I'm not clear on how the .yaml in manifests/ is generated so I may be missing something.

It's been manually maintained

ewolinetz:
@alanconway squash your commits and then we can :shipit:

alanconway:
squashed

periklis:
I think it's not mandatory for this PR, but we should consider switching to Go >1.13 and Go Modules after this lands in. The operator sdk dropped dep suport in 0.12: https://github.com/operator-framework/operator-sdk/pull/1949

PR #380, opened 2020-02-25
Updated log source type names to use `-` instead of `.`

vimalk78:
OpenDistro security plugin parses role index patterns wrongly if it contains `.`, so using `-` instead.

Updated `logs.app` to `logs-app`
Updated `logs.infra` to `logs-infra`
Updated `logs.audit` to `logs-audit`

for https://issues.redhat.com/browse/LOG-600


lukas-vlcek:
> OpenDistro security plugin parses role index patterns wrongly if ...

@vimalk78 did you consider opening a ticket for OpenDistro plugin?

PR #388, opened 2020-02-28
WIP: LOG-653: GA API for cluster forwarding.

alanconway:
GA API for ClusterLogForwarder.

This will be implemented on the tech preview forwarding implementation: 
https://issues.redhat.com/browse/LOG-653

Known issue: deploy/crds is not in the correct form for OLM. This will be fixed, meantime please use deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml for review purposes.


pkg/apis/logging/v1/output_types.go
```diff
@@ -0,0 +1,109 @@
...
+// OutputType is a union of supported output types.
+//
+// It always includes the required a `type` string and optionally a struct field
+// with type-specific configuration.
+type OutputType struct {
+	// Type of output plugin, for example 'syslog'
+	//
+	// +unionDiscriminator
+	// +required
+	Type string `json:"type"`
```
periklis: Why do we need a string as union discriminator here? The types `Syslog` or whatever provide enough specificity, aren't they?. Following the k8s api for [corev1.Volume/VolumeSource](https://github.com/kubernetes/api/blob/master/core/v1/types.go#L38) here should be a good example.

pkg/apis/logging/v1/output_types.go
```diff
@@ -0,0 +1,109 @@
...
+	// 'tls' are used for output types that don't define their own URL scheme.
+	// Example:
+	//
+	//     { type: syslog, destination: tls://syslog.example.com:1234 }
+	//
+	// TLS with server authentication is enabled by the URL scheme, for
+	// example 'tls' or 'https'.  See `secretRef` for TLS client authentication.
+	//
+	// +required
+	Destination string `json:"destination"`
```
periklis: IMHO, destination is a configuration property of the output type. Although, it seems that we will have a connection string for all our output types, I can imagine more complicated destination properties, e.g. connecting through a sidecar to the target destination where your connection string could be a ref to a config map. Furthermore, it makes the YAML output more coherent, e.g.:
```
ouput:
   name: "my-target-syslog"
   syslog:
      destination: "tls://syslog.example.com:1234"
      facility: "whatever"
```

ewolinetz:
@alanconway reminder to rebase :)

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,203 @@
...
+                          secret name must be unique.
+                        type: string
+                    type: object
+                  syslog:
+                    description: Syslog optional extra syslog-specific properties.
+                    properties:
+                      facility:
+                        description: Facility for outgoing syslog records. Special
+                          value 'COPY' means copy from incoming log record.
+                        enum:
```
ewolinetz: is the enum really necessary here?

pkg/apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -0,0 +1,76 @@
...
+
+	Spec   ClusterLogForwarderSpec   `json:"spec,omitempty"`
+	Status ClusterLogForwarderStatus `json:"status,omitempty"`
+}
+
+// ClusterLogForwarderSpec defines the desired state of ClusterLogForwarder
+type ClusterLogForwarderSpec struct {
+	// Outputs are named output destinations for log messages.
+	//
+	// +optional
```
ewolinetz: from an implementation perspective, what does having a `pipeline` defined without `outputs` net users?

pkg/apis/logging/v1/output_types.go
```diff
@@ -0,0 +1,109 @@
...
+
+// Output defines a destination for log messages.
+type Output struct {
+	// Name used to refer to the output from a `pipeline`.
+	//
+	// +required
+	Name string `json:"name"`
+
+	// OutputType is inlined with a required `type` and optional extra configuration.
+	OutputType `json:",inline"`
```
ewolinetz: i thought the k8s pattern is to have something like
```
Type           OutputType `json:"type"`
OneOfTheTypes  TheType `json:"oneOfTheTypes"`
AnotherType    ...
```

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,203 @@
...
+                      secret contains keys `tls.crt`, `tls.key` and `ca.crt`. Output
+                      types with password authentication will use keys `password`
+                      and `username`, not the exposed 'username@password' part of
+                      the `destination` URL."
+                    properties:
+                      name:
+                        description: Name is unique within a namespace to reference
+                          a secret resource.
+                        type: string
+                      namespace:
```
jcantrill: Consider omitting until such time as we actually support it

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,203 @@
...
+                        description: Namespace defines the space within which the
+                          secret name must be unique.
+                        type: string
+                    type: object
+                  syslog:
```
jcantrill: @ewolinetz any elasticsearch or forward specific things we need in the initial release?

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,203 @@
...
+                      By default, all available log streams are forwarded.
+                    items:
+                      enum:
+                      - Application
+                      - Infrastructure
+                      - Audit
+                      type: string
+                    type: array
+                  name:
+                    description: Name of the pipeline. Optional, needed for patch
```
jcantrill: Consider making this required.  it makes it easier to identify in status.

deploy/crds/logging.openshift.io_clusterloggings_crd.yaml
```diff
@@ -0,0 +1,795 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
jcantrill: we dont need this from the sdk as we don't deploy using this mechanism but via OLM.  the content needs to be merged into the OLM manifest

deploy/crds/logging.openshift.io_collectors_crd.yaml
```diff
@@ -0,0 +1,107 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
jcantrill: same here

deploy/crds/logging.openshift.io_logforwardings_crd.yaml
```diff
@@ -0,0 +1,188 @@
+apiVersion: apiextensions.k8s.io/v1beta1
```
jcantrill: same

pkg/apis/logging/v1/output_types.go
jcantrill: disagree here.  the idea is top level fields are generic and should apply to any output followed by type specific fields.  In this case 'destination' is applicable to every output

pkg/apis/logging/v1/output_types.go
jcantrill: Agree.  this seems odd to me... The pattern i have seen elsewhere is as described by @ewolinetz comment though carrying the type in the containing struct may allow easier decisions of its type and the things i can/need to do

pkg/apis/logging/v1/output_types.go
```diff
@@ -0,0 +1,109 @@
...
+	// Has no function other than a marker to help avoid configuration mistakes.
+	//
+	// +optional
+	Insecure bool `json:"insecure,omitempty"`
+
+	// Reconnect configures how the output handles connection failures.
+	// Auto-reconnect is enabled by default.
+	//
+	// +optional
+	Reconnect *Reconnect `json:"reconnect,omitempty"`
```
jcantrill: Consider omitting if we are unable to support initially with the GA release

pkg/apis/logging/v1/output_types.go
jcantrill: One argument would be that a given output is only one type. Which one is correct if multiple pointers are defined?

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
ewolinetz: I don't believe so unless we know for sure there is something specific we _need_ to be able to configure out of the gate

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
ewolinetz: :+1: also if its needed for patch updates it would make sense to make it required

pkg/apis/logging/v1/output_types.go
ewolinetz: @jcantrill I believe that's what the `+unionDiscriminator` is for.

I can see both sides of this -- we initially followed this pattern [1] but following the VolumeSource pattern makes this cleaner


[1] https://github.com/kubernetes/api/blob/master/apps/v1/types.go#L78

pkg/apis/logging/v1/output_types.go
alanconway: Strongly disagree - +1 Jeff. The goal is to avoid the anti-pattern adopted by most logging solutions where you can't do anything without learning type-specific details.  E.g.
```
output:
  type: kafka
  url: tls://foobar.com/my-topic
  secretRef: {...}
output:
  type: fluent
  url:tls://xyz.com:1234
  secretRef: {...} 
```  
You should not be required to learn that kafka expects TLS secrets as env. vars. naming PEM format cert files, but fluent expects blah blah blah. Things that are effectively the same for all output types should be configured the same in our API. Likewise for reconnect configuration when we add it - are reconnect intervals seconds or milliseconds or blah blah. It's all the same and we can provide common config for it.

pkg/apis/logging/v1/output_types.go
alanconway: @periklis there is a bunch of discussion on unions in the k8s world and contradictory examples, I *think* this is the "winning" pattern.  For our purposes I like it because it handles the empty case (which is our common case) better, i.e. I think this:
```
output:
  type: fluent
  url: tcp://foobar.org
```
Reads better than this:
```
output: 
  fluent: {}
  url: tcp://foobar.org
```

pkg/apis/logging/v1/output_types.go
alanconway: The two are equivalent from the perspective of the resulting k8s API.  Output is the most complex and interesting type in the package so to me it makes sense for it to be its own type. If we want to re-use it in future (e.g. for a free-standing shared Output API object) then this also avoids duplication. However, I can inline it for now if that's what folks are more comfortable with, and we can reconsider if/when we need to re-use it. Thoughts?

alanconway: Also I think a separate type clarifies that the "type" field and the following "optional" fields are in fact a single "union" object, and facilitates writing methods that treat it as such - you can of course glom such methods onto the containing object but to me it feels less clear.

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,203 @@
...
+          type: object
+        spec:
+          description: ClusterLogForwarderSpec defines the desired state of ClusterLogForwarder
+          properties:
+            outputs:
+              description: Outputs are named output destinations for log messages.
+              items:
+                description: Output defines a destination for log messages.
+                properties:
+                  destination:
```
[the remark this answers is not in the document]
alanconway: Yes, out of date - will fix.

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,203 @@
...
+                          re-connect.  This is simple short-hand for configuring the
+                          output to a given level of reliability.  The exact meaning
+                          depends on the output `type`.  The default is determined
+                          by the output type.
+                        enum:
+                        - Unreliable
+                        - Retry
+                        type: string
+                    type: object
+                  secret:
```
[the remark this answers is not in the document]
alanconway: secretRef, fixed.

pkg/apis/logging/v1/cluster_log_forwarder_types.go
alanconway: Nothing, will make that required non-empty

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: It gives us explicit openapi validation, which I think is better than hidden validation code or failing late when something tries to parse the string as a facility. Did you have some other approach in mind?

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: done.

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: +1

deploy/crds/logging.openshift.io_clusterloggings_crd.yaml
alanconway: +1 my use of operator-sdk was naive, will figure out how to properly integrate this with OLM manifests before I merge. https://issues.redhat.com/browse/LOG-664

pkg/apis/logging/v1/output_types.go
alanconway: +1

pkg/apis/logging/v1/output_types.go
jcantrill: > The two are equivalent from the perspective of the resulting k8s API. Output is the most complex and interesting type in the package so to me it makes sense for it to be its own type. If we want to re-use it in future (e.g. for a free-standing shared Output API object) then this also avoids duplication. However, I can inline it for now if that's what folks are more comfortable with, and we can reconsider if/when we need to re-use it. Thoughts?

Can you provide a yaml sample to what this looks like?  I can't seem to wrap my brain around the assembly of the golang structs.  Keeping Output as a separate struct makes sense to me given our intention to maybe roll it out in future as standalone

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
jcantrill: I will ask the question given what I saw working with eclipse plugins and apis:  If we get out of sync with the possible values could that not lead to a validation failure because we are behind a version for a given syslog spec. This means we would unnecessarily reject a value.  Maybe not likely but just a thought.

pkg/apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -0,0 +1,97 @@
...
+	State   string `json:"state,omitempty"`
+	Reason  string `json:"reason,omitempty"`
+	Message string `json:"message,omitempty"`
+
+	Conditions  []Condition `json:"conditions,omitempty"`
+	LastUpdated metav1.Time `json:"lastUpdated,omitempty"`
+}
+
+type Condition struct {
+	Type        string      `json:"typ,omitempty"`
```
jcantrill: typo.... 'type' not 'typ'

pkg/apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// FIXME(alanconway) review status - should have state/reason enums?
+
+type ClusterLogForwarderStatus struct {
+	State       string      `json:"state,omitempty"`
+	Reason      string      `json:"reason,omitempty"`
+	Message     string      `json:"message,omitempty"`
+	LastUpdated metav1.Time `json:"lastUpdated,omitempty"`
+	Inputs      []SubStatus `json:"outputs,omitempty"`
```
jcantrill: inputs

pkg/apis/logging/v1/output_types.go
```diff
@@ -0,0 +1,114 @@
...
+	// +optional
+	// Reconnect *Reconnect `json:"reconnect,omitempty"`
+}
+
+// +kubebuilder:validation:Enum=Unreliable;Retry
+type Reliability string
+
+const (
+	// Unreliable may drop data after a reconnect (at-most-once).
+	Unreliable = "Unreliable"
```
jcantrill: I thought the pattern for enums was like https://github.com/openshift/cluster-logging-operator/blob/master/pkg/apis/logging/v1alpha1/forwarding_types.go#L12-L15

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
ewolinetz: i'm not sure we even need to validate this, if a user specifies a value we should expect that they know which value they actually want

pkg/apis/logging/v1/cluster_log_forwarder_types.go
ewolinetz: s/json:"outputs,/json:"inputs,/

pkg/apis/logging/v1/output_types.go
ewolinetz: +1 that is the pattern we have seen used in k8s

pkg/apis/logging/v1/output_types.go
alanconway: YAML is the same either way, the difference only exists in the Go code, to group the fields that are part of the "union" separately from the rest of the Output struct. They don't have to be grouped in the YAML:
```
output:
  type: syslog
  url: tcp://myhost:1234
  syslog:
     severity: critical
```

deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: I did hesitate over such a large enum - you've convinced me. 

Syslog severities and facilities are defined in the standard as integers, different syslog impls use a mostly-but-not-entirely-consistent set of spellings for names.  Instead of an advertised enum, we should probably accept integers and all capitalizations the common symbolic names.

IMO we should still check this during validation and normalize to whatever representation is used by the syslog library that we use internally, but I'll kill the openapi enum.

pkg/apis/logging/v1/output_types.go
alanconway: I can switch to lower-hyhenated if that's the way we have been doing things. I suspect it's an older convention, grepping around it looks like almost everything that uses the kubebuilder:validation:Enum marker uses MixedCaps. What's best for us?

The  kubebuilder book and k8s examples use MixedCaps
```
sigs.k8s.io/controller-tools/testData/pkg/apis/fun/v1alpha1/toy_types.go:	// +kubebuilder:validation:Enum=Lion,Wolf,Dragon
./kubernetes-sigs/controller-tools/pkg/crd/testdata/cronjob_types.go:// +kubebuilder:validation:Enum=Allow;Forbid;Replace
````
so does the elasticsearch operator:
```
github.com/openshift/elasticsearch-operator/pkg/apis/logging/v1/elasticsearch_types.go:// +kubebuilder:validation:Enum=FullRedundancy;MultipleRedundancy;SingleRedundancy;ZeroRedundancy
```
Grepping my local ~/go/src I find a few lowercase_underscore (including bazai) but no lowercase-hypen. The majority of validation:Enum use MixedCaps:
```
[aconway@grommit src]$ find -name '*.go' -type f | xargs grep 'kubebuilder:validation:Enum:*=[a-z]' 2> /dev/null
./github.com/banzaicloud/logging-operator/pkg/sdk/model/output/format.go:	// +kubebuilder:validation:Enum=out_file;json;ltsv;csv;msgpack;hash;single_value
./github.com/openshift/installer/vendor/github.com/metal3-io/baremetal-operator/pkg/apis/metal3/v1alpha1/baremetalhost_types.go:	// +kubebuilder:validation:Enum=registration error;inspection error;provisioning error;power management error
./github.com/knative/eventing-contrib/github/pkg/apis/sources/v1alpha1/githubsource_types.go:	// +kubebuilder:validation:Enum=check_suite,commit_comment,create,delete,deployment,deployment_status,fork,gollum,installation,integration_installation,issue_comment,issues,label,member,membership,milestone,organization,org_block,page_build,ping,project_card,project_column,project,public,pull_request,pull_request_review,pull_request_review_comment,push,release,repository,status,team,team_add,watch

[aconway@grommit src]$ find -name '*.go' -type f | xargs grep 'kubebuilder:validation:Enum:*=[A-Z]' 2> /dev/null | wc -l
24
```

Comment (jcantrill) [reply pkg/apis/logging/v1/output_types.go]: "Lower hyphenated" is not the issue; Its defining struct values using an aliased type like, note LogSourceType: 
```
type LogSourceType string

const (
	LogForwardingKind string = "LogForwarding"

	//LogSourceTypeApp are container logs from non-infra structure containers
	LogSourceTypeApp LogSourceType = "logs-app"
```

Comment (alanconway) [reply pkg/apis/logging/v1/output_types.go]: Aha, sorry. Yes will fix.

File: deploy/crds/logging.openshift.io_clusterlogforwarders_crd.yaml [hunk 18]
```diff
@@ -0,0 +1,203 @@
...
+                    type: string
+                  insecure:
+                    description: Insecure must be true for intentionally insecure
+                      outputs. Has no function other than a marker to help avoid configuration
+                      mistakes.
+                    type: boolean
+                  name:
+                    description: Name used to refer to the output from a `pipeline`.
+                    type: string
+                  reconnect:
```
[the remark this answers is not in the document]
alanconway: I believe we can, but I've removed for now to be added when implemented. It may be we can just implement a default reconnect strategy that satisfies most cases, and leave anything more complicated to the specifics.

pkg/apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -0,0 +1,97 @@
...
+// ClusterLogForwarderList is a list of ClusterLogForwarders
+//
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+type ClusterLogForwarderList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []ClusterLogForwarder `json:"items"`
+}
+
+// FIXME(alanconway) review status - should have state/reason enums?
```
periklis: I would go for an enum even if we have one. If you use the type system from the start, it makes it easy to add switch/case blocks when you extend the enum.

pkg/apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -0,0 +1,97 @@
...
+type ClusterLogForwarderStatus struct {
+	State       string      `json:"state,omitempty"`
+	Reason      string      `json:"reason,omitempty"`
+	Message     string      `json:"message,omitempty"`
+	LastUpdated metav1.Time `json:"lastUpdated,omitempty"`
+	Inputs      []SubStatus `json:"inputs,omitempty"`
+	Outputs     []SubStatus `json:"outputs,omitempty"`
+}
+
+type SubStatus struct {
```
periklis: Why not just calling it Status?

pkg/apis/logging/v1/output_types.go
```diff
@@ -0,0 +1,114 @@
...
+
+	// Resend "in doubt" data after a reconnect. May cause duplicates (at-least-once).
+	// May enable buffering, blocking and/or acknowledgment features of the output type.
+	Resend = "Resend"
```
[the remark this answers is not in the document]
alanconway: Yes, fixed.

pkg/apis/logging/v1/cluster_log_forwarder_types.go
alanconway: Agreed, will define this explicitly in the API as I do the implementation - the existing impl already has some lists of values.

pkg/apis/logging/v1/cluster_log_forwarder_types.go
alanconway: I can't think of a good reason. Done.

alanconway:
Closed by mistake, I don't have permission to re-open. Started a new PR: https://github.com/openshift/cluster-logging-operator/pull/427

PR #394, opened 2020-03-05
Bug 1810709: fix Dockerfile to be buildable by buildah

jcantrill:
fix Dockerfile to be buildable by buildah

https://bugzilla.redhat.com/show_bug.cgi?id=1810709

@bparees


jcantrill:
Ref https://bugzilla.redhat.com/show_bug.cgi?id=1809567

.dockerignore
```diff
@@ -0,0 +1,4 @@
+.git/*
+test/*
+*/*_test.go
```
adambkaplan: Since you're using multistage Dockerfile builds, I would not exclude test files. These may be inadvertently ignored by our CI (now or in the future). Ignoring .git and _output makes sense.

Dockerfile
```diff
@@ -12,12 +12,14 @@ RUN INSTALL_PKGS=" \
...
-COPY --from=builder _output/bin/cluster-logging-operator /usr/bin/
+COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/_output/bin/cluster-logging-operator /usr/bin/
 COPY scripts/* /usr/bin/scripts/
 RUN mkdir -p /usr/share/logging/
 COPY files/ /usr/share/logging/
-COPY manifests/4* /manifests/
-COPY manifests/cluster-logging.package.yaml /manifests/
+
+COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/manifests /manifests
+RUN rm /manifests/art.yaml
```
adambkaplan: You can keep the glob since it looks like having a flat directory structure in `/manifests` is the desired behavior. The COPY command can be squeezed into one instruction:

```
COPY manifests/cluster-logging.package.yaml manifests/4* /manifests/
```

jcantrill: we actually dont want a flat directory structure

adambkaplan: then disregard - what you have here makes sense.

hack/testing/test-010-deploy-via-olm-minimal.sh
```diff
@@ -22,7 +23,7 @@ version=$(basename $(find $manifest -type d | sort -r | head -n 1))
 cleanup(){
   local return_code="$?"
   set +e
-  oc logs -n ${NAMESPACE} deployment/cluster-logging-operator > $ARTIFACT_DIR/cluster-logging-operator.log 2>&1 ||:
+  gather_logging_resources ${NAMESPACE} $test_artifactdir
```
alanconway: The lines after gather_logging_resources are executed with `set -e` whereas in the old code they were executed with `set +e`. Not sure if this is intentional or relevant?

hack/testing/utils
```diff
@@ -76,6 +76,25 @@ try_until_text() {
   log::error "try_until_text expired for '$cmd' == '$expected'"
   return 1
 }
+
+gather_logging_resources() {
+  set +e
```
alanconway: Why bother with `set +e` when each line ends "|| :" which ignores errors anyway?

Dockerfile
```diff
@@ -12,12 +12,14 @@ RUN INSTALL_PKGS=" \
...
     chmod og+w /tmp/ocp-clo
-COPY --from=builder _output/bin/cluster-logging-operator /usr/bin/
+COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/_output/bin/cluster-logging-operator /usr/bin/
 COPY scripts/* /usr/bin/scripts/
 RUN mkdir -p /usr/share/logging/
 COPY files/ /usr/share/logging/
-COPY manifests/4* /manifests/
-COPY manifests/cluster-logging.package.yaml /manifests/
+
+COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/manifests /manifests
```
alanconway: Hard-coded path is a bit ugly, will break if the surrounding directory tree is ever moved. Not familiar with Dockerfiles, if there is  something like $PWD or $CURDIR or $(realpath ...) that works inside a Dockerfile that would be better - if not then may be the hard-coded path is the best we can do.

jcantrill: You can define WORKDIR but this is the behavior difference between the previous build tool and this one; it is honored differently between layers.  Im less concerned as the likelyhood of this paths changes are small.

hack/testing/utils
jcantrill: only to ensure that any error in this logic does not fail a test if it otherwise passes.

PR #401, opened 2020-03-10
minor updates to collector reconciliation

jcantrill:
This PR cleans up some inconsistencies in collector reconciliation

### DIFF
--- a/pkg/k8shandler/collection.go
+++ b/pkg/k8shandler/collection.go
@@ -94,8 +94,8 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection(proxyConfi
 
 			// remove our finalizer from the list and update it.
 			collectorServiceAccount.ObjectMeta.Finalizers = utils.RemoveString(collectorServiceAccount.ObjectMeta.Finalizers, metav1.FinalizerDeleteDependents)
-			err = clusterRequest.Create(collectorServiceAccount)
-			if err != nil && !errors.IsAlreadyExists(err) {
+			if err = clusterRequest.Update(collectorServiceAccount); err != nil {
+				logger.Warnf("Unable to update the collector serviceaccount %q finalizers: %v", collectorServiceAccount.Name, err)
 				return nil
 			}
 		}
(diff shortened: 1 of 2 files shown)


pkg/k8shandler/reconciler.go
```diff
@@ -78,7 +78,7 @@ func ReconcileForLogForwarding(forwarding *logforwarding.LogForwarding, requestC
 
 	// Reconcile Collection
 	if err = clusterLoggingRequest.CreateOrUpdateCollection(proxyConfig); err != nil {
-		return fmt.Errorf("Unable to create or update collection for %q: %v", clusterLoggingRequest.cluster.Name, err)
+		return fmt.Errorf("Unable reconcile collection for %q: %v", clusterLoggingRequest.cluster.Name, err)
```
syedriko: to reconcile?

pkg/k8shandler/collection.go
```diff
@@ -94,8 +94,8 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection(proxyConfi
 
 			// remove our finalizer from the list and update it.
 			collectorServiceAccount.ObjectMeta.Finalizers = utils.RemoveString(collectorServiceAccount.ObjectMeta.Finalizers, metav1.FinalizerDeleteDependents)
-			err = clusterRequest.Create(collectorServiceAccount)
-			if err != nil && !errors.IsAlreadyExists(err) {
+			if err = clusterRequest.Update(collectorServiceAccount); err != nil {
+				logger.Warnf("Unable to update the collector serviceaccount %q finalizers: %v", collectorServiceAccount.Name, err)
```
ewolinetz: do we know what this block is actually trying to accomplish? if it wasn't correct before and fluentd worked, do we even need any of this at all?

jcantrill: I believe this was originally added by @nhosoi to address a bug where fluentd was not being terminated in a timely fashion and it was related to the finalizers.  We would need to investigate the commit hash to find the bug.

PR #403, opened 2020-03-10
Removed duplicate configv1.AddToScheme

syedriko:
Removed duplicate configv1.AddToScheme()


ewolinetz:
will fail until we have the updated eo manifest deployment methods
/lgtm

PR #407, opened 2020-03-11
Updating EO deployment to not require vendor dir

ewolinetz:
Updating our vendoring of EO to only pull `.go` files and pull the manifest file from github rather than rely on the vendored manifest.


jcantrill:
closing in favor of #408

PR #411, opened 2020-03-13
Bug 1810710: Make useable by buildah

jcantrill:
https://bugzilla.redhat.com/show_bug.cgi?id=1810710

backport of https://github.com/openshift/cluster-logging-operator/pull/394

@bparees


bparees:
overriding BZ req, the changes to this dockerfile are sufficiently validated by the tests to know we didn't totally break it (another option would have been to verify the master BZ based on that.)

PR #412, opened 2020-03-13
LOG-448 Convert pod labels to array of key=value

jcantrill:
This PR flattens labels into array of key/value pairs to avoid key field explosion in ES

Ref: https://issues.redhat.com/browse/LOG-448

### DIFF
--- a/hack/testing/test-050-collection.sh
+++ b/hack/testing/test-050-collection.sh
@@ -1,0 +1,85 @@
+#!/bin/bash
+
+set -e
+
+source "$(dirname $0)/../common"
+
+start_seconds=$(date +%s)
+
+LOG_LEVEL=${LOG_LEVEL:-}
+TEST_DIR=${TEST_DIR:-'./test/e2e/collection/*/'}
+test_name=$(basename ${BASH_SOURCE[0]} .sh)
+test_artifact_dir="${ARTIFACT_DIR:-"$repo_dir/_output"}/$test_name"
+if [ ! -d $test_artifact_dir ] ; then
+  mkdir -p $test_artifact_dir
+fi
+
+cleanup(){
+  local return_code="$?"
+  set +e
+  log::info "Running cleanup"
+  end_seconds=$(date +%s)
+  runtime="$(($end_seconds - $start_seconds))s"
+  
+  for item in "ns/openshift-logging" "ns/openshift-operators-redhat"; do
+    oc delete $item --wait=true --ignore-not-found --force --grace-period=0
+  done
+  for item in "ns/openshift-logging" "ns/openshift-operators-redhat"; do
+    try_until_failure "oc get ${item}" "$((1 * $minute))"
+  done
+  
+  set -e
+  exit ${return_code}
+}
+
+if [ "${DO_CLEANUP:-true}" == "true" ] ; then
+  trap cleanup exit
+fi
+
+failed=0
+for dir in $(ls -d $TEST_DIR); do
+  log::info "=========================================================="
+  log::info "Starting $test_name: $dir"
+  log::info "=========================================================="
+  
+  if [ "${DO_SETUP:-true}" == "true" ] ; then
+    log::info "Deploying elasticsearch-operator"
+    deploy_elasticsearch_operator
+    
+    log::info "Deploying cluster-logging-operator"
+    deploy_clusterlogging_operator
+  fi
+  artifact_dir=$test_artifact_dir/$(basename $dir)
+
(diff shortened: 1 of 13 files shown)


pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -359,6 +359,16 @@ var _ = Describe("Generating fluentd config", func() {
 					alt_key kubernetes.event.metadata.uid
 					alt_tags "#{ENV['GENID_ALT_TAG'] || 'kubernetes.var.log.containers.logging-eventrouter-*.** kubernetes.var.log.containers.eventrouter-*.** kubernetes.var.log.containers.cluster-logging-eventrouter-*.** kubernetes.journal.container._default_.kubernetes-event'}"
 				</filter>
+				#flatten labels to prevent field explosion in ES
+				<filter ** >
+					@type record_transformer
+					enable_ruby true
+					<record>
+						kubernetes ${!record['kubernetes'].nil? ? record['kubernetes'].merge({"labels_alt": (record['kubernetes']['labels']||{}).map{|k,v| "#{k}=#{v}"}}) : {} }
```
periklis: Does `labels_alt` reflect the final key name?

test/e2e/collection/fluentd/flatten_labels_test.go
```diff
@@ -0,0 +1,104 @@
...
+			e2e.Cleanup()
+		}
+	}, helpers.DefaultCleanUpTimeout)
+
+	It("should remove 'kubernetes.labels' and create 'kubernetes.labels_alt' with an array of 'kubernetes.labels'", func() {
+		Expect(e2e.LogStore.HasApplicationLogs(helpers.DefaultWaitForLogsTimeout)).To(BeTrue(), "Expected to find stored application logs")
+
+		//verify infra namespaces are not stored to their own index
+		store := helpers.FluentReceiverLogStore{Framework: e2e}
+		if response, err := store.ApplicationLogs(helpers.DefaultWaitForLogsTimeout); err != nil {
```
periklis: IMHO it improves readability by moving the call to `ApplicationLogs` before the If statement so that you can drop the else clause.

test/helpers/fluentd.go
```diff
@@ -17,9 +17,9 @@ import (
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
 )
 
-type fluentReceiverLogStore struct {
+type FluentReceiverLogStore struct {
```
periklis: Why not keeping this private and enhance the LogStore interface by `ApplicationLogs()`. Just put `panic("Not implemented")` for the missing implementations in Elastic and Syslog

pkg/generators/forwarding/fluentd/fluent_conf_test.go
jcantrill: I think it has to for atleast a release (or longer).  I defer to @lukas-vlcek but after and index is created we are unable to modify the field mapping.  Additionally, depending on time of upgrade we may have records with the map and others with the array so unless we intend to reindex and transform there is way to support both 'types'.  I'm open to alternate suggestions for naming

hack/testing/test-050-collection.sh
```diff
@@ -0,0 +1,85 @@
...
+}
+
+if [ "${DO_CLEANUP:-true}" == "true" ] ; then
+  trap cleanup exit
+fi
+
+failed=0
+for dir in $(ls -d $TEST_DIR); do
+  log::info "=========================================================="
+  log::info "Staring $test_name: $dir"
```
periklis: Nit. But a *Staring* test is neither a *stargazing* nor a *starting* test ;)

test/e2e/collection/fluentd/flatten_labels_test.go
```diff
@@ -0,0 +1,104 @@
...
+			Fail(fmt.Sprintf("Failed waiting for component %s to be ready: %v", helpers.ComponentTypeCollector, err))
+		}
+
+	})
+
+	AfterEach(func() {
+		doCleanup := strings.TrimSpace(os.Getenv("DO_CLEANUP"))
+		if doCleanup == "" || strings.ToLower(doCleanup) == "true" {
+			e2e.Cleanup()
+		}
```
periklis: IMHO moving the `DO_CLEANUP` check inside `e2e.Cleanup` would enable this behaviour for all test suites

jcantrill: a version of it is already there.  Ill make the adjustments

PR #416, opened 2020-03-16
Bug 1813085: Enable monitoring of logging ns

jcantrill:
https://bugzilla.redhat.com/show_bug.cgi?id=1813085

modified backport of https://github.com/openshift/cluster-logging-operator/pull/324

cc @bparees


bparees:
overriding, bug is valid, it was fixed in 4.3.0 not 4.3.z.

PR #420, opened 2020-03-17
Fix for LOG-676, prevent a hotlooping reconcile for a terminated curator job

syedriko:
This fix replaces the changing lastTransitionTime for Terminated container statuses with the actual termination time (FinishedAt), thus preventing a hotloop reconcile over status that is known to not change.

### DIFF
--- a/pkg/k8shandler/status.go
+++ b/pkg/k8shandler/status.go
@@ -291,7 +291,7 @@ func (clusterRequest *ClusterLoggingRequest) getPodConditions(component string) 
 						Status:             v1.ConditionTrue,
 						Reason:             containerStatus.State.Terminated.Reason,
 						Message:            containerStatus.State.Terminated.Message,
-						LastTransitionTime: metav1.Now(),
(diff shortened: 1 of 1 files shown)


pkg/k8shandler/status.go
```diff
@@ -291,7 +291,7 @@ func (clusterRequest *ClusterLoggingRequest) getPodConditions(component string)
 						Status:             v1.ConditionTrue,
 						Reason:             containerStatus.State.Terminated.Reason,
 						Message:            containerStatus.State.Terminated.Message,
-						LastTransitionTime: metav1.Now(),
+						LastTransitionTime: containerStatus.State.Terminated.FinishedAt,
```
jcantrill: what happens when the state is other then "terminated"? Isn't it possible we are checking this condition when the pod is still running

ewolinetz: @jcantrill this block is only when the container status state is terminated. there are other blocks for other conditions

ewolinetz:
@syedriko what about the waiting condition?

PR #421, opened 2020-03-17
Bug 1813810: Updating resource update policy and adding unit test to verify

ewolinetz:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1813810

### DIFF
--- a/pkg/utils/resources.go
+++ b/pkg/utils/resources.go
@@ -14,27 +14,24 @@ func CompareResources(current, desired v1.ResourceRequirements) (bool, v1.Resour
 
 	changed := false
 
+	// Check CPU limits
 	if desired.Limits.Cpu().Cmp(*current.Limits.Cpu()) != 0 {
-		current.Limits[v1.ResourceCPU] = *desired.Limits.Cpu()
 		changed = true
 	}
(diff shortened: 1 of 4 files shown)


pkg/utils/resources.go
```diff
@@ -14,27 +14,24 @@ func CompareResources(current, desired v1.ResourceRequirements) (bool, v1.Resour
 
 	changed := false
 
+	// Check CPU limits
 	if desired.Limits.Cpu().Cmp(*current.Limits.Cpu()) != 0 {
-		current.Limits[v1.ResourceCPU] = *desired.Limits.Cpu()
```
ewolinetz: @jcantrill discussion around this... 
i think the reason we did this previously is because we wanted to keep some sane values in the case of ES -- this does away with that and now a user would need to be explicit if they wanted to have a default value for limit or request... thoughts?

jcantrill: I recall in places we explicitly set the cpu limit but I don't know why;  we have always had the stance of:
* Allow unbounded cpu usage because cpu helps with ES processing records
* Peg memory limit and request to ensure we are able to land the ES node and that it doesn't get booted

PR #422, opened 2020-03-17
Adding conditions for when log collection is missing destinations

ewolinetz:
Example output for when we have no logstore nor logforward defined

when invalid:
```
status:
  clusterConditions:
  - lastTransitionTime: "2020-03-17T19:42:25Z"
    message: Collectors are defined but there is no defined LogStore of LogForward
      destinations
    reason: No defined logstore destination
    status: "True"
    type: CollectorDeadEnd
```

when valid we remove the condition

### DIFF
--- a/pkg/k8shandler/forwarding.go
+++ b/pkg/k8shandler/forwarding.go
@@ -13,6 +13,7 @@ import (
 	"github.com/openshift/cluster-logging-operator/pkg/generators/forwarding"
 	"github.com/openshift/cluster-logging-operator/pkg/logger"
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
+	v1 "k8s.io/api/core/v1"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/k8shandler/forwarding.go
```diff
@@ -48,7 +49,26 @@ func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig() (config s
...
 	clusterRequest.ForwardingSpec = clusterRequest.normalizeLogForwarding(clusterRequest.cluster.Namespace, clusterRequest.cluster)
 	generator, err := forwarding.NewConfigGenerator(clusterRequest.cluster.Spec.Collection.Logs.Type, clusterRequest.includeLegacyForwardConfig(), clusterRequest.includeLegacySyslogConfig())
-	return generator.Generate(&clusterRequest.ForwardingSpec)
+	generatedConfig, err := generator.Generate(&clusterRequest.ForwardingSpec)
+
+	if err != nil {
+		return "",
+			clusterRequest.UpdateCondition(
+				logging.CollectorDeadEnd,
+				"Collectors are defined but there is no defined LogStore of LogForward destinations",
```
jcantrill: Isn't this condition specifically related the our managed logStore not being defined?  This would mean this is misleading in any other context as generally for LogForwarding, we  have no idea if the destination is a dead end.  IMO, we should explicitly reference it is CLO managed logstore and maybe rename the ClusteConditionType accordingly

ewolinetz: to get an error here we have neither a LogForward destination defined nor a CLO logstore...

ewolinetz:
/hold
since we're spinning

PR #425, opened 2020-03-19
Add kafka output support for ClusterLogForwarder API

periklis:
### Motivation
This PR provides initial support for LogForwarding (atm avail. in Tech Preview) to a third-party Kafka deployment. ~The following limitations apply due to the capabilities of the Logforwarding tech preview:~
- ~Use of a single Topic to write all logs (infra, audit, app).~
- ~Server side TLS for kafka brokers only.~
- Requires a kafka server side configuration for `message.max.bytes` above 8 megabytes.

Moreover, the E2E kafka resources are inspired by [Yolean/kubernetes-kafka](/Yolean/kubernetes-kafka).

**Depends on:** openshift/origin-aggregated-logging#1853 and  #427


sichvoge:
@periklis What's the follow up on this to close on the limitations if necessary? Is it something Alan and Jeff (and maybe me) need to make a decision?

periklis:
@sichvoge The limitations will be closed when #388 will land.

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -596,3 +597,26 @@ const storeSyslogTemplate = `{{- define "storeSyslog" }}
...
+    brokers {{.Brokers}}
+    topic_key {{.Topic}}
+    default_topic {{.Topic}}
+    <format>
+       @type json
+    </format>
+    <buffer {{.Topic}}>
+        @type file
+        path '{{.BufferPath}}'
+        flush_interval "#{ENV['ES_FLUSH_INTERVAL'] || '1s'}"
```
jcantrill: Remove ref to env vars

test/helpers/elasticsearch.go
```diff
@@ -132,6 +132,10 @@ func (es *ElasticLogStore) HasAuditLogs(timeToWait time.Duration) (bool, error)
 	return true, err
 }
 
+func (es *ElasticLogStore) ClusterLocalEndpoint() string {
```
jcantrill: What does this mean "ClusterLocalEndpoint"

periklis: It means you provide a service object for your store and this method gives you the cluster local DNS name e.g. `kafka.openshift-logging.svc.cluster.local`

pkg/generators/forwarding/fluentd/templates.go
periklis: All env vars or `ES_FLUSH_INTERVAL` only?

periklis: All `ES_` prefixed env vars are removed. I kept the `BUFFER_` prefixed env vars which are set in run.sh

alanconway:
/hold
need to hold merging till the base PR merges.

pkg/generators/forwarding/fluentd/kafka_conf.go
```diff
@@ -0,0 +1,74 @@
...
+//first connention only. The collector's kafka client fetches constantly an updated list
+//from kafka. These updates are not reconciled back to the collector configuration.
+//The list of brokers are populated from the Kafka OutputSpec `Brokers` field, a list of
+//valid URLs. If none provided the target URL from the OutputSpec is used as fallback.
+//Finally, if neither approach works the current collector process will be terminated.
+func (conf *outputLabelConf) Brokers() string {
+	parseBroker := func(b string) string {
+		url, err := url.Parse(b)
+		if err != nil {
+			log.Errorf("Failed to parse Kafka broker from output spec %q: %s", b, err)
```
alanconway, line 23: We should return errors back to the forwarding code so that informative status conditions can be set for the Output.
This is the first output type that generates errors so the code may not be well factored for this. Shout if you want me to help with designing a good error handling pattern for this.

pkg/generators/forwarding/fluentd/kafka_conf.go
```diff
@@ -0,0 +1,74 @@
...
+//The kafka topic is either extracted from the kafka OutputSpec `Topic` field in a multiple broker
+//setup or as a fallback from the OutputSpec URL if provided as a host path. Defaults to `topic`.
+func (conf *outputLabelConf) Topic() string {
+	if conf.Target.Kafka != nil && conf.Target.Kafka.Topic != "" {
+		return conf.Target.Kafka.Topic
+	}
+
+	url, err := url.Parse(conf.Target.URL)
+	if err != nil {
+		log.Fatalf("Failed to extract Kafka topic from output spec url %q: %s", conf.Target.URL, err)
```
alanconway, line 64: We definitely can't crash the operator for errors in an output - see the previous comment about error handling.

pkg/generators/forwarding/fluentd/kafka_conf.go
periklis: Good catch to re-evaluate this point with an extra test case (See [here]()). Actually we don't need to factor anything out because golang's `text/template` catches every error and will pass it back to the `template.Execute(...)` point of the caller. However, the `log.Fatal` below did not trigger this but it's possible with a `panic`.

Furthermore, regarding the line of code you point to, we should discuss how to handle informative errors. In the highlighted case the helper `parserBroker` function is applied to every broker. We can have valid brokers but also non valid ones on the list. Should we return a error if any of the brokers is broken or just log the error and proceed with a partial list?

pkg/generators/forwarding/fluentd/kafka_conf.go
periklis: See above. Let's close this thread.

pkg/generators/forwarding/fluentd/kafka_conf.go
alanconway: log.Fatal does os.Exit which is unstoppable :) Panic is OK, we should add a comment somewhere explaining how we use it,  so we don't get a newbie (or forgetful oldie) cut-pasting a panic into a place where it will cause a crash.
 
Outputs are evaluated at reconcile, marked with Conditions, and excluded if they don't have condition Ready=true.  So I think this code (and output-type-specific code in general) should return a Conditions object that can be applied or merged into the Output conditions.

The returned conditions might be like:
* [Ready=true,Degraded=false] - everything works
* [Ready=true,Degraded=true] - have some good and some bad brokers
* [Ready=false,Degraded=false] - total bust

Degrated=true and Ready=False conditions need a Reason and Message. I've been using a small, fixed set of general reasons and putting details (broker names etc.) in the Message. The Reasons are part of the API (Enum-like) so we should be careful with expanding that set. 

I've been composing compound messages when I have multiple things to report, e.g. a Pipeline Message might be something like "not all outputs ready: unready: [foo,bar,baz], degraded:[fred,bill,bob]" 

If we return Conditions they would be merged into the overall status of the Output.

pkg/generators/forwarding/fluentd/kafka_conf.go
alanconway: Added comment on error handling, not necessary to address to merge this PR, but something to revisit - I should probably add an interface that all output types need to implement so the error reporting rules are obvious.

alanconway: /lgtm
no need to act on the condition remarks to merge this, we can improve it later.

alanconway:
@periklis I saw this "Not mergeable" state on a previous PR,  @jcantrill  had to manually merge to get around it.

PR #423, opened 2020-03-17
normalize artifact dirs to write to separate places

jcantrill:
This PR attempts to normalize the artifact directories so artifacts are separated by test

PR #426, opened 2020-03-19
Adding default RetentionPolicies when nil

ewolinetz:


ewolinetz:
@jcantrill note: i wasn't able to still see indices created with this change... not sure if there's something else still missing but this does cover what we talked about on slack re: https://github.com/openshift/cluster-logging-operator/pull/423/files#diff-30b25a29a6d4bfa41c4be502b2ae34b6R54

PR #413, opened 2020-03-13
LOG-653: Generate CRDs from Go source.

alanconway:
Generate CRDs from Go source, in preparation for adding the ClusterLogForwarding CRD.
At this point we have too many CRDs to maintain by hand.

manifest/4.5/*.crd.yaml are now generated by `make generate` and checked in.
They should not have any semantic changes, but this is hard to see with `diff`.
I used https://github.com/homeport/dyff to verify the changes.

INTENTIONAL CHANGE: `proxySpec` -> `proxy`, this was an error in the original
hand-written CRD. The generated CRD matches the Go code.

Summary:
* Removed deploy/ directory - CRDs are generated into manifests/4.5
* `make generate` updates generated code and CRDs if sources have changed.
* `make regenerate` forces re-generation regardless of timestamps.
* clusterlogging_types.go: add comments and directives to produce an equivalent CRD
* forwarding_types.go: add comments and directives to produce an equivalent CRD

Signed-off-by: Alan Conway <aconway@redhat.com>

PR #427, opened 2020-03-20
LOG-653: Implement GA ClusterLogForwarder API

alanconway:
commit aaa44e6f58e633faa752d8f2c730730bf2be9568 (ga-cluster-log-forwarder, alanconway/ga-cluster-log-forwarder)
Author: Alan Conway <aconway@redhat.com>
Commit: Alan Conway <aconway@redhat.com>

    LOG-653: Define ClusterLogForwarder API
    
    ClusterLogForwarder API to replace tech. preview LogForwarding API.
    
    - Main schema: pkg/apis/logging/v1/cluster_log_forwarder_types.go
    - Output struct: pkg/apis/logging/v1/output_types.go
    - Plugin details: pkg/apis/logging/v1/outputs/syslog.go (others to come)
    - Generated CRD: manifests/4.5/cluster-loggings.crd.yaml


alanconway:
/hold 
Wait for https://github.com/openshift/cluster-logging-operator/pull/413 to merge and rebase.

ewolinetz:
@alanconway which commit should i be looking at for this work?

alanconway:
This is not complete - look for FIXME and TODO comments. It passes the Go unit tests but I don't expect it to pass e2e. Its a big merge so offering for review early.

alanconway:
Status: 
- refactored & buliding clean: code, unit tests, e2e tests
- unit tests: all pass
- e2e tests: expecting failures

Todo:
- pass e2e tests
- incorporate recent forwarding work (syslog, kafka) 
- additional tests: upgrade from tech preview, check coverage
- API doc, notes for doc writers.
- Clean up unused old code.

pkg/generators/forwarding/fluentd/generators.go
```diff
@@ -14,9 +15,10 @@ import (
 //ConfigGenerator is a config generator for fluentd
 type ConfigGenerator struct {
 	*generators.Generator
-	includeLegacyForwardConfig bool
-	includeLegacySyslogConfig  bool
-	useOldRemoteSyslogPlugin   bool
+	includeLegacyForwardConfig    bool
```
vimalk78, line 18: I hope we can refactor this like the GA API. Using a common part and a type specific part. Otherwise this gets overcrowded. For syslog there will be few more things added here.

pkg/constants/constants.go
```diff
@@ -19,7 +19,9 @@ const (
 	ElasticsearchPort = "9200"
 	FluentdName       = "fluentd"
 	KibanaName        = "kibana"
-	LogStoreService   = ElasticsearchFQDN + ":" + ElasticsearchPort
+	LogStoreURL       = "tls://" + ElasticsearchFQDN + ":" + ElasticsearchPort
```
jcantrill, line 22: My ignorance to protocol but is this equivalent to accessing ES using http?

pkg/constants/constants.go
```diff
@@ -19,7 +19,9 @@ const (
...
+	// Secret used by cluster logging components (collector and store).
```
jcantrill, line 23: This is not exactly correct.  The content of the secret may have the same entries but the ES output does not mount this secret

vimalk78:
```
waiting for deployment openshift-operators-redhat/elasticsearch-operator - remaining attempts: 2
waiting for deployment openshift-operators-redhat/elasticsearch-operator - remaining attempts: 1
waiting for deployment openshift-operators-redhat/elasticsearch-operator - remaining attempts: 0
[ERROR] PID 239: hack/test-e2e-olm.sh:11: `ELASTICSEARCH_OPERATOR_NAMESPACE=openshift-operators-redhat olm_deploy/scripts/operator-install.sh` exited with status 1.
[INFO] 		Stack Trace: 
[INFO] 		  1: hack/test-e2e-olm.sh:11: `ELASTICSEARCH_OPERATOR_NAMESPACE=openshift-operators-redhat olm_deploy/scripts/operator-install.sh`
[INFO]   Exiting with code 1.
```

/retest

pkg/constants/constants.go
ewolinetz: out of curiosity, do we need to specify `tls` doesn't fluentd already build its connector to ES using tls?

pkg/generators/forwarding/fluentd/generators.go
alanconway: These are deprecated features that will eventually go away and be replaced by the new-API equivalent. We don't want to spend time on them, but if the clutter gets in the way of developing new features it might be worth refactoring as you suggest.

test/matchers_test.go
```diff
@@ -0,0 +1,17 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	. "github.com/openshift/cluster-logging-operator/test"
+)
+
+var _ = Describe("Matchers", func() {
+	It("matches by line", func() {
+		Expect(" foo\nbar  \n\nbaz\n").To(MatchLines("foo \n\n bar\nbaz"))
```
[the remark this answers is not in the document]
alanconway: +1

test/helpers/framework.go
```diff
@@ -22,17 +22,17 @@ import (
...
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	e2eutil "github.com/openshift/cluster-logging-operator/test/e2e"
 )
 
 const (
-	clusterLoggingURI     = "apis/logging.openshift.io/v1/namespaces/openshift-logging/clusterloggings"
-	logforwardingURI      = "apis/logging.openshift.io/v1alpha1/namespaces/openshift-logging/logforwardings"
-	DefaultCleanUpTimeout = 60.0 * 2
+	clusterLoggingURI      = "apis/logging.openshift.io/v1/namespaces/openshift-logging/clusterloggings"
+	clusterlogforwarderURI = "apis/logging.openshift.io/v1alpha1/namespaces/openshift-logging/clusterlogforwarders"
```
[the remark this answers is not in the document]
alanconway: Not sure, open a new task for that. I'm only changing the version of the forwarder in this PR.

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
```diff
@@ -24,45 +24,45 @@ const (
...
 					},
-					Pipelines: []logforward.PipelineSpec{
-						logforward.PipelineSpec{
+					Pipelines: []logging.PipelineSpec{
+						logging.PipelineSpec{
 							Name:       "test-infra",
-							OutputRefs: []string{helpers.SyslogReceiverName},
-							SourceType: logforward.LogSourceTypeInfra,
+							OutputRefs: []string{"out"},
+							InputRefs:  []string{"Infrastructure"},
```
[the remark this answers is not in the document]
alanconway: +1

pkg/k8shandler/deprecated.go
```diff
@@ -0,0 +1,28 @@
...
+const (
+	//ForwardingAnnotation enable preview instance with value "enabled"
+	PreviewForwardingAnnotation = "clusterlogging.openshift.io/logforwardingtechpreview"
+	// UseOldRemoteSyslogPlugin use old syslog plugin (docebo/fluent-plugin-remote-syslog)
+	UseOldRemoteSyslogPlugin = "clusterlogging.openshift.io/useoldremotesyslogplugin"
+)
+
+// IsPreviewForwardingEnabled check if the tech-preview annotation is enabled.
+// For use in upgrade code to check if an existing preview deployment needs to be upgraded.
+func IsPreviewForwardingEnabled(cluster *logging.ClusterLogging) bool {
```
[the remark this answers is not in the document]
alanconway: No, but we do need a test case for upgrading a cluster that has deployed the preview. I have a separate story for that.

pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -56,54 +51,54 @@ var _ = Describe("Generating fluentd config", func() {
...
 
 	It("should exclude source to pipeline labels when there are no pipelines for a given sourceType (e.g. only logs-app)", func() {
-		forwarding = &logging.ForwardingSpec{
+		forwarder = &logging.ClusterLogForwarderSpec{
 			Outputs: []logging.OutputSpec{
 				{
-					Type:     logging.OutputTypeForward,
-					Name:     "secureforward-receiver",
-					Endpoint: "es.svc.messaging.cluster.local:9654",
+					Type: "fluentForward",
```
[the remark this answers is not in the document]
alanconway: I'm in two minds about it - once we document these strings to users, changing the constants is an API break. Literal strings in the tests will catch that mistake. We should be consistent one way or the other but I will defer that cleanup to after the initial merge.

pkg/constants/constants.go
alanconway: If ES is using HTTP then we should definitely use http: and https: - anyonne know for sure?

I think we do need to allow TLS and non-TLS connections, unless ES requires TLS only. Dev environments and trusted clusters will want to be able to avoid the burden of managing certs.

pkg/constants/constants.go
alanconway: Yes, you explained this to me before - I'll update the comment.

vimalk78:
```
waiting for deployment openshift-operators-redhat/elasticsearch-operator - remaining attempts: 2
waiting for deployment openshift-operators-redhat/elasticsearch-operator - remaining attempts: 1
waiting for deployment openshift-operators-redhat/elasticsearch-operator - remaining attempts: 0
[ERROR] PID 250: hack/test-e2e-olm.sh:11: `ELASTICSEARCH_OPERATOR_NAMESPACE=openshift-operators-redhat olm_deploy/scripts/operator-install.sh` exited with status 1.
[INFO] 		Stack Trace: 
[INFO] 		  1: hack/test-e2e-olm.sh:11: `ELASTICSEARCH_OPERATOR_NAMESPACE=openshift-operators-redhat olm_deploy/scripts/operator-install.sh`
[INFO]   Exiting with code 1.
make: *** [test-e2e-olm] Error 1
---
```

/retest

olm_deploy/scripts/operator-install.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
 set -eou pipefail
+set -x
```
bparees, line 3: this tends to make for hideous output.  Do we really need to commit it?

Elsewhere we usually wrap this in an if $DEBUG block so you can turn it on when testing locally.  (Of course there is no easy way to turn it on in a CI run on your PR, but in that case you can make the temporary change in your PR while you are debugging, also).

olm_deploy/scripts/operator-uninstall.sh
```diff
@@ -9,6 +9,7 @@ oc delete --wait --ignore-not-found ns ${CLUSTER_LOGGING_OPERATOR_NAMESPACE}
 oc delete --wait --ignore-not-found crd clusterloggings.logging.openshift.io
 oc delete --wait --ignore-not-found crd collectors.logging.openshift.io
 oc delete --wait --ignore-not-found crd logforwardings.logging.openshift.io
+oc delete --wait --ignore-not-found crd clusterlogforwarders.logging.openshift.io
```
bparees, line 12: do we still need the delete of logforwardings or is that dead code now?

pkg/controller/collector/collector_controller.go
```diff
@@ -4,7 +4,7 @@ import (
 	"context"
 	"time"
 
-	loggingv1 "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
```
bparees, line 7: there's an established practice here of using this naming convention to make it clear you're referencing the external types (and of course if we ever introduce a v2 api), i don't see a good reason to change that practice.

pkg/apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -10,7 +10,6 @@ const ClusterLogForwarderKind = "ClusterLogForwarder"
 // ClusterLogForwarder is the schema for the `clusterlogforwarder` API.
 //
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
-// +kubebuilder:resource:scope=cluster
 // +kubebuilder:subresource:status
 type ClusterLogForwarder struct {
```
periklis, line 14: Why is scope removed? Although it had a type and would have worked anyway, right syntax is:
```
// +kubebuilder:resource:scope=Cluster
```

olm_deploy/scripts/operator-uninstall.sh
periklis: For a N+1 release we will need it, since the CRD will be in place for people to migrate from tech preview to GA.

pkg/k8shandler/forwarding.go
```diff
@@ -75,197 +46,196 @@ func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig() (config s
...
+	logger.Debugf("Normalizing logforwarding from request: %v", clusterRequest)
+	logger.Debugf("ForwarderRequest: %v", clusterRequest.ForwarderRequest)
+	logger.Debugf("ForwarderSpec: %v", clusterRequest.ForwarderSpec)
+
+	// Default configuration for empty/missing forwarder, forward to the default store
+	if len(clusterRequest.ForwarderSpec.Pipelines) == 0 {
+		logger.Debug("Configuring forwarder to use the default log store")
+		clusterRequest.ForwarderSpec.Pipelines = []logging.PipelineSpec{
+			{
+				InputRefs:  []string{logging.InputNameApplication, logging.InputNameInfrastructure},
```
periklis, line 75: Is there a specific reason we skip the audit logs for the default pipeline.

pkg/k8shandler/forwarding.go
```diff
@@ -75,197 +46,196 @@ func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig() (config s
...
+		return
+	}
+	name := output.Secret.Name
+	if name == "" {
+		setInvalid(conds, "secretRef must have a name")
 		return
 	}
-	if _, exists := secret.Data["shared_key"]; !exists {
-		status.AddCondition(logforward.OutputConditionTypeSecret, logforward.OutputConditionReasonSecretMissingSharedKey, "")
+	if _, err := clusterRequest.GetSecret(name); err != nil {
```
periklis, line 238: The err check should be limit to `errors.IsNotFound` or else we could just err on transient errors, e.g. networking errors. OTOH if we don't limit the err check we should reconcile the whole thing again. WDYT?

pkg/k8shandler/forwarding_test.go
```diff
@@ -114,371 +76,258 @@ var _ = Describe("Normalizing Forwarding", func() {
...
-						SourceType: logging.LogSourceTypeApp,
+						InputRefs:  []string{logging.InputNameApplication},
 					},
 				}
-				normalizedForwardingSpec := request.normalizeLogForwarding(namespace, cluster)
-				Expect(len(normalizedForwardingSpec.Pipelines)).To(Equal(0), "Exp. all pipelines to be dropped")
-				Expect(request.ForwardingRequest.Status.LogSources).To(Equal([]logging.LogSourceType{}), "Exp. no log sources")
+				spec, status := request.normalizeForwarder()
+				Expect(spec.Pipelines).To(BeEmpty(), "Exp. all pipelines to be dropped")
+				Expect(status.Inputs).To(BeEmpty())
```
periklis, line 96: The test description implies the opposite of this assertion. Is this an error or just confusing?

pkg/k8shandler/forwarding.go
ewolinetz: the `GetSecret` function already does that check :)
https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/secret.go#L67-L77

ewolinetz: This does bring up another concern that we aren't doing the same comparison as before, where we ensure the `shared_key` exists within the secret...

manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml
```diff
@@ -67,33 +67,32 @@ metadata:
             }
           },
           {
-            "apiVersion": "logging.openshift.io/v1alpha1",
-            "kind": "LogForwarding",
+            "apiVersion": "logging.openshift.io/v1",
+            "kind": "ClusterLogForwarder,
```
ewolinetz, line 71: we're changing the kind as part of the API going GA? is there a particular pattern we're following by doing this? wouldn't the `apiVersion` alone be enough to distinguish the two?

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -9,7 +9,7 @@ spec:
     listKind: ClusterLogForwarderList
     plural: clusterlogforwarders
     singular: clusterlogforwarder
-  scope: cluster
+  scope: Namespaced
```
ewolinetz, line 12: we have scope as `Namespaced` here but in the manifest we are missing a namespace for the metadata

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -49,22 +80,20 @@ spec:
...
-                    description: "SecretRef refers to a `secret` object for secure
-                      communication. \n Client-authenticated TLS is enabled if the
-                      secret contains keys `tls.crt`, `tls.key` and `ca.crt`. Output
-                      types with password authentication will use keys `password`
-                      and `username`, not the exposed 'username@password' part of
-                      the `url`."
+                  secret:
+                    description: "Secret for secure communication. Secrets must be
+                      stored in the namespace containing the cluster logging operator.
+                      \n Client-authenticated TLS is enabled if the secret contains
```
ewolinetz, line 86: nit but should likely be `\nClient-authenticated`

pkg/apis/logging/v1/input_types.go
```diff
@@ -0,0 +1,44 @@
...
+}
+
+// InputTypeSpec is a union of optional type-specific extra specs.
+//
+type InputTypeSpec struct {
+	// Filter for application logs.
+	// +optional
+	Application *inputs.ApplicationType `json:"application,omitempty"`
+
+	// TODO(alanconway) in future we may add infra, audit or other types of filter.
```
ewolinetz, line 43: this comment is a little confusing to me... do we support infra and audit as well?

pkg/apis/logging/v1/input_types.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+var ReservedInputNames = sets.NewString(InputNameApplication, InputNameInfrastructure, InputNameAudit)
+
+// InputSpec defines a source of log messages.
+type InputSpec struct {
+	// Name used to refer to the input of a `pipeline`.
+	//
+	// +required
+	Name string `json:"name"`
```
ewolinetz, line 22: in the past we've had an alias Type for fields that expected specific values [1]. IIRC this follows an existing k8s pattern.

[1] https://github.com/openshift/cluster-logging-operator/blob/master/pkg/apis/logging/v1/clusterlogging_types.go#L252-L270

pkg/apis/logging/v1/output_test.go
```diff
@@ -2,28 +2,67 @@ package v1_test
 
 import (
```
ewolinetz, line 3: it looks like the `pkg/apis/logging/v1/output.go` file has been deleted...

test/helpers/clusterlogging.go
```diff
@@ -28,7 +28,7 @@ func NewClusterLogging(componentTypes ...LogComponentType) *cl.ClusterLogging {
 			Name:      ClusterLoggingName,
 			Namespace: OpenshiftLoggingNS,
 			Annotations: map[string]string{
-				k8shandler.ForwardingAnnotation: "enabled",
+				k8shandler.PreviewForwardingAnnotation: "enabled",
```
ewolinetz, line 31: is this annotation still required if we're going GA?

alanconway:
No

pkg/controller/forwarding/forwarder_controller.go
```diff
@@ -31,30 +27,30 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileForwarding{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	return &ReconcileForwarder{client: mgr.GetClient(), scheme: mgr.GetScheme()}
```
ewolinetz, line 30: simply out of curiosity, why are we renaming this? It looks like a lot of changes in this pr are related to this renaming and it doesn't change functionality...

alanconway: CRD Kinds are supposed to be nouns according to the k8s convention doc. It turned out to be a bigger change than I expected, but then I was already into it so I kept going...

alanconway: Also I thought it might be important to distinguish the TP from the GA API, but since they are v1 and v1alpha1 that wasn't really necessary.

pkg/apis/logging/v1/input_types.go
alanconway: I use that convention for all strings that are closed Enums, but these names are open-ended.
I dithered over using for input/output names since it provides some type-safety, where you try to use an InputName where an OutputName is expected. However it requires a bunch of duplication in code that deals with both generically.  It's probably worth doing, maybe raise a PR to rectify this later.

pkg/apis/logging/v1/output_test.go
alanconway: No, the output stuff is now more cleanly split between output_types.go and the output/ sub-package. There wasn't anything left for output.go to do.

test/helpers/clusterlogging.go
alanconway: No, needs cleaning up.

pkg/apis/logging/v1/input_types.go
alanconway: It is poorly worded. The idea is that we have an input selection spec that is specific to Applicaiton logs today, and uses namespaces. In future we may add other types of of selectors that are specific to infra/audit and use some other criteria. That's why the Application is separated as a "Type" to anticipate having other types in future, even though there aren't any yet.

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: Agreed, but it's the CRD generator mangling Go comments so I'm not sure how to fix. The go comment looks fine:
```
	// SecretRef refers to a `secret` object for secure communication.
	//
	// Client-authenticated TLS is enabled if the secret contains keys
	// `tls.crt`, `tls.key` and `ca.crt`. Output types with password
```

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: Excellent point, need to fix the manifest.

manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml
alanconway: Thats what we agreed during various discussions regarding clarity and future proofing (we may have a NamespaceLogForwarder in future) You are correct that tthe apiVersion would have been enough.

pkg/k8shandler/forwarding_test.go
alanconway: Don't follow. That line could be more neatly expressed as To(BeEmpy()) but it's still testing for emptiness - equality with an empty slice, "expect no sources". I started using some of Gomega's more expressive matchers, so the code is a little inconsistent in that regard but I think correct.

pkg/apis/logging/v1/inputs/application.go
```diff
@@ -0,0 +1,35 @@
...
+type ApplicationType struct {
+	// FIXME(alanconway) revisit nesting
+
+	// Only collect logs from applications in these namespaces. If empty, all application container logs will be collected.
+	//
+	// +optional
+	Namespaces []string `json:"namespaces"`
+}
+
+// FIXME(alanconway) fix deepcopy codegen.
```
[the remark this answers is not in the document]
alanconway: Not sure, Vimal had a problem with the code generator and wrote them by hand in frustration - it's on my list to investigate & fix.

pkg/k8shandler/forwarding.go
alanconway: Dunno, just following the TP forwarder.

pkg/k8shandler/forwarding.go
alanconway: I'll restore the original logic - there's no good reason for the change.

alanconway: The secret can be used for other output types, not just secure fluent forwarding. Not all will have a shared_secret. Perhaps we add some validation routines per output-type to vet the secret. One for after the merge.

olm_deploy/scripts/operator-install.sh
alanconway: No, fully agreed, embarassing oversight. ALready fixed.

pkg/controller/collector/collector_controller.go
alanconway: Those aren't external types - that's why I changed it. Its our own logging API. I think this was just a typo, every other source file uses "logging"

pkg/apis/logging/v1/cluster_log_forwarder_types.go
alanconway: I thought that it was the reason for some problems I saw during tests - it was unable to determine the correct operatorgroup because it's not in a namespace. Going back to namespaced seems less risky if it changes the deployment - but we should probably talk about it.

pkg/apis/logging/v1/inputs/application.go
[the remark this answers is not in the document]
alanconway: Vimal had trouble with the code generator and wrote them by hand. I'm planning to investigate after the initial merge is done.

pkg/k8shandler/forwarding.go
alanconway: I'm following the original logforwarding impl here, not sure what the reason is.

pkg/k8shandler/forwarding_test.go
alanconway: Seems ok to me: it could be expressed more neatly as To(BeEmpty) but equal to an empty slice is the same thing. I started using some of the neater gomega assertions bud dint' clean up all the existing ones.

manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml
alanconway: That's what we agreed on after much discussion. One reason is future proofing as someday we may have a NamespacedLogForwarder, so this name is more specific.

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: Will fix.

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: Agreed, but that's the CRD generator mangling Go comments. Not sure if we can fix it.

pkg/apis/logging/v1/input_types.go
alanconway: It is poorly worded. Today we only support selecting by namespace for application logs. In future we _may_ support filtering infra and/or audit logs by other criteria (namespace doesn't make much sense) For future-proofing we split out the ApplicationType as if there was going to be an enum with InfraType etc. - even though there isn't yet.

pkg/apis/logging/v1/input_types.go
alanconway: Raise a PR to address that later, I think I should have done that for type-safety. I took the lazy way out as there's code that deals with both - but it would be easy enough to address.

pkg/apis/logging/v1/output_test.go
alanconway: Nope, the output code is now split between output_types,go and the output/ sub-package.

test/helpers/clusterlogging.go
alanconway: No, will remove.

pkg/controller/forwarding/forwarder_controller.go
alanconway: We had long discussions about it, I think the main reasons are:
- clarity: we may have a NamespaceLogForwarder in future, so the ClusterLogForwarder is unambiguous if that happens.
- k8s convention says that these names should be nouns, so forwarder instead of forwarding

pkg/constants/constants.go
alanconway: If that's the case we should definitely use https: instead of tcp/tls for clarity.

pkg/apis/logging/v1/cluster_log_forwarder_types.go
jcantrill: Namespaced is fine given the arguments presented by @bparees:
* CLO is deployed to a single namespace
* CLO is only watching a single namespace
* Namespace is not the global namespace

pkg/apis/logging/v1/input_types.go
ewolinetz: nit, but why not fix it while this pr is open? we can deal with duplication of code for now and clean that up later...

pkg/apis/logging/v1/output_test.go
ewolinetz: my point i guess is more that `output_test.go` seems out of place.. should we rename this and create the `output_*_test.go` files as well?

alanconway:
/hold
This will not land in 4.5. It will land on master-post-release, or 4.6 master, whichever is available first.

alanconway:
> Nit, we should probably rename all `Log Fowarder` refs here `Cluster Log Forwarder` to skip errors when we introduce the namespaced version `Log Fowarder`. WDYT?
Fixed your nit - can you give me a new LGTM please?

alanconway:
Closing, this PR will be merged to branch master-post-release.

alanconway:
/reopen
Re-open, will re-target this PR instead of creating a new one

PR #428, opened 2020-03-22
Add Loki output support for LogForwarding

periklis:
### Motivation
This PR provides initial support for LogForwarding (atm avail. in Tech Preview) to a third-party Loki deployment. The following design decisions apply to the implementation:
- The collector field `docker` and `kubernetes` are translated to loki labels and dropped from the log payload to minimize chunk size. 
- The collector `pipeline_metadata` data are dropped entirely to prevent Loki label bloat.
- A LogForwarding Output need to be declared for each tenant.
- Ingenster rate is configured to support at min. 8MB payloads. Same applies for GRPC server configs.

**Depends on:** openshift/origin-aggregated-logging#1856


pkg/generators/forwarding/fluentd/output_conf_loki_test.go
```diff
@@ -0,0 +1,82 @@
...
+                namespace_id $.kubernetes.namespace_id
+
+                # general
+                index_name $.viaq_index_name
+                hostname $.hostname
+            </label>
+            <buffer>
+                @type file
+                path '/var/lib/fluentd/loki_receiver'
+                flush_interval "#{ENV['ES_FLUSH_INTERVAL'] || '1s'}"
```
jcantrill: I would set these to a value and skip the ability to utilize an ENV var.  This is a pattern left over from when you could modify the env var and their was no operator

test/helpers/loki/single_server.go
```diff
@@ -0,0 +1,157 @@
...
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+}
+
+func NewService(namespace string) *v1.Service {
```
jcantrill: We may already have this utility method in util?  Consider reusing it.

test/helpers/loki/single_server.go
```diff
@@ -0,0 +1,157 @@
...
+				},
+			},
+			Selector: map[string]string{
+				"app": DeploymentName,
+			},
+		},
+	}
+}
+
+func NewConfigMap(namespace string) *v1.ConfigMap {
```
jcantrill: We may already have this utility method in util?  Consider reusing it.

test/helpers/framework.go
```diff
@@ -60,6 +60,8 @@ type LogStore interface {
 	HasInfraStructureLogs(timeToWait time.Duration) (bool, error)
 
 	HasAuditLogs(timeToWait time.Duration) (bool, error)
+
+	ClusterLocalEndpoint() string
```
jcantrill: What does this represent? The name is not obvious to me

periklis: It means you provide a service object for your store and this method gives you the cluster local DNS name e.g. kafka.openshift-logging.svc.cluster.local

pkg/generators/forwarding/fluentd/output_conf_loki_test.go
periklis: All `ES_` prefixed env vars are removed. I kept the `BUFFER_` prefixed env vars which are set in run.sh

pkg/generators/forwarding/fluentd/loki_conf.go
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	return fmt.Sprintf("%s://%s", url.Scheme, url.Host)
+}
+
+// LokiTenant returns the loki tenant ID
+func (conf *outputLabelConf) LokiTenant() string {
+	url, err := url.Parse(conf.Target.Endpoint)
+	if err != nil {
+		log.Errorf("Failed to extract Kafka topic from endpoint url %q: %s", conf.Target.Endpoint, err)
```
cyriltovena, line 26: I think something is wrong here.

cyriltovena: Or may be there is a relation with kafka :) no idea.

pkg/generators/forwarding/fluentd/output_conf_loki_test.go
```diff
@@ -0,0 +1,82 @@
...
+	BeforeEach(func() {
+		generator, err = NewConfigGenerator(false, false)
+		Expect(err).To(BeNil())
+	})
+
+	Context("for a loki endpoint", func() {
+		lokiConf := `<label @LOKI_RECEIVER>
+        <match **>
+            @type loki
+            url https://logs-us-west1.grafana.net
```
cyriltovena, line 26: This cluster will be removed at some point.

pkg/generators/forwarding/fluentd/loki_conf.go
periklis: No this really me switching between #425 and this PR :)

pkg/generators/forwarding/fluentd/output_conf_loki_test.go
periklis: This is just a unit test for the config generator. It is now using the cluster at all. Probably I have copied the URL from the plugin's readme to right the test in the first place.

jcantrill:
@sichvoge what do you want to do with this change?

pkg/apis/logging/v1/output_types.go
```diff
@@ -153,6 +156,14 @@ type Kafka struct {
 	Brokers []string `json:"brokers,omitempty"`
 }
 
+// Loki provides optional extra properties for `type: loki`
+type Loki struct {
+	// Tenant is the name of the tenant of the log records.
```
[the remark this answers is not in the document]
alanconway: Would it make sense to use the username part of the URL as the tenant ID? Not pushing it, only if the tenannat ID is semantically a username and this wouldn't create confusion. ie.

alanconway (approved):

pkg/apis/logging/v1/output_types.go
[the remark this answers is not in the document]
periklis: Actually Loki is making a distinction between username and TenantID, e.g. when using basic auth. TenantID is for Loki only an extra HTTP Header `X-Scope-OrgID`.

PR #429, opened 2020-03-23
LOG-682: Remove _output directory, simplify Makefile

alanconway:
Changes:
- trivial source file changes from gofmts and to satisfy the linter.
- drop _output file, not necessary and causes confusion.  
- remove unused variables, simplify targets.  
- add dependencies to make targets self-sufficient.

Signed-off-by: Alan Conway <aconway@redhat.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-logging-operator
 COPY . .
-RUN make
+RUN make build
 
 FROM centos:centos7
 RUN INSTALL_PKGS=" \
@@ -12,7 +12,7 @@ RUN INSTALL_PKGS=" \
     yum clean all && \
     mkdir /tmp/ocp-clo && \
     chmod og+w /tmp/ocp-clo
-COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/_output/bin/cluster-logging-operator /usr/bin/
+COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/bin/cluster-logging-operator /usr/bin/
 COPY scripts/* /usr/bin/scripts/
 RUN mkdir -p /usr/share/logging/
 COPY files/ /usr/share/logging/
--- a/Makefile
+++ b/Makefile
@@ -1,62 +1,34 @@
-TARGET_DIR=$(CURDIR)/_output
 KUBECONFIG?=$(HOME)/.kube/config
 
-GOBUILD=go build
-BUILD_GOPATH=$(TARGET_DIR):$(TARGET_DIR)/vendor:$(CURDIR)/cmd
+export GOBIN=$(CURDIR)/bin
+export PATH:=$(GOBIN):$(PATH)
 
-IMAGE_BUILDER_OPTS=
-IMAGE_BUILDER?=imagebuilder
-IMAGE_BUILD=$(IMAGE_BUILDER)
 export IMAGE_TAG_CMD?=docker tag
 
 export APP_NAME=cluster-logging-operator
-APP_REPO=github.com/openshift/$(APP_NAME)
-TARGET=$(TARGET_DIR)/bin/$(APP_NAME)
-IMAGE_TAG?=quay.io/openshift/origin-$(APP_NAME):latest
-export IMAGE_TAG
-MAIN_PKG=cmd/manager/main.go
-export OCP_VERSION?=$(shell basename $(shell find manifests/  -maxdepth 1  -not -name manifests -type d))
-export CSV_FILE=$(CURDIR)/manifests/$(OCP_VERSION)/cluster-logging.v$(OCP_VERSION).0.clusterserviceversion.yaml
+export IMAGE_TAG?=quay.io/openshift/origin-$(APP_NAME):latest
+
+export OCP_VERSION?=4.5
 export NAMESPACE?=openshift-logging
-export EO_CSV_FILE=$(CURDIR)/vendor/github.com/openshift/elasticsearch-operator/manifests/$(OCP_VERSION)/elasticsearch-operator.v$(OCP_VERSION).0.clusterserviceversion.yaml
 
 FLUENTD_IMAGE?=quay.io/openshift/origin-logging-fluentd:latest
 
-PKGS=$(shell go list ./... | grep -v -E '/vendor/|/test|/examples')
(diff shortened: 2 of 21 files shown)


Makefile
```diff
@@ -64,38 +43,34 @@ run:
...
 
-image: imagebuilder
+image: bin/imagebuilder build
 	@if [ $${USE_IMAGE_STREAM:-false} = false ] && [ $${SKIP_BUILD:-false} = false ] ; \
 	then hack/build-image.sh $(IMAGE_TAG) $(IMAGE_BUILDER) $(IMAGE_BUILDER_OPTS) ; \
 	fi
 
-lint:
+lint: fmt
 	golangci-lint run -c golangci.yaml
```
igor-karpukhin: @alanconway `make build` will fail if this `golangci-lint` is missing. I guess we should add this as a prerequisite to README.md, or add the way to download it automatically if it is missing.

Makefile
```diff
@@ -1,58 +1,37 @@
...
-IMAGE_TAG?=quay.io/openshift/origin-$(APP_NAME):latest
-export IMAGE_TAG
-MAIN_PKG=cmd/manager/main.go
-export OCP_VERSION?=$(shell basename $(shell find manifests/  -maxdepth 1  -not -name manifests -type d))
+export IMAGE_TAG?=quay.io/openshift/origin-$(APP_NAME):latest
+
+export OCP_VERSION?=4.5
 export CSV_FILE=$(CURDIR)/manifests/$(OCP_VERSION)/cluster-logging.v$(OCP_VERSION).0.clusterserviceversion.yaml
 export NAMESPACE?=openshift-logging
 export EO_CSV_FILE=$(CURDIR)/vendor/github.com/openshift/elasticsearch-operator/manifests/$(OCP_VERSION)/elasticsearch-operator.v$(OCP_VERSION).0.clusterserviceversion.yaml
```
periklis: IIRC `CSV_FILE` and `EO_CSV_FILE` are unused.

Makefile
alanconway: +1, fixing.

Makefile
alanconway: +1 removed.

Dockerfile
```diff
@@ -12,12 +12,12 @@ RUN INSTALL_PKGS=" \
     yum clean all && \
     mkdir /tmp/ocp-clo && \
     chmod og+w /tmp/ocp-clo
-COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/_output/bin/cluster-logging-operator /usr/bin/
+COPY cmd/manager /usr/bin/cluster-logging-operator
```
ewolinetz: this defeats the whole purpose of using a builder image

ewolinetz: i think this would break ART -- they do not have access to our `Makefile`

jcantrill: This will break the ART process because this copies the binary from local instead of the building layer.

alanconway: OK, will fix. My bad.

alanconway:
@jcantrill @ewolinetz Fixed my blunder with the Dockerfile, I think this is good now assuming CI passes.

Makefile
```diff
@@ -112,10 +93,10 @@ deploy-elasticsearch-operator:
 deploy-example: deploy
 	oc create -n $(NAMESPACE) -f hack/cr.yaml
 
-test-unit: fmt
-	@LOGGING_SHARE_DIR=$(CURDIR)/files go test $(TEST_OPTIONS) $(PKGS)
+test-unit: generate
+	go test ./pkg/...
 
-test-e2e:
+test-e2e: build
```
jcantrill: We should rethink the dependency upon `build` as its unnecessary.  In a CI context, the operator will be deployed using the image that was built by a separate CI job.  Locally the deployment will reference an image that only exists if you run `make deploy-image`.  This means the primary place this  is used we are building the operator twice.

README.md
```diff
@@ -46,24 +46,23 @@ Running locally outside an OKD cluster:
...
+See also: [sdk_setup.sh](https://raw.githubusercontent.com/openshift/origin-aggregated-logging/master/hack/sdk_setup.sh) to set up your environment.
 
 The deployment can be optionally modified using any of the following:
 
 | Env Var | Default | Description|
 |---------|---------|------------|
-|`IMAGE_BUILDER`|`imagebuilder`| The command to build the container image|
 |`EXCLUSIONS`|none|The list of manifest files that should will be ignored|
-|`OC`|`oc` in `PATH`| The openshift binary to use to deploy resources|
 |`REMOTE_REGISTRY`|false|`true` if you are running the cluster on a different machine than the one you are developing on|
```
jcantrill: While you are updating, can you remove the ref here and below to "REMOTE_REGISTRY" as they are not needed atm

README.md
```diff
@@ -80,7 +79,7 @@ create a user and assign it rights:
 ```
 
 If you used the new `openshift-installer`, it creates a user named `kubeadmin`
-    with the password in the file `installer/auth/kubeadmin_password`.
+with the password in the file `installer/auth/kubeadmin_password`.
 
 To push the image to the remote registry, set `REMOTE_REGISTRY=true`, `PUSH_USER=kubeadmin`,
```
jcantrill: `make deploy-image` does this

Makefile
alanconway: +1 I only left it there because I assumed it was a CI requirement. Will remove.

README.md
alanconway: +1 check updated README.

alanconway:
Fixed all comments
/lgtm

ewolinetz:
/hold
so it doesn't keep spinning

Makefile
```diff
@@ -1,62 +1,38 @@
...
-APP_REPO=github.com/openshift/$(APP_NAME)
-TARGET=$(TARGET_DIR)/bin/$(APP_NAME)
-IMAGE_TAG?=quay.io/openshift/origin-$(APP_NAME):latest
-export IMAGE_TAG
-MAIN_PKG=cmd/manager/main.go
-export OCP_VERSION?=$(shell basename $(shell find manifests/  -maxdepth 1  -not -name manifests -type d))
-export CSV_FILE=$(CURDIR)/manifests/$(OCP_VERSION)/cluster-logging.v$(OCP_VERSION).0.clusterserviceversion.yaml
+export IMAGE_TAG?=quay.io/openshift/origin-$(APP_NAME):latest
+
+export OCP_VERSION?=4.5
```
jcantrill, line 11: It needs a rebase but we are trying to move away from hard coding the version? Why is this here?

PR #434, opened 2020-03-26
Bug 1813209: Adding oauth-openshift route url to noProxy list for kibana proxy

ewolinetz:
when using proxy config, we should add the oauth-openshift route url to the `NO_PROXY` and `no_proxy` env vars so that when we get a callback from oauth we can properly not go through the proxy.

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1813209

### DIFF
--- a/pkg/k8shandler/visualization.go
+++ b/pkg/k8shandler/visualization.go
@@ -236,7 +236,14 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaDeployment(prox
 		return
 	}
 
-	kibanaPodSpec := newKibanaPodSpec(clusterRequest.cluster, "kibana", "elasticsearch.openshift-logging.svc.cluster.local", proxyConfig, kibanaTrustBundle)
+	routerCanonicalHostname, err := clusterRequest.GetRouterCanonicalHostname("kibana")
+	if err != nil {
+		logrus.Error("Unable to get routerCanonicalHostname for kibana")
+	}
+
+	oauthIssuer := fmt.Sprintf("%s.%s", "oauth-openshift", routerCanonicalHostname)
+
+	kibanaPodSpec := newKibanaPodSpec(clusterRequest.cluster, "kibana", "elasticsearch.openshift-logging.svc.cluster.local", proxyConfig, kibanaTrustBundle, oauthIssuer)
 	kibanaDeployment := NewDeployment(
 		"kibana",
 		clusterRequest.cluster.Namespace,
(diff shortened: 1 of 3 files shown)


pkg/k8shandler/visualization.go
```diff
@@ -236,7 +236,14 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaDeployment(prox
 		return
 	}
 
-	kibanaPodSpec := newKibanaPodSpec(clusterRequest.cluster, "kibana", "elasticsearch.openshift-logging.svc.cluster.local", proxyConfig, kibanaTrustBundle)
+	routerCanonicalHostname, err := clusterRequest.GetRouterCanonicalHostname("kibana")
+	if err != nil {
+		logrus.Error("Unable to get routerCanonicalHostname for kibana")
+	}
+
+	oauthIssuer := fmt.Sprintf("%s.%s", "oauth-openshift", routerCanonicalHostname)
```
ewolinetz: @bparees is it appropriate to always assume the oauth issuer will be "oauth-openshift" ? Are users able to configure this at all?

bparees: yes, i would expect even if the cluster has an alternate IDP configured, you're still talking to openshift-oauth and i'd be surprised if the value is cluster-admin configurable.  But @deads2k or someone from the apiserver team would be best to consult with.

deads2k:
> when using proxy config, we should add the oauth-openshift route url to the `NO_PROXY` and `no_proxy` env vars so that when we get a callback from oauth we can properly not go through the proxy.
> 
> Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1813209

we have faced this question a couple times and consistently made the decision *not* to place routes (cluster ingress) into no_proxy unconditionally.  We do this because having external (outside the proxy bubble) cluster-ingress.  If a customer is trying to isolate their cluster, the cluster-ingress outside is valuable.

To support this use-case, we allow a customer to explicitly place routes in the no_proxy list, but we do not automatically do it because there is no way to remove something we automatically place in the list.

pkg/k8shandler/visualization.go
sttts: Keycloak will replace oauth-server by using their own authn webhook. I miss the context here to tell whether this matters for the question. Maybe @stlaz has more insights.

ewolinetz:
closing per https://github.com/openshift/cluster-logging-operator/pull/434#issuecomment-605037018
fix will be a docs update

pkg/k8shandler/visualization.go
stlaz: Like David said, this approach should not be used.

Also, you should by no means assume the issuer of the user-tokens uses the oauth-openshift route, we are working on abstracting the token validation to any service capable of dealing with kubernetes webhook authentication flow, meaning users will be able to configure the oauth-server to whatever they please.

PR #435, opened 2020-03-26
Remove default storageclass from example

cblecker:
This removes a specified `storageClassName` from the pre-populated example from the CSV.

From a usability perspective, if you are provisioning a new `ClusterLogging` and you are using OCP on AWS, you can click "Create" right after bringing the example up and it will provision. If you are on a non-AWS cloud, you won't have the `gp2` storage class, and things will stall out provisioning. Recovery from this is a little complex, as you need to manually remove all the PVCs (as they won't get deleted, even if you delete the `ClusterLogging` CR).

### DIFF
--- a/manifests/4.6/cluster-logging.v4.6.0.clusterserviceversion.yaml
+++ b/manifests/4.6/cluster-logging.v4.6.0.clusterserviceversion.yaml
@@ -40,7 +40,6 @@ metadata:
                   "nodeCount": 3,
                   "redundancyPolicy": "SingleRedundancy",
                   "storage": {
-                    "storageClassName": "gp2",
                     "size": "200G"
                   }
                 },


cblecker:
/hold
The test is failing constantly. Unsure as to why.

ewolinetz:
/hold
we may have a ci capacity issue right now, no need to needlessly spin

cblecker:
All looks well now. Thanks @ewolinetz !

@jcantrill ping for re-LGTM

manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml
```diff
@@ -36,7 +36,6 @@ metadata:
                   "nodeCount": 3,
                   "redundancyPolicy": "SingleRedundancy",
                   "storage": {
-                    "storageClassName": "gp2",
```
ewolinetz: in this case the `"size": "200G"` does nothing when the elasticsearch operator ends up resolving this... so by default you'll end up with ephemeral storage. There isn't an easy way to then move any captured log records into a cluster backed with storage. So i'm not sure which is the lesser evil here... loss of logs or needing to manually remove a pvc once the correctly backed ones are created...

cblecker: Oh, the fall back is ephemeral, rather than the default storage class? Hmm

jcantrill: > in this case you'll end up with ephemeral storage. 

It will not be ephemeral as  the cluster has a default class so you will get whatever that is.  The fact you define a size will beget you a PVC

jcantrill:
@cblecker please rebase to master or submit a new PR which is likely easier

cblecker:
@jcantrill rebased

manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml
jcantrill: Note this is dependent also on a forth coming change in the EO

PR #436, opened 2020-03-27
Bug 1817592: fix Makefile in order to fix ART builds

jcantrill:
This PR removes `make` dependency on generate to satisfy the ART building process

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,7 +46,7 @@ imagebuilder:
 	then go get -u github.com/openshift/imagebuilder/cmd/imagebuilder ; \
 	fi
 
-build: generate fmt
+build: fmt
 	@mkdir -p $(TARGET_DIR)/src/$(APP_REPO)
 	@cp -ru $(CURDIR)/pkg $(TARGET_DIR)/src/$(APP_REPO)
 	@cp -ru $(CURDIR)/vendor/* $(TARGET_DIR)/src


Makefile
```diff
@@ -46,7 +46,7 @@ imagebuilder:
 	then go get -u github.com/openshift/imagebuilder/cmd/imagebuilder ; \
 	fi
 
-build: generate fmt
+build: fmt
```
periklis: Why does it fail here and not (yet) for EO: https://github.com/openshift/cluster-logging-operator/blob/master/Makefile#L49

jcantrill: it will fail for EO.  This is the output from brew not CI ```2020-03-26 08:48:28,993 - atomic_reactor.plugins.imagebuilder - INFO - --> WORKDIR /go/src/github.com/openshift/cluster-logging-operator
2020-03-26 08:48:29,127 - atomic_reactor.plugins.imagebuilder - INFO - --> COPY . .
2020-03-26 08:48:33,015 - atomic_reactor.plugins.imagebuilder - INFO - --> RUN make
2020-03-26 08:48:34,201 - atomic_reactor.plugins.imagebuilder - INFO - curl -f -L -o operator-sdk-v0.15.2 https://github.com/operator-framework/operator-sdk/releases/download/v0.15.2/operator-sdk-v0.15.2-x86…```

jcantrill: it looks also like EO does not have the same dependency or mechanism so thats why it doesnt break

jcantrill:
closing as @alanconway assures me https://github.com/openshift/cluster-logging-operator/pull/429 will resolve the problem

alanconway:
/lgtm
I think auto-generate will when #429 goes in, but lets get things passing now and I'll put it back with#429.

Makefile
periklis: Actually I linked the wrong repository above. EO should have the same issue: https://github.com/openshift/elasticsearch-operator/blob/master/Makefile#L77

However, I cannot find any EO builds for 4.5 on brew neither looking through the web nor searching via CLI.

periklis: IMHO, the correct fix is to vendor the operator-sdk.

ewolinetz: We already vendor the sdk code
https://github.com/openshift/cluster-logging-operator/blob/master/Gopkg.toml#L63-L67

ewolinetz: I think `generate` should have already been called before dev has started since it is doing the k8s generation

PR #438, opened 2020-03-27
remove dots in index management types

jcantrill:
This PR further address https://github.com/openshift/enhancements/pull/261 and changes to remove dots from indexmanagement names


jcantrill:
> do we want to change these names in
> https://github.com/openshift/cluster-logging-operator/blob/master/manifests/4.5/cluster-loggings.crd.yaml#L126
> and
> https://github.com/openshift/cluster-logging-operator/blob/master/deploy/crds/logging.openshift.io_clusterloggings_crd.yaml#L293

yes.  though these will go away with LF GA. I'll update accordingly

Issue #439, opened 2020-03-28
Invalid fields in Logging Cluster Components status


anshulvermapatel:
When the Logging Cluster is installed properly and I am able to see the logs. But when I try to see the statuses of the different components, I see that the messages seen in the ScreenCapture below. The Screen Capture is from OpenShift 4.3 but I saw the same in 4.5 also.

After digging into it, I saw that the path being accessed in the CSV to check the status is incorrect in  [this file](https://github.com/openshift/cluster-logging-operator/blob/8d0d82ff4b6943f652b1be4114d798e5256cd32d/manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml#L354-L451).
The below path are arrays -
~~~
- visualization.kibanaStatus
- logStore.elasticsearchStatus
- logStore.elasticsearchStatus
- logStore.elasticsearchStatus
~~~
But they are being referred at they are single objects.
~~~
- visualization.kibanaStatus.pods
- logStore.elasticsearchStatus.pods.client
- logStore.elasticsearchStatus.pods.data
- logStore.elasticsearchStatus.pods.master
~~~
Rather they should have been as below -
~~~
- visualization.kibanaStatus[0].pods
- logStore.elasticsearchStatus[0].pods.client
- logStore.elasticsearchStatus[0].pods.data
- logStore.elasticsearchStatus[0].pods.master
~~~

I saw this issue in 4.2, 4.3, 4.4 and 4.5.
It's just a small change is there which needs to be done.
I plan to create a PR for 4.5,

The Screen Capture mentioned above -

![Invalid_field_clusterlogging](https://user-images.githubusercontent.com/26687535/77832948-a5759600-715f-11ea-93ed-cdcdafe2ba95.png)

anshulvermapatel:
This should get closed as the PR for this has been merged.
Closing for now.

PR #440, opened 2020-03-28
fixes the issue #439 where the status is not shown and gives a warning message that the field is invalid

anshulvermapatel:
Team,

I created [this issue](https://github.com/openshift/cluster-logging-operator/issues/439) and created this PR to have the correct path to get the status of ES and Kibana.

Do let me know if any modification is required.

### DIFF
--- a/manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml
+++ b/manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml
@@ -418,27 +418,27 @@ spec:
       statusDescriptors:
       - description: The status for each of the Kibana pods for the Visualization component
         displayName: Kibana Status
-        path: visualization.kibanaStatus.pods
+        path: visualization.kibanaStatus[0].pods
         x-descriptors:
         - 'urn:alm:descriptor:com.tectonic.ui:podStatuses'
       - description: The status for each of the Elasticsearch Client pods for the Log Storage component
         displayName: Elasticsearch Client Pod Status
-        path: logStore.elasticsearchStatus.pods.client
+        path: logStore.elasticsearchStatus[0].pods.client
         x-descriptors:
         - 'urn:alm:descriptor:com.tectonic.ui:podStatuses'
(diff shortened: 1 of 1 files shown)


ewolinetz:
@anshulvermapatel can you please change the target of this to `master` instead? We will then cherry-pick backwards

manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml
```diff
@@ -418,27 +418,27 @@ spec:
       statusDescriptors:
       - description: The status for each of the Kibana pods for the Visualization component
         displayName: Kibana Status
-        path: visualization.kibanaStatus.pods
+        path: visualization.kibanaStatus[0].pods
```
jcantrill: What happens when there is more then one status?  What happens when there is no status?

jcantrill:
> @anshulvermapatel can you please change the target of this to `master` instead? We will then cherry-pick backwards

I rebased it on master

manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml
ewolinetz: @jcantrill unfortunately we may be past this update from the API perspective.. but I don't think the CLO will ever need to worry about more than one Kibana. This is leftover from the `ops` pattern. so it should only be `[0]`...

anshulvermapatel: Yes, I too think I should be only [0]

jcantrill: @ewolinetz which is fine but what happens when there are no kibana status or any of the others for that matter.  I'm trying to ensure we don't introduce some error condition

ewolinetz: @jcantrill seems to be no worse than what's already seen https://github.com/openshift/cluster-logging-operator/issues/439

PR #442, opened 2020-03-30
fix elasticsearch unmanaged test

jcantrill:
This PR fixes the elasticsearch unmanaged test:

* Adding indexmanagement for the desired indices that are failing because of https://github.com/uken/fluent-plugin-elasticsearch/issues/734 and blocking autocreate of indices

### DIFF
--- a/test/helpers/elasticsearch.go
+++ b/test/helpers/elasticsearch.go
@@ -15,6 +15,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/wait"
 
 	k8shandler "github.com/openshift/cluster-logging-operator/pkg/k8shandler"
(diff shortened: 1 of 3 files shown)


test/helpers/elasticsearch.go
```diff
@@ -202,6 +205,11 @@ func (tc *E2ETestFramework) DeployAnElasticsearchCluster(pwd string) (cr *elasti
 			Nodes:            []elasticsearch.ElasticsearchNode{node},
 			ManagementState:  elasticsearch.ManagementStateManaged,
 			RedundancyPolicy: elasticsearch.ZeroRedundancy,
+			IndexManagement: indexmanagement.NewSpec(&clusterlogging.RetentionPoliciesSpec{
```
ewolinetz: we don't need mappings too?

PR #447, opened 2020-04-01
Fix pretty print of generated fluent.conf

vimalk78:
Fix pretty print of generated fluent.conf. Improve readability.

Example output:

```

<system>
    @log_level "#{ENV['LOG_LEVEL'] || 'warn'}"
</system>

<source>
    @type prometheus
    bind ''
    <ssl>
        enable true
        certificate_path "#{ENV['METRICS_CERT'] || '/etc/fluent/metrics/tls.crt'}"
        private_key_path "#{ENV['METRICS_KEY'] || '/etc/fluent/metrics/tls.key'}"
    </ssl>
</source>
```
Output before this change:
```
## CLO GENERATED CONFIGURATION ###    
# This file is a copy of the fluentd configuration entrypoint    
# which should normally be supplied in a configmap.    
    
<system>    
@log_level "#{ENV['LOG_LEVEL'] || 'warn'}"    
</system>    
    
# In each section below, pre- and post- includes don't include anything initially;    
# they exist to enable future additions to openshift conf as needed.    
    
## sources    
## ordered so that syslog always runs last...    
<source>    
@type prometheus    
bind ''    
    <ssl>    
    enable true    
    certificate_path "#{ENV['METRICS_CERT'] || '/etc/fluent/metrics/tls.crt'}"    
    private_key_path "#{ENV['METRICS_KEY'] || '/etc/fluent/metrics/tls.key'}"    
</ssl>    
</source>    
```

### DIFF
--- a/pkg/generators/forwarding/fluentd/generators.go
+++ b/pkg/generators/forwarding/fluentd/generators.go
@@ -110,7 +110,7 @@ func mapSourceTypesToPipelineNames(pipelines []logforward.PipelineSpec) map[logf
 }
 
 func pretty(in string) string {
-	stack := -1
+	stack := 0
 	out := bytes.NewBufferString("")
 	for _, line := range strings.Split(in, "\n") {
 		stack = prettyLine(out, line, stack)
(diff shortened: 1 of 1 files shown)


ewolinetz:
@vimalk78 to better visualize, do you have an example of what the output was before this change too?

vimalk78:
> @vimalk78 to better visualize, do you have an example of what the output was before this change too?

@ewolinetz Updated above.

PR #450, opened 2020-04-02
LOG-653: GA cluster log forwarder API

alanconway:
ClusterLogForwarder GA API Go and generated .yaml


Makefile
```diff
@@ -15,8 +15,10 @@ FLUENTD_IMAGE?=quay.io/openshift/origin-logging-fluentd:latest
 
 .PHONY: all build clean fmt generate regenerate deploy-setup deploy-image image deploy deploy-example test-unit test-e2e test-sec undeploy run operator-sdk imagebuilder golangci-lint
 
-# Run all local pre-merge tests. CI runs additional system tests on the image.
-all: generate fmt test-unit lint
+all: test-unit
+
+# For developers: updated, formatted, linted code and unit test results.
+dev: generate lint test-unit
```
ewolinetz, line 21: no `fmt` on here?

hack/get-operator-sdk.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Download latest release of operator-sdk, intended for use by ../Makefile
+
+set -e
+BASE=https://github.com/operator-framework/operator-sdk
+VERSION="$(curl -fLsS -o /dev/null -w '%{url_effective}' $BASE/releases/latest | sed 's-.*/--')"
+URL=$BASE/releases/download/$VERSION/operator-sdk-$VERSION-$(uname -i)-linux-gnu
+echo downloading $URL
+curl -fL -o bin/operator-sdk $URL && chmod +x bin/operator-sdk
+bin/operator-sdk version > /dev/null # Verify executable works
```
ewolinetz, line 10: shouldn't this be `$GOPATH/bin/operator-sdk` ? thats where i currently have my operator-sdk binary...

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,275 @@
...
+          type: object
+        status:
+          properties:
+            conditions:
+              additionalProperties:
+                properties:
+                  lastTransitionTime:
+                    description: LastTransitionTime is the last time this condition
+                      changed.
+                    format: date-time
```
ewolinetz, line 148: do we think we should give the formatting it'll show up as? e.g. `%Y.%M.%d... etc`

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,275 @@
...
+              description: Conditions of the log forwarder.
+              type: object
+            inputs:
+              additionalProperties:
+                additionalProperties:
+                  properties:
+                    lastTransitionTime:
+                      description: LastTransitionTime is the last time this condition
+                        changed.
+                      format: date-time
```
ewolinetz, line 177: similar ask here

pkg/apis/logging/v1/output.go
```diff
@@ -0,0 +1,72 @@
...
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+
+	"github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1/outputs"
+)
+
+const OutputTypeDefault = "default"
```
ewolinetz, line 11: is there a reason we have "default" as a default type rather than it being the same as `OutputTypeElasticsearch` ?

ewolinetz:
since @alanconway is an approver putting /hold so we give people enough time to review (it has auto approve currently)

@alanconway please remove it when we're ready to merge

Makefile
alanconway: lint has fmt as a dependency so same thing.

hack/get-operator-sdk.sh
alanconway: The makefile will pick it up from PATH or download a released binary into bin/ - way back when I started using it to generate CRDs - this tool doesn't work with normal `go get` and I ran into issues building it from master. We should revisit this, when we update our operator-sdk library dependencies to a modern version, we can just build or  `go run` the same version of the tool as we are using for the libraries and avoid all the downloading muck.

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: Yep, done. Its the RFC3339 format that meta.time uses, but god to mention it in the description.

pkg/apis/logging/v1/output.go
alanconway: Yep - someday elasticsearch may not be the default. OutputTypeElasticSearch means "I want to forward to an elastic instance (and I'll provide the URL and other necessary config)" whereas "default" means "I want to forward to the openshift logging default store, whatever that is, however it's configured".

hack/get-operator-sdk.sh
ewolinetz: sorry, i realize i wrote my comment on the execution line, but this applies to where we pull it down from `curl` `-o bin/operator-sdk`

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
ewolinetz: done where?

Makefile
```diff
@@ -64,12 +64,16 @@ fmt:
 
 # Do all code/CRD generation at once, with timestamp file to check out-of-date.
 GEN_TIMESTAMP=.zz_generate_timestamp
+MANIFESTS=manifests/$(OCP_VERSION)
```
jcantrill, line 67: Why do we have to hardcode the version if we can figure it out by inspecting the manifests dir?  This inhibits our desire to make it easier to embrace feature branches with openshift build CI

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,275 @@
...
+                        type: string
+                    type: object
+                  syslog:
+                    description: 'Syslog provides optional extra properties for `type:
+                      syslog`'
+                    properties:
+                      facility:
+                        description: "Facility to set on outgoing syslog records.
+                          \n The string value may contain an integer or one of these
+                          names:     kernel user mail daemon auth syslog lpr news
```
vimalk78, line 76: is it better to give a reference to RFC? rfc5424 section 6.2.1

vimalk78: as per rfc and also syslog_protocol ruby plugin, the values are `audit, alert` instead of `security,console`
as per rfc the value is `clock daemon` instead of `cron`

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,275 @@
...
+                        description: Namespace defines the space within which the
+                          secret name must be unique.
+                        type: string
+                    type: object
+                  syslog:
+                    description: 'Syslog provides optional extra properties for `type:
+                      syslog`'
+                    properties:
+                      facility:
+                        description: "Facility to set on outgoing syslog records.
```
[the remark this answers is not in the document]
vimalk78: we can have a convention that if the value of `facility` is prefixed by `#` the remaining content will be treated as a key in the record to extract the `facility` value.
same for `severity`

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,275 @@
...
+                      payloadKey:
+                        description: PayloadKey specifies record field to use as payload.
+                        type: string
+                      severity:
+                        description: "Severity to set on outgoing syslog records.
+                          \n The string value may contain an integer or one of these
+                          names:     emergency alert critical error warning notice
+                          informational debug"
+                        type: string
+                      tagKey:
```
vimalk78, line 89: as per rfc 5424, the TAG has been split into APP-NAME, PROCID, and MSGID.  should we expose these?

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,275 @@
...
+                      name:
+                        description: Name is unique within a namespace to reference
+                          a secret resource.
+                        type: string
+                      namespace:
+                        description: Namespace defines the space within which the
+                          secret name must be unique.
+                        type: string
+                    type: object
+                  syslog:
```
vimalk78, line 69: as per rfc, the header contains a VERSION, but the ruby plugin doesnt seem to support it. The syslog implementation it uses conforms to old rfc
rfc: https://tools.ietf.org/html/rfc5424#page-8
ruby gem:https://github.com/eric/syslog_protocol/blob/master/lib/syslog_protocol/packet.rb#L14

Makefile
alanconway: OK, I'll revert that for this PR. I have some questions/thoughts but will raise them separately.

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: +1 to referencing RFC. The only truly "standard" values are the integers - the string names and descriptions vary slightly by implementation & historical version of the RFC. I guess we should follow rfc5424 for documentation, but maybe we should also  be forgiving and accept common variations, e.g. treat "audit" and "security" as equivalent.

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: Yep, the Tag/Payload/Trim stuff should be aligned with your implementation. I'll leave that for you to update since you're closer to the details.

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: Again you should align this to your work.

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,275 @@
...
+          description: ClusterLogForwarderSpec defines the desired state of ClusterLogForwarder
+          properties:
+            outputs:
+              description: Outputs are named destinations for log messages.
+              items:
+                description: Output defines a destination for log messages.
+                properties:
+                  elasticsearch:
+                    type: object
+                  fluent:
```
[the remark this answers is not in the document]
alanconway: Renamed fluentForward

hack/get-operator-sdk.sh
alanconway: Yep, currently we're caching those binaries in the local bin/ directory. I don't want to put something in GOPATH/bin that wasn't locally built  from GOPATH. We should revisit this once we move to `go mod`  which has a much better story for locally caching multiple versions of libs & binaries without trampling on the users
 GOPATH.

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,275 @@
...
+                      secret contains keys `tls.crt`, `tls.key` and `ca.crt`. Output
+                      types with password authentication will use keys `password`
+                      and `username`, not the exposed 'username@password' part of
+                      the `url`."
+                    properties:
+                      name:
+                        description: Name is unique within a namespace to reference
+                          a secret resource.
+                        type: string
+                      namespace:
```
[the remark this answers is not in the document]
alanconway: This is the k8s.io.api.core.v1.SecretReference, which seems to be a very common way to refer to a secret in an API. I can replace it with a string, but I wasn't confident we'll never need the namespace, in particular since the ClusterLogForwarder is deployed in cluster scope.  What do you think?

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -0,0 +1,275 @@
...
+                        description: Namespace defines the space within which the
+                          secret name must be unique.
```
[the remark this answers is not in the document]
alanconway: That's comment is from the k8s.io code :)

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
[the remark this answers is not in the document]
alanconway: @vimalk78 I'll let you document that when we incorporate your syslog code with this API.

alanconway:
/hold
have fixes for review comments.

PR #452, opened 2020-04-03
turn off ES verification at start to not stop log collection

jcantrill:
This PR disables verifying ES version at startup so that other collections and outputs may proceed

### DIFF
--- a/pkg/generators/forwarding/fluentd/fluent_conf_test.go
+++ b/pkg/generators/forwarding/fluentd/fluent_conf_test.go
@@ -914,6 +914,7 @@ var _ = Describe("Generating fluentd config", func() {
 						@id retry_infra_es
 						host es.svc.infra.cluster
 						port 9999
+						verify_es_version_at_startup false
 						scheme https
 						ssl_version TLSv1_2
 						target_index_key viaq_index_name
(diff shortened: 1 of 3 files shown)


pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -914,6 +914,7 @@ var _ = Describe("Generating fluentd config", func() {
 						@id retry_infra_es
 						host es.svc.infra.cluster
 						port 9999
+                        verify_es_version_at_startup false
```
ewolinetz, line 917: formatting?

ewolinetz:
@jcantrill so this only stops version verification? it doesn't also check that ES is available?

jcantrill:
> @jcantrill so this only stops version verification?  it doesn't also check that ES is available?
Yes and yes.  It has to contact ES in order to verify the version.  This blocks fluent initialization entirely https://github.com/uken/fluent-plugin-elasticsearch#fluentd-seems-to-hang-if-it-unable-to-connect-elasticsearch-why  as I observed why testing out buffers and queue limits.  This would be problematic for LF if a user defined an ES output that was down or not reachable

ewolinetz:
ack, so its redundant with the retry logic anyways...

PR #460, opened 2020-04-07
LOG-668: Remove unused code in fluent run.sh

jcantrill:
This PR:
* cleans up fluent run.sh to remove code that is no longer used

Ref: https://issues.redhat.com/browse/LOG-668

### DIFF
--- a/files/fluentd/run.sh
+++ b/files/fluentd/run.sh
@@ -1,9 +1,6 @@
 #!/bin/bash
 
-export MERGE_JSON_LOG=${MERGE_JSON_LOG:-false}
 CFG_DIR=/etc/fluent/configs.d
-ENABLE_PROMETHEUS_ENDPOINT=${ENABLE_PROMETHEUS_ENDPOINT:-"true"}
-OCP_OPERATIONS_PROJECTS=${OCP_OPERATIONS_PROJECTS:-"default openshift openshift- kube-"}
 
 fluentdargs="--no-supervisor"
 # find the sniffer class file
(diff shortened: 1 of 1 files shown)


pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -296,7 +282,7 @@ var _ = Describe("Generating fluentd config", func() {
 					elasticsearch_index_prefix_field 'viaq_index_name'
 					default_keep_fields CEE,time,@timestamp,aushape,ci_job,collectd,docker,fedora-ci,file,foreman,geoip,hostname,ipaddr4,ipaddr6,kubernetes,level,message,namespace_name,namespace_uuid,offset,openstack,ovirt,pid,pipeline_metadata,rsyslog,service,systemd,tags,testcase,tlog,viaq_msg_id
 					extra_keep_fields "#{ENV['CDM_EXTRA_KEEP_FIELDS'] || ''}"
-					keep_empty_fields "#{ENV['CDM_KEEP_EMPTY_FIELDS'] || 'message'}"
+					'message'
```
ewolinetz, line 285: shouldn't this still be `keep_empty fields 'message'` ?

files/fluentd/run.sh
```diff
@@ -15,49 +12,20 @@ if [ -n "$sniffer" -a -f "$sniffer" ] ; then
...
 
 issue_deprecation_warnings() {
     : # none at the moment
 }
 
-if [ -z "${JOURNAL_SOURCE:-}" ] ; then
-    if [ -d /var/log/journal ] ; then
-        export JOURNAL_SOURCE=/var/log/journal
-    else
-        export JOURNAL_SOURCE=/run/log/journal
```
ewolinetz, line 38: what was the case that it would be `/run/log/journal` again?

pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -321,7 +307,7 @@ var _ = Describe("Generating fluentd config", func() {
...
 						type k8s_journal
-						remove_keys "#{ENV['K8S_FILTER_REMOVE_KEYS'] || 'log,stream,MESSAGE,_SOURCE_REALTIME_TIMESTAMP,__REALTIME_TIMESTAMP,CONTAINER_ID,CONTAINER_ID_FULL,CONTAINER_NAME,PRIORITY,_BOOT_ID,_CAP_EFFECTIVE,_CMDLINE,_COMM,_EXE,_GID,_HOSTNAME,_MACHINE_ID,_PID,_SELINUX_CONTEXT,_SYSTEMD_CGROUP,_SYSTEMD_SLICE,_SYSTEMD_UNIT,_TRANSPORT,_UID,_AUDIT_LOGINUID,_AUDIT_SESSION,_SYSTEMD_OWNER_UID,_SYSTEMD_SESSION,_SYSTEMD_USER_UNIT,CODE_FILE,CODE_FUNCTION,CODE_LINE,ERRNO,MESSAGE_ID,RESULT,UNIT,_KERNEL_DEVICE,_KERNEL_SUBSYSTEM,_UDEV_SYSNAME,_UDEV_DEVNODE,_UDEV_DEVLINK,SYSLOG_FACILITY,SYSLOG_IDENTIFIER,SYSLOG_PID'}"
+						remove_keys log,stream,MESSAGE,_SOURCE_REALTIME_TIMESTAMP,__REALTIME_TIMESTAMP,CONTAINER_ID,CONTAINER_ID_FULL,CONTAINER_NAME,PRIORITY,_BOOT_ID,_CAP_EFFECTIVE,_CMDLINE,_COMM,_EXE,_GID,_HOSTNAME,_MACHINE_ID,_PID,_SELINUX_CONTEXT,_SYSTEMD_CGROUP,_SYSTEMD_SLICE,_SYSTEMD_UNIT,_TRANSPORT,_UID,_AUDIT_LOGINUID,_AUDIT_SESSION,_SYSTEMD_OWNER_UID,_SYSTEMD_SESSION,_SYSTEMD_USER_UNIT,CODE_FILE,CODE_FUNCTION,CODE_LINE,ERRNO,MESSAGE_ID,RESULT,UNIT,_KERNEL_DEVICE,_KERNEL_SUBSYSTEM,_UDEV_SYSNAME,_UDEV_DEVNODE,_UDEV_DEVLINK,SYSLOG_FACILITY,SYSLOG_IDENTIFIER,SYSLOG_PID
```
ewolinetz, line 310: single quotes aren't required?

pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -727,7 +699,7 @@ var _ = Describe("Generating fluentd config", func() {
 					elasticsearch_index_prefix_field 'viaq_index_name'
 					default_keep_fields CEE,time,@timestamp,aushape,ci_job,collectd,docker,fedora-ci,file,foreman,geoip,hostname,ipaddr4,ipaddr6,kubernetes,level,message,namespace_name,namespace_uuid,offset,openstack,ovirt,pid,pipeline_metadata,rsyslog,service,systemd,tags,testcase,tlog,viaq_msg_id
 					extra_keep_fields "#{ENV['CDM_EXTRA_KEEP_FIELDS'] || ''}"
-					keep_empty_fields "#{ENV['CDM_KEEP_EMPTY_FIELDS'] || 'message'}"
+					'message'
```
ewolinetz, line 702: same ask

files/fluentd/run.sh
jcantrill: pre/post atomic? :man_shrugging:   I removed it because we explicitly rely on one path with the generator so this block is of no value unless we are going to modify the generator

pkg/generators/forwarding/fluentd/fluent_conf_test.go
jcantrill: not if this is an example: https://github.com/openshift/cluster-logging-operator/pull/460/files/7263b5a6deef9be538019de0e1b733139ba380d8#diff-8a739dd813c84684d327a66b7b62df58L335

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -165,20 +165,6 @@ const fluentConfTemplate = `{{- define "fluentConf" }}
...
-		json_fields "#{ENV['JSON_FIELDS'] || 'MESSAGE,log'}"
-	</filter>
-		
-	<filter kubernetes.var.log.containers.**>
-		@type parse_json_field
-		merge_json_log "#{ENV['MERGE_JSON_LOG'] || 'false'}"
-		preserve_json_log "#{ENV['PRESERVE_JSON_LOG'] || 'true'}"
-		json_fields "#{ENV['JSON_FIELDS'] || 'log,MESSAGE'}"
-	</filter>
-
```
ewolinetz, line 181: can you remind me again what this was doing and why its now alright to remove it? is this only for the `merge_json` case?

jcantrill: I'll add these back in since 'preserve_json_log' is true and we set the env to true.  I dont recall what this plugin does

tux-o-matic: From the [Fluend plugin doc](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter):
NOTE: As of the release 2.1.x of this plugin, it no longer supports parsing the source message into JSON and attaching it to the payload. The following configuration options are removed:
- merge_json_log
- preserve_json_log

jcantrill: resolving as I reduced the scope of the PR

PR #461, opened 2020-04-08
Formatted cm/fluentd:fluent.conf so it is usable with oc get cm/fluentd -oyaml

syedriko:
- Formatted cm/fluentd:fluent.conf so it is usable with ```oc get cm/fluentd -oyaml```
- Added build-debug and run-debug targets to the Makefile
- Added error handling during fluent.conf generation
- Made lint happy in test/framework.go

Before:
```
$ oc get cm/fluentd -oyaml
apiVersion: v1
data:
  fluent.conf: "## CLO GENERATED CONFIGURATION ###\n# This file is a copy of the fluentd
    configuration entrypoint\n# which should normally be supplied in a configmap.\n\n<system>\n
    \ @log_level \"#{ENV['LOG_LEVEL'] || 'warn'}\"\n</system>\n\n# In each section
```

After:
```
$ oc get cm/fluentd -oyaml
apiVersion: v1
data:
  fluent.conf: |+
    ## CLO GENERATED CONFIGURATION ###
    # This file is a copy of the fluentd configuration entrypoint
    # which should normally be supplied in a configmap.

    <system>
      @log_level "#{ENV['LOG_LEVEL'] || 'warn'}"
    </system>
...
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,9 +26,11 @@ imagebuilder:
 golangci-lint:
 	@type -p golangci-lint > /dev/null || go get github.com/golangci/golangci-lint/cmd/golangci-lint
 
-
 build:
-	go build -o bin/cluster-logging-operator ./cmd/manager
+	go build $(BUILD_OPTS) -o bin/cluster-logging-operator ./cmd/manager
+
+build-debug:
+	$(MAKE) build BUILD_OPTS='-gcflags=all="-N -l"'
 
 run:
 	ELASTICSEARCH_IMAGE=quay.io/openshift/origin-logging-elasticsearch6:latest \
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -26,11 +26,15 @@ imagebuilder:
...
+build-impl:
+	go build $(BUILD_OPTS) -o bin/cluster-logging-operator ./cmd/manager
 
-build:
-	go build -o bin/cluster-logging-operator ./cmd/manager
+build: build-impl
 
-run:
+build-debug: BUILD_OPTS=-gcflags=all="-N -l"
+build-debug: build-impl
```
periklis, line 35: This could be simplified to keep the number of make targets exposed to the user small.
```
build:
 go build $(BUILD_OPTS)

build-debug:
  $(make) build BUILD_OPTS=...
```

WDYT?

Makefile
```diff
@@ -42,7 +46,13 @@ run:
...
 	WORKING_DIR=$(TARGET_DIR)/ocp-clo \
 	LOGGING_SHARE_DIR=$(CURDIR)/files \
-	go run cmd/manager/main.go
+	$(RUN_WHAT)
+
+run: RUN_WHAT=go run cmd/manager/main.go
+run: run-impl
+
+run-debug: RUN_WHAT=dlv exec -- bin/cluster-logging-operator
+run-debug: run-impl
```
periklis, line 55: Same here as with the build targets.

Makefile
syedriko: I'm usually allergic to recursive make, but in this case it doesn't really matter. Let's go with fewer make targets.

PR #467, opened 2020-04-13
Updated syslog LF testcase for each syslog rfc

vimalk78:
Updated syslog LF testcase for each syslog rfc
 - the syslog receiver is created with specific rfc parser (rfc3164, rfc5424)
 - the testcase Expect clause checks for specific fields in the received syslog messages
 - increased timeout for logforwarding testcase completion, and added `-ginkgo.noColor`
 - capturing syslog-receiver logs and configuration in artifacts

### DIFF
--- a/pkg/generators/forwarding/fluentd/syslog_conf.go
+++ b/pkg/generators/forwarding/fluentd/syslog_conf.go
@@ -6,3 +6,9 @@ func (conf *outputLabelConf) SyslogPlugin() string {
 	}
 	return "syslog_buffered"
 }
+
+func (conf *outputLabelConf) Rfc() string {
+	//return "rfc3164"
+	// TODO update it whem merged with new v1 api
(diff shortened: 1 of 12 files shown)


pkg/generators/forwarding/fluentd/syslog_conf.go
```diff
@@ -6,3 +6,9 @@ func (conf *outputLabelConf) SyslogPlugin() string {
 	}
 	return "syslog_buffered"
 }
+
+func (conf *outputLabelConf) Rfc() string {
+	//return "rfc3164"
+	// TODO update it whem merged with new v1 api
+	return "rfc5424"
```
vimalk78, line 13: have tested with both "rfc5424" and "rfc3164", and both tests pass with changes in https://github.com/openshift/origin-aggregated-logging/pull/1874

PR #469, opened 2020-04-14
Ensure LogForwarding works with multiple outputs

periklis:
This PR addresses the case of a LogForwarding CR with multiple outputs. It ensures that all available outputs receive the requests logs according to the defined pipeline.

Depends on: #427


alanconway:
I think we can merge this now and resolve separately when we merge to master, or earlier on master-post-release if that takes a long time.

PR #473, opened 2020-04-15
Uniform environment variable to skip cleanup for E2E tests

periklis:
This PR uniforms the know to skip clean-up scripts for E2E tests. The environemt variable `DO_CLEANUP` needs to be set explicitly to `false` for this purpose.


jcantrill:
@periklis please flip these changes to do `DO_CLEANUP` and `DO_SETUP` which is more affirmative  IMO

PR #476, opened 2020-04-16
make manifests dir world writable in operator-registry image

bparees:


bparees:
/hold

this should not be needed, if this is needed there is a bug in crio.

bparees:
manually merged on red to unblock the team.

PR #478, opened 2020-04-17
Bug 1825275: Make useable by buildah

jcantrill:
Ref https://bugzilla.redhat.com/show_bug.cgi?id=1825275


bparees:
overriding bz req, we're juggling a few different bzs related to buildah compatibility issues but this PR addresses them and references one.

PR #479, opened 2020-04-17
Bug 1780698: Fix disk pressure for fluentd buffer chunks

periklis:
This change limits the fluentd queue length to match the number of chunks on the filesystem.

To address: https://bugzilla.redhat.com/show_bug.cgi?id=1780698

Needs backport for: release-4.4 and release 4.3.

### DIFF
--- a/pkg/generators/forwarding/fluentd/templates.go
+++ b/pkg/generators/forwarding/fluentd/templates.go
@@ -546,7 +546,7 @@ tls_cert_path {{ .SecretPath "ca-bundle.crt"}}
 <buffer>
   @type file
   path '{{.BufferPath}}'
-  queued_chunks_limit_size "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
+  queue_limit_length "#{ENV['BUFFER_QUEUE_LIMIT'] || '32' }"
   chunk_limit_size "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
   flush_interval "#{ENV['FORWARD_FLUSH_INTERVAL'] || '5s'}"
   flush_at_shutdown "#{ENV['FLUSH_AT_SHUTDOWN'] || 'false'}"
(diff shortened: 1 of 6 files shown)


pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -546,7 +546,7 @@ tls_cert_path {{ .SecretPath "ca-bundle.crt"}}
 <buffer>
   @type file
   path '{{.BufferPath}}'
-  queued_chunks_limit_size "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
+  queue_limit_length "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
```
jcantrill, line 549: What should be the correct fallback here?  In one place we say 32 and the other 1024?  seems like the default is maybe conflating two distinct concepts?

periklis: You need to help me out here. This is the buffer for secure-forward/forward.

periklis: AFAICS per git blame, this was introduced with LogForwarding.

jcantrill: It's a hold over from earlier releases.

jcantrill: I think 32 is probably the "correct" value since its a number of files is it not?

periklis: Yes queue_limit_length refers to number of chunk files.

test/files/secure-forward.conf
```diff
@@ -11,7 +11,7 @@
   <buffer>
     @type file
     path '/var/lib/fluentd/secureforwardlegacy'
-    queued_chunks_limit_size "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
+    queue_limit_length "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
```
ewolinetz, line 14: should this be `32` as well?

periklis: This is used only for an E2E test, so basically arbitrary data here.

periklis: To be precise the file is used for an unmanaged fluentd receiver to test log forwarding.

ewolinetz: ack, thanks

PR #482, opened 2020-04-18
bug 1826095: make schema valid for clusterlogging

jcantrill:
Update CLO CRD to be a valid schema


hack/testing-olm/assertions
```diff
@@ -14,3 +11,18 @@ assert_resources_exist(){
...
+
+assert_kibana_instance_exists() {
+	# verify kibana crd
+	try_until_success "oc -n $NAMESPACE get kibana instance" ${TIMEOUT_MIN}
+}
+
+is_crd_installed() {
+  crd_name=$1
+  cmd="oc get crd $crd_name"
+  if $cmd; then
```
syedriko, line 23:
```
$cmd
return $?
```

hack/testing-olm/utils
```diff
@@ -193,3 +193,44 @@ EOL
...
 }
+
+get_github_directory() {
+    local outputdir=$1
+    local ghdir=$2
+    local comp=$3
+    local repo=${4:-openshift}
+    local branch=${5:-master}
+    local name
+    local type
```
syedriko, line 204: type is a bash built-in, can we use another name please?

jcantrill: fixed.  this was added from whence it was already commited

PR #485, opened 2020-04-20
Addressing error being printed out when we try to create a namespace …

ewolinetz:
…that already exists for tests


ewolinetz:
/hold
lower priority fix

jcantrill:
@ewolinetz any idea if this is still valid?

PR #489, opened 2020-04-21
remove obsolete test logic

bparees:


hack/lib/cleanup.sh
```diff
@@ -36,7 +36,6 @@ function os::cleanup::all() {
 	os::cleanup::truncate_large_logs
 
 	if [[ -z "${SKIP_TEARDOWN:-}" ]]; then
-		os::cleanup::containers
```
bparees, line 39: i think there is a lot more that can/should be removed from this cleanup routine given that this is used completely differently from how it used to be, but this is one i'm reasonably confident of and takes a bunch of time to run locally.

hack/testing/utils
```diff
@@ -1,410 +0,0 @@
...
-      return 0
-    fi  
-    sleep $interval
-    now=$(date +%s%3N)
-  done
-  log::error "try_until_text expired for '$cmd' == '$expected'"
-  return 1
-}
-
-gather_logging_resources() {
```
vimalk78, line 80: The corresponding method in `hack/testing-olm/utils` is not this detailed.
Copy to `testing-olm` before deleting.

bparees: done, thanks for catching it.

hack/testing-olm/utils
```diff
@@ -83,15 +83,38 @@ gather_logging_resources() {
   set +e
   local LOGGING_NS=$1
   local outdir=${2:-$ARTIFACT_DIR}
+  local runtime=${3:-$(date +%s)}
+  outdir=$outdir/$runtime
```
ewolinetz, line 87: why are we setting `outdir` when we have it set above as well?

bparees: i don't know, i copied this from the hack/testing/utils script:
https://github.com/openshift/cluster-logging-operator/blob/master/hack/testing/utils#L83-L85

seems like it's trying to build up a path.

PR #490, opened 2020-04-22
Migrate to go modules

periklis:
### Motivation
This PR migrates dependency management from dep to go modules to enable forward versioned support for k8s >1.18 api, client-go, library-go. In addition the Makefile includes following enhancements:
- Pin `GOFLAGS` to `-mod=vendor` for vendoring all go module dependencies.


jcantrill:
Leave hold in place until either we are comfortable landing LF GA in 4.5 or push to 4.6

periklis:
@jcantrill This also need a last review please.

alanconway:
/lgtm
I don't think this will cause any problems merging LF GA (I could be wrong) but I'll leave the hold for @jcantrill to remove in case I've missed something...

jcantrill:
> /lgtm
> I don't think this will cause any problems merging LF GA (I could be wrong) but I'll leave the hold for @jcantrill to remove in case I've missed something...

In the interest of priorities, would prefer to land LF GA into master and know for certain

PR #491, opened 2020-04-22
[release-4.4] Bug 1826861: Fix disk pressure for fluentd buffer chunks

periklis:
This PR backports a serious issue we have according to multiple customer cases regarding fluentd buffers taking up all the node filesystem until crashing the node itself.

Manual cherry-pick for: #479 because of [reason](https://github.com/openshift/cluster-logging-operator/pull/479#issuecomment-617887451)

To address: https://bugzilla.redhat.com/show_bug.cgi?id=1826861

### DIFF
--- a/test/files/secure-forward.conf
+++ b/test/files/secure-forward.conf
@@ -11,7 +11,7 @@
   <buffer>
     @type file
     path '/var/lib/fluentd/secureforwardlegacy'
-    queued_chunks_limit_size "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
+    queue_limit_length "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
     chunk_limit_size "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
     flush_interval "#{ENV['FORWARD_FLUSH_INTERVAL'] || '5s'}"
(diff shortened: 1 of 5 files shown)


test/files/secure-forward.conf
```diff
@@ -11,7 +11,7 @@
   <buffer>
     @type file
     path '/var/lib/fluentd/secureforwardlegacy'
-    queued_chunks_limit_size "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
+    queue_limit_length "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
```
jcantrill, line 14: @periklis missed on here

periklis: Actually this is a test configuration for a third-party managed fluentd receiver, which we only use on the corresponding E2E logfowarding test. See more [here](https://github.com/openshift/cluster-logging-operator/pull/479#discussion_r412995527)

periklis:
> Looks good. But maybe it makes sense to make this magic number '32' a constant somewhere? Like `const default_queue_limit_length = 32`

Let's keep it on 4.5/4.6 backlog because these constants are easily spread across fluentd's `run.sh` and the configuration templates. IIRC Jeff would like to see a cleanup/refactoring of all the environment variables that override fluentd buffer parameters because the environment variables cannot be overridden in the deamonset anymore, expect in the unmanaged case. TBH, if we need these knobs we should expose them in the CRD and not per env.

PR #496, opened 2020-04-27
Added a wrapper over oc to run oc commands from golang

vimalk78:
Added a wrapper over oc to run oc commands from golang

 - supports oc.Exec() for 'oc exec'
 - supports oc.Get() for 'oc get' 
 - supports oc.Literal() for other oc commads which do not have a DSL yet. can be added later.
- added unit test cases

### DIFF
--- a/test/helpers/oc/exec.go
+++ b/test/helpers/oc/exec.go
@@ -1,0 +1,139 @@
+package oc
+
+import (
+	"fmt"
+	"strings"
+)
+
+// Execer is interface for collecting arguments for Exec command
+type Execer interface {
+	Command
+
+	// argument for option --config
+	WithConfig(string) Execer
+	// argument for option -n
+	WithNamespace(string) Execer
+
+	// argument for podname
+	Pod(string) Execer
+	// a command can be composed to get the pod name. e.g. oc.Get or oc.Literal
+	WithPodGetter(Command) Execer
+
+	// argument for option -c
+	Container(string) Execer
+	// a command can be composed to get the container name. e.g. oc.Get or oc.Literal
+	WithContainerGetter(Command) Execer
+
+	// command and argument to be executed in pod
+	// arguments for option --
+	WithCmd(string, ...string) Execer
+}
+
+type exec struct {
+	*runner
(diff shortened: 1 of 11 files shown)


test/helpers/oc/exec.go
```diff
@@ -0,0 +1,139 @@
...
+	Container(string) Execer
+	// a command can be composed to get the container name. e.g. oc.Get or oc.Literal
+	WithContainerGetter(Command) Execer
+
+	// command and argument to be executed in pod
+	// arguments for option --
+	WithCmd(string, ...string) Execer
+}
+
+type exec struct {
```
jcantrill, line 32: @alanconway I think we should consider refraining from introducing command specific structs so that we can stay away from creating a bunch of these to wrapper oc; I suspect we will have a very limited number of them.  The code in origin has a `cli` struct that keeps all the bits needed:  subcommand, args, global options (e.g. namespace).  It then uses them to build up the final call.  

I want to make sure we are not introducing too much here

vimalk78: we could create a common struct and embed it in type specific structs

jcantrill: @vimalk78 I spoke with @alanconway and he is good with your approach.  Ignore my feedback

bparees:
this is a more specific implementation than i would have gone with.  The origin tests literally just use a "oc $args" style, they don't care which specific command (e.g. `exec`) you are trying to run.   

You're going to have a lot of work to do to maintaiin/extend this for each command people want to use it with.

alanconway:
> this is a more specific implementation than i would have gone with. The origin tests literally just use a "oc $args" style, they don't care which specific command (e.g. `exec`) you are trying to run.
> 
> You're going to have a lot of work to do to maintaiin/extend this for each command people want to use it with.

oc.Literal("oc args...") gives you that - so we can have a type-safe API for the most common commands but we don't have to implement them all.

vimalk78:
ping @bparees , requesting approval

bparees:
overriding BZ req because these are test code only changes

PR #501, opened 2020-05-01
Make a faster build for development

jcantrill:
This PR reduces the build time for local development

### DIFF
--- a/Dockerfile.fast
+++ b/Dockerfile.fast
@@ -1,0 +1,20 @@
+FROM centos:centos7
+RUN INSTALL_PKGS=" \
+      openssl \
+      " && \
+    yum install -y $INSTALL_PKGS && \
+    rpm -V $INSTALL_PKGS && \
+    yum clean all && \
+    mkdir /tmp/ocp-clo && \
+    chmod og+w /tmp/ocp-clo
+COPY scripts/* /usr/bin/scripts/
+RUN mkdir -p /usr/share/logging/
+COPY files/ /usr/share/logging/
+
+COPY manifests /manifests
+RUN rm /manifests/art.yaml
+
+# this is required because the operator invokes a script as `bash scripts/cert_generation.sh`
+WORKDIR /usr/bin
+ENTRYPOINT ["/usr/bin/cluster-logging-operator"]
+COPY bin/cluster-logging-operator /usr/bin/
--- a/Makefile
+++ b/Makefile
@@ -51,10 +51,13 @@ clean:
 	rm -rf tmp
 	go clean -cache -testcache ./...
 
-image:
+image: build
 	@if [ $${SKIP_BUILD:-false} = false ] ; then \
-		podman build -t $(IMAGE_TAG) . ; \
+		podman build -t $(IMAGE_TAG) -f Dockerfile.fast . ; \
 	fi
+
+image-ci:
+	podman build -t $(IMAGE_TAG) . ; \
 
 lint: fmt
 	@$(MAKE) golangci-lint


jcantrill:
cc @bparees

jcantrill:
> I believe this is not the right approach that results in maintaining multiple Dockerfiles - in total 3 by now - for building the same project. 

This is not exactly true.  One of these files is for doing e2e tests which is unrelated to the needs of building the operator

> Fixes/Additions could easily slip out of one of the Dockerfiles. The more consistent approach is to define exactly the copys and the order in the Dockerfile to make use of proper usage of layer caching. Latter is unfortunately disabled by this PR. My proposal is to go with:
> 
> ```
> FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
> WORKDIR /go/src/github.com/openshift/cluster-logging-operator
> 
> #
> # Order of COPY statements based on frequency of change.
> # This enables better caching of each layer.
> #
> 
> # Configuration
> COPY config config
> COPY deploy deploy
> COPY files files
> 
> # manifests
> COPY manifests manifests
> COPY olm_deploy olm_deploy
> 
> # vendored dependencies
> COPY Gopkg.lock Gopkg.lock
> COPY Gopkg.toml Gopkg.toml
> COPY vendor vendor
> 
> # source code
> COPY cmd cmd
> COPY pkg pkg
> COPY scripts scripts
> COPY version version
> 
> # make targets
> COPY Makefile Makefile
> RUN make build
> 
> FROM centos:centos7
> RUN INSTALL_PKGS=" \
>       openssl \
>       " && \
>     yum install -y $INSTALL_PKGS && \
>     rpm -V $INSTALL_PKGS && \
>     yum clean all && \
>     mkdir /tmp/ocp-clo && \
>     chmod og+w /tmp/ocp-clo
> COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/bin/cluster-logging-operator /usr/bin/
> COPY scripts/* /usr/bin/scripts/
> RUN mkdir -p /usr/share/logging/
> COPY files/ /usr/share/logging/
> 
> COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/manifests /manifests
> RUN rm /manifests/art.yaml
> 
> # this is required because the operator invokes a script as `bash scripts/cert_generation.sh`
> WORKDIR /usr/bin
> ENTRYPOINT ["/usr/bin/cluster-logging-operator"]
> LABEL io.k8s.display-name="OpenShift cluster-logging-operator" \
>       io.k8s.description="This is a component of OpenShift Container Platform that manages the lifecycle of the Aggregated logging stack." \
>       io.openshift.tags="openshift,logging,cluster-logging" \
>       com.redhat.delivery.appregistry=true \
>       maintainer="AOS Logging <aos-logging@redhat.com>"
> ```
> 
>  The COPY directives are grouped and ordered in the estimated order of usage frequency. This is enabled faster rebuilds by using layer caching properly. WDYT?

I think there needs to be ALOT more consideration how to utilize a single Dockerfile.  There is conflation in your suggestion of what is needed for testing and what is needing for the actual deployment of the operator.  I defer to the rest of the team here but my suggestion in the interest of accelerating development would be to merge this change and resolve your specific issues later.  The amount of duplication in the fast and non-fast files is minimal IMO.  We've cared multiple files in other repos for many years and have rarely had issues.

jcantrill:
I can see in this case however.. maybe the copy bin should be last

alanconway:
I would be for merging Dockerfile.fast in the short term, as it serves an immediate purpose; but also with merging that back into a single Dockerfile in the future if we find a way to do that that meets all requirements.
Note that the current problem is a podman bug - podman takes an obscenely long time to copy large numbers of small files compared to docker. I/whenf podman fixes that issue then we can throw away Dockerfile.fast and go back to a single build process for all (with the improvements Peri suggests) which would definitely be the best situation.

alanconway:
We are not the only ones affected by the podman bug: https://github.com/containers/buildah/issues/1849

periklis:
> I think there needs to be ALOT more consideration how to utilize a single Dockerfile. There is conflation in your suggestion of what is needed for testing and what is needing for the actual deployment of the operator. I defer to the rest of the team here but my suggestion in the interest of accelerating development would be to merge this change and resolve your specific issues later. The amount of duplication in the fast and non-fast files is minimal IMO. We've carried multiple files in other repos for many years and have rarely had issues.

IMHO it is not an issue to have the test files into the single Dockerfile until we untangle them. Since we use a multi-stage builds in the same Dockerfile we need only to ensure that the COPY statements in the final build are the smallest possible footprint namely the operator binary and the scripts.

Furthermore, single precise COPY directives that enable layer caching work like incremental builds. Since the project order of an operator stays the same for a long time one can see in my proposed Dockerfile that this will also benefit CI, e.g.:
- If CRDs change then everything below `COPY manifests manifests` will be rebuild both on CI and on our local builds
- If vendors change then everything below `COPY Gopkg.lock Gopkg.lock` will be rebuild.
- Usual code change in pkg will be fast on both CI and local builds

IMHO, if the container build is broken for what ever reason as #394 I want to see this locally and not on CI or way later in ART builds.

TBH, the proposed Dockerfile is already fast enough. However multiple Dockerfile defeat the idea of multi-staged builds in the first place, don't you think?

periklis:
> We are not the only ones affected by the podman bug: [containers/buildah#1849](https://github.com/containers/buildah/issues/1849)

Good to know. However, as my nature commands I digged through it. The proposed/merged fix is a `.dockerignore` fix which landed in buildah 1.4.8 that in turn is already included in podman >1.9.0. Using podman 1.9.1 the timings did not improve:
1. cold cache after `podman system prune -a -f`:
```
make image  
332.12s user 
239.55s system 
135% cpu 
7:02.22 total
```
2. warm cache:
```
make image  
321.07s user 
237.82s system
145% cpu 
6:24.72 total
```
Of course YMMV on each machine/filesystem/etc. combination. These benchmarks were taken on a Fedora 32 fully refreshed on ext4/luks-encrypted device and with podman 1.9.1

jcantrill:
> > I think there needs to be ALOT more consideration how to utilize a single Dockerfile. There is conflation in your suggestion of what is needed for testing and what is needing for the actual deployment of the operator. I defer to the rest of the team here but my suggestion in the interest of accelerating development would be to merge this change and resolve your specific issues later. The amount of duplication in the fast and non-fast files is minimal IMO. We've carried multiple files in other repos for many years and have rarely had issues.
> 
> IMHO it is not an issue to have the test files into the single Dockerfile until we untangle them. 

You are completely missing what I am stating; one of our  Dockerfile's has no relevance to this discussion as its not used to build: https://github.com/openshift/cluster-logging-operator/blob/master/Dockerfile.e2e

> Since we use a multi-stage builds in the same Dockerfile we need only to ensure that the COPY statements in the final build are the smallest possible footprint namely the operator binary and the scripts.
> 
> Furthermore, single precise COPY directives that enable layer caching work like incremental builds. Since the project order of an operator stays the same for a long time one can see in my proposed Dockerfile that this will also benefit CI, e.g.:
> 
> * If CRDs change then everything below `COPY manifests manifests` will be rebuild both on CI and on our local builds
> * If vendors change then everything below `COPY Gopkg.lock Gopkg.lock` will be rebuild.
> * Usual code change in pkg will be fast on both CI and local builds
> 
> IMHO, if the container build is broken for what ever reason as #394 I want to see this locally and not on CI or way later in ART builds.
> 
> TBH, the proposed Dockerfile is already fast enough. However multiple Dockerfile defeat the idea of multi-staged builds in the first place, don't you think?

alanconway:
On Tue, May 5, 2020 at 3:41 AM Periklis Tsirakidis <notifications@github.com>
wrote:

> We are not the only ones affected by the podman bug:
> containers/buildah#1849
> <https://github.com/containers/buildah/issues/1849>
>
> Good to know. However, as my nature commands I digged through it. The
> proposed/merged fix is a .dockerignore fix which landed in buildah 1.4.8
> that in turn is already included in podman >1.9.0. Using podman 1.9.1 the
> timings did not improve:
>
In our case the problem is not that we're copying including unnecessary
files, so .dockerignore won't help. It's the gigantic vendor tree. Without
that we would never have noticed any problem.

PR #509, opened 2020-05-08
Added proper synchronization for oc port-forward to the remote image registry

syedriko:


hack/deploy-image.sh
```diff
@@ -3,16 +3,18 @@
 set -euo pipefail
 
 echo "Setting up port-forwarding to remote registry ..."
-oc -n openshift-image-registry port-forward service/image-registry 5000:5000 > pf.log 2>&1 &
-forwarding_pid=$!
-trap "kill -15 ${forwarding_pid}" EXIT
+coproc oc -n openshift-image-registry port-forward service/image-registry 5000:5000
```
jcantrill, line 6: This does not appear to be a utility commonly available on someone's dev machine. If we are to proceed to use it, you may wish to add it to the README

jcantrill:
cc @bparees

PR #511, opened 2020-05-08
adding back PUSH_USER mechanism for deploy-image

ewolinetz:


hack/deploy-image.sh
```diff
@@ -2,15 +2,54 @@
...
+        oc login -u "$username" -p "$PUSH_PASSWORD" > /dev/null
+        token=$( oc whoami -t 2> /dev/null || : )
+        oc config use-context "$savectx"
+    else
+        # see if current context has a token
+        token=$( oc whoami -t 2> /dev/null || : )
+        if [ -n "$token" ] ; then
+            username=$( oc whoami )
+        else
+            # get the first user with a token
```
syedriko, line 23: Should we be looking for the first user with a token in the right cluster?

syedriko:
@ewolinetz Just wanted to mention an adjacent PR #509

ewolinetz:
Will document proper "how to" using `oc login` instead

PR #494, opened 2020-04-24
fix up make deploy to use catalog

jcantrill:
fix up make deploy to use catalog

PR #514, opened 2020-05-08
LOG-764: Improved `make run` with self setup, consolidated README.

alanconway:
Improved the `make run` target to be self-contained and idempotent.
Consolidated HACKING into README and re-orgainzed to reflect
recent dev experience.

Ref https://bugzilla.redhat.com/show_bug.cgi?id=1833520

### DIFF
--- a/HACKING.md
+++ b/HACKING.md
@@ -1,5 +1,139 @@
-# Hacking on the `cluster-logging-operator`
+# Hacking on the Cluster Logging Operator
+
+## Preparation
+
+Clone this repository into `$GOPATH/github.com/openshift/cluster-logging-operator`
+
+Install `podman`. For example on  Fedora:
+```
+sudo dnf install -y podman
+```
+
+Check out some other required repositories in your $GOPATH:
+```
+go get -d -u github.com/openshift/elasticsearch-operator/... github.com/openshift/origin-aggregated-logging/...
+```
+
+**Note:** `podman` copies our source tree very slowly. You can use `docker`
+instead, this will not exactly reproduce how images are built for release or by
+CI, but is close enough for many purposes. Put a symlink early in your path:
+```
+ln -s $(which docker) /somewhere/early/in/my/PATH/podman`
+```
+
+## Testing Overview
+
+There are two types of test:
+
+* *Unit tests:* run in seconds, no cluster needed, verify internal behavior - behind the API. New code *should* be reasonably well covered by unit tests but we don't have a formal requirement.
+* *E2E tests:* run in minutes to hour(s), requires a cluster. There are several ways to run these:
+   - *CI tests:* run automatically when you raise a PR in a controlled cluster environment. Can take hour(s) to get results.
+   - *Deployed tests:* Build and deploy CLO image to your own cluster, allows you to run tests selectively.
+   - *Local process:* requires a cluster with dependencies deployed (elasticsearch), but runs CLO as a *local process*. Faster and easier to debug.
+
+All PRs should pass unit tests. CI runs all tests automatically and won't merge a PR till they pass, so it is OK to create a PR and let CI run the e2e tests if you've done enough testing to be confident. This allows for human review and feedback while CI tests are running.
+
+To get a code review of unfinished work, create a PR with "[WIP]" at the start of the description. The CI system will not merge it until the "[WIP]" is removed and all tests are passing.
+
+## Setting up a test cluster
+
+You can use a real cluster (for example an AWS cluster) or create a virtual cluster on your development box with [Code Ready Containers](https://developers.redhat.com/products/codeready-containers/download)
+
+If you use CRC, you may need to start with more memory than the default, e.g.:
+```
+crc start -m 12288
+```
+
+Log in as a user with the role `cluster-admin` - user `kubeadmin` is often predefined with this role.
+
+You can `export KUBECONFIG=/path/to/cluster/config` or copy/add your config to `$HOME/.kube/config`. Depending on how you set it up, you may also need to `oc login` to your cluster and add pull secrets to your credentials.
+
+## Building and running tests
+
+Unit tests can be run without a cluster:
+* `make check`: Generate and format code, run unit tests and lint.
+   You should run *at least* this set of checks before a commit.
+
+To build, deploy and test the CLO image in your own cluster:
+* `make deploy`: Build CLO image and deploy to cluster with elasticsearch.
+* `make undeploy`: Undo `make deploy`.
+
+To run CLO as a local process:
+* `make run`: Run CLO as a local process. Does not require `make deploy`.
+* `make debug`: Run CLO in the `dlv` debugger.
+
+After running `make deploy`, `make run` or `make debug`, you can run e2e tests:
+* `make test-cluster` Run all e2e tests against the running CLO.
+* `go run ./test/e2e/something`: Run tests individually, in a debugger or whatever.
+
+To do a complete deploy, test, cleanup cycle exactly as CI does (this must start from a clean cluster e.g. after `make undeploy`):
+* `make test-olm-e2e`: Run e2e tests the same way CI does. See below for more.
+
+Some other useful targets (read the Makefile for more) 
+* `make image`: Build the image as $IMAGE_TAG, do not deploy it.
+* `make` or `make all`: Run `make check` and build the CLO image.
+
+## More about `make run`
+
+You can run the CLO as a local process, outside the cluster. This is *not* the
+normal way to run an operator, and does not test all aspects of the CLO
+(e.g. problems with the CSV or OLM interactions), but it has advantages:
+
+* Fast edit/run/test cycle - runs from source code, skips some slow build/deploy steps.
+* Directly access the CLO logs on stdout/stderr
+* Control logging levels and other environment variables, e.g. `export LOG_LEVEL=debug`
+* Run CLO in a debugger, profiler or other development tools.
+
+*How it works*: An operator is actually a cluster *client*. It watches for
+changes to its own custom resources, and creates/updates other resources
+accordingly. This can all be done from outside the cluster.
+
+Examples:
+```
+make run  # Run the CLO locally
+make run-debug  # Run CLO under the dlv debugger
+LOG_LEVEL=debug make run  # Run CLO with debug logging
+RUN_CMD=foo make run # Run CLO under imaginary "foo" debugger/profiler.
+```
+
+Note `make run` will not return until you terminate the CLO.
+
+### More about * `make test-olm-e2e`
+
+This test assumes:
+* the cluster-logging-catalog image is available
+* the cluster-logging-operator image is available
+* the cluster-logging component images are available (i.e. $docker_registry_ip/openshift/$component)
+
+**Note:** This test will fail if the images are not pushed to the cluster
+on which the operator runs or can be pulled from a visible registry.
+
+**Note:** It is necessary to set the `IMAGE_CLUSTER_LOGGING_OPERATOR` environment variable to a valid pull spec
+in order to run this test against local changes to the `cluster-logging-operator`. For example:
+```
+$ make deploy-image && IMAGE_CLUSTER_LOGGING_OPERATOR=image-registry.openshift-image-registry.svc:5000/openshift/origin-cluster-logging-operator:latest make test-e2e
+``**
+
+**Note:** To skip cleanup of resources while hacking/debugging an E2E test apply `DO_CLEANUP=false`.
+
+## Building a Universal Base Image (UBI) based image
+
+You must first `oc login api.ci.openshift.org`.  You'll need these credentials in order
+to pull images from the UBI registry.
+
+The image build process for UBI based images uses a private yum repo.
+In order to use the private yum repo, you will need access to
+https://github.com/openshift/release/blob/master/ci-operator/infra/openshift/release-controller/repos/ocp-4.1-default.repo
+and
+https://github.com/openshift/shared-secrets/blob/master/mirror/ops-mirror.pem
+Note that the latter is private and requires special permission to access.
+
+The best approach is to clone these repos under `$GOPATH/src/github.com/openshift`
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,22 +1,85 @@
...
-## Quick Start
+# Hacking
+
+Note: In the rest of this doc "CLO" stands for Cluster Logging Operator
+
+## Preparation
+
+Clone this repository into $GOPATH/github.com/openshift/cluster-logging-operator
+
+Install `podman` and `skopeo`. For example on  Fedora:
```
periklis, line 14: We don't use skopeo to push images anymore because using rootless podman stores all images under `$HOME/.local/containers` and AFAIK skopeo does not support rootless container storage. Anyway `podman push` does the work well enough!

README.md
```diff
@@ -1,22 +1,85 @@
...
 
-## Hacking
+After running  either `make deploy` or `make run`, you can run non-unit tests:
+* `make test-cluster` Run non-unit tests against an existing CLO.
+* `go run test/e2e/something`: Run tests individually.
 
-### Building a Universal Base Image (UBI) based image
+Some other useful targets (read the Makefile for more) 
+* `make image`: Build the image as $IMAGE_TAG, do not deploy it.
+* `make test-olm-e2e`: Run e2e the way the CI server does.
```
periklis, line 45: We should mention `make deploy-image` here.

README.md
```diff
@@ -1,22 +1,85 @@
...
+role `cluster-admin`.  Often user `kubeadmin` is predefined with this role.
+
+You can use [Code Ready Containers](https://developers.redhat.com/products/codeready-containers/download)
+to create a test cluster in a VM on your development machine.
+You may need to start your cluster with more memory than the default, for example:
+```
+crc start -m 12288
+```
+
+## Fast testing and debugging with `make run`
```
periklis, line 59: This whole approach works if either `KUBECONFIG` or `$HOME/.kube/config` are set. A local running operator on a dev machine needs a way to connect to the remote cluster. We should mention these steps.

README.md
```diff
@@ -39,77 +102,8 @@ under `$GOPATH/github.com/openshift/origin-aggregated-logging` you can use its `
...
-* Logged into an OKD cluster with cluster role `cluster-admin`
-* Clone of `openshift/elasticsearch-operator` repository in `$GOPATH`
-
-The deployment can be optionally modified using any of the following:
-
-| Env Var | Default | Description|
-|---------|---------|------------|
-|`EXCLUSIONS`|none|The list of manifest files that should will be ignored|
-|`PUSH_USER`|none|The name of the user e.g. `kubeadmin` used to push images to the remote registry|
-|`PUSH_PASSWORD`|none|The password for `PUSH_USER`|
```
periklis, line 72: These variables will be essential after merging #511. We need a section for them.

Makefile
```diff
@@ -31,12 +31,12 @@ build:
...
 	$(MAKE) build BUILD_OPTS='-gcflags=all="-N -l"'
 
-# Run the CLO locally. Requires a cluster login, sets up all other pre-requisites.
-# Can be run multiple times without any undeploy/uninstall steps.
+# Run the CLO locally - see HACKING.md
 RUN_CMD?=go run
 run:
-	$(MAKE) deploy-elasticsearch-operator
 	@oc get ns $(NAMESPACE) &> /dev/null || oc create ns $(NAMESPACE)
+	@oc delete -n $(NAMESPACE) all --all
```
periklis, line 38: IIRC `oc delete all` will not delete CRDs or instances of those. Same applies for ClusterRoles and ClusterRoleBindings.

README.md
alanconway: I thought that was the idea, but we still  have https://github.com/openshift/cluster-logging-operator/blob/master/hack/common#L185 - what's the equivalent podman command?

README.md
alanconway: I'll improve the README with this and your earlier comments

README.md
periklis: I believe this function is not used. `hack/deploy-image.sh` and `olm_deploy/scripts/catalog-build.sh` were replaced to use `podman push`.

syedriko: > I thought that was the idea, but we still have https://github.com/openshift/cluster-logging-operator/blob/master/hack/common#L185 - what's the equivalent podman command?

It's ```podman image (pull, tag, push)``` which makes ```scopeo copy``` so much more intuitive.

README.md
alanconway: Can you LGTM this and add the section for the PUSH variables in a later PR? I don't want to mis-state the story.

Makefile
alanconway: Correct - it's sufficient for re-starting a `make run` CLO but it won't clean up leftovers from aborted tests. I've found I have to clean up a whole list of service addresses and other things. See the `test-cleanup` target for a start at doing a proper house-cleaning. I haven't documented it yet because it doesn't really work and I'm not sure it's the best approach. The tests have code to do their own cleanup, but it appears not to always work if tests are interrupted. I think the tests themselves shoudl do a pre-clean to prepare their own preconditions rather than assuming that previous tests left things clean. Anyhow - all that will need to be the subject of another PR.

README.md
alanconway: done.

README.md
periklis: The only extra requirement that made skopeo previously mandatory was that of copy an image tag to a target registry and relabel this tag. To be precise, we used to build local tags like:
```
image-registry.openshift-image-registry.svc:5000/openshift/origin-cluster-logging-operator:latest
```
and by using skopeo pushing them as to the dev cluster registry as:
```
127.0.0.1:5000/openshift/origin-cluster-logging-operator:latest
```
However, since [PR-494](https://github.com/openshift/cluster-logging-operator/pull/494/files#diff-b67911656ef5d18c4ae36cb6741b7965R7) we use the latter tag for local tags.

In addition skopeo is not able to copy from a local oci storage used by a rootless podman.

HACKING.md
```diff
@@ -1,5 +1,127 @@
...
+```
+
+Note `make run` will not return until you terminate the CLO.
+
+## Running a full integration test with a 4.x cluster
+
+If you have a local clone of [origin-aggregated-logging](https://github.com/openshift/origin-aggregated-logging)
+under `$GOPATH/github.com/openshift/origin-aggregated-logging` you can use its `hack/get-cluster-run-tests.sh`
+to build the logging, elasticsearch-operator, and cluster-logging-operator images; deploy a 4.x cluster;
+push the images to the cluster; deploy logging; and launch the logging CI tests.
```
periklis, line 105: Are you using these commands to run the tests? I encounter them for the first time by reading this section

alanconway: No, that section is from before my changes. I was reluctant to remove it in case it's still useful, @vimalk78 I think you tried this path - is it worth keeping or better to drop since other methods are now documented?

HACKING.md
```diff
@@ -1,5 +1,127 @@
...
+Check out some other required repositories in your $GOPATH:
+```
+go get -d -u github.com/openshift/elasticsearch-operator/... github.com/openshift/origin-aggregated-logging/...
+```
+
+**Note:** `podman` copies our source tree very slowly. You can use `docker`
+instead, this will not exactly reproduce how images are built for release or by
+CI, but is close enough for many purposes. Put a symlink early in your path:
+```
+ln -s /usr/bin/podman /somewhere/early/in/my/PATH`
```
bparees, line 21: shouldn't this symlink be targeting the docker binary?  and include podman in the linkname?

HACKING.md
```diff
@@ -1,5 +1,127 @@
...
+```
+
+## Testing Overview
+
+There are two types of test:
+
+* *Unit tests:* run in seconds, no cluster needed, verify internal behavior - behind the API. New could *should* be reasonably well covered by unit tests but we don't have a formal requirement.
```
bparees, line 28:
```suggestion
* *Unit tests:* run in seconds, no cluster needed, verify internal behavior - behind the API. New code *should* be reasonably well covered by unit tests but we don't have a formal requirement.
```

HACKING.md
```diff
@@ -1,5 +1,127 @@
...
+```
+
+Log in as a user with the role `cluster-admin` - user `kubeadmin` is often predefined with this role.
+
+You can `export KUBECONFIG=/path/to/cluster/config` or copy/add your config to `$HOME/.kube/config`. Depending on how you set it up, you may also need to `oc login` to your cluster and add pull secrets to your credentials.
+
+## Running tests
+
+Unit tests can be run without a cluster:
+* `make`: Run unit tests, lint and format code. Do this before submitting a PR.
```
bparees, line 54: the default target for the makefile does "unit tests, link, and code formatting" (doesn't actually build anything?)  seems odd.

HACKING.md
```diff
@@ -1,5 +1,127 @@
...
+* `make test-unit`: Just run unit tests.
+
+To build, deploy and test the CLO image in your own cluster:
```
bparees, line 57: what if i just want to build it?

HACKING.md
```diff
@@ -1,5 +1,127 @@
...
+* `make deploy`: Build CLO image and deploy to cluster with elasticsearch.
```
bparees, line 58: does this do an OLM based install of the operator?

HACKING.md
```diff
@@ -1,5 +1,127 @@
...
+
+To run CLO as a local process:
+* `make run`: Run CLO as a local process. Does not require `make deploy`.
+* `make debug`: Run CLO in the `dlv` debugger.
+
+After running `make deploy`, `make run` or `make debug`, you can run e2e tests:
+* `make test-cluster` Run all e2e tests against the running CLO.
+* `go run ./test/e2e/something`: Run tests individually, in a debugger or whatever.
+
+To do a complete deploy, test, cleanup cycle exactly as CI does (this must  start from a clean cluster e.g. after `make undeploy`):
```
bparees, line 69:
```suggestion
To do a complete deploy, test, cleanup cycle exactly as CI does (this must start from a clean cluster e.g. after `make undeploy`):
```

HACKING.md
```diff
@@ -1,5 +1,127 @@
...
+* `make test-olm-e2e`: Run e2e tests the same way CI does
```
bparees, line 70: this requires some env vars to be set if you want it to actually run your code and not just the latest images on api.ci, right?

HACKING.md
```diff
@@ -1,5 +1,127 @@
...
+
+If you have a local clone of [origin-aggregated-logging](https://github.com/openshift/origin-aggregated-logging)
+under `$GOPATH/github.com/openshift/origin-aggregated-logging` you can use its `hack/get-cluster-run-tests.sh`
+to build the logging, elasticsearch-operator, and cluster-logging-operator images; deploy a 4.x cluster;
+push the images to the cluster; deploy logging; and launch the logging CI tests.
+
+## Building a Universal Base Image (UBI) based image
+
+You must first `oc login api.ci.openshift.org`.  You'll need these credentials in order
+to pull images from the UBI registry.
```
bparees, line 110: really you need to be logged into registry.svc.ci.openshift.org in your dockercfg.  which you can do via `oc registry login --to ~/.docker/config.json` once you are oc login'd to api.ci.

HACKING.md
```diff
@@ -1,5 +1,127 @@
...
+Note that the latter is private and requires special permission to access.
+
+The best approach is to clone these repos under `$GOPATH/src/github.com/openshift`
+which the build scripts will pick up automatically.  If you do not, the build script
+will attempt to clone them to a temporary directory.
 
 ## Updating Dependencies
-Bumping the release and manifest versions typically require udating the `elasticsearch-operator`
+
+Bumping the release and manifest versions typically require udating the `elasticsearch-operator` as well.
```
bparees, line 125:
```suggestion
Bumping the release and manifest versions typically require updating the `elasticsearch-operator` as well.
```

HACKING.md
alanconway: Yes, good catch.

HACKING.md
alanconway: `make image` is mentioned further down.

HACKING.md
alanconway: It's the fastest and most frequent developer action - I do this every time I stop typing. Running  unit tests implicitly builds all the code that has unit tests, and the tests themselves. Building the image takes longer and is only  useful when you're ready to do a full-on cluster test.

I agree it is traditional for the default make target to do "everything" - if you feel strongly I'm OK with turning this into `make check` and having `make` build the image.

HACKING.md
alanconway: Yes, @jcantrill rewrote the "deploy" and "undeploy" targets to use your targets, and your targets are still there if someone want's finer control.

HACKING.md
alanconway: Yes, I think this is documented in the Makefile itself, if not we should add it in a future PR.

HACKING.md
alanconway: This section is unchanged since before my time and I haven't tried it myself. If you want to correct it I'd appreciate that.

HACKING.md
bparees: i don't feel that strongly, but it did catch me off guard and may surprise new contributors.

alanconway: I agree, will change that.

HACKING.md
alanconway: I'm going to take that section out as I suspect it's no longer needed and I"m not sure if it works.
We can put it back later if I'm wrong.

alanconway:
@bparees @periklis  I think I've addressed your feedback, can I get a LGTM?

bparees:
overriding BZ req since this doesn't touch product code.

will let @openshift/openshift-team-logging  lgtm it.

README.md
ewolinetz: i closed out #511 -- i think we should be using the `oc login -u kubeadmin -p <password>` approach

PR #516, opened 2020-05-11
Log 656 syslog on ga api

vimalk78:
- Implemeted Syslog LogForwarding on LF GA API implementation
- added e2e test cases


vimalk78:
/hold

need https://github.com/openshift/origin-aggregated-logging/pull/1882 to merge before this gets merged

PR #521, opened 2020-05-12
Add test for certificate generation

periklis:
This PR extends the unit test coverage for the entire CLO certificate generation.

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -22,6 +22,7 @@ import (
 
 const (
 	DefaultWorkingDir = "/tmp/ocp-clo"
+	DefaultScriptsDir = "./scripts"
 	OsNodeLabel       = "kubernetes.io/os"
 	LinuxValue        = "linux"
 )
@@ -184,6 +185,14 @@ func GetShareDir() string {
 		return "/usr/share/logging"
 	}
 	return shareDir
+}
+
+func GetScriptsDir() string {
+	scriptsDir := os.Getenv("SCRIPTS_DIR")
+	if scriptsDir == "" {
+		return DefaultScriptsDir
+	}
+	return scriptsDir
 }
 
 func GetWorkingDirFileContents(filePath string) []byte {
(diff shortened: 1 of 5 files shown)


pkg/utils/utils.go
```diff
@@ -186,6 +187,14 @@ func GetShareDir() string {
 	return shareDir
 }
 
+func GetScriptsDir() string {
+	scriptsDir := os.Getenv("SCRIPTS_DIR")
```
ewolinetz, line 191: i don't think there's a need/want to let users override this value

ewolinetz:
holding to review -- the CLO never generated certificates for metrics, it was auto generated because if we create/sign it ourselves then it will not be trusted (using our own self signed certs)

ewolinetz:
per the bz - they are using the incorrect port:
```
Get https://10.128.2.51:60000/_prometheus/metrics: x509: certificate is valid for localhost, elasticsearch, elasticsearch.cluster.local, elasticsearch.openshift-logging.svc, elasticsearch.openshift-logging.svc.cluster.local, not elasticsearch-metrics.openshift-logging.svc
```

per EO:
```
			// HTTPs listener for metrics
			"--metrics-listening-address=:60001",
			"--metrics-tls-cert=/etc/proxy/secrets/tls.crt",
			"--metrics-tls-key=/etc/proxy/secrets/tls.key",
```

pkg/utils/utils.go
jcantrill: this is inline with our current pattern for workingdir... user's could only override if they go unmanaged, edit the deployment, kick the pod

ewolinetz: i still think its unnecessary to provide this, it was never configurable before and not sure what it actually provides...

periklis: The issue here is that method under test `CreateOrUpdateCertificates` relied on a hardcoded string. However, we need to be able to setup a proper dir for unit and e2e tests independent of container fs layout.

PR #523, opened 2020-05-12
Bug 1835268: validate metadata name in crd

blockloop:
Ensure that the metadata name is always `instance` using validation spec

### DIFF
--- a/manifests/4.5/cluster-loggings.crd.yaml
+++ b/manifests/4.5/cluster-loggings.crd.yaml
@@ -15,6 +15,14 @@ spec:
     openAPIV3Schema:
       type: object
       properties:
+        metadata:
+          description: Specification of the metadata for the cluster
+          properties:
+            name:
+              description: The name of the clusterlogging object
+              type: string
+              enum:
+              - "instance"
         spec:
           description: Specification of the desired behavior of the Logging cluster.
           type: object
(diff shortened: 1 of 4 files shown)


manifests/4.5/cluster-loggings.crd.yaml
```diff
@@ -19,6 +19,14 @@ spec:
           description: Specification of the desired behavior of the Logging cluster.
           type: object
           properties:
+            metadata:
+              description: Specification of the metadata for the cluster
+              properties:
+                name:
+                  description: The name of the clusterlogging object
+                  enum:
+                    - instance
```
ewolinetz, line 28: nit -- not sure if this will affect anything but we may want to un-indent this to match formatting (not sure if `go fmt` would catch this)

ewolinetz:
/lgtm
@jcantrill i'll leave this up to you whether we want to just have this wait for 4.6 or open a bz to get it in 4.5

jcantrill:
@ewolinetz aside from the obvious, is there an issue this resolves?  If so, open a bz.

ewolinetz:
I can think of something ... I'll open a bz to track this.

ewolinetz:
/hold
there seems to be an issue setting up the cluster via the installer

PR #525, opened 2020-05-13
Fix lint errors (ignored return values)

alanconway:
Also fixed mis-formatted files and updated generated CRDs.


bparees:
nothing in this PR ships, overriding BZ req

PR #527, opened 2020-05-14
Bug 1836451: Fix for removing Kibana CR

igor-karpukhin:
Fixed nil pointer error when removing Kibana CR

### DIFF
--- a/pkg/k8shandler/visualization.go
+++ b/pkg/k8shandler/visualization.go
@@ -130,6 +130,7 @@ func (clusterRequest *ClusterLoggingRequest) removeKibana() (err error) {
 	if clusterRequest.isManaged() {
 		name := "kibana"
 		proxyName := "kibana-proxy"
+
 		if err = clusterRequest.removeKibanaCR(); err != nil {
 			return
 		}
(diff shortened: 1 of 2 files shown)


pkg/k8shandler/visualization.go
```diff
@@ -389,7 +390,17 @@ func isKibanaCRDDifferent(current *es.Kibana, desired *es.Kibana) bool {
 }
 
 func (clusterRequest *ClusterLoggingRequest) removeKibanaCR() error {
-	cr := newKibanaCustomResource(clusterRequest.cluster, constants.KibanaName)
+	cr := &es.Kibana{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "Kibana",
+			APIVersion: es.SchemeGroupVersion.String(),
+		},
```
periklis, line 397: You don't need to set this up, because we use client-go for the request.

pkg/k8shandler/visualization.go
```diff
@@ -389,7 +390,17 @@ func isKibanaCRDDifferent(current *es.Kibana, desired *es.Kibana) bool {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      constants.KibanaName,
+			Namespace: clusterRequest.cluster.Namespace,
+		},
+	}
+	utils.AddOwnerRefToObject(cr, utils.AsOwner(clusterRequest.cluster))
```
periklis, line 403: The owner ref is not used on delete.

pkg/k8shandler/visualization.go
igor-karpukhin: Fixed

pkg/k8shandler/visualization.go
igor-karpukhin: Fixed

PR #529, opened 2020-05-14
WIP: LOG-772: Go tests clean up existing resources for themselves.

alanconway:
In test/helpers/framework.go, see foreCreate and forceCreateREST - these two functions do a generic "try to create, if resource exists delete and then create."

forceCreate uses reflection, forceCreateREST uses the rest client, there are type-safe CreateFoo functions for all the resource types used by test code.

WORK IN PROGRESS: The tests are still not well-behaved, but they are getting past
most of the "cannot create already exists" errors that I was seeing before.


test/helpers/framework.go
```diff
@@ -120,12 +121,89 @@ func (tc *E2ETestFramework) CreateTestNamespace() string {
...
+		result := createMethod.Call([]reflect.Value{reflect.ValueOf(config)})
+		return result[0], valueToError(result[1])
+	}
+	delete := func() error {
+		result := deleteMethod.Call([]reflect.Value{name, reflect.ValueOf(deleteOpts)})
+		return valueToError(result[0])
+	}
+	v, err := create()
+	if errors.IsAlreadyExists(err) {
+		if err = delete(); err == nil {
```
syedriko, line 155: Don't we need to wait for the delete() to take effect before we attempt the create()?

pkg/k8shandler/visualization.go
```diff
@@ -241,6 +241,9 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateKibanaSecret() error
 
 func newKibanaCustomResource(cluster *logging.ClusterLogging, kibanaName string) *es.Kibana {
 	visSpec := cluster.Spec.Visualization
+	if visSpec == nil {
+		visSpec = &logging.VisualizationSpec{}
+	}
```
periklis, line 246: This is fixed by #527

test/helpers/framework.go
```diff
@@ -120,12 +121,89 @@ func (tc *E2ETestFramework) CreateTestNamespace() string {
...
+	rclient := reflect.ValueOf(client)
+	createMethod := rclient.MethodByName("Create")
+	deleteMethod := rclient.MethodByName("Delete")
+	if !createMethod.IsValid() || !deleteMethod.IsValid() {
+		panic(fmt.Errorf("not a kube client interface: %#v", client))
+	}
+	name := reflect.ValueOf(config).Elem().FieldByName("Name")
+	if !name.IsValid() {
+		panic(fmt.Errorf("not a valid config object, no Name: %#v", config))
+	}
```
periklis, line 143: Why do we use `interface{}` and reflection. The k8s client-go/api-machinery offers `runtime.object` for this.

test/helpers/framework.go
```diff
@@ -230,43 +308,46 @@ func (tc *E2ETestFramework) SetupClusterLogging(componentTypes ...LogComponentTy
 }
 
 func (tc *E2ETestFramework) CreateClusterLogging(clusterlogging *cl.ClusterLogging) error {
-	body, err := json.Marshal(clusterlogging)
+	return tc.forceCreateREST(clusterlogging)
+}
+
+// forceCreateREST creates config using the REST client, deleting existing instance if necessary.
+// It adds the new object to the cleanup list.
+func (tc *E2ETestFramework) forceCreateREST(config interface{}) error {
```
periklis, line 316: TBH we don't need to use the REST API if the client registers the scheme for our CRD. This is easily done by:
```
import (
     loggingv1 "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
     "k8s.io/client-go/kubernetes/scheme"
)

_ = loggingv1.SchemeBuilder.AddToScheme(scheme.Scheme)

File: test/helpers/framework.go (hunk 1 above)
Comment (alanconway) [reply test/helpers/framework.go:155]: I was assuming that *Interface.Delete() is synchronous, if not then we do need to add something here. Needs to be checked. Another option is to use Update() instead of Delete()+Create() -  I'm unclear if that's equivalent - also needs to be checked.

File: pkg/k8shandler/visualization.go (hunk 2 above)
Comment (alanconway) [reply pkg/k8shandler/visualization.go:246]: Will merge across & rebase.

File: test/helpers/framework.go (hunk 3 above)
Comment (alanconway) [reply test/helpers/framework.go:143]: Was not aware, thanks! Not sure how it allows us to avoid reflection, can you provide code?
```
type Object interface {
    GetObjectKind() schema.ObjectKind
    DeepCopyObject() Object
}
```
That gives us a clean way to get Kind, but not Name. We also need a way to call the Create, Delete and Update methods on the client, which don't have the same signatures for different kinds of object.
I'm sure we can find a cleaner way to do this somewhere in the k8s libraries though.

test/helpers/framework.go
alanconway: Cool, I was following existing code that uses REST, but this looks more consistent. We do already add these to the scheme, we just weren't taking advantage.

alanconway:
PR is too far out of date, will bring in a new PR

PR #530, opened 2020-05-15
LOG-653: Generate all CRDs from Go source.

alanconway:
Generate all CRDs from Go source. Summary:

* Removed deploy/ directory - CRDs are generated into manifests/4.5
* clusterlogging_types.go: add comments and directives to produce an equivalent CRD
* forwarding_types.go: add comments and directives to produce an equivalent CRD

manifest/4.5/*.crd.yaml are now generated by `make generate` and checked in.
They should not have any semantic changes from the hand-written CRDs, but this
is hard to see with `diff` since the order of fields has changed.
I used https://github.com/homeport/dyff to verify the changes.

INTENTIONAL CHANGE: `proxySpec` -> `proxy`, this was an error in the original
hand-written CRD. The generated CRD matches the Go code.


alanconway:
Fixed by @periklis in another PR

PR #532, opened 2020-05-15
Rename fluentForward to fluentdForward.

alanconway:
Following the lead of the fluent project, their protocol spec:
https://github.com/fluent/fluentd/wiki/Forward-Protocol-Specification-v1 calls
it the "fluentd forward protocol" so we should too.


jcantrill:
Everything is green.  Manually merging because of https://issues.redhat.com/browse/DPTP-1209

PR #533, opened 2020-05-15
Use a real URL parser, only accept well-formed absolute URLs for output.url.

alanconway:
Use url.Parse() for all URL parsing, require that URLs be well formed and absolute

### DIFF
--- a/docs/output_url_field.md
+++ b/docs/output_url_field.md
@@ -1,0 +1,30 @@
+# Using the clusterlogforwarder.output.url field
+
+The `clusterlogforwarder.output.url` field requires a valid absolute URL.  An
+'absolute' URL is one with non-empty `scheme` and `host:port` parts, in other
+words it must start with "something://hostname". User and password parts
+(e.g. http://user:pass@hostname) are *not allowed*, credentials should be
+provided in the `output.secret` field.
+
+In some cases an output type may provide an alternative way to configure
+connections, e.g. the `output.kafka.brokers` field. In such cases the output.url
+can be omitted. Using the output.url field is preferred whenever possible, but
+some output features may not be available via URL (e.g. kafka allows multiple
+failover addresses via `output.kafka.brokers`).
+
+Some output types define a URL scheme, for example elasticsearch uses `http` and
+`https`. Other output types (e.g. syslog) don't define a "natural" URL
+scheme. For those output types we use these special schemes:
+
+* tcp: insecure TCP connection.
+* tls: secure TLS over TCP connection.
+* udp: insecure UDP packets.
+* udps: secure TLS over UDP packets.
+
+If the url scheme is a TLS secure scheme (https, tls, udps) then the
+`output.secret` MUST NOT be empty, it provides the TLS certificates. If the URL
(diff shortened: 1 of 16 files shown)


alanconway:
/hold 
fix test failures and review default/allowed URL schemes in code.

alanconway:
Replace hand-parsing of URLs with url.Parse(), enforce valid absolute URLs with scheme and host.

/uncc @lukas-vlcek 
/cc @periklis @igor-karpukhin

docs/output_url_field.md
```diff
@@ -0,0 +1,30 @@
+# Using the clusterlogforwarder.output.url field
+
+The `clusterlogforwarder.output.url` field requires a valid absolute URL.  An
+'absolute' URL is one with non-empty `scheme` and `host:port` parts, in other
+words it must start with "something://hostname". User and password parts
```
periklis, line 5: nit. rename `something` to `scheme`

pkg/url/url.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	u, err := url.Parse(s)
+	switch {
+	case err != nil:
+		return nil, err
+	case u.Scheme == "":
+		return nil, fmt.Errorf("no scheme: %v", u)
+	case u.Host == "":
+		return nil, fmt.Errorf("no host: %v", u)
+	}
```
periklis, line 24: What about `case u.Port != 0` ?

alanconway: Could do that but it's crossing from syntactic validation to semantic, which I wasn't intending here.
http://foo:0 is a valid URL even if not a useful one. Actually http://foo:blahblah is a valid URL, since symbolic port names are allowed and the URL spec doesn't define the list of legal values. The actual resolvable port names depends on the local OS, contents of /etc/services etc.

alanconway:
Rebased on Igor's validation changes, should be ready to go.

alanconway:
@jcantrill @periklis  can I get another LGTM here? I've fixed the e2e failures.

docs/output_url_field.md
```diff
@@ -0,0 +1,30 @@
...
+connections, e.g. the `output.kafka.brokers` field. In such cases the output.url
+can be omitted. Using the output.url field is preferred whenever possible, but
+some output features may not be available via URL (e.g. kafka allows multiple
+failover addresses via `output.kafka.brokers`).
+
+Some output types define a URL scheme, for example elasticsearch uses `http` and
+`https`. Other output types (e.g. syslog) don't define a "natural" URL
+scheme. For those output types we use these special schemes:
+
+* tcp: insecure TCP connection.
```
jcantrill, line 19: Do we really need or want to support insecure?

alanconway: I'm inclined to - developers who are testing forwarding to an external server will appreciate not having to screw around with setting up certificates in early development. Unlike our own elastic instance, we won't be setting that up for them. Also it seems likely that some users will rely on provider security and treat their cluster like a private network so again might not want to hassle with certs.

alanconway: We can review that it if you think it's against the opinionated policy. Secure is more likely in production, and almost nobody uses plain HTTP anymore.

PR #536, opened 2020-05-20
[WIP] Bug 1823305: Append logout url to oauth container if user provides one for idp

ewolinetz:
Part of fix to address https://bugzilla.redhat.com/show_bug.cgi?id=1823305

Depends on https://github.com/openshift/origin-aggregated-logging/pull/1898


ewolinetz:
is there a way we can bypass the need to have a cherry pick for this? this functionality no longer exists for 4.5 so there would be nothing to cherrypick...

PR #537, opened 2020-05-22
Update LF GA api for syslog rfc5424 parameters

vimalk78:
- updated template for new syslog parameters
 - added test cases

Fixed LogForwarding artifacts collection

### DIFF
--- a/manifests/4.6/logging.openshift.io_clusterlogforwarders_crd.yaml
+++ b/manifests/4.6/logging.openshift.io_clusterlogforwarders_crd.yaml
@@ -126,6 +126,10 @@ spec:
                       description: Syslog provides optional extra properties for output
                         type `syslog`
                       properties:
+                        appName:
+                          description: "AppName is APP-NAME part of the syslog-msg
+                            header \n AppName needs to be specified if using rfc5424"
+                          type: string
                         facility:
                           description: "Facility to set on outgoing syslog records.
                             \n Facility values are defined in https://tools.ietf.org/html/rfc5424#section-6.2.1.
@@ -136,9 +140,27 @@ spec:
                             cron authpriv ftp ntp security console solaris-cron     local0
                             local1 local2 local3 local4 local5 local6 local7"
                           type: string
+                        msgID:
+                          description: "MsgID is MSGID part of the syslog-msg header
+                            \n MsgID needs to be specified if using rfc5424"
+                          type: string
                         payloadKey:
                           description: PayloadKey specifies record field to use as
                             payload.
+                          type: string
+                        procID:
+                          description: "ProcID is PROCID part of the syslog-msg header
+                            \n ProcID needs to be specified if using rfc5424"
+                          type: string
+                        rfc:
+                          default: RFC5424
+                          description: "Rfc specifies the rfc to be used for sending
+                            syslog \n Rfc values can be one of:  - RFC3164 (https://tools.ietf.org/html/rfc3164)
+                            \ - RFC5424 (https://tools.ietf.org/html/rfc5424) \n If
+                            unspecified, RFC5424 will be assumed."
+                          enum:
+                          - RFC3164
+                          - RFC5424
                           type: string
                         severity:
                           description: "Severity to set on outgoing syslog records.
@@ -147,9 +169,8 @@ spec:
                             keywords: \n     Emergency Alert Critical Error Warning
                             Notice Informational Debug"
                           type: string
-                        tagKey:
-                          description: TagKey specifies a record field  to  use as
-                            tag.
+                        tag:
+                          description: Tag specifies a record field to use as tag.
                           type: string
                         trimPrefix:
                           description: TrimPrefix is a prefix to trim from the tag.
--- a/pkg/apis/logging/v1/output_types.go
+++ b/pkg/apis/logging/v1/output_types.go
@@ -55,15 +55,49 @@ type Syslog struct {
 	// +optional
 	TrimPrefix string `json:"trimPrefix,omitempty"`
 
-	// TagKey specifies a record field  to  use as tag.
+	// Tag specifies a record field to use as tag.
 	//
 	// +optional
-	TagKey string `json:"tagKey,omitempty"`
+	Tag string `json:"tag,omitempty"`
 
 	// PayloadKey specifies record field to use as payload.
 	//
 	// +optional
 	PayloadKey string `json:"payloadKey,omitempty"`
+
+	// Rfc specifies the rfc to be used for sending syslog
+	//
+	// Rfc values can be one of:
+	//  - RFC3164 (https://tools.ietf.org/html/rfc3164)
+	//  - RFC5424 (https://tools.ietf.org/html/rfc5424)
+	//
+	// If unspecified, RFC5424 will be assumed.
+	//
+	// +kubebuilder:validation:Enum:=RFC3164;RFC5424
+	// +kubebuilder:default:=RFC5424
+	// +optional
+	RFC string `json:"rfc,omitempty"`
+
+	// AppName is APP-NAME part of the syslog-msg header
+	//
+	// AppName needs to be specified if using rfc5424
+	//
+	// +optional
+	AppName string `json:"appName,omitempty"`
+
+	// ProcID is PROCID part of the syslog-msg header
+	//
+	// ProcID needs to be specified if using rfc5424
+	//
+	// +optional
+	ProcID string `json:"procID,omitempty"`
+
+	// MsgID is MSGID part of the syslog-msg header
+	//
+	// MsgID needs to be specified if using rfc5424
+	//
+	// +optional
+	MsgID string `json:"msgID,omitempty"`
 }
 
 // Kafka provides optional extra properties for `type: kafka`
--- a/pkg/generators/forwarding/fluentd/output_conf_syslog_test.go
+++ b/pkg/generators/forwarding/fluentd/output_conf_syslog_test.go
@@ -22,6 +22,12 @@ var _ = Describe("Generating external syslog server output store config blocks",
 		})
 
 		tcpConf := `<label @SYSLOG_RECEIVER>
+		<filter **>
+		  @type parse_json_field
+		  json_fields  message
+		  merge_json_log false
+		  replace_json_log true
+		</filter>
 		<match **>
 		@type copy
 		<store>
@@ -37,6 +43,12 @@ var _ = Describe("Generating external syslog server output store config blocks",
 	</label>`
 
 		udpConf := `<label @SYSLOG_RECEIVER>
+		<filter **>
+		  @type parse_json_field
+		  json_fields  message
+		  merge_json_log false
+		  replace_json_log true
+		</filter>
 		<match **>
 		@type copy
 		<store>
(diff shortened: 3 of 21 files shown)


manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -112,18 +116,35 @@ spec:
...
                         description: PayloadKey specifies record field to use as payload.
                         type: string
+                      procID:
+                        description: "ProcID is PROCID part of the syslog-msg header
+                          \n ProcID needs to be specified if using rfc5424"
+                        type: string
+                      rfc:
+                        description: "Rfc specifies the rfc to be used for sending
+                          syslog \n Rfc values can be one of:  - RFC3164 (https://tools.ietf.org/html/rfc3164)
+                          \ - RFC5424 (https://tools.ietf.org/html/rfc5424) \n If
```
alanconway, line 133: Put an explicit default with `// +kubebuilder:validation:Default` - check the exact syntax, I may not have it right.

pkg/apis/logging/v1/output_types.go
```diff
@@ -124,15 +124,48 @@ type Syslog struct {
...
+
+	// Rfc specifies the rfc to be used for sending syslog
+	//
+	// Rfc values can be one of:
+	//  - RFC3164 (https://tools.ietf.org/html/rfc3164)
+	//  - RFC5424 (https://tools.ietf.org/html/rfc5424)
+	//
+	// If unspecified, RFC5424 will be assumed.
+	//
+	// +kubebuilder:validation:Enum:=RFC3164;RFC5424
```
alanconway, line 145: add `// +kubebuilder:validation:Default=...` here to make the default clear.

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -112,18 +116,35 @@ spec:
                           cron authpriv ftp ntp security console solaris-cron     local0
                           local1 local2 local3 local4 local5 local6 local7"
                         type: string
+                      msgID:
+                        description: "MsgID is MSGID part of the syslog-msg header
+                          \n MsgID needs to be specified if using rfc5424"
+                        type: string
                       payloadKey:
```
alanconway, line 123: Should be `payload` to be consistent with `tag`. Need to document the $key convention.

pkg/generators/forwarding/fluentd/syslog_conf.go
```diff
@@ -1,14 +1,199 @@
...
 
+import (
+	"fmt"
+	"regexp"
+	"strings"
+)
+
+// The Syslog output fields can be set to an expression of the form $.abc.xyz
+// If an expression is used, its value will be taken from corresponding key in the record
+var keyre = regexp.MustCompile(`^\$(\.[[:word:]]*)+$`)
```
alanconway, line 11: I like $ as the key character - safe and widely used as a "variable expansion" indicator. 
I was thrown at first by the '.' after the $, but I guess the leading '.' is common convention for field paths in JSON paths etc. Perhaps it could be optional, so $abc.xyz means the same as $.abc.xyz? E.g if it's a top-level field `foo` then  $foo reads a bit more naturally than $.foo - it's subjective, I leave it to your discretion.

pkg/generators/forwarding/fluentd/syslog_conf.go
```diff
@@ -1,14 +1,199 @@
...
+
+var tagre = regexp.MustCompile(`\${tag\[-??\d+\]}`)
```
alanconway, line 13: Why is this different from payloadRe? I thought they were doing the same thing.

pkg/generators/forwarding/fluentd/syslog_conf.go
```diff
@@ -1,14 +1,199 @@
...
+	}
+	return severity
+}
+
 func (conf *outputLabelConf) Rfc() string {
-	//return "rfc3164"
-	// TODO update it whem merged with new v1 api
-	return "rfc5424"
+	switch conf.Target.Syslog.RFC {
+	case "RFC3164":
```
alanconway, line 46: Make this case-insensitive.

pkg/generators/forwarding/fluentd/syslog_conf.go
vimalk78: yes, this is jsonpath syntax. fluentd and plugins already support that.

Avoiding initial `.` seems appropriate we can discuss what syntax we expose, and then convert it internally to a jsonpah syntax. leaving it as jsonpath syntax could also be beneficial.

alanconway: Yes, the more I think of it, just ignore my comment - use strict jsonpath and document as such. It won't surprise anyone who has used jsonpath filters with oc or kubectl, and for those who are surprised there's a well known standard syntax they can refer to.

pkg/generators/forwarding/fluentd/syslog_conf.go
vimalk78: `keyre`  takes values from record by jsonpath, and `tagre` takes values from fluentd tag

pkg/generators/forwarding/fluentd/syslog_conf.go
vimalk78: done

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
vimalk78: added

pkg/apis/logging/v1/output_types.go
vimalk78: added

manifests/4.5/logging.openshift.io_clusterlogforwarders_crd.yaml
vimalk78: `payload` can only refer to a path in record, where as `tag` or `appname` can refer to path in record or fluentd tag

pkg/generators/forwarding/fluentd/syslog_conf.go
```diff
@@ -1,14 +1,199 @@
...
+	if conf.isFacilityKeyExpr() {
+		return fmt.Sprintf("${%s}", facility)
+	}
+	return conf.Target.Syslog.Facility
+}
+
+func (conf *outputLabelConf) Severity() string {
+	severity := strings.ToLower(conf.Target.Syslog.Severity)
+	if severity == "" {
+		return "debug"
```
periklis, line 36: Is this default a general default for syslog or a tradeoff for our API?

pkg/k8shandler/fluentd.go
```diff
@@ -276,7 +276,7 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, elasticsearchAppName str
 			},
 		}
 	}
-	fluentdContainer := NewContainer("fluentd", "fluentd", v1.PullIfNotPresent, *resources)
+	fluentdContainer := NewContainer("fluentd", "fluentd", v1.PullAlways, *resources)
```
periklis, line 279: Keeping `v1.PullIfNotPresent` is the correct value? Probably a miss from testing this PR.

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
```diff
@@ -33,90 +35,135 @@ var _ = PDescribe("LogForwarder", func() {
 		e2e              = helpers.NewE2ETestFramework()
 		testDir          string
 		forwarder        *logging.ClusterLogForwarder
+		generatorPayload map[string]string
+		waitlogs         string
+		grepappname      string
+		grepprocid       string
+		grepmsgid        string
+		greptag          string
```
periklis, line 43: nit. golang style is camel case instead of all small caps

test/helpers/framework.go
```diff
@@ -115,6 +114,64 @@ func (tc *E2ETestFramework) DeployLogGenerator() error {
 	return tc.waitForDeployment(namespace, "log-generator", defaultRetryInterval, defaultTimeout)
 }
 
+func (tc *E2ETestFramework) DeployJsonLogGenerator(vals map[string]string) (string, string, error) {
+	namespace := tc.CreateTestNamespace()
+	pycode := `
```
periklis, line 119: nit. it would be nice to have this python code on a separate file, e.g. `test/files/jsonlogger.py` to enable better editing on editors for python.

test/helpers/framework.go
```diff
@@ -139,13 +196,38 @@ func (tc *E2ETestFramework) WaitFor(component LogComponentType) error {
...
 	}
 	return fmt.Errorf("Unable to waitfor unrecognized component: %v", component)
 }
 
+func (tc *E2ETestFramework) waitForFluentDaemonSet(retryInterval, timeout time.Duration) error {
+	// daemonset should have non-zero number of instances for maxtimes consecutive retryInterval to detect a CrashLoopBackOff pod
+	maxtimes := 5
+	times := 0
+	return wait.Poll(retryInterval, timeout, func() (bool, error) {
+		numReady, err := oc.Literal().From("oc -n openshift-logging get daemonset/fluentd -o jsonpath={.status.numberReady}").Run()
```
periklis, line 211: Consider using the kubeclient here as other wait functions in our test framework code. Re-using one pattern is easier to maintain, e.g. `tc.KubeClient.AppsV1().Daemonset(..)`

test/helpers/oc/runner.go
```diff
@@ -134,3 +138,14 @@ func (r *runner) setArgs(args []string) {
 func (r *runner) setArgsStr(argstr string) {
 	r.args = sanitizeArgs(argstr)
 }
+
+func truncateString(str string, num int) string {
+	bnoden := str
```
periklis, line 143: bnoden? What is the meaning of this variable?

pkg/k8shandler/fluentd.go
vimalk78: yes, a leftover from the testing.

test/helpers/framework.go
vimalk78: for querying daemonset there is no difference between oc and kubectl. Also the literal command begins with `oc` because it is an `ochelper`

periklis: Sure there is no difference in functionality terms, but we have multiple wait methods written using the kubeclient. Would be nice to keep the pattern here.

vimalk78: the expression is querying the daemonset, and is written using `oc.helper` , a kubectl will look rather odd here

test/helpers/oc/runner.go
vimalk78: had copied it from go-playground, dont know what it means. will change.

test/helpers/framework.go
vimalk78: i think it is easier to see the generator code in-place rather than in a different file. the function accepts arguments and uses them to generate the logic. in a different file, there is a disconnect.

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
vimalk78: golang style may be camel case, but these variables are actually bash expressions. these expressions are used with `oc exec` , thats why these are all smallcaps. If these variables were keeping the program state etc, i will use automatically use camel case. but these are different than those.

pkg/k8shandler/fluentd.go
vimalk78: but it is a pain to change it everytime you test with changes in fluent image.
k8s will always pull the image if we dont specify the pull policy, and the image tag is `latest`.
In production, we will use sha based tags, so image will be pulled only once if we leave it unspecified.
So perhaps we should consider removing the pull policy.

pkg/generators/forwarding/fluentd/syslog_conf.go
vimalk78: this is for our API. there is no default in syslog spec.

manifests/4.6/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -126,6 +126,10 @@ spec:
                       description: Syslog provides optional extra properties for output
                         type `syslog`
                       properties:
+                        appName:
```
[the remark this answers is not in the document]
periklis: Nope the CRD generation case target only **ClusterLogging**. The ClusterLogForwarder is already generated from golang source code.

pkg/generators/forwarding/fluentd/output_conf_syslog_test.go
```diff
@@ -22,6 +22,12 @@ var _ = Describe("Generating external syslog server output store config blocks",
 		})
 
 		tcpConf := `<label @SYSLOG_RECEIVER>
+		<filter **>
+		  @type parse_json_field
```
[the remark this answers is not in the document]
vimalk78: if the record is not json, the record is unchanged
https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/lib/filter_parse_json_field/lib/filter_parse_json_field.rb#L84-L91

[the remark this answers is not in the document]
vimalk78: if the record is not json, or doesnt contain the asked jsonpath expression, the message is pushed to syslog. but `appname`, `procid`, `msgid` goes as the expressions itself. 
This is as per fluentd behavior. If the expression does not evaluate to a value, the expression string is returned.

example: message is not json string
(log from infra container)
```
<15>1 2020-07-12T22:45:46.779841+00:00 fluentd-wqqb6 ${$.message.appname_key} ${$.message.procid_key} ${$.message.msgid_key} - docker:{"container_id"=>"671d64acddb836667c0736c9d0a260dc692a0a34c213a8c557013b5d5c3e2436"}#011kubernetes:{"container_name"=>"olm-operator", "namespace_name"=>"openshift-operator-lifecycle-manager", "pod_name"=>"olm-operator-775d5486df-glk5q", "pod_id"=>"f0331fe2-65c8-4070-a776-4dbde541363a", "host"=>"crc-m4tnv-master-0", "master_url"=>"https://kubernetes.default.svc", "namespace_id"=>"8f0f6778-edd0-4b12-ad08-a54000d394eb", "namespace_labels"=>{"openshift_io/cluster-monitoring"=>"true", "openshift_io/scc"=>"anyuid"}, :flat_labels=>["app=olm-operator", "pod-template-hash=775d5486df"]}#011message:time="2020-07-12T22:44:42Z" level=info msg="checking packageserver"#011level:unknown#011hostname:crc-m4tnv-master-0#011pipeline_metadata:{"collector"=>{"ipaddr4"=>"192.168.126.11", "inputname"=>"fluent-plugin-systemd", "name"=>"fluentd", "received_at"=>"2020-07-12T22:44:43.364387+00:00", "version"=>"1.9.2 1.6.0"}}#011@timestamp:2020-07-12T22:44:42.531303+00:00#011viaq_index_name:infra-write#011viaq_msg_id:YmEzMTYwMjUtNWJhOS00NDg0LThkMDUtMjVlOWQxMjMzMTNh
```

example: message is json string, and returns a value for the jsonpath expression
(log from testcase controlled python container)
```
<15>1 2020-07-12T22:46:04.347499+00:00 fluentd-wqqb6 rec_appname rec_procid rec_msgid - docker:{"container_id"=>"394e1f40ac1ec559ece10a6866f8e9e990a7050454e314d6e68368563da57678"}#011kubernetes:{"container_name"=>"log-generator", "namespace_name"=>"clo-test-16824", "pod_name"=>"log-generator-6bcd987b67-cnfp2", "container_image"=>"docker.io/library/centos:centos7", "container_image_id"=>"docker.io/library/centos@sha256:c2f1d5a9c0a81350fa0ad7e1eee99e379d75fe53823d44b5469eb2eb6092c941", "pod_id"=>"57fb0f6f-891e-4131-81bd-de3406f2f59a", "host"=>"crc-m4tnv-master-0", "master_url"=>"https://kubernetes.default.svc", "namespace_id"=>"7dd801cd-a1ae-4013-a6ac-f2f3f116e2b9", :flat_labels=>["component=test", "logging-infra=log-generator", "pod-template-hash=6bcd987b67", "provider=openshift"]}#011message:{"index"=>36, "appname_key"=>"rec_appname", "msgcontent"=>"My life is my message", "timestamp"=>"2020-07-12 22:46:03", "msgid_key"=>"rec_msgid", "procid_key"=>"rec_procid"}#011level:unknown#011hostname:crc-m4tnv-master-0#011pipeline_metadata:{"collector"=>{"ipaddr4"=>"192.168.126.11", "inputname"=>"fluent-plugin-systemd", "name"=>"fluentd", "received_at"=>"2020-07-12T22:46:04.067336+00:00", "version"=>"1.9.2 1.6.0"}}#011@timestamp:2020-07-12T22:46:03.526355+00:00#011viaq_index_name:app-write#011viaq_msg_id:N2Q5YmVjNGUtZDk3MS00OWI4LTg4NTMtNzAxNWY5YTYwMzg2
```

PR #547, opened 2020-06-01
Speedup build by COPYing less

vimalk78:
buildah takes more time to execute COPY if there is a .dockerignore file
 - add individual COPY statements for copying individual files and folders required for `make build`
 - delete .dockerignore

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,16 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-logging-operator
-COPY . .
+
+# COPY steps are in the reverse order of frequency of change
+COPY cmd ./cmd
+COPY version ./version
+COPY scripts ./scripts
+COPY files ./files
+COPY vendor ./vendor
+COPY manifests ./manifests
+COPY Makefile ./Makefile
+COPY pkg ./pkg
+
 RUN make build
 
 FROM centos:centos7
@@ -13,9 +23,9 @@ RUN INSTALL_PKGS=" \
     mkdir /tmp/ocp-clo && \
     chmod og+w /tmp/ocp-clo
 COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/bin/cluster-logging-operator /usr/bin/
-COPY scripts/* /usr/bin/scripts/
+COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/scripts/* /usr/bin/scripts/
 RUN mkdir -p /usr/share/logging/
-COPY files/ /usr/share/logging/
+COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/files/ /usr/share/logging/
 
 COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/manifests /manifests
 RUN rm /manifests/art.yaml
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,3 +1,0 @@
-.git/*
-_output/*
-tmp/*


Dockerfile
```diff
@@ -1,6 +1,14 @@
...
-COPY . .
+
+COPY cmd ./cmd
+COPY manifests ./manifests
+COPY pkg ./pkg
+COPY vendor ./vendor
+COPY version ./version
+COPY Dockerfile ./Dockerfile
+COPY Makefile ./Makefile
+
```
periklis, line 11: Ordering and grouping was also my proposal back then when `Dockerfile.fast` was introduced:
```
#
# Order of COPY statements based on frequency of change.
# This enables better caching of each layer.
#

# Configuration
COPY config config
COPY files files

# manifests
COPY manifests manifests
COPY olm_deploy olm_deploy

# vendored dependencies
COPY Gopkg.lock Gopkg.lock
COPY Gopkg.toml Gopkg.toml
COPY vendor vendor

# source code
COPY cmd cmd
COPY pkg pkg
COPY scripts scripts
COPY version version

# make targets
COPY Makefile Makefile
RUN make build
```

Dockerfile
```diff
@@ -1,6 +1,14 @@
 FROM registry.svc.ci.openshift.org/openshift/release:golang-1.12 AS builder
 WORKDIR /go/src/github.com/openshift/cluster-logging-operator
-COPY . .
+
+COPY cmd ./cmd
+COPY manifests ./manifests
+COPY pkg ./pkg
+COPY vendor ./vendor
+COPY version ./version
+COPY Dockerfile ./Dockerfile
```
jcantrill, line 9: No need to copy itself right?

Dockerfile
jcantrill: Looks like @periklis has identified additional directories to copy

Dockerfile
periklis: We don't need this

Dockerfile
periklis: You can take my list if you like. I am working with this based on git stash for weeks now. The order of the groups rely on the frequency of change and dependency order (e.g. a manifests change means rebuild most of the code because CRD go and yaml files changed)

Dockerfile
vimalk78: yes, can remove this

vimalk78:
> Looks like @periklis has identified additional directories to copy
so we need all the directories for building the executable? i only copied which are needed for `RUN make build` step
Additional directories are being copied in later steps

Dockerfile
vimalk78: we dont need all the directories to be copied. The list i have is sufficient for building the executable at the `RUN make build` step 
other directories are getting copied later in the Dockerfile.

vimalk78: should the order matter? as of now i see that it maters. As soon as there is a step which has a change, buildah stops using cache for steps which appear later, even though it is unrelated.

periklis: The list is missing the `scripts`. Because `make build` works it doesn't tell that the operator works. We need the `scripts` dir for the `cert_generation.sh`.

periklis: In addition the order provide in the PR is akward. If `pkg` changes the big `vendor` will be rebuild again.

periklis: Furthermore how can your Dockerfile work without the dirs `files` and `manifests`. They are used in the next build stage.

vimalk78: `scripts` are being copied later than the `make build` step
updated the order
`files` are being copied later than the `make build` step. `manifests` copying is there

periklis: :hear_no_evil: by mistake, you made the copies directly in the runner build stage. IMHO this is not bad, but also not the best approach. One reason why the second stage was using `--from` in COPY statements is to make the build context only attached to the first stage. So that subsequent build stages only relate to artifacts on the cached layers of the previous ones.

vimalk78: for `manifests` and outputs of `make build`, it makes sense to use `--from` COPY statements, as they are produced/changed in first stage.
for copying from filesystem, is it not ok to use plain COPY statements?

Dockerfile
```diff
@@ -1,6 +1,16 @@
...
+
+# COPY steps are in the reverse order of frequency of change
+COPY cmd ./cmd
+COPY version ./version
+COPY scripts ./scripts
+COPY files ./files
+COPY manifests ./manifests
+COPY vendor ./vendor
+COPY Makefile ./Makefile
+COPY pkg ./pkg
```
periklis, line 12: Can we regroup these according to dependency and frequency order please? For example cmd, version, scripts and files should not affect vendor. You check my sample.

vimalk78: we can either have relative order based on frequency of change OR based on dependency, not both. I have ordered in based on decreasing order of frequency of change.

My intention behind this PR is to replace `COPY . . ` with more reasonable set of `COPY`, without impacting correctness, nothing else.  Relative order of COPY affects usage of cache, affects efficiency not correctness as far as i understand. Thats why this order of `COPY` statements. `cmd, version, scripts, files` is less likely to change, so these are at top. `pkg` is most like to change, so it is at bottom. `vendor` is more likely to change than `cmd, version, scripts, files` so somewhere in middle.

periklis: Sorry for the misunderstanding, I fully understand your intention and would also see this `COPY . .` vanish sooner than later. 

My statement is deps order first so that each group is depending on each other for correctness and then within each group frequency order. IMHO if something in vendor changes everything in `cmd/version/pkg` should be copied anew. For example, a bump of the operator-sdk in vendor affects immediately `cmd/manager/main.go`.

vimalk78: if `vendor` changes, and there is a corresponding change in `cmd/pkg/version` wouldnt that be detected by buildah? and if there is no change, it need not be copied again.
As far as i can see, there is no impact on correctness w.r.t. the order of COPY. correctness is impacted if we miss to copy something totally.
If it is so sensitive on order, i would suggest we go back to `COPY . . ` then we have taken care of all conditions.

PR #551, opened 2020-06-03
Bug 1844445: Fixed CLO crash when reconciling Fluentd

igor-karpukhin:
When CLO's CR created only with the `collection` section and then removed after some time, CLO will crash trying to reconcile Fluentd. This patch fixes that.
The root cause is that when CLO is triggered to restart fluentd, the `collection` section may not be configured in the CLO's CR.

### DIFF
--- a/pkg/k8shandler/forwarding.go
+++ b/pkg/k8shandler/forwarding.go
@@ -41,6 +41,11 @@ func isForwardingEnabled(cluster *logging.ClusterLogging) bool {
 
 func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig() (config string, err error) {
 
+	if clusterRequest.cluster == nil || clusterRequest.cluster.Spec.Collection == nil {
+		logger.Warnf("skipping collection config generation as 'collection' section is not specified in the CLO's CR")
+		return "", nil
+	}
+
 	switch clusterRequest.cluster.Spec.Collection.Logs.Type {
 	case logging.LogCollectionTypeFluentd:
 		break
--- a/pkg/k8shandler/forwarding_test.go
+++ b/pkg/k8shandler/forwarding_test.go
@@ -2,6 +2,8 @@ package k8shandler
 
 import (
 	"fmt"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -10,9 +12,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	cl "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
-	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
+	loggingv1 "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	loggingv1alpha1 "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
 	"github.com/openshift/cluster-logging-operator/pkg/logger"
 	helpers "github.com/openshift/cluster-logging-operator/test"
+	"k8s.io/client-go/kubernetes/scheme"
 )
 
 const (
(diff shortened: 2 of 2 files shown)


pkg/k8shandler/forwarding.go
```diff
@@ -43,6 +43,11 @@ func isForwardingEnabled(cluster *logging.ClusterLogging) bool {
 
 func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig() (config string, err error) {
 
+	if clusterRequest.cluster == nil || clusterRequest.cluster.Spec.Collection == nil {
+		logger.Warnf("unable to restart fluentd because of cluster collection is not specified")
```
jcantrill, line 47: I would just make this debug... and it may be more appropriate to say something like "Collection is unspecified. Skipping collector config generation."

jcantrill: Please see if you can add a test for this as well

vimalk78: we should add `clusterRequest.cluster == nil` check in all `ClusterLoggingRequest` methods + method specific nil checks

igor-karpukhin: @vimalk78 I think the actual code has to be refactored to provide a single entry point where the CLO creates `clusterRequest`, like it's done for `kibana` in the CLO.

pkg/k8shandler/forwarding_test.go
```diff
@@ -10,9 +12,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	cl "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
-	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
+	loggingv1 "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	loggingv1Alpha "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1alpha1"
```
periklis, line 16: small caps and complete version: `loggingv1alpha1`

pkg/k8shandler/forwarding_test.go
```diff
@@ -483,3 +487,94 @@ var _ = Describe("Normalizing Forwarding", func() {
 		})
 	})
 })
+
+func TestClusterLoggingRequest_generateCollectorConfig(t *testing.T) {
```
periklis, line 491: Why not having this test also in Gingko/Gomega style?

pkg/k8shandler/forwarding_test.go
igor-karpukhin: Fixed

pkg/k8shandler/forwarding_test.go
igor-karpukhin: It's a simple unit test, why not to keep it as a simple test?

periklis: In general we migrate tests to Gingko/Gomega where possible.

jcantrill: In general, we should be moving towards Ginko as they are more expressive IMO as their intent.  Writing them should not be more difficult; it truly is just a different pattern over the existing test library.

pkg/k8shandler/forwarding.go
```diff
@@ -41,6 +41,11 @@ func isForwardingEnabled(cluster *logging.ClusterLogging) bool {
 
 func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig() (config string, err error) {
 
+	if clusterRequest.cluster == nil || clusterRequest.cluster.Spec.Collection == nil {
+		logger.Warnf("skipping collection config generation as 'collection' section is not specified in the CLO's CR")
```
ewolinetz, line 45: is this really a "warn" ? i think this is a debug statement...

igor-karpukhin: The second thing, I guess. The condition sounds good, makes sense to add a tech debt ticket @ewolinetz  ?

pkg/k8shandler/forwarding_test.go
igor-karpukhin: @jcantrill @periklis  Ok, but what's the advantage this Ginko gives to us over the standard table-driven testing?

PR #552, opened 2020-06-03
[WIP] Bug 1842732: Declare required CRDs from elasticsearch-operator

periklis:
To address: https://bugzilla.redhat.com/show_bug.cgi?id=1842732

### DIFF
--- a/manifests/4.6/cluster-logging.v4.6.0.clusterserviceversion.yaml
+++ b/manifests/4.6/cluster-logging.v4.6.0.clusterserviceversion.yaml
@@ -322,6 +322,17 @@ spec:
                   - name: PROMTAIL_IMAGE
                     value: "quay.io/openshift/origin-promtail:latest"
   customresourcedefinitions:
+    required:
+    - name: elasticsearches.logging.openshift.io
+      version: v1
+      kind: Elasticsearch
+      displayName: Elasticsearch
+      description: Represents an Elasticsearch instance
+    - name: kibanas.logging.openshift.io
+      version: v1
+      kind: Kibana
+      displayName: Kibana
+      description: Represents a Kibana instance
     owned:
     - name: clusterloggings.logging.openshift.io
       version: v1
@@ -348,6 +359,8 @@ spec:
       - kind: Route
         version: v1
       - kind: Elasticsearch
+        version: v1
+      - kind: Kibana
         version: v1
       - kind: LogForwarding
         version: v1alpha1


manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml
```diff
@@ -322,6 +322,17 @@ spec:
...
+    - name: elasticsearches.logging.openshift.io
+      version: v1
+      kind: Elasticsearch
+      displayName: Elasticsearch
+      description: Represents an Elasticsearch instance
+    - name: kibanas.logging.openshift.io
+      version: v1
+      kind: Kibana
+      displayName: Kibana
+      description: Represents a Kibana instance
```
ewolinetz, line 335: does this mean that CLO will not start up without these CRDs that EO installs?
does this break/complicate user cases where they are just using CLO for log forwarding and will not install EO?

njhale: yes, this creates a prerequisite on these APIs being present.

periklis: @ewolinetz Fact is we have a hard dependency on elasticsearch-operator owned CRDs. Even if users use only LF they may decide to use the in-cluster store later. In this case we would need to teach them to install the elasticsearch-operator and kill the cluster-logging-operator pod to register the Elasticsearch/Kibana CRDs. (putting on cc @jcantrill @ik as dicussed this on a call yesterday).

Nevertheless, I hoped that this change makes the missing CRDs visible on the subscription or clusterserviceversion requirements on the web console. However, I can't make it work (See https://coreos.slack.com/archives/C3VS0LV41/p1591191245202700)

ewolinetz: i thought we have docs that state that if you are to use "elasticsearch" or "kibana" you need to install EO... to me that makes more sense than putting a hard requirement in place.

We shouldn't need to kill the CLO pod for it to know how to do anything with the kibana or elasticsearch CRs... the only requirement is k8s knows about them before we try to do anything with them. So the docs would read that EO must be installed and then afterwards they can update their ClusterLogging instance.

periklis: My understanding is that the CRD are a hard dependency because we register them once in the lifetime in `cmd/manager/main.go`. Can you please confirm/deny to add to my understanding it the following call is not requiring CRDs present on the APIserver:
```
	if err := elasticsearch.AddToScheme(mgr.GetScheme()); err != nil {
		log.Error(err, "")
		os.Exit(1)
	}
```

alanconway: On reflection I think @periklis  is right in principle. Elastic is our current `default` store,  and we need a  forwarder with `pipeline: { output: default }` to work reliably. 
In practice @ewolinetz  if this might break existing customer use, then we need the forwarder mark `default` outputs with a degraded or ready:false condition and a helpful message. As it stands it will just sit in a retry loop failing to connect to a URL that the user knows nothing about.

ewolinetz:
/hold
just to continue discussion for now (since it is approved and has lgtm)

manifests/4.5/cluster-logging.v4.5.0.clusterserviceversion.yaml
ewolinetz: @periklis i thought that is simply registering the api type for usage, i didn't think it had anything to do with CRDs themselves. in my mind the CRD is so k8s knows that it is an object type to use and what sort of validations it requires.

ewolinetz: a simple test would be to set up CLO without EO and just have a logforwarding/fluentd installation...

periklis: @ewolinetz the logforwarding/fluend in an installation without EO should work by today. But in my understanding if you install the ES/Kibana CRD later the registration I point above should be broken.

ewolinetz: it _shouldn't_ so long as the CRD is out there before CLO tries to resolve the kibana or elasticsearch CRs it should be fine. even if it does fail, it should requeue and reconcile again (at which point, if the CRD is installed it should go through). this is based on what i've seen in the past

syedriko: > @ewolinetz My understanding is that the CRD are a hard dependency because we register them once in the lifetime in `cmd/manager/main.go`. Can you please confirm/deny to add to my understanding it the following call is not requiring CRDs present on the APIserver:
> 
> ```
> if err := elasticsearch.AddToScheme(mgr.GetScheme()); err != nil {
>     log.Error(err, "")
>     os.Exit(1)
> }
> ```

@periklis Where it the case that Elastic CRDs where requried for this code to work, it would've been impossible to start CLO without Elastic CRDs present but in the BZ CLO runs.

syedriko: I agree with @ewolinetz on how this should behave. The error in the BZ needs to be root-caused and then we can see how it can be addressed.

periklis:
Based on the discussion in [1], using the required field for CRDs in our CSV is not resulting in a nice status message that a CRD is missing. Thus, sticking with the workaround per docs for now, I close this PR.
[1] https://coreos.slack.com/archives/C3VS0LV41/p1593087231269300

PR #553, opened 2020-06-03
Bug 1808167: Update crypto dependency

jcantrill:
Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1808167


syedriko:
@jcantrill How can I tell if the updated golang.org/x/crypto fixes the CVE?

jcantrill:
> Looks we miss the `Gopkg.toml` or did this not require an update?

I did not see a diff for it but I was surprised it wasn't their either

periklis:
> > Looks we miss the `Gopkg.toml` or did this not require an update?
> 
> I did not see a diff for it but I was surprised it wasn't their either

Then it is an indirect dependency.

/lgtm

PR #570, opened 2020-06-22
Add -cover and -race flags to unit tests.

alanconway:
Prints a coverage percent per package.
Runs the race detector on tests.


alanconway:
This was merged as part of  #574

PR #575, opened 2020-06-24
Bug 1850546: Add support label for s390x & ppc64le

andymcc:
In order for the cluster-logging-operator to be correctly filtered in
the OperatorHub for s390x and ppc64le we need to add the correct arch
label as supported.

### DIFF
--- a/manifests/4.6/cluster-logging.v4.6.0.clusterserviceversion.yaml
+++ b/manifests/4.6/cluster-logging.v4.6.0.clusterserviceversion.yaml
@@ -6,6 +6,10 @@ metadata:
   # The version value is substituted by the ART pipeline
   name: clusterlogging.v4.6.0
   namespace: placeholder
+  labels:
+    "operatorframework.io/arch.amd64": supported
+    "operatorframework.io/arch.ppc64le": supported
+    "operatorframework.io/arch.s390x": supported
   annotations:
     "operatorframework.io/suggested-namespace": openshift-logging
     "operatorframework.io/cluster-monitoring": "true"


periklis:
@andymcc Is this targeting only 4.6? Or needs backports?

andymcc:
@periklis - we'll need backports, I'll create BZ's for this (and the other 2 operators that have MA support)

periklis:
> @periklis - we'll need backports, I'll create BZ's for this (and the other 2 operators that have MA support)

Please create only one BZ for both operators with target release version 4.6.0 and link both PRs with it. The backport BZ can be taken over by the cherry-pick-bot.

andymcc:
To address: https://bugzilla.redhat.com/show_bug.cgi?id=1850546

Needs backport: release-4.3

periklis:
@andymcc AFAICS this change can only manually get backported to 4.5, 4.4 and 4.3 because it touches the version-specific CSV for each release. Same applies for openshift/elasticsearch-operator#406

andymcc:
@periklis ahh yeah of course that makes sense. Thanks for pointing it out - I can make the manual backports, but should I wait for this to get merged first? additionally, I assume that means I need to create a BZ for each release?

periklis:
@andymcc According to our BZ rules for OpenShift Org, you will need to wait until this gets merged. Basically you do the backport one at a time. Be patient, going down 4 release will take some time.

andymcc:
@periklis - that lines up with what I expected. Thanks for the pointers.

PR #578, opened 2020-06-25
Clean up the ClusterLogForwarder API to follow recommendations

alanconway:
- Move methods to separate files, `*_type.go` only has type definitions.
- Move all the main CRD spec types (InputSpec, OutputSpec) into one file 
  - OutputTypeSpec and children (Kafka, Syslog etc.) still in output_types.go 
- Add 'clf' as short name for ClusterLogForwarder.
- Improved API documentation, fixed some minor API issues (+required vs. +optional)
- Dropped the `insecure` field, not useful and is confusing.

Note apart from removing the insecure field, this is just moving code around and adding doc comments and some kubebuilder annotation comments.


alanconway:
/assign @jcantrill 
Can you give this a quick LGTM? Its just moving code around, no  changes.

ewolinetz:
/hold
just because it gets auto-approved

manifests/4.6/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -35,29 +42,32 @@ spec:
...
+                  \n There are three built-in inputs named `application`, `infrastructure`
+                  and `audit`. You don't need to define inputs here if those are sufficient
+                  for your needs. See `inputRefs` for more."
                 items:
                   description: InputSpec defines a selector of log messages.
                   properties:
                     application:
-                      description: 'Enable `application` logs. Use `application: {}`
-                        to enable with no filter.'
+                      description: 'Application enables `application` logs. `application:
```
ewolinetz, line 53: the first part is a little confusing to read simply because the property name is self documenting, what about something like "This property enables...."

manifests/4.6/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -35,29 +42,32 @@ spec:
...
+                        {}` enables all application logs.'
                       properties:
                         namespaces:
-                          description: Only collect logs from applications in these
-                            namespaces. If empty, all application container logs will
-                            be collected.
+                          description: Namespaces selects logs only from pods in the
```
ewolinetz, line 57: maybe something like "List of specific namespaces to collect pod logs from. If the list is empty..."

manifests/4.6/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -35,29 +42,32 @@ spec:
...
+                            specified namespaces. If the list is empty, all application
+                            logs are collected.
                           items:
                             type: string
                           type: array
                       type: object
                     audit:
-                      description: 'Enable `audit` logs. Use `infrastructure: {}`
-                        to enable with no filter.'
+                      description: 'Audit enables `audit` logs. `audit: {}` enable
```
ewolinetz, line 65: similar ask as above

manifests/4.6/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -35,29 +42,32 @@ spec:
...
+                        all audit logs'
                       type: object
                     infrastructure:
-                      description: 'Enable `infrastructure` logs. Use `infrastructure:
-                        {}` to enable with no filter.'
+                      description: 'Infrastructure enables `infrastructure` logs.
```
ewolinetz, line 69: same ask

ewolinetz:
in general looks good, few nits regarding descriptions

manifests/4.6/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: Reworded thusly: 
	// Application if present enables `application` logs.

This also complies with the Go convention of starting a doc comment with the field name. 
It seems like most Go-based CRDs follow the Go comment convention even though it uses the wrong case (leading caps) for the generated CRD field name. I'll follow the herd on this one, I guess people are used to seeing CRD descriptions with capitalized field names.

manifests/4.6/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: +1

manifests/4.6/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: +1

PR #579, opened 2020-06-26
Bug 1851381: Add condition when log collection is missing outputs

periklis:
### Description
This change set addresses a manual cherry pick for 8d0d82ff to handle cluster logging setups w/o a log store and log forwarding instances. In addition it improves the missing error handling for the else branch in the original commit.

### Notes to the patch-manger
This PR for release 4.4 is different to the parent PR for 4.5 because 8d0d82ff was introduced by a sprint enhancement task w/o noticing the broken feature. In addition the patch for 4.5 as well for 4.6 is a consequence of introducing Elastiscsearch 6. The target release 4.4 is still based on Elasticsearch 5. 

To address: https://bugzilla.redhat.com/show_bug.cgi?id=1851381

### DIFF
--- a/pkg/k8shandler/conditions.go
+++ b/pkg/k8shandler/conditions.go
@@ -1,0 +1,91 @@
+package k8shandler
+
+// the purpose of this file is give an easy means to update/add clusterconditions
+
+import (
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func (clusterRequest *ClusterLoggingRequest) UpdateCondition(conditionType logging.ClusterConditionType, message, reason string, status v1.ConditionStatus) error {
+
+	condition := clusterRequest.GetCondition(conditionType)
+
+	condition.Message = message
+	condition.Reason = reason
+	condition.Status = status
+
+	found := false
+	updated := false
+	for index, clusterCondition := range clusterRequest.cluster.Status.Conditions {
+		if clusterCondition.Type == condition.Type {
+			found = true
+			if condition.Status == v1.ConditionFalse {
+				clusterRequest.cluster.Status.Conditions = removeCondition(clusterRequest.cluster.Status.Conditions, index)
+				updated = true
+			} else {
+				if isConditionDifferent(clusterCondition, condition) {
+					condition.LastTransitionTime = metav1.Now()
+					clusterRequest.cluster.Status.Conditions[index] = condition
+					updated = true
+				}
+			}
+			break
+		}
+	}
+
+	if !found {
+		if condition.Status == v1.ConditionTrue {
+			condition.LastTransitionTime = metav1.Now()
+			clusterRequest.cluster.Status.Conditions = append(clusterRequest.cluster.Status.Conditions, condition)
+			updated = true
+		}
+	}
+
+	if updated {
+		return clusterRequest.UpdateStatus(clusterRequest.cluster)
+	}
+
+	return nil
+}
+
+func (clusterRequest *ClusterLoggingRequest) GetCondition(conditionType logging.ClusterConditionType) logging.ClusterCondition {
+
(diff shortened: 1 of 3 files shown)


pkg/k8shandler/forwarding.go
```diff
@@ -48,7 +49,26 @@ func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig() (config s
 
 	clusterRequest.ForwardingSpec = clusterRequest.normalizeLogForwarding(clusterRequest.cluster.Namespace, clusterRequest.cluster)
 	generator, err := forwarding.NewConfigGenerator(clusterRequest.cluster.Spec.Collection.Logs.Type, clusterRequest.includeLegacyForwardConfig(), clusterRequest.includeLegacySyslogConfig())
-	return generator.Generate(&clusterRequest.ForwardingSpec)
+	generatedConfig, err := generator.Generate(&clusterRequest.ForwardingSpec)
+
+	if err != nil {
+		return "",
+			clusterRequest.UpdateCondition(
```
ewolinetz, line 56: as we talked about on slack, we should not have the return here since we will obscure the original `err`.
please update the if/else to just call `clusterRequest.UpdateCondition` as it is currently structured, and leave the return for the end of the function as `return generatedConfig, err`

periklis: Late evening discussion missed to get persistent in my brain

pkg/k8shandler/conditions.go
```diff
@@ -0,0 +1,91 @@
...
+	condition.Reason = reason
+	condition.Status = status
+
+	found := false
+	updated := false
+	for index, clusterCondition := range clusterRequest.cluster.Status.Conditions {
+		if clusterCondition.Type == condition.Type {
+			found = true
+			if condition.Status == v1.ConditionFalse {
+				clusterRequest.cluster.Status.Conditions = removeCondition(clusterRequest.cluster.Status.Conditions, index)
```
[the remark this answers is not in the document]
ewolinetz: since we're doing it based on the index it is at, yes. also we break after we've matched the type

pkg/k8shandler/forwarding.go
periklis: > i thought we agreed on slack, that we should not have the return here since we will obscure the original `err`.

Regarding returning the error back to the reconciliation loop, I believe the reason no to return was a good one. We basically log out the error as a warning and add the DeadEnd Condition to the CR. We stop/pause the reconciliation loop because manual user intervention is in need here.

FYI, i put back the old behavior for this backport, because otherwise we are keeping reconciling the DeadEnd Condition endlessly. WDYT?

In addition, I put the missing error handling for creating the generator.

ewolinetz: the concern i have with us returning the result of the `clusterRequest.UpdateCondition` error instead of `err` is that we are hiding an error that who ever is calling this may want to handle in a specific way. I think we should still return the error, but also adjust the error handling of whomever calls the `generateCollectorConfig()` function.

periklis: I believe the following things are orthogonal:
* Returning the original error because generateConfig did not yield a valid configuration
* Reconcile fluentd with an empty configuration w/o crashlooping and in addition having a DeadEnd condition that indicates manual user intervention on the CR.

Returning the original error will keep the reconciliation trying to generate a config while fluentd remains on crashlooping w/o a config.

ewolinetz: what about doing this:
1. return the original error, because that is a better code practice than letting the caller think that everything was alright
2. adjust the caller to not keep retrying in the case that it was unable to generate a config (if we don't have a config, then don't spin up a fluentd daemonset, and have a condition explain this)

periklis: @ewolinetz Exactly the opposite is the case of the linked BZ. According to @sichvoge we support a fluentd-only setup in clusterlogging and that used to work before LF introduction. However until now the destiny of spining up a fluentd daemonset and generating a configuration is tightly bound. Therefore if later does not succeed - as it is the case of the BZ not providing any LF or logstore bits - I believe we need to shallow the original error into a log entry and in addition singal to the user the status condition "DeadEnd".

PR #583, opened 2020-07-01
Provide an exhaustive desciption for fluentd init container

periklis:
This PR provides an exhaustive description why the fluentd init container is bound to LogFowarding and/or a ClusterLogging logStore stanza.

Proof why needed: https://coreos.slack.com/archives/CB3HXM2QK/p1593617729192400

/cc @ewolinetz

### DIFF
--- a/pkg/k8shandler/fluentd.go
+++ b/pkg/k8shandler/fluentd.go
@@ -414,6 +414,38 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, elasticsearchAppName str
 	// Shorten the termination grace period from the default 30 sec to 10 sec.
 	fluentdPodSpec.TerminationGracePeriodSeconds = utils.GetInt64(10)
 
+	// FIXME: The following conditional branch is a refactoring candidate
+	// To address here https://issues.redhat.com/browse/LOG-833
+	//
+	// Why do we need this branch at all?
+	//
+	// The ClusterLogging CR supports three additional use cases without providing
+	// a logStore stanza namely:
+	// Case 1: A collection only stanza to use with secure forward
+	// Case 2: A collection only stanza to use with syslog
+	// Case 3: A collection only stanza to use with ClusterLogFowarder API
+	//
+	// Supporting all three cases implies:
+	// 1. We don't want to add the init container if `cluster.Spec.LogStore == nil`.
+	// 2. We don't want to add the init container if `cluster.Spec.LogStore == nil`
+	//    and a ClusterLogFowarder CR and does not provide a default output (only
+	//    relevant for case 3)
+	//
+	// What is the init container bound to a log store stanza?
+	//
+	// The init container implementation is only relevant for the default ES log store.
+	// It checks for the version of an ES cluster being >=6. In upgrade scenarios
+	// an ES cluster can eventually provide nodes with multiple version (5.x and 6.x)
+	// until all nodes are upgraded. The data model used in 6.x is not backward
+	// compatible to 5.x, thus preventing fluentd from starting avoids data model
+	// inconsistencies during ES upgrades from 5.x to 6.x.
+	//
+	// Why does this work w/o a ClusterLogForwarder CR provided?
+	//
+	// ClusterLogging and ClusterLogFowarder CR reconciliation share the same logic
+	// for collection config generation. If first is provided without the latter a
+	// ClusterLogFowarder with default fields is assumed.
+	// (See ClusterLoggingRequest#getLogForwarder)
 	if pipelineSpec.HasDefaultOutput() && cluster.Spec.LogStore != nil {
 		fluentdPodSpec.InitContainers = []v1.Container{
 			newFluentdInitContainer(cluster),


pkg/k8shandler/fluentd.go
```diff
@@ -414,6 +414,37 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, elasticsearchAppName str
 	// Shorten the termination grace period from the default 30 sec to 10 sec.
 	fluentdPodSpec.TerminationGracePeriodSeconds = utils.GetInt64(10)
 
+	// FIXME: The following conditional branch is a refactoring candidate
+	//
+	// Why do we need this branch at all?
+	//
+	// The ClusterLogging CR supports three additional use cases without providing
+	// a logStore stanza namely:
+	// Case 1: A collection only stanza to use with secure forward
```
jcantrill, line 423: Why do we specifically need to call out the target (e.g fluent or secure forward)?  It is only needed for the default output we provide; this is the only case

pkg/k8shandler/fluentd.go
```diff
@@ -414,6 +414,37 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, elasticsearchAppName str
...
+	// Case 2: A collection only stanza to use with syslog
+	// Case 3: A collection only stanza to use with ClusterLogFowarder API
+	//
+	// Supporting all three cases implies:
+	// 1. We don't want to add the init container if clsuter.Spec.LogStore == nil.
```
jcantrill, line 428: cluster

pkg/k8shandler/fluentd.go
```diff
@@ -414,6 +414,37 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, elasticsearchAppName str
...
+	// 2. We don't want to add the init container if cluster.Spec.LogStore == nil
+	//    and a ClusterLogFowarder CR and does not provide a default output (only
+	//    relevant for case 3)
+	//
+	// What is the init container bound to a log store stanza?
+	//
+	// The current init container implementation is only relevant for ES log stores.
```
jcantrill, line 435: only relevent for our, default log store

pkg/k8shandler/fluentd.go
periklis: Describing only the positive case where a log store is used with/without LF in place leaves a lot of space to forget why we support a collection-only stanza in ClusterLogging. According to docs we support only three use cases. In addition calling them out here puts more pressure to the reader to refactor this code branch, don't you think?

periklis:
> Maybe this should go in `docs/SOMEFILE` instead of as a code comment? FAQ?

The whole comment block targets the urgency to refactor this code branch. Moving it to an FAQ will leave the reader unaffected and unaware why we place code condintionals based on `PipelineSpec` and `LogStoreSpec` inside a fluentd pod factory method. The comment and the written disclaimer should be the scream in our heads to tackle this soon!

ewolinetz:
+1 to moving this to a readme/troubleshooting section rather than code

pkg/k8shandler/fluentd.go
```diff
@@ -414,6 +414,37 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, elasticsearchAppName str
 	// Shorten the termination grace period from the default 30 sec to 10 sec.
 	fluentdPodSpec.TerminationGracePeriodSeconds = utils.GetInt64(10)
 
+	// FIXME: The following conditional branch is a refactoring candidate
```
ewolinetz, line 417: is there a tech debt card associated with this?

periklis:
> +1 to moving this to a readme/troubleshooting section rather than code

I believe moving it to a troubleshooting section is loosing the scary effect this should have to get addressed sooner than later. In addition it is not troubleshooting anything, it provides only hints to the code reader. If someone like @bparees felt that this code branch is unreadable to give his blessing for a backport PR (See discussion thread on https://coreos.slack.com/archives/CB3HXM2QK/p1593617729192400), it is an indicator that something is wrong.

pkg/k8shandler/fluentd.go
```diff
@@ -414,6 +414,38 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, elasticsearchAppName str
...
+	// 2. We don't want to add the init container if `cluster.Spec.LogStore == nil`
+	//    and a ClusterLogFowarder CR and does not provide a default output (only
+	//    relevant for case 3)
+	//
+	// What is the init container bound to a log store stanza?
+	//
+	// The init container implementation is only relevant for the default ES log store.
+	// It checks for the version of an ES cluster being >=6. In upgrade scenarios
+	// an ES cluster can eventually provide nodes with multiple version (5.x and 6.x)
+	// until all nodes are upgraded. The data model used in 5.x is not backward
```
bparees, line 439: 6.x?

pkg/k8shandler/fluentd.go
```diff
@@ -414,6 +414,38 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, elasticsearchAppName str
...
+	// compatible to 5.x, thus preventing fluentd from starting avoids data model
+	// inconsistencies during ES upgrades from 5.x to 6.x.
```
bparees, line 441: so it prevents fluentd from starting until all ES nodes are at ES6.x?

pkg/k8shandler/fluentd.go
```diff
@@ -414,6 +414,38 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, elasticsearchAppName str
...
+	// Why do we need this branch at all?
+	//
+	// The ClusterLogging CR supports three additional use cases without providing
+	// a logStore stanza namely:
+	// Case 1: A collection only stanza to use with secure forward
+	// Case 2: A collection only stanza to use with syslog
+	// Case 3: A collection only stanza to use with ClusterLogFowarder API
+	//
+	// Supporting all three cases implies:
+	// 1. We don't want to add the init container if `cluster.Spec.LogStore == nil`.
```
bparees, line 429: "because spec.logstore==nil" means there is no elastic and thus there is no need to wait"?  (which covers case 1 and 2)?

pkg/k8shandler/fluentd.go
```diff
@@ -414,6 +414,38 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, elasticsearchAppName str
...
+	// compatible to 5.x, thus preventing fluentd from starting avoids data model
+	// inconsistencies during ES upgrades from 5.x to 6.x.
+	//
+	// Why does this work w/o a ClusterLogForwarder CR provided?
+	//
+	// ClusterLogging and ClusterLogFowarder CR reconciliation share the same logic
+	// for collection config generation. If first is provided without the latter a
+	// ClusterLogFowarder with default fields is assumed.
+	// (See ClusterLoggingRequest#getLogForwarder)
 	if pipelineSpec.HasDefaultOutput() && cluster.Spec.LogStore != nil {
```
bparees, line 449: i guess the case i'm still struggling with is "piplinespec that does not have default output, and spec.logstore is not nil"

why does it matter if the pipelinespec has a defaultoutput?  If logstore is not nil, then we have elastic and need the initcontainer, right?  Or is spec.logstore ignored elsewhere(for deploying elastic) when the pipeline has a non-default output?

pkg/k8shandler/fluentd.go
periklis: Exactly.

pkg/k8shandler/fluentd.go
periklis: Yes this is correct.

pkg/k8shandler/fluentd.go
periklis: One needs to put the default log store in perspective when a User creates a ClusterLogging with `spec.LogStore` set and a LogForwarding CR together. Regardless of any additional outputs put in the LogFowarding CR, the user will get always one additional output namely the default logstore from `spec.LogStore`. This automatic addition is there to ensure that a running logging stack and especially the logStore is not destroyed by creating a LogForwarding CR on the first usage. The user can opt-out later from using the default logstore.

bparees: i'm still not seeing what you'd do any differently if the LogForwarding CR(pipelinespec?) does not have a defaultoutput, but cluster.Spec.LogStore is set.

seems like you would still need to add the initcontainer because you're still deploying elastic.

periklis: Even if the user does not provide a default output in the LogForwarding CR, the `spec.LogStore` is inherited automatically for the user so in turn we have a default output.

However, if the user sets [DisableDefaultForwarding](https://github.com/openshift/cluster-logging-operator/blob/release-4.5/pkg/apis/logging/v1alpha1/forwarding_types.go#L39), this will be translated as an opt-out and anything defined in `spec.LogStore` is not taken into account.

bparees: > However, if the user sets DisableDefaultForwarding, this will be translated as an opt-out and anything defined in spec.LogStore is not taken into account.

got it, this is the piece i was missing i think.

PR #593, opened 2020-07-10
Add validation to ClusterLogForwarder name required to be "instance"

periklis:
**Description**
This PR addresses the required schema validation for `ClusterLogForwarder` resources to allow where `metadata.name` has value `instance`.

**Solution**
The approach selected is based on patching the generated CRD file with the on-board openshift client utility `oc kustomize`. The reasoning is that CRD openAPI schema generation is based on kubebuilder annotations, which is limited to explicitly declared fields in the CRD go struct fields. Since metadata fields (i.e. `TypeMeta` and `ObjectMeta`) are declared as embedded structs in the CRD go struct, their fields cannot be annotation via comments.

Kustomize is the onboard utility to patch manifests with kubectl/oc, e.g. to add common labels, namespaces and in extend apply patches.

/cc @jcantrill

### DIFF
--- a/hack/get-openshift-client.sh
+++ b/hack/get-openshift-client.sh
@@ -1,0 +1,16 @@
+#!/usr/bin/bash
+# Download latest release of openshift-client, intended for use by ../Makefile
+
+set -e
+VERSION=$(curl -s -L  https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/4.6.0-0.ci/latest | jq --raw-output '.name')
+
+DOWNLOAD_URL=$(curl -s -L  https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/4.6.0-0.ci/latest | jq --raw-output '.downloadURL')
+
+NAME="openshift-client-linux-$VERSION.tar.gz"
+pushd bin/
+echo "Extracting openshift client binary...."
+curl -sSfL "$DOWNLOAD_URL/$NAME" -O > "$NAME"
+tar xfz "$NAME" oc
+rm "$NAME"
+echo "Done"
+popd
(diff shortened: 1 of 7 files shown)


ewolinetz:
/hold
quick question on why we have the additional manifest files as opposed to just having the crd update?

periklis:
> /hold
> quick question on why we have the additional manifest files as opposed to just having the crd update?

:point_up: to the description, but TL;DR; The CRD is generated from golang source code and kubebuilder is not limited to do so.

manifests/4.6/kustomization.yaml
```diff
@@ -0,0 +1,9 @@
+resources:
+- logging.openshift.io_clusterlogforwarders_crd.yaml
+patchesJson6902:
+- path: crd-v1-singelton-patch.yaml
```
ewolinetz, line 4: "singleton"

hack/get-openshift-client.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/bash
+# Download latest release of openshift-client, intended for use by ../Makefile
+
+set -e
+version=$(curl -s -L  https://openshift-release.svc.ci.openshift.org/api/v1/releasestream/4.6.0-0.ci/latest | jq --raw-output '.pullSpec')
+
+pushd bin/
+echo "Extracting install binary...."
+oc adm release extract --command=oc "$version"
```
ewolinetz, line 9: this seems wonky... we need to have `oc` already on our machine to pull `oc` ?

ewolinetz:
> The CRD is generated from golang source code and kubebuilder is limited to do so.

it doesn't even seem to be in scope for kubebuilder... i wonder if we can (in parallel to this) open up an issue/RFE to add this similar to how we have the `resource` field

PR #594, opened 2020-07-10
LOG-719: Enable clusterlogging CRD generation

periklis:
### Description
Ref: https://issues.redhat.com/browse/LOG-719

This PR addresses the required annotations in `clusterlogging_types.go` to enable an _almost_ identical but backward-compatible generated CRD by operator-sdk's kubebuilder facility. It is based on `operator-sdk generate crds` and the `oc kustomize` of #593 to suppress validation that did not exist in the former CRD (e.g. status field) and in turn maintain backward compatibility. The remaining diff hunks in [crd-generation-diff.txt](https://github.com/openshift/cluster-logging-operator/files/4903194/crd-generation-diff.txt) represent non-breaking changes and are discussed in the following section. 

In detail, the changes included:
* The CRD generation ensures apiVersion `v1beta1` for `ClusterLogging` and `v1` for `ClusterLogForwarder`
* Transfers all descriptions from yaml file into `clusterlogging_types.go`
* Maintains all `nullable` and `optional` field validations.
* Maintains all `enum` field validations where provided in yaml file, i.e. only for `ManagementState`.
* Maintains singleton name `instance` validation by applying kustomize patch
* Maintains generic status validation as `type: object`  by applying kustomize patch

~~**Notes to the reviewer:** Ignore the first commit, this PR needs a rebase on master after #593 is merged.~~

### About remaining non-breaking CRD validation differences
1. Field `spec` includes a new section `versions` which represents the defaults for this field and thus not breaking the API. :heavy_check_mark: 
2. Field `openAPIV3Schema` includes the defaults for `apiVersion`, `kind` and `description`. :heavy_check_mark: 
3. Field `metadata.properties.nama` drops the description field because it is not allowed due to [CRD structural schema.](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#specifying-a-structural-schema). :heavy_check_mark: 
4. Field `kibana.properties.proxySpec` is replaced with the correct `kibana.properties.proxy`, was broken before. :heavy_check_mark: 
5. Fields `resources.properties.{limits,requests}` include additional validations for the string schema (e.g. `1Gi`, `100m`). This change is an addition and is not breaking existing API objects, because `resource.MustParse(...)` ensured this already. Affects `kibana`, `kibana.proxy`, `elastiscearch`, `fluentd`, `curator`. :heavy_check_mark: 
[...]

PR #595, opened 2020-07-10
Add output fluentd tuning parameters to clusterlogging API and CRD

periklis:
### Description
Ref: https://issues.redhat.com/browse/LOG-742

This PR represents the API and CRD implementation for the [fluentd tuning enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/fluentd-tuning.md). It provides an optional field for the ClusterLogging spec named `output.fluentd`. This PR is **not** addressing reconciliation of the field's contents to the collector configuration.

~~**Notes to the reviewers:** Ignore the first **two** commits, this PR needs a rebase on master after #593 and #594 are merged.~~

### Discussion
All new non-enumeration fields introduced by the `FluentdBufferSpec` struct include validation annotations for data types accepted by Fluentd (See [Support Data Types](https://docs.fluentd.org/configuration/config-file#supported-data-types-for-values)). In detail:
- For type `time` the validation only accepts values that indicate seconds, minutes, hours or days, i.e. `1s`, `2m`, `3h` and `4d` or simple integers representing number of seconds. Using golang's `time.duration` is too powerful and in turn would need conversions between the stacks.
- For type `size` the validation only acceps values that indicate kilobytes, megabytes, gigabytes and terabytes, i.e. `8k or 8K`, `16m  or 16M`, `32g or 32G` and `64g or 64G` or simple integers representing bytes. Using k8s api-machinery `resource.Quantity` is not possible, because it allows declaring values for CPU cores and memory. In addition api-machinery allows switching the base between 1024 (`Mi`) and 1000 (`M`) and using exponent notation. 

The validation-only approach for both types ensures avoiding conversion errors and user confusion between the API spec and the final generated configuration.

/cc @jcantrill @alanconway @vimalk78


pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -210,6 +215,147 @@ type CuratorSpec struct {
...
+	// ChunkLimitSize represents the maximum size of each chunk. Events will be
+	// written into chunks until the size of chunks become this size. For data
+	// type pattern see:
+	// https://docs.fluentd.org/configuration/config-file#supported-data-types-for-values
+	//
+	// Notice: The OpenAPI validation pattern is an ECMA262 regular expression
+	// (See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#properties)
+	//
+	// +kubebuilder:validation:Pattern:="^([0-9]+)([kmgtKMGT]{0,1})$"
+	ChunkLimitSize string `json:"chunkLimitSize"`
```
jcantrill, line 287: these should be underscore instead of camelcase per the enhancement spec: 
```The field names are copied exactly from fluentd 1.0, including the use of "_" rather than camelCase. This emphasizes that these settings are not a general API, they set defaults directly for the underlying fluentd.```

File: pkg/apis/logging/v1/clusterlogging_types.go [hunk 2]
```diff
@@ -210,6 +215,147 @@ type CuratorSpec struct {
...
+// apply to all buffer configurations of each fluentd output.
+type FluentdOutputSpec struct {
+	Buffer *FluentdBufferSpec `json:"buffer,omitempty"`
+}
+
+const (
+	// ThrowExceptionAction raises an exception when output buffer is full
+	ThrowExceptionAction OverflowActionType = "throw_exception"
+	// BlockAction blocks processing inputs when output buffer is full
+	BlockAction OverflowActionType = "block"
```
jcantrill, line 236: Do we need to mention this is the default for us?

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -210,6 +215,147 @@ type CuratorSpec struct {
...
+	// DropOldestChunkAction drops oldest chunk to accept newly incoming chunks
+	// when buffer is full
+	DropOldestChunkAction OverflowActionType = "drop_oldest_chunk"
+)
+
+type OverflowActionType string
+
+const (
+	// Flush one chunk per time key
```
jcantrill, line 245: Can we expand what "time key" represents?

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -38,6 +38,11 @@ type ClusterLoggingSpec struct {
 	//
 	// +nullable
 	Curation *CurationSpec `json:"curation,omitempty"`
+
+	// Specification for output tuning parameters
+	//
+	// +nullable
+	Output *OutputConfSpec `json:"output,omitempty"`
```
ewolinetz, line 45: shouldn't this be nested under the fluentd spec?

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -210,6 +215,147 @@ type CuratorSpec struct {
...
+	// https://docs.fluentd.org/configuration/config-file#supported-data-types-for-values
+	//
+	// Notice: The OpenAPI validation pattern is an ECMA262 regular expression
+	// (See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#properties)
+	//
+	// +kubebuilder:validation:Pattern:="^([0-9]+)([smhd]{0,1})$"
+	FlushInterval string `json:"flushInterval"`
+
+	// FlushAtShutdown represents the switch to flush all buffers on fluentd shutdown.
+	FlushAtShutdown bool `json:"flushAtShutdown"`
```
ewolinetz, line 328: i don't think this is necessary to expose to customers. we should pick an opinionated value (likely true)

pkg/apis/logging/v1/clusterlogging_types.go
ewolinetz: as we expand other collectors we would nest specific output specs under those as well -- also we may not need to do that for other collectors

periklis:
Closing this in favor of delivering API and reconciliation together in #596

PR #596, opened 2020-07-13
Add fluentd buffer config generation with tuning parameters

periklis:
### Description
Ref: https://issues.redhat.com/browse/LOG-742

This PR is follow up of#595 addressing fluentd config generation based on user-provided fluentd buffer configuration. In detail this PR will:
* Integrate buffer configuration with `fluentd/run.sh` provided environment variables. Precedence: API fields -> environment variables -> defaults.
* Add explicit default values for `flush_mode`, `retry_type` and `retry_wait` that match former defaults and to enable override via API.
* Drop environment variable `BUFFER_QUEUE_FULL_ACTION` from `fluentd/run.sh` in favor of `output.fluentd.buffer.OverflowAction`

**Notes to the reviewers:** Ignore the first ~~three~~ commit~~s~~, this PR needs a rebase on master after ~~#593, #594~~ and #595 are merged.


pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+const (
+	// Buffer size defaults
+	defaultOverflowAction = "block"
+
+	// Flush buffer defaults
+	defaultFlushThreadCount = "2"
+	defaultFlushMode        = "interval"
+	defaultFlushInterval    = "1s"
```
jcantrill, line 16: In the interest of being safe, I recommend reusing `TimeUnit` which has some attached conversion methods (or it did) and converting all of these to seconds without unit

pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
```diff
@@ -0,0 +1,122 @@
+package fluentd
```
jcantrill, line 1: Asking because I dont know the right answer... should we introduce an "output" package?

pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
```diff
@@ -0,0 +1,122 @@
...
+	defaultFlushAtShutdown  = false
+
+	// Retry buffer to output defaults
+	defaultRetryWait        = "1s"
+	defaultRetryType        = "exponential_backoff"
+	defaultRetryMaxInterval = "300s"
+
+	// Output fluentdForward default
+	fluentdForwardChunkLimitSize = "1m"
+	fluentdForwardOverflowAction = "throw_exception"
```
jcantrill, line 26: I believe if you look at the resolution of the env vars  that this would also be "block" and and not throw

pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
```diff
@@ -0,0 +1,122 @@
...
+	defaultFlushInterval    = "1s"
+	defaultFlushAtShutdown  = false
+
+	// Retry buffer to output defaults
+	defaultRetryWait        = "1s"
+	defaultRetryType        = "exponential_backoff"
+	defaultRetryMaxInterval = "300s"
+
+	// Output fluentdForward default
+	fluentdForwardChunkLimitSize = "1m"
```
jcantrill, line 25: Same as below I believe this would also be 8m

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -568,17 +568,28 @@ tls_cert_path {{ .SecretPath "ca-bundle.crt"}}
   @type file
   path '{{.BufferPath}}'
   queued_chunks_limit_size "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
-  chunk_limit_size "#{ENV['BUFFER_SIZE_LIMIT'] || '1m' }"
+  {{ if .TotalLimitSize }}
```
jcantrill, line 571: @alanconway mentioned this to me while writing the proposal and I believe it has merit.  We should consider passing the changes in via env vars as it would significantly simplify the config generation and reduce the number of:
* if/else checks
* helper methods needed
I like the idea of being explicit in config but in the interest of mentally embracing tuning it may make more sense

pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
periklis: It is the right question, we have already 4 go files with `output_` prefix. This is a good indication. I would put this as a follow-up PR/commit. WDYT?

pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
periklis: Good catch, looking back I was misguided by looking on `templates.go` for the default but should have looked on `run.sh`.

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -38,6 +38,11 @@ type ClusterLoggingSpec struct {
 	//
 	// +nullable
 	Curation *CurationSpec `json:"curation,omitempty"`
+
+	// Specification for output tuning parameters
+	//
+	// +nullable
+	Output *OutputConfSpec `json:"output,omitempty"`
```
ewolinetz, line 45: i feel this should be nested under the fluentd spec instead, that's what its there for

as we expand other collectors we would nest specific output specs under those as well -- also we may not need to do that for other collectors

to me, this is awkward -- we are introducing another field that is specific to our collectors when we already have a field for this, and we also have a field under that specific for fluentd. to me this is redundant

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -210,6 +215,147 @@ type CuratorSpec struct {
...
+	// https://docs.fluentd.org/configuration/config-file#supported-data-types-for-values
+	//
+	// Notice: The OpenAPI validation pattern is an ECMA262 regular expression
+	// (See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#properties)
+	//
+	// +kubebuilder:validation:Pattern:="^([0-9]+)([smhd]{0,1})$"
+	FlushInterval string `json:"flushInterval"`
+
+	// FlushAtShutdown represents the switch to flush all buffers on fluentd shutdown.
+	FlushAtShutdown bool `json:"flushAtShutdown"`
```
ewolinetz, line 328: i don't think this is necessary to expose to customers. we should pick an opinionated value (likely true)

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -210,6 +215,147 @@ type CuratorSpec struct {
...
+// OutputConfSpec represents the output configuration for tuning parameters per
+// output type, e.g. Fluentd. The tuning parameters are applied for all outputs
+// of the same type. Currently supported: `fluentd`.
+type OutputConfSpec struct {
+	Fluentd *FluentdOutputSpec `json:"fluentd,omitempty"`
+}
+
+// FluentdOutputSpec represents the configuration for outputs of type fluentd.
+// It supports overriding settings for fluentd buffer configurations. The settings
+// apply to all buffer configurations of each fluentd output.
```
periklis, line 227: Take over comment from #595 [comment](https://github.com/openshift/cluster-logging-operator/pull/595#discussion_r455066469):
> Nit. I think this is more obvious in the CRD generation but I think this particular comment should be less specific should it not? These settings are for overriding Fluent Output configuration . The included spec are specifically for buffer

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -210,6 +215,147 @@ type CuratorSpec struct {
...
+	// FlushInterval represents the time duration to wait between two consecutive flush
+	// operations. Takes only effect used together with `flushMode: interval`. For data
+	// type pattern see:
+	// https://docs.fluentd.org/configuration/config-file#supported-data-types-for-values
+	//
+	// Notice: The OpenAPI validation pattern is an ECMA262 regular expression
+	// (See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#properties)
+	//
+	// +kubebuilder:validation:Pattern:="^([0-9]+)([smhd]{0,1})$"
+	FlushInterval string `json:"flushInterval"`
```
periklis, line 325: Take over from #595 [comment](https://github.com/openshift/cluster-logging-operator/pull/595#discussion_r455070554):
> Consider reusing or abstracting the type for reuse: pkg/apis/logging/v1/index_management_types.go#TimeUnit

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -210,6 +215,147 @@ type CuratorSpec struct {
...
+	// ChunkLimitSize represents the maximum size of each chunk. Events will be
+	// written into chunks until the size of chunks become this size. For data
+	// type pattern see:
+	// https://docs.fluentd.org/configuration/config-file#supported-data-types-for-values
+	//
+	// Notice: The OpenAPI validation pattern is an ECMA262 regular expression
+	// (See https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md#properties)
+	//
+	// +kubebuilder:validation:Pattern:="^([0-9]+)([kmgtKMGT]{0,1})$"
+	ChunkLimitSize string `json:"chunkLimitSize"`
```
periklis, line 287: Take over from #595 [comment](https://github.com/openshift/cluster-logging-operator/pull/595#discussion_r455076527):
>these should be underscore instead of camelcase per the enhancement spec:
The field names are copied exactly from fluentd 1.0, including the use of "_" rather than camelCase. This emphasizes that these settings are not a general API, they set defaults directly for the underlying fluentd.

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -210,6 +215,147 @@ type CuratorSpec struct {
...
+// apply to all buffer configurations of each fluentd output.
+type FluentdOutputSpec struct {
+	Buffer *FluentdBufferSpec `json:"buffer,omitempty"`
+}
+
+const (
+	// ThrowExceptionAction raises an exception when output buffer is full
+	ThrowExceptionAction OverflowActionType = "throw_exception"
+	// BlockAction blocks processing inputs when output buffer is full
+	BlockAction OverflowActionType = "block"
```
periklis, line 236: Carry over from #595 [comment](https://github.com/openshift/cluster-logging-operator/pull/595#discussion_r455077348):
>Do we need to mention this is the default for us?

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -210,6 +215,147 @@ type CuratorSpec struct {
...
+	// DropOldestChunkAction drops oldest chunk to accept newly incoming chunks
+	// when buffer is full
+	DropOldestChunkAction OverflowActionType = "drop_oldest_chunk"
+)
+
+type OverflowActionType string
+
+const (
+	// Flush one chunk per time key
```
periklis, line 245: Carry over from #595 [comment](https://github.com/openshift/cluster-logging-operator/pull/595#discussion_r455077641):
>Can we expand what "time key" represents?

pkg/apis/logging/v1/clusterlogging_types.go
ewolinetz: cc @alanconway

alanconway: @ewolinetz: This has been a difficult point all along, we need to hammer it out till we're all happy.

Here's a summary of the debate:

My view: Fluent is **not just a collector**. It is a collector *and a forwarder*. We are trying to escape from the CLO==fluentd mindset, and present a logical API that can control other (possibly distinct) collector/forwarder impls in future. We need to be clear (to ourselves and users) about the difference between _collection_ (which logs get scraped/received from the node) and _forwarding_ (where/how logs get sent off the node)

These tuning parameters affect fluentd *as a forwarder* - they affect fluent "output" config sections, which are created by the ClusterLogForwarder from `output` specs. They have no effect on which/how logs are collected, only on how they are sent to remote targets.

IMO this logically belongs on the ClusterLogForwarder API (my original proposal) which would also allow us to specify different tuning per output. I yielded to @christianvogt and @jcantrill argument (as I understand it): implementing per-output on the forwarder is adding new functionality, we don't want to sink so much effort. We already have global config via env-vars, so global config via API is a low-risk way to do the same thing in managed mode.

The Compromise: OK but we need a new section. This is not "collector" config, it can't be in a section called "collector" and described as "Specification of the Collection component for the cluster". We considered `forwarder` and `tuning` before settling on `output`.

@ewolinetz thoughts? We should keep at this till we're all happy with the result.

pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
alanconway: It's already under "fowarding" so I would expect that everything in this package is related to "output". Maybe the output_ prefix is unnecessary?

pkg/generators/forwarding/fluentd/templates.go
alanconway: TLDR: do what's easiest :) My thinking was that as long as we have the env vars in run.sh anyway, we might as well take advantage of them. In future it might be better to simplify run.sh and get rid of those env-vars so there's only one place where we set the config - but we can do that when we get there, there's no benefit in anticipating it before we have actually simplified run.sh, which isn't for this PR.

pkg/apis/logging/v1/clusterlogging_types.go
alanconway: +1 this was overshoot by me copying from the fluent docs.

pkg/apis/logging/v1/clusterlogging_types.go
alanconway: Possibly, I was getting paranoid and probably over-explaining things when I wrote that. Feel free improve the wording :)

pkg/apis/logging/v1/clusterlogging_types.go
alanconway: So I think there are 3 options on the table:
- use existing collector.fluentd section
- use new sectoin output.fluentd (naming? forwarder, tuning)
- add clusterlogforwarder.output.tuning.fluentd section (output.fluentd is already used to define a fluentd *target*)

pkg/generators/forwarding/fluentd/templates.go
periklis: We concluded to go with the present approach using template helper methods.

pkg/apis/logging/v1/clusterlogging_types.go
periklis: We concluded to use `spec.forwarder.fluentd` for grouping the parameters.

pkg/apis/logging/v1/clusterlogging_types.go
periklis: We concluded to delete this parameter and default it to `true` for all buffers.

pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
periklis: As mentioned below, `TimeUnit` is related to elasticsearch and fluentd has its on limitation on the time config parameter type. We are safe from an API standpoint, because the validation is using the same pattern as fluentd's parser. Any wrong input will never make it into our API stored ClusterLogging CRs.

pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
periklis: Will follow-up on this in a separate PR.

pkg/apis/logging/v1/clusterlogging_types.go
periklis: TimeUnit is relevant for Elasticsearch, while the pattern here is Fluentd's version of time config parameters. The selected pattern is aligned with fluentd's parser. IMHO we should not mix these and keep them separate. Thus, I created `FluentdTimeUnit` to abstract the validation on a single place.

pkg/apis/logging/v1/clusterlogging_types.go
periklis: AFAIK, K8S APIs default to use camelCase. IMHO sticking to this standard makes our API more standard.

alanconway: I conceed this one, will change the enhancement doc. The k8s API conventions doc mandates camelCase https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#naming-conventions

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -210,6 +215,145 @@ type CuratorSpec struct {
 	Schedule string `json:"schedule"`
 }
 
+// OutputConfSpec represents the output configuration for tuning parameters per
+// output type, e.g. Fluentd. The tuning parameters are applied for all outputs
+// of the same type. Currently supported: `fluentd`.
```
alanconway, line 220:
```suggestion
// ForwarderSpec contains global tuning parameters for specific forwarder implementations.
// This field is not required for general use, it allows performance tuning by users 
// familiar with the underlying forwarder technology.
// Currently supported: `fluentd`.
```

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -688,25 +710,37 @@ const storeSyslogTemplate = `{{- define "storeSyslog" -}}
   keep_alive_cnt 9
   keep_alive_intvl 7200
 {{ end -}}
+
 	{{if .PayloadKey -}}
 	<format>
 	  @type single_value
 	  message_key {{.PayloadKey}}
 	</format>
 	{{end -}}
```
ewolinetz, line 719: non-blocker but while you're fixing `<buffer>` indentation can you also fix the indentation in the rest of this `<store>`? it can be an additional commit if you want..

ewolinetz:
i think we want to err on the side of caution with giving users too much ability to configure things. most of our users aren't going to be fluentd tuning experts nor will they want to have to go through the fluentd docs to know the correct buffer structure. also, when we decide to ship a different version of fluentd we control the schedule for that and can also update the buffer structure if need be. if we leave it entirely up to the customer then they need to update yet another file while performing a log forwarding upgrade.

periklis:
> just a thought, since the config parameters we are addding here map one-to-one to flunet’s config parameter, why not let user define a buffer conf, and supply the actual conf in the CR? We adding another layer is actually not making anything easier for customer, nor are we providing any added functionality. Customers have to know the detail fluent buffer parameters anyway. Moreover if we use a different fluent version where these parameters change how will we handle?
> 
> Users could define the buffer config in place like below:
> 
> ```
> spec:      
>   managementState: "Managed"      
>   collection:      
>     logs:      
>       type: "fluentd"      
>       fluentd: {}      
>   forwarder:       
>     fluentd:      
>       buffer:         
>         name: buffer-tuner1      
>         buffer: |+      
>           <buffer>          
>             @type file      
>             path '/var/lib/fluentd/default'      
>             flush_interval "10s"      
>             flush_thread_count "5"      
>             flush_at_shutdown "true"      
>             retry_max_interval "1000"      
>             retry_forever true      
>             queue_limit_length "1024"      
>             chunk_limit_size "32m"      
>             overflow_action "block"      
>           </buffer>      
> ```
> 
> and we can generate new config in the config map with the name given `buffer-tuner1.conf`
> _buffer-tune1.conf_:
> 
> ```
> <buffer>          
>   @type file      
>   path '/var/lib/fluentd/default'      
>   flush_interval "10s"      
>   flush_thread_count "5"      
>   flush_at_shutdown "true"      
>   retry_max_interval "1000"      
>   retry_forever true      
>   queue_limit_length "1024"      
>   chunk_limit_size "32m"      
>   overflow_action "block"      
> </buffer>  
> ```
> 
> and include this config in `fluent.conf`
> 
> ```
> <match **>    
>   @type copy                    
>   <store>        
>     @type elasticsearch    
>     @id default          
>     host elasticsearch.openshift-logging.svc.cluster.local    
>     port 9200    
>     // lot more params
>     request_timeout 2147483648    
>     @include buffer-tuner1.conf
>   </store>    
> </match>    
> ```
> 
> @jcantrill @alanconway

Thank you very much for this enhancement proposal. It is worth broadening the discussion on such topics in the team and hear multiple perspectices. :tada: :+1: 
However, I believe the proposed solution defeats two important aspects in our logging stack
1. It breaks with the philosophy that we provide an opinionated and tested set of defaults for an openshift logging solution. In addition we expose only the knobs that really need to be defined by users and nothing more or less.
2. The solution to provide a prose multi line yaml configuration breaks in two ways:
2.1 The prose of the buffer config cannot be easily validated and since our CRD is an API, we would open a big hole for arbitary content if we take into account fluentd's evolution. An API should be typed, simple, testable and very very concise, because we need to support it for years to come.
2.2 The buffer config is mixing API content with implementation details from fluentd. What if the implementation details are broken in the next fluentd release? How to come up with non-breaking fix in our API to upgrade customers? Exposing such details w/o a translation/transformation layer makes the API very fragile.

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -210,6 +215,146 @@ type CuratorSpec struct {
...
 
+// ForwarderSpec contains global tuning parameters for specific forwarder implementations.
+// This field is not required for general use, it allows performance tuning by users
+// familiar with the underlying forwarder technology.
+// Currently supported: `fluentd`.
+type ForwarderSpec struct {
+	Fluentd *FluentdForwarderSpec `json:"fluentd,omitempty"`
+}
+
+// FluentdForwarderSpec represents the configuration for forwarders of type fluentd.
```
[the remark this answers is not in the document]
periklis: Golang's comment style guide follows actually always the pattern:
```
// NameOfTheStruct ....
```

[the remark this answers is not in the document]
vimalk78: My point was about what comes out in the CRD description. Perhaps can use some other method to specify it.

PR #598, opened 2020-07-13
Add development command to generate logforwarder configuration

jcantrill:
This PR:

* Adds a development command to generate fluent config from the CLI

### DIFF
--- a/pkg/k8shandler/clusterloggingrequest.go
+++ b/pkg/k8shandler/clusterloggingrequest.go
@@ -14,8 +14,8 @@ import (
 )
 
 type ClusterLoggingRequest struct {
-	client  client.Client
-	cluster *logging.ClusterLogging
+	Client  client.Client
+	Cluster *logging.ClusterLogging
 
 	// ForwarderRequest is a logforwarder instance
 	ForwarderRequest *logging.ClusterLogForwarder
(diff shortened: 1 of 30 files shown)


pkg/k8shandler/clusterloggingrequest.go
```diff
@@ -16,7 +16,7 @@ import (
 
 type ClusterLoggingRequest struct {
 	client  client.Client
```
ewolinetz, line 18: while we're at it, why not also rename this and change its case? something like `K8sClient`? (since `Client` is very overloaded)

ewolinetz:
what do you forsee the ability to gen fluentd configs from CLI will buy us?

jcantrill:
> what do you forsee the ability to gen fluentd configs from CLI will buy us?

I really only see us being able to use this for collector testing.  The ability to manually generate a config and then edit it as needed.  Potentially one could use this to test the collector without needing to stand up a full cluster.  I would prefer to leave this is a "dev tool" then a full on command

cmd/forwarder-generator/main.go
```diff
@@ -0,0 +1,83 @@
...
+			logrus.Errorf("Unable to evaluate the LOG_LEVEL: %s %v", logLevel, err)
+			os.Exit(1)
+		}
+		logrus.SetLevel(level)
+	}
+
+	if len(os.Args) < 2 {
+		logrus.Error("Need to pass the logging forwarding yaml as an arg")
+		os.Exit(1)
+	}
```
periklis, line 32: A good option for flags is the already included `github.com/spf13/pflag` as in the operator main.go

alanconway:
I've been thinking about something similar for unit testing - instead of a command it would be some test helper functions so a unit test can create a config, `podman run` a fluentd image, write some log files to imitate conmon and verify that e.g. a local syslog server in a goroutine gets the forwarded message.

alanconway:
/lgtm
Will try to get my podman run stuff into shape so I can use this to build a test.

PR #599, opened 2020-07-14
LOG-839: Remove unused fluent env vars

jcantrill:
This PR:
* Removes unused env vars
* Replaces env vars in config in favor of generating defaults where appropriate

The change of interest is the commti for LOG-839

Ref: https://issues.redhat.com/browse/LOG-839

### DIFF
--- a/pkg/generators/forwarding/fluentd/fluent_conf_test.go
+++ b/pkg/generators/forwarding/fluentd/fluent_conf_test.go
@@ -316,32 +316,32 @@ var _ = Describe("Generating fluentd config", func() {
 				</match>
 				<filter kubernetes.**>
 					@type kubernetes_metadata
-					kubernetes_url "#{ENV['K8S_HOST_URL']}"
-					cache_size "#{ENV['K8S_METADATA_CACHE_SIZE'] || '1000'}"
-					watch "#{ENV['K8S_METADATA_WATCH'] || 'false'}"
-					use_journal "#{ENV['USE_JOURNAL'] || 'nil'}"
-					ssl_partial_chain "#{ENV['SSL_PARTIAL_CHAIN'] || 'true'}"
+					kubernetes_url 'https://kubernetes.default.svc'
+					cache_size '1000'
+					watch 'false'
+					use_journal 'nil'
+					ssl_partial_chain 'true'
 				</filter>
 
 				<filter kubernetes.journal.**>
 					@type parse_json_field
-					merge_json_log "#{ENV['MERGE_JSON_LOG'] || 'false'}"
-					preserve_json_log "#{ENV['PRESERVE_JSON_LOG'] || 'true'}"
-					json_fields "#{ENV['JSON_FIELDS'] || 'MESSAGE,log'}"
+					merge_json_log 'false'
+					preserve_json_log 'true'
+					json_fields 'log,MESSAGE'
 				</filter>
 
 				<filter kubernetes.var.log.containers.**>
 					@type parse_json_field
-					merge_json_log "#{ENV['MERGE_JSON_LOG'] || 'false'}"
-					preserve_json_log "#{ENV['PRESERVE_JSON_LOG'] || 'true'}"
-					json_fields "#{ENV['JSON_FIELDS'] || 'log,MESSAGE'}"
+					merge_json_log 'false'
+					preserve_json_log 'true'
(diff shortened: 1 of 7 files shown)


periklis:
@jcantrill After taking a closer look on this, I can attest that it is not conflicting with #596 except that both drop `BUFFER_QUEUE_FULL_ACTION`. But this is not a big deal.

jcantrill:
> @jcantrill After taking a closer look on this, I can attest that it is not conflicting with #596 except that both drop `BUFFER_QUEUE_FULL_ACTION`. But this is not a big deal.

I need to push a change as the test log shows the value for this field is not a one of the valid enum values

periklis:
> > @jcantrill After taking a closer look on this, I can attest that it is not conflicting with #596 except that both drop `BUFFER_QUEUE_FULL_ACTION`. But this is not a big deal.
> 
> I need to push a change as the test log shows the value for this field is not a one of the valid enum values

Yes the fluentdForward case is using `exception` and that should be `throw_exception`. Noticed the same on my tuning spec PR.

jcantrill:
Blocked pending the outcome of https://github.com/openshift/cluster-logging-operator/pull/596  This PR may need to be significantly altered

pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -316,32 +316,32 @@ var _ = Describe("Generating fluentd config", func() {
...
 				</filter>
 
 				<filter kubernetes.journal.**>
 					@type parse_json_field
-					merge_json_log "#{ENV['MERGE_JSON_LOG'] || 'false'}"
-					preserve_json_log "#{ENV['PRESERVE_JSON_LOG'] || 'true'}"
-					json_fields "#{ENV['JSON_FIELDS'] || 'MESSAGE,log'}"
+					merge_json_log 'false'
+					preserve_json_log 'true'
+					json_fields 'log,MESSAGE'
```
periklis, line 330: Is the order of `log` and `MESSAGE` important for the filter plugin?

pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -950,11 +950,11 @@ var _ = Describe("Generating fluentd config", func() {
...
 						type_name _doc
 						write_operation create
-						reload_connections "#{ENV['ES_RELOAD_CONNECTIONS'] || 'true'}"
+						reload_connections 'true'
 						# https://github.com/uken/fluent-plugin-elasticsearch#reload-after
-						reload_after "#{ENV['ES_RELOAD_AFTER'] || '200'}"
+						reload_after '200'
 						# https://github.com/uken/fluent-plugin-elasticsearch#sniffer-class-name
-						sniffer_class_name "#{ENV['ES_SNIFFER_CLASS_NAME'] || 'Fluent::Plugin::ElasticsearchSimpleSniffer'}"
+
```
periklis, line 957: Is the `sniffer_class_name` drop intentionally?

jcantrill:
> LGTM. Besides the question below, why are on some places boolean, number and reserved words value quoted with single quotes and on other not?, e.g. `'false', '-1', 'nil'`

I replaced the values as they would have been before with the env evaluation.  In the interest of not breaking what works I would prefer to leave it but I think the question stands and is valid:  when is quoting needed

pkg/generators/forwarding/fluentd/fluent_conf_test.go
jcantrill: yes.  this should be the default.  I believe this is here for when the change was not part of the original plugin

pkg/generators/forwarding/fluentd/fluent_conf_test.go
jcantrill: I don't believe so.  we have examples of where this is flipped the other way but it is the same fields

pkg/generators/forwarding/fluentd/fluent_conf_test.go
jcantrill: @periklis looks as I was wrong and should actually check the default https://bugzilla.redhat.com/show_bug.cgi?id=1864397  https://github.com/uken/fluent-plugin-elasticsearch#sniffer-class-name

PR #600, opened 2020-07-14
Bug 1833486: Collector config changes to address performance versus 4.4

jcantrill:
This PR provides additional collector config changes to address performance versus 4.4:

* Bumps rotate_wait to allow fluent to continue to collect rotated logs
* Drops the new file interval
* Extends the pos_file_compact_interval which seems to affect dropped messages for short runs with high load

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1833486

### DIFF
--- a/pkg/generators/forwarding/fluentd/fluent_conf_test.go
+++ b/pkg/generators/forwarding/fluentd/fluent_conf_test.go
@@ -108,9 +108,9 @@ var _ = Describe("Generating fluentd config", func() {
   path "/var/log/containers/*_project1-namespace_*.log", "/var/log/containers/*_project2-namespace_*.log"
   exclude_path ["/var/log/containers/fluentd-*_openshift-logging_*.log", "/var/log/containers/elasticsearch-*_openshift-logging_*.log", "/var/log/containers/kibana-*_openshift-logging_*.log"]
   pos_file "/var/log/es-containers.log.pos"
-  pos_file_compaction_interval 300
-  refresh_interval 5
-  rotate_wait 1
+  pos_file_compaction_interval 1800
+  refresh_interval 1
+  rotate_wait 300
   tag kubernetes.*
   read_from_head "true"
   @label @CONCAT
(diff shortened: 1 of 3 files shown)


jcantrill:
> LGTM overall. One tiny recommendation is to use the time unit for values. In this case we speak about seconds right?

I tried doing this for the compact interval and the code drops the unit and assumes seconds; i do not know if this is a bug or by design.  It doesn't actually honor anything but seconds so 5m = 5 sec, 5000h = 5000 sec.  I would otherwise prefer time units.

periklis:
> > LGTM overall. One tiny recommendation is to use the time unit for values. In this case we speak about seconds right?
> 
> I tried doing this for the compact interval and the code drops the unit and assumes seconds; i do not know if this is a bug or by design. It doesn't actually honor anything but seconds so 5m = 5 sec, 5000h = 5000 sec. I would otherwise prefer time units.

OMG. This is an important finding that needs vetting on the new API tuning knobs, because they units are not part of the API. Good to know.

jcantrill:
> > > LGTM overall. One tiny recommendation is to use the time unit for values. In this case we speak about seconds right?
> > 
> > 
> > I tried doing this for the compact interval and the code drops the unit and assumes seconds; i do not know if this is a bug or by design. It doesn't actually honor anything but seconds so 5m = 5 sec, 5000h = 5000 sec. I would otherwise prefer time units.
> 
> OMG. This is an important finding that needs vetting on the new API tuning knobs, because they units are not part of the API. Good to know. Did you notice this also with other fields, e.g. `retry_wait` that is actually a field that we expose?

I only discovered it for the pos_file_compact_interval and don't know if it applies to other fields.  In theory, this is defined as a `:time` field that would have common characteristics for all the plugins but I have no confirmation

periklis:
> > LGTM overall. One tiny recommendation is to use the time unit for values. In this case we speak about seconds right?
> 
> I tried doing this for the compact interval and the code drops the unit and assumes seconds; i do not know if this is a bug or by design. It doesn't actually honor anything but seconds so 5m = 5 sec, 5000h = 5000 sec. I would otherwise prefer time units.

Did you notice this also for 1.11.x? Because it was an issue with 1.9.2 according to https://github.com/fluent/fluentd/issues/2917

PR #602, opened 2020-07-16
Bug 1856940: Add kafka mTLS conditional support

periklis:
### Description
This PR addresses the missing conditional configuration for kafka mTLS support. In detail the fluentd configuration sets the client cert and key only if the calculated paths from the secretRef exists on runtime.

To address: https://bugzilla.redhat.com/show_bug.cgi?id=1856940

/cc @igor-karpukhin @jcantrill

### DIFF
--- a/pkg/generators/forwarding/fluentd/output_conf_kafka_test.go
+++ b/pkg/generators/forwarding/fluentd/output_conf_kafka_test.go
@@ -81,8 +81,8 @@ var _ = Describe("Generating external kafka server output store config block", f
            brokers broker1-kafka.svc.messaging.cluster.local:9092
            default_topic topic
            ssl_ca_cert '/var/run/ocp-collector/secrets/some-secret/ca-bundle.crt'
-           ssl_client_cert '/var/run/ocp-collector/secrets/some-secret/tls.crt'
-           ssl_client_cert_key '/var/run/ocp-collector/secrets/some-secret/tls.key'
+           ssl_client_cert "#{File.exist?('/var/run/ocp-collector/secrets/some-secret/tls.crt') ? '/var/run/ocp-collector/secrets/some-secret/tls.crt' : use_nil}"
+           ssl_client_cert_key "#{File.exist?('/var/run/ocp-collector/secrets/some-secret/tls.key') ? '/var/run/ocp-collector/secrets/some-secret/tls.key' : use_nil}"
            <format>
                @type json
            </format>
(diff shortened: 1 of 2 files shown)


pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -715,9 +715,11 @@ const storeKafkaTemplate = `{{- define "storeKafka" -}}
 brokers {{.Brokers}}
 default_topic {{.Topic}}
 {{ if .Target.Secret -}}
+{{ $tlsCert := .SecretPath "tls.crt" }}
+{{ $tlsKey := .SecretPath "tls.key" }}
 ssl_ca_cert '{{ .SecretPath "ca-bundle.crt"}}'
-ssl_client_cert '{{ .SecretPath "tls.crt"}}'
-ssl_client_cert_key '{{ .SecretPath "tls.key"}}'
+ssl_client_cert "#{File.exist?('{{ $tlsCert }}') ? '{{ $tlsCert }}' : use_nil}"
```
jcantrill, line 721: Isn't there code elsewhere where we check for the secret and its keys prior to using the template which means we remove one more conditional from configuration at run time.  This should be our goal: make the configuration "what you see is what you get" as much as is possible

periklis: We actually don't check the secret or its contents when generating TLS configuration for our outputs. We enable the TLS config block if the forwarder spec has a sercetRef. The normalizer is only checking if the sercetRef exists as a sercret resource nothing more.

pkg/generators/forwarding/fluentd/output_conf_kafka_test.go
```diff
@@ -81,8 +81,8 @@ var _ = Describe("Generating external kafka server output store config block", f
            brokers broker1-kafka.svc.messaging.cluster.local:9092
            default_topic topic
            ssl_ca_cert '/var/run/ocp-collector/secrets/some-secret/ca-bundle.crt'
-           ssl_client_cert '/var/run/ocp-collector/secrets/some-secret/tls.crt'
-           ssl_client_cert_key '/var/run/ocp-collector/secrets/some-secret/tls.key'
+           ssl_client_cert "#{File.exist?('/var/run/ocp-collector/secrets/some-secret/tls.crt') ? '/var/run/ocp-collector/secrets/some-secret/tls.crt' : use_nil}"
```
alanconway, line 84: Don't like repeating the filename, opportunity for future error. Suggest wraping this expression in a func existing_filename() or similar. Like the rubyites say, keep your code DRY :)

periklis: It's already the case in `templates.go` with `{{ .SecretPath }}`. You look here on the expected output of a test case ;)

alanconway: Doh!
/lgtm

PR #604, opened 2020-07-20
Bug 1858200: Disable LogFowarding when legacy protocols in use

periklis:
### Description
This PR addresses a fix to disable [log forwarding](https://docs.openshift.com/container-platform/4.5/logging/cluster-logging-external.html#cluster-logging-collector-log-forward_cluster-logging-external) config generation when one of the two legacy methods is used, i.e. [fluentd legacy forward](https://docs.openshift.com/container-platform/4.5/logging/cluster-logging-external.html#cluster-logging-collector-fluentd_cluster-logging-external), [syslog legacy forward](https://docs.openshift.com/container-platform/4.5/logging/cluster-logging-external.html#cluster-logging-collector-syslog_cluster-logging-external)).

To address: https://bugzilla.redhat.com/show_bug.cgi?id=1858200

**Needs backport:** release-4.5 and release-4.4

/cc @jcantrill @alanconway @vimalk78

### DIFF
--- a/pkg/generators/forwarding/fluentd/generators.go
+++ b/pkg/generators/forwarding/fluentd/generators.go
@@ -41,35 +41,72 @@ func NewConfigGenerator(includeLegacyForwardConfig, includeLegacySyslogConfig, u
 }
 
 //Generate the fluent.conf file using the forwarding information
-func (engine *ConfigGenerator) Generate(pipelines *logging.ClusterLogForwarderSpec, forwarderSpec *logging.ForwarderSpec) (string, error) {
-	logger.DebugObject("Generating fluent.conf using %s", pipelines)
-	logger.DebugObject("Generating fluent.conf using buffer config: %s", forwarderSpec)
+func (engine *ConfigGenerator) Generate(clfSpec *logging.ClusterLogForwarderSpec, fwSpec *logging.ForwarderSpec) (string, error) {
+	logger.DebugObject("Generating fluent.conf using %s", clfSpec)
+	logger.DebugObject("Generating fluent.conf using buffer config: %s", fwSpec)
 
 	//sanitize here
-	var sourceInputLabels []string
-	var sourceToPipelineLabels []string
-	var pipelineToOutputLabels []string
-	var outputLabels []string
-	var err error
+	var (
+		inputs                 sets.String
+		namespaces             sets.String
+		routeMap               logging.RouteMap
+		sourceInputLabels      []string
+		sourceToPipelineLabels []string
+		pipelineToOutputLabels []string
+		outputLabels           []string
+		err                    error
+	)
 
-	inputs, namespaces := gatherSources(pipelines)
+	// Provide inputs and inputsPipelines for legacy forwarding protocols
+	// w/o a user-provided ClusterLogFowarder instance to enable inputs and
+	// inputs-to-pipelines a.k.a. route map template generation.
+	if engine.includeLegacyForwardConfig || engine.includeLegacySyslogConfig {
+		inputs = sets.NewString(
+			string(logging.InputNameInfrastructure),
+			string(logging.InputNameApplication),
+			string(logging.InputNameAudit),
+		)
 
-	if sourceInputLabels, err = engine.generateSource(inputs, namespaces); err != nil {
+		routeMap = logging.RouteMap{
+			logging.InputNameInfrastructure: sets.NewString(),
+			logging.InputNameApplication:    sets.NewString(),
+			logging.InputNameAudit:          sets.NewString(),
+		}
+	} else {
+		inputs, namespaces = gatherSources(clfSpec)
+		routeMap = inputsToPipelines(clfSpec.Pipelines)
+	}
+
+	sourceInputLabels, err = engine.generateSource(inputs, namespaces)
+	if err != nil {
+
 		logger.Tracef("Error generating source blocks: %v", err)
 		return "", err
 	}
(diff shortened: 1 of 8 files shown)


jcantrill:
> This PR addresses a fix to disable [log forwarding](https://docs.openshift.com/container-platform/4.5/logging/cluster-logging-external.html#cluster-logging-collector-log-forward_cluster-logging-external) config generation when one of the two legacy methods is used,

What is the reasoning for disabling? Why can the configuration not support both?

periklis:
> > This PR addresses a fix to disable [log forwarding](https://docs.openshift.com/container-platform/4.5/logging/cluster-logging-external.html#cluster-logging-collector-log-forward_cluster-logging-external) config generation when one of the two legacy methods is used,
> 
> What is the reasoning for disabling? Why can the configuration not support both?

According to this [thread](https://coreos.slack.com/archives/GGUR75P60/p1595230491162500), our docs should be understood as exclusive alternatives.

test/e2e/logforwarding/fluentlegacy/forward_to_fluent_test.go
```diff
@@ -84,7 +84,7 @@ var _ = Describe("Backwards compatibility prior to ClusterLogForwarder", func()
 					return e2e.KubeClient.CoreV1().Secrets(fluentDeployment.Namespace).Delete(context.TODO(), secret.ObjectMeta.Name, opts)
 				})
 
-				components := []helpers.LogComponentType{helpers.ComponentTypeCollector, helpers.ComponentTypeStore}
```
vimalk78, line 87: why removed this?

pkg/generators/forwarding/fluentd/generators.go
```diff
@@ -44,30 +44,67 @@ func NewConfigGenerator(includeLegacyForwardConfig, includeLegacySyslogConfig, u
...
+		outputLabels           []string
+		err                    error
+	)
 
-	inputs, namespaces := gatherSources(forwarder)
+	// Provide inputs and inputsPipelines for legacy forwarding protocols
+	// w/o a user-provided ClusterLogFowarder instance to enable inputs and
+	// inputs-to-pipelines a.k.a. route map template generation.
+	if engine.includeLegacyForwardConfig || engine.includeLegacySyslogConfig {
+		inputs = sets.NewString(
```
vimalk78, line 62: If legacy LF is used, it must collect all inputs?

pkg/generators/forwarding/fluentd/output_conf_legacy_test.go
```diff
@@ -0,0 +1,1376 @@
...
+
+            <store>
+              @type relabel
+              @label @_LEGACY_SECUREFORWARD
+            </store>
+
+          </match>
+        </label>
+
+        # Relabel specific pipelines to multiple, outputs (e.g. ES, kafka stores)
```
vimalk78, line 445: can we get this comment removed, since we are avoiding the pipeline itself?

pkg/generators/forwarding/fluentd/output_conf_legacy_test.go
```diff
@@ -0,0 +1,1376 @@
...
+        </label>
+      `
+
+      results, err := generator.Generate(forwarderSpec)
+      Expect(err).Should(Succeed())
+      Expect(results).To(EqualTrimLines(legacyConf))
+    })
+  })
+
+  Context("based on legacy syslog and fluentd forward method", func() {
```
vimalk78, line 911: can we keep just this one test case, other two are included in this one it seems

test/e2e/logforwarding/fluentlegacy/forward_to_fluent_test.go
periklis: The test is designated to forward logs only to a fluentd receiver and is actually asserting only logs for that output. So why installing an ES at all and churning resources w/o reason?

pkg/generators/forwarding/fluentd/generators.go
periklis: For legacy LF there was no config option to declare which logs to collect or not. Users could only configure which logs they accepted on fluentd/syslog and drop all others.

pkg/generators/forwarding/fluentd/output_conf_legacy_test.go
periklis: The test cases are distinct from each other. They cover all the if-conditionals in the template code, a) only fluent, b) only syslog, c) both. The configuration looks different for each three and it should be tested in isolation each time.

pkg/generators/forwarding/fluentd/generators.go
```diff
@@ -41,35 +41,72 @@ func NewConfigGenerator(includeLegacyForwardConfig, includeLegacySyslogConfig, u
...
 		return "", err
 	}
 	sort.Strings(sourceToPipelineLabels)
-	if pipelineToOutputLabels, err = engine.generatePipelineToOutputLabels(pipelines.Pipelines); err != nil {
-		logger.Tracef("Error generating pipeline to output labels blocks: %v", err)
-		return "", err
+
+	// Omit generation for missing pipelines, i.e. legacy methods don't provide any
+	if len(clfSpec.Pipelines) > 0 {
+		pipelineToOutputLabels, err = engine.generatePipelineToOutputLabels(clfSpec.Pipelines)
```
vimalk78, line 96: the len() check is necessary? the method is going to loop on the same array.

PR #605, opened 2020-07-20
Bug 1857206: Use fluentd event time for processing logs to kafka

periklis:
### Description
This PR addresses a fix to use Fluentd EventTime instead of ruby's time for nanosecond precision.

To address: https://bugzilla.redhat.com/show_bug.cgi?id=1857206

/cc @jcantrill @vimalk78

### DIFF
--- a/pkg/generators/forwarding/fluentd/output_conf_kafka_test.go
+++ b/pkg/generators/forwarding/fluentd/output_conf_kafka_test.go
@@ -25,6 +25,7 @@ var _ = Describe("Generating external kafka server output store config block", f
            @type kafka2
            brokers broker1-kafka.svc.messaging.cluster.local:9092
            default_topic topic
+           use_event_time true
            <format>
                @type json
            </format>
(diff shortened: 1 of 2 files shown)


pkg/generators/forwarding/fluentd/output_conf_kafka_test.go
```diff
@@ -80,6 +81,7 @@ var _ = Describe("Generating external kafka server output store config block", f
            @type kafka2
            brokers broker1-kafka.svc.messaging.cluster.local:9092
            default_topic topic
+           use_event_time true
```
jcantrill, line 84: How does 'event time' relate to the time parsed from the log message?  Do we still get both?

periklis: The parsed time is converted from Fluentd EventTime instead of using `Time.now` from ruby in: https://github.com/fluent/fluent-plugin-kafka/blob/master/lib/fluent/plugin/out_kafka2.rb#L265

jcantrill: I believe this is only relevent if the time is unable to be determined from the incoming message when sourced.  Does that matter?  seems we may not end up with the precision we are trying to introduce as a log's timestamp is supposed to be from when it was generated not when it was ingested by fluent

vimalk78: it seems using `use_event_time true` gives us log's timestamp from when it was generated.
perhaps `use_event_time` default value should be true in the plugin ?

PR #610, opened 2020-07-22
Bug 1856502: Added Elasticsearch proxy resources

vimalk78:
fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1856502

### DIFF
--- a/manifests/4.6/logging.openshift.io_clusterloggings_crd.yaml
+++ b/manifests/4.6/logging.openshift.io_clusterloggings_crd.yaml
@@ -245,6 +245,41 @@ spec:
                       description: Define which Nodes the Pods are scheduled on.
                       nullable: true
                       type: object
+                    proxy:
+                      description: Specification of the Elasticsearch Proxy component
+                      properties:
+                        resources:
+                          description: ResourceRequirements describes the compute
+                            resource requirements.
+                          nullable: true
+                          properties:
+                            limits:
+                              additionalProperties:
+                                anyOf:
+                                - type: integer
+                                - type: string
+                                pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                                x-kubernetes-int-or-string: true
+                              description: 'Limits describes the maximum amount of
+                                compute resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/'
+                              type: object
+                            requests:
+                              additionalProperties:
+                                anyOf:
+                                - type: integer
+                                - type: string
+                                pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
+                                x-kubernetes-int-or-string: true
+                              description: 'Requests describes the minimum amount
+                                of compute resources required. If Requests is omitted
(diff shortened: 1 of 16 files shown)


pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -150,6 +150,12 @@ type ElasticsearchSpec struct {
 
 	// +optional
 	RedundancyPolicy elasticsearch.RedundancyPolicyType `json:"redundancyPolicy"`
+
+	// The resource requirements for ElasticsearchProxy
+	//
+	// +nullable
+	// +optional
+	ProxyResources *v1.ResourceRequirements `json:"proxyResources"`
```
periklis, line 158: To keep following the pattern (e.g. See `KibanaSpec.ProxySpec`), splitting to a separate `ProxySpec` struct is more appropriate. Grouping proxy configuration in one new struct enables backward compatible additions. (cc @ewolinetz )

vimalk78: updated

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -150,6 +150,9 @@ type ElasticsearchSpec struct {
 
 	// +optional
 	RedundancyPolicy elasticsearch.RedundancyPolicyType `json:"redundancyPolicy"`
+
+	// Specification of the Kibana Proxy component
+	ProxySpec `json:"proxy,omitempty"`
```
periklis, line 155: This reuses unfortunately the same proxy spec from kibana. The proxy component refers for kibana to `oauth-proxy`, but for elasticsearch to `elasticsearch-proxy`. The two ProxySpecs should be different structs. In addition please align the comment block for this field to match the purpose.

manifests/4.6/logging.openshift.io_clusterloggings_crd.yaml
```diff
@@ -245,6 +245,41 @@ spec:
                       description: Define which Nodes the Pods are scheduled on.
                       nullable: true
                       type: object
+                    proxy:
+                      description: Specification of the Kibana Proxy component
```
ewolinetz, line 249: s/Kibana/Elasticsearch/

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -150,6 +150,9 @@ type ElasticsearchSpec struct {
 
 	// +optional
 	RedundancyPolicy elasticsearch.RedundancyPolicyType `json:"redundancyPolicy"`
+
+	// Specification of the Kibana Proxy component
```
ewolinetz, line 154: s/Kibana/Elasticsearch/

pkg/apis/logging/v1/clusterlogging_types.go
ewolinetz: i disagree, i think it is acceptable to reuse this spec given its just:
```
type ProxySpec struct {
	// +nullable
}
```

i think it is unnecessary to create another struct that represents this same thing but just for elasticsearch's proxy

ewolinetz:
please address the comment referencing kibana otherwise :+1:

pkg/apis/logging/v1/clusterlogging_types.go
periklis: What if we want to extend this struct with kibana's oauth-proxy only fields?

ewolinetz: we can add fields specific to oauth at that time

periklis: But the struct will be a union of fields for oauth-proxy and fields for both proxies. Isn't this awkward?

ewolinetz: i dont think so, its for a proxy... it's delineated by function. similar to logstore or collection or visualization or forwarder

PR #611, opened 2020-07-22
Bug 1858811: syslog fixed support for tls protocol

vimalk78:
fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1858811

/cc: @jcantrill @alanconway

### DIFF
--- a/pkg/generators/forwarding/fluentd/syslog_conf.go
+++ b/pkg/generators/forwarding/fluentd/syslog_conf.go
@@ -72,6 +72,19 @@ func (conf *outputLabelConf) Tag() string {
 	return tag
 }
 
+func (conf *outputLabelConf) SyslogProtocol() string {
+	switch strings.ToLower(conf.Protocol()) {
+	case "":
+		fallthrough
+	case "tls":
+		return "tcp"
+	case "udp":
+		return "udp"
+	default:
+		return conf.Protocol()
+	}
+}
+
 func (conf *outputLabelConf) PayloadKey() string {
 	return conf.Target.Syslog.PayloadKey
 }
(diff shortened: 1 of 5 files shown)


pkg/generators/forwarding/fluentd/syslog_conf.go
```diff
@@ -72,6 +72,18 @@ func (conf *outputLabelConf) Tag() string {
...
 
+func (conf *outputLabelConf) SyslogProtocol() string {
+	switch strings.ToLower(conf.Protocol()) {
+	case "tls":
+		return "tcp"
+	case "":
+		return "tcp"
+	case "udp":
+		return "udp"
+	}
```
periklis, line 83: nit. Consider using constants from golang for protocol names.

pkg/generators/forwarding/fluentd/syslog_conf.go
```diff
@@ -72,6 +72,18 @@ func (conf *outputLabelConf) Tag() string {
 	return tag
 }
 
+func (conf *outputLabelConf) SyslogProtocol() string {
+	switch strings.ToLower(conf.Protocol()) {
```
ewolinetz, line 76: can we also add a `default:` here?

pkg/generators/forwarding/fluentd/syslog_conf.go
```diff
@@ -72,6 +72,18 @@ func (conf *outputLabelConf) Tag() string {
...
+	case "tls":
+		return "tcp"
+	case "":
+		return "tcp"
+	case "udp":
+		return "udp"
+	}
+	return conf.Protocol()
```
ewolinetz, line 84: remove this in favor for `default:` case

pkg/generators/forwarding/fluentd/syslog_conf.go
```diff
@@ -72,6 +72,18 @@ func (conf *outputLabelConf) Tag() string {
 	return tag
 }
 
+func (conf *outputLabelConf) SyslogProtocol() string {
+	switch strings.ToLower(conf.Protocol()) {
+	case "tls":
+		return "tcp"
```
ewolinetz, line 78: nit, but since `"tls"` and `""` return the same value we can instead have:
```
case "tls":
  fallthrough
case "":
  return "tcp"
```

pkg/generators/forwarding/fluentd/syslog_conf.go
vimalk78: i didnt find which defines tcp/udp/tls all in same

PR #612, opened 2020-07-22
Remove generated tech-preview logforwardings.crd.yaml

alanconway:
Remove the generated YAML file which was left behind when the Go source was removed.

/cc @periklis


alanconway:
@periklis  I'm ok with holding off on this, but I was assuming we would just do a YAML tree transform from the TP to GA formats, in which case (I think) we wouldn't need the old API code or the CRD. If we do it by unmarshalling into the old API struct and hand coding a copy to the new then we do need it, but I was guessing (hoping) that a tree transform would be simpler to code.

periklis:
> @periklis I'm ok with holding off on this, but I was assuming we would just do a YAML tree transform from the TP to GA formats, in which case (I think) we wouldn't need the old API code or the CRD. If we do it by unmarshalling into the old API struct and copying field-by-field to the new then we do need it, but I was guessing (hoping) that a tree transform would be simpler to code.

I had a similar assumption until last planning meeting, where you were off. @jcantrill and @sichvoge would like to have live transformation on upgrading the operator. I am investigating this now and I believe we need the CRD for the APIserver to allow the operator read the old resource and transform it to the new ClusterLogFowarder and save it back.

alanconway:
Closing - we're going to keep the old API around to (possibly) enable live conversions to the new API after upgrade.

PR #613, opened 2020-07-22
DO NOT MERGE: Test for 'empty config with namespaces'

alanconway:
According to https://bugzilla.redhat.com/show_bug.cgi?id=1857673 we generate an empty fluentd config if the 'namespaces' feature is used.  This unit test shows that the generator works correctly with the suspect YAML file, so we must be dropping the configuration somewhere else.

/hold
DO NOT MERGE - @vimalk78 owns this and will merge the final result, this is just a temporary work area.


alanconway:
@vimalk78 do you still need this PR? If not I'll close it.

vimalk78:
@alanconway you can close it

alanconway:
Has been included in #626

PR #615, opened 2020-07-23
Clean up E2E test suite scripts and junit reporting

periklis:
### Description
This PR addresses a series of minor improvements for the E2E test suites. I detail:
* Adds the ginkgo junit reporter for all golang-based tests in addition to the default reporter.
* Renames ginkgo top-level `Describe` with an appropriate prefix per API, e.g. `[ClusterLogForwarder]...`, `[ClusterLogging]...` to enhance grouping in junit reports.
* Re-activates junit reporting for all bash-based tests by using the `os::cmd::*` commands.
* Activates the OLM operator upgrade test suite `test-020-olm-upgrade`.
* Cleans up duplicate functionality in `testing-olm/utils` in favor of the junit-reporting `os::cmd::*` functions.


ewolinetz:
for posterity can you show the before and after outputs?

periklis:
> for posterity can you show the before and after outputs?

@ewolinetz Here we go:

### Before
You see only the junit report from the openshift-tests suite, e.g. pulling images, etc.:
image

### After
You can see each Gingko testcase on top of the openshift-tests suite:
image

hack/lib/test/junit.sh
```diff
@@ -183,15 +183,15 @@ function os::test::junit::generate_report() {
 #  export JUNIT_REPORT_NUM_FAILED
 function os::test::junit::internal::generate_report() {
     local report_type="$1"
-    os::util::ensure::built_binary_exists 'junitreport'
+    # os::util::ensure::gopath_binary_exists 'junitreport'
```
ewolinetz, line 186: why do we not want to ensure this?

periklis: Good catch!

PR #617, opened 2020-07-23
Fix lint errors, refuse to build unlinted code.

alanconway:
/assign @jcantrill
/cc @periklis

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,22 +20,25 @@ FLUENTD_IMAGE?=quay.io/openshift/origin-logging-fluentd:latest
 # Update code (generate, format), run unit tests and lint. Make sure e2e tests build.
 check: generate fmt test-unit
 	go test ./test/... -exec true > /dev/null # Build but don't run e2e tests.
-	$(MAKE) lint
+	$(MAKE) lint				  # Only lint if code builds.
 
-openshift-client:
+bin:
+	mkdir -p bin
+
+openshift-client: bin
 	@type -p oc > /dev/null || bash hack/get-openshift-client.sh
 
 # Download tools if not available on local PATH.
-operator-sdk:
+operator-sdk: bin
 	@type -p operator-sdk > /dev/null || bash hack/get-operator-sdk.sh
 
-golangci-lint:
+golangci-lint: bin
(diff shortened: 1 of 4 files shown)


pkg/k8shandler/defaults.go
```diff
@@ -19,7 +19,4 @@ var (
 
 	defaultFluentdMemory     resource.Quantity = resource.MustParse("736Mi")
 	defaultFluentdCpuRequest resource.Quantity = resource.MustParse("100m")
-
```
vimalk78, line 22: unrelated change?

Makefile
```diff
@@ -32,7 +32,7 @@ operator-sdk:
 golangci-lint:
 	@type -p golangci-lint > /dev/null || GOFLAGS="" GO111MODULE=off go get github.com/golangci/golangci-lint/cmd/golangci-lint
 
-build:
+build: lint		# Reject unlinted code in CI.
```
periklis, line 35: Can we re-evaluate the set of linters and the golangci-lint version we use in a follow-up PR please? I tend to believe that we are using really old linters.

alanconway: Sure, open a JIRA or just start a PR with proposal.

pkg/k8shandler/defaults.go
alanconway: Nope, those are unused variables, they cause lint to fail. They are the reason I was pushed to do this - tired of getting other peoples lint in my PRs :)

alanconway:
> Is this going to work with our CI building images
> 
> /lgtm cancel

@periklis That's the idea. I want CI to fail if someone's PR has linty code so they have fix it before it infects master.
I originally suggested changing the CI script, but @jcantrill felt it would be easier to just ad to build: target.

alanconway:
Looks like CI is faliing to `go get` the golangci-lint tool:

alanconway:
@periklis @igor-karpukhin rebased and improved a little, needs LGTM

periklis:
> CI was failing in `go get`. I switched to the module version of `go get` since gloagnci-lint now supports that.
> It required some module changes and a surprising number of vendor changes - @periklis can you say if it makes sense?

Since golangci-lint is a go modules project, doing `go get` w/o prepending `GOFLAGS=""...` would result in adding this tool into our `go.mod` definitions and don't want this for now. At least as long as we don't have a conclusive decision how we want to include tools.

PR #622, opened 2020-07-24
Remove need for LOGGING_SHARE_DIR in tests.

alanconway:
Use the working directory for tests to locate /cluster-logging-operator/files, drop the env var LOGGING_SHARE_DIR. Allows all tests to run directly with `go test`

### DIFF
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -10,6 +10,7 @@ import (
 	"path"
 	"path/filepath"
 	"reflect"
+	"strings"
 	"time"
 
 	"github.com/sirupsen/logrus"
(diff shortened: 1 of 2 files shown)


pkg/utils/utils.go
```diff
@@ -180,11 +181,15 @@ func GetFileContents(filePath string) []byte {
...
 func GetShareDir() string {
-	shareDir := os.Getenv("LOGGING_SHARE_DIR")
-	if shareDir == "" {
-		return "/usr/share/logging"
-	}
-	return shareDir
+	// shareDir is <repo-root>/files - try to find from working dir.
+	const sep = string(os.PathSeparator)
+	const repoRoot = sep + "cluster-logging-operator" + sep
+	wd, _ := os.Getwd()
```
igor-karpukhin, line 187: @alanconway why suppress errors?

alanconway: The outcome is the same - failure of Getwd() will return "", and we definitely won't find /cluster-logging-operator/ in that.
However you are right that it would be clearer to handle the error, so I'll fix that and put a postit on my screen saying "Shortcuts Make Long Delays"  ;)

alanconway:
/lgtm
connection failure, looks like a flake

alanconway:
/retest
connection failure ,looks like a flake

PR #621, opened 2020-07-24
Bug 1860393:  LogForwarding GA on 4.5

jcantrill:
This PR brings LogForwarding GA into 4.5

Ref: 
* https://issues.redhat.com/browse/LOG-829
* https://bugzilla.redhat.com/show_bug.cgi?id=1860393

PR #624, opened 2020-07-27
Add one-time conversion for LogForwarding to ClusterLogForwarder CR

periklis:
### Description
This PR adds a migration path for API users of LogForwarding Tech Preview to ClusterLogFowarder GA. The migration takes place after upgrading the operator and the controller syncs with the apiserver all LogFowarding CRs. The outcome is a single ClusterLogForwarder instance according to the requirements defined in the [enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/cluster-logging-log-forwarding.md).


manifests/4.6/cluster-logging.v4.6.0.clusterserviceversion.yaml
```diff
@@ -353,6 +353,8 @@ spec:
         version: v1
       - kind: ClusterLogForwarder
         version: v1
+      - kind: LogForwarding
+        version: v1alpha1
```
periklis, line 357: These changes will be dropped once this PR is re-targeted to 4.5.z

manifests/4.6/cluster-logging.v4.6.0.clusterserviceversion.yaml
```diff
@@ -483,3 +485,40 @@ spec:
...
+      - description: The status of forwarding pipelines
+        displayName: Pipelines Status
+        path: pipelines
+        x-descriptors:
+        - 'urn:alm:descriptor:com.tectonic.ui:pipelineStatuses'
+      - description: The status of log forwarding resourece
+        displayName: Log Forwarding Status
+        path: status
+        x-descriptors:
+        - 'urn:alm:descriptor:com.tectonic.ui:logforwardingStatuses'
```
periklis, line 524: Same here, will be dropped once the PR is re-targeted to 4.5.z.

pkg/controller/logforwarding/logforwarding_controller.go
```diff
@@ -0,0 +1,102 @@
...
+		return reconcile.Result{}, err
+	}
+
+	clf := &loggingv1.ClusterLogForwarder{}
+	if err := instance.ConvertTo(clf); err != nil {
+		return reconcile.Result{}, err
+	}
+
+	opts := []client.CreateOption{}
+	err = r.client.Create(context.TODO(), clf, opts...)
```
jcantrill, line 90: There should be some status of update of both the GA and TP object here to indicate one is no longer usable and the other was created out of the other

pkg/apis/logging/v1alpha1/conversion.go
```diff
@@ -0,0 +1,153 @@
...
+
+import (
+  "fmt"
+
+  metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+  loggingv1 "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+)
+
+func (src *LogForwarding) ConvertTo(dst *loggingv1.ClusterLogForwarder) error {
```
ewolinetz, line 11: how do you feel about "TranslateTo" ?

periklis: Conversion is close to k8s an api-machinery terminology when a resource is translated/transformed/whatever to a different form and/or version.

pkg/controller/logforwarding/logforwarding_controller.go
```diff
@@ -0,0 +1,134 @@
...
+	if err != nil {
+		return err
+	}
+
+	// Watch for creates only
+	pred := predicate.Funcs{
+		UpdateFunc:  func(e event.UpdateEvent) bool { return false },
+		DeleteFunc:  func(e event.DeleteEvent) bool { return false },
+		CreateFunc:  func(e event.CreateEvent) bool { return true },
+		GenericFunc: func(e event.GenericEvent) bool { return false },
```
ewolinetz, line 48: i think we would also want to process for the `GenericEvent` too. IIRC this will trigger when the operator first starts up and there is already an existing object

pkg/controller/logforwarding/logforwarding_controller.go
```diff
@@ -0,0 +1,134 @@
...
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+
+		return reconcile.Result{}, err
+	}
+
+	clf := &loggingv1.ClusterLogForwarder{}
+	if err := instance.ConvertTo(clf); err != nil {
+		return reconcile.Result{}, err
```
ewolinetz, line 117: can we also update the TP status to indicate that we were unable to convert it? this way if it gets "lost" in the CLO logs a customer can still easily find this message

pkg/controller/logforwarding/logforwarding_controller.go
```diff
@@ -0,0 +1,134 @@
...
+	createOpts := []client.CreateOption{}
+	err = r.client.Create(context.TODO(), clf, createOpts...)
+	if err != nil {
+		if errors.IsAlreadyExists(err) {
+			// ClusterLogForwarder already exists, don't requeue
+			reqLogger.Info("Skip reconcile: ClusterLogFowarder already exists")
+			return reconcile.Result{}, nil
+		}
+
+		return reconcile.Result{}, err
```
ewolinetz, line 129: shouldn't we requeue the event to make sure we eventually create the GA LF object?

pkg/controller/logforwarding/logforwarding_controller.go
```diff
@@ -0,0 +1,134 @@
...
+	err := r.client.Get(context.TODO(), request.NamespacedName, instance)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
```
ewolinetz, line 92: this doesn't requeue

pkg/controller/logforwarding/logforwarding_controller.go
periklis: AFAIK, returning an err will always requeue:
```
// The Controller will requeue the Request to be processed again if the returned error is non-nil or
// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
```

pkg/controller/logforwarding/logforwarding_controller.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+
+	updateOpts := []client.UpdateOption{}
+	err = r.client.Status().Update(context.TODO(), instance, updateOpts...)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return reconcile.Result{}, nil
+		}
+
+		return reconcile.Result{}, err
```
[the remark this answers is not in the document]
periklis: AFAIK, returning an err will always requeue:
```
// The Controller will requeue the Request to be processed again if the returned error is non-nil or
// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
```

pkg/controller/logforwarding/logforwarding_controller.go
periklis: AFAIK, returning an err will always requeue:
```
// The Controller will requeue the Request to be processed again if the returned error is non-nil or
// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
```

pkg/controller/logforwarding/logforwarding_controller.go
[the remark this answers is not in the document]
alanconway: Wouldn't hurt to explicitly set `Result.Requeue = true` as well, as an "executable comment" that this will be requeued. Not essential though if there are no other changes to make.

ewolinetz:
/hold
isn't this pointing to the wrong branch?

pkg/controller/logforwarding/logforwarding_controller.go
ewolinetz: interesting, the error thing is news to me

pkg/controller/logforwarding/logforwarding_controller.go
```diff
@@ -0,0 +1,153 @@
...
+func add(mgr manager.Manager, r reconcile.Reconciler) error {
+	// Create a new controller
+	c, err := controller.New("logforwarding-controller", mgr, controller.Options{Reconciler: r})
+	if err != nil {
+		return err
+	}
+
+	// Watch for creates only
+	pred := predicate.Funcs{
+		UpdateFunc:  func(e event.UpdateEvent) bool { return true },
```
ewolinetz, line 45: did we decide that we're going to update every time? I thought we wanted to encourage users to stop updating the TP api and instead move to the GA one and this PR was to save them the headache of log loss upon upgrade.

to me, if we do this on updates then this isn't a "one-time conversion"

periklis: I re-adjusted this one, because we stop reconciling to GA if a TP resource is mallformed. This allows users to fix their TP resource and let the controller convert it to a GA resource. The controller won't process any further update evengts if a GA resource was successfully created.

periklis:
> /hold
> isn't this pointing to the wrong branch? or are we now going to support this for 4.6?
> @jcantrill

This is true and you are right to put an hold label on this. I started developing this against master to have CI running. But this one needs to get re-targeted to `release-4.5` after #621 is merged.

pkg/controller/logforwarding/logforwarding_controller.go
ewolinetz: if a TP resource is malformed does that mean that they wouldn't have logs flowing in as well? or is it possible that it can be malformed and work in TP?

pkg/controller/logforwarding/logforwarding_controller.go
```diff
@@ -0,0 +1,153 @@
...
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			return reconcile.Result{}, nil
+		}
+		// Error reading the object - requeue the request.
+		return reconcile.Result{}, err
+	}
+
+	clf := &loggingv1.ClusterLogForwarder{}
+	if err := instance.ConvertTo(clf); err != nil {
```
ewolinetz, line 96: does it make sense to save time converting (and checking for errors and unnecessarily updating status) by checking if a `ClusterLogForwarder` already exists?

jcantrill:
> > /hold
> > isn't this pointing to the wrong branch? or are we now going to support this for 4.6?
> > @jcantrill
> 
> This is true and you are right to put an hold label on this. I started developing this against master to have CI running. But this one needs to get re-targeted to `release-4.5` after #621 is merged.

I actually take back my comment.  @sichvoge we need this in master correct since we would be supporting LFTP for 1 release?  Potentially we could have a customer who has 4.5 pre GA upgrading to 4.6GA?

ewolinetz:
> I actually take back my comment. @sichvoge we need this in master correct since we would be supporting LFTP for 1 release? Potentially we could have a customer who has 4.5 pre GA upgrading to 4.6GA?

but we had decided our upgrade story to 4.6 was that we are okay with a break since it is a minor version bump (vs a release stream bump) and a customer can choose not to move to 4.6 right away

PR #625, opened 2020-07-28
Bug 1861013: Disable LogFowarding when legacy protocols in use

periklis:
### Description
This PR is a manual cherry-pick of #604 to address: https://bugzilla.redhat.com/show_bug.cgi?id=1861013

### Notes to the patch-manager
The manual cherry-pick is justified here, because the underlying issue conflicts with different releases of the log forwarding API. In release 4.5 the API is called `LogFowarding` and is TechPreview. OTOH in release 4.6 the API is called `ClusterLogForwarder` and will be GA when the release lands for public consumption. However, the fix addresses how both APIs work together with our legacy support of ConfigMap based forwarding methods, i.e. `fluent forward`and `syslog`. Both methods are declared legacy since release 4.3 and should be used exclusive to either `LogFowarding` TechPreview or `ClusterLogFowarder` GA. 

The issue addressed here is re-establishing this exclusive usage to allow users to use the legacy methods w/o the need to provide a log forwarding custom resource.

/cc @jcantrill @vimalk78

### DIFF
--- a/pkg/generators/forwarding/fluentd/fluent_conf_test.go
+++ b/pkg/generators/forwarding/fluentd/fluent_conf_test.go
@@ -16,7 +16,7 @@ var _ = Describe("Generating fluentd config", func() {
 	)
 	BeforeEach(func() {
 		var err error
-		generator, err = NewConfigGenerator(true, true, true)
+		generator, err = NewConfigGenerator(false, false, true)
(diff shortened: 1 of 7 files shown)


igor-karpukhin (approved):
/lgtm

PR #626, opened 2020-07-28
Bug 1857673: Fixed fluentd conf generation, namespace filtering

vimalk78:
fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1857673

/cc @jcantrill @alanconway

### DIFF
--- a/pkg/generators/forwarding/fluentd/generators.go
+++ b/pkg/generators/forwarding/fluentd/generators.go
@@ -74,7 +74,7 @@ func (engine *ConfigGenerator) Generate(clfSpec *logging.ClusterLogForwarderSpec
 		}
 	} else {
 		inputs, namespaces = gatherSources(clfSpec)
-		routeMap = inputsToPipelines(clfSpec.Pipelines)
+		routeMap = inputsToPipelines(clfSpec)
 	}
 
 	sourceInputLabels, err = engine.generateSource(inputs, namespaces)
@@ -161,11 +161,20 @@ func gatherSources(forwarder *logging.ClusterLogForwarderSpec) (types sets.Strin
 	return types, namespaces
 }
 
-func inputsToPipelines(pipelines []logging.PipelineSpec) logging.RouteMap {
+func inputsToPipelines(fwdspec *logging.ClusterLogForwarderSpec) logging.RouteMap {
 	result := logging.RouteMap{}
-	for _, pipeline := range pipelines {
+	inputs := fwdspec.InputMap()
+	for _, pipeline := range fwdspec.Pipelines {
 		for _, inRef := range pipeline.InputRefs {
-			result.Insert(inRef, pipeline.Name)
+			if input, ok := inputs[inRef]; ok {
+				// User defined input spec, unwrap.
+				for t := range input.Types() {
+					result.Insert(t, pipeline.Name)
+				}
+			} else {
+				// Not a user defined type, insert direct.
+				result.Insert(inRef, pipeline.Name)
+			}
 		}
 	}
 	return result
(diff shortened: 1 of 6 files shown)


pkg/generators/forwarding/fluentd/generators.go
```diff
@@ -161,11 +161,27 @@ func gatherSources(forwarder *logging.ClusterLogForwarderSpec) (types sets.Strin
 	return types, namespaces
 }
 
-func inputsToPipelines(pipelines []logging.PipelineSpec) logging.RouteMap {
+func inputsToPipelines(fwdspec *logging.ClusterLogForwarderSpec) logging.RouteMap {
 	result := logging.RouteMap{}
-	for _, pipeline := range pipelines {
+	for _, pipeline := range fwdspec.Pipelines {
 		for _, inRef := range pipeline.InputRefs {
```
alanconway, line 167: This is correct, but I propose the following refactor. Getting the types from an InputSpec seem like it might be a common operation so I put it into the logging package. This simplifies the generator code which can also handle more than one type in an InputSpec (not sure if we will ever need that but it's easy so might as well play safe.)
```
diff --git a/pkg/apis/logging/v1/cluster_log_forwarder.go b/pkg/apis/logging/v1/cluster_log_forwarder.go
index ec98429f..dbd84db9 100644
--- a/pkg/apis/logging/v1/cluster_log_forwarder.go
+++ b/pkg/apis/logging/v1/cluster_log_forwarder.go
@@ -108,3 +108,18 @@ func (spec *ClusterLogForwarderSpec) HasDefaultOutput() bool {
 	_, ok := spec.OutputMap()[OutputNameDefault]
 	return ok
 }
+
+// Types returns the set of input types that are used to by the input spec.
+func (input *InputSpec) Types() sets.String {
+	result := sets.NewString()
+	if input.Application != nil {
+		result.Insert(InputNameApplication)
+	}
+	if input.Infrastructure != nil {
+		result.Insert(InputNameInfrastructure)
+	}
+	if input.Audit != nil {
+		result.Insert(InputNameAudit)
+	}
+	return result
+}
diff --git a/pkg/generators/forwarding/fluentd/generators.go b/pkg/generators/forwarding/fluentd/generators.go
index c28c0184..42cfd0d3 100644
--- a/pkg/generators/forwarding/fluentd/generators.go
+++ b/pkg/generators/forwarding/fluentd/generators.go
@@ -163,24 +163,15 @@ func gatherSources(forwarder *logging.ClusterLogForwarderSpec) (types sets.Strin
 
 func inputsToPipelines(fwdspec *logging.ClusterLogForwarderSpec) logging.RouteMap {
 	result := logging.RouteMap{}
+	inputs := fwdspec.InputMap()
 	for _, pipeline := range fwdspec.Pipelines {
 		for _, inRef := range pipeline.InputRefs {
-			if len(fwdspec.Inputs) > 0 {
-				// look up type of input
-				for _, input := range fwdspec.Inputs {
-					if input.Name == inRef {
-						// ideally input should only be of one type
-						if input.Application != nil {
-							result.Insert(logging.InputNameApplication, pipeline.Name)
-						} else if input.Infrastructure != nil {
-							result.Insert(logging.InputNameInfrastructure, pipeline.Name)
-						} else if input.Audit != nil {
-							result.Insert(logging.InputNameAudit, pipeline.Name)
-						}
-					}
+			if input, ok := inputs[inRef]; ok { // User defined input spec, unwrap.
+				for t := range input.Types() {
+					result.Insert(t, pipeline.Name)
 				}
 			} else {
-				result.Insert(inRef, pipeline.Name)
+				result.Insert(inRef, pipeline.Name) // Not a user defined type, insert direct.
 			}
 		}
 	}
```

vimalk78: updated.

PR #628, opened 2020-07-29
Add must-gather plugin image for cluster-logging stack

periklis:
### Description
This PR provides a must-gather plugin image for the `oc adm must-gather` facility to align cluster-logging's debugging practices within the openshift organization. The implementation is basically a break out of the established [origin-aggregated-logging/logging-dump.sh](https://github.com/openshift/origin-aggregated-logging/blob/master/hack/logging-dump.sh) but with several improvements and benefits:
1. The dump of standard k8s resources and their logs (e.g. deployments, services, secrets, pods, cronjobs, etc.) is automated via `oc adm inspect` and follows a well-known filesystem structure.
2. The former `logging-dump.sh` checks and dumps are grouped in separate `gather_$COMPONENT_resources` scripts and in turn improve maintenance and extensibility.
3. Users can easily identify that our stacks provides a `must-gather` plugin image by looking on the top-level repo directory or the docs. The can use it directly instead of requesting the `logging-dump` script in BZs or forum channels.
4. The packaging model in a container hardens the gathering process from polluted user environments.
5. The must-gather output for standard k8s resource either namespace-scoped or cluster-scoped ensures that users can easily navigate through the output as they do with other operators.

### Open questions
- Where to place on the docs? [HERE???](https://docs.openshift.com/container-platform/4.5/support/gathering-cluster-data.html#gathering-data-specific-features_gathering-cluster-data)
- Where to host the image on quay.io? proposed: `quay.io/cluster-logging-dev/cluster-logging-must-gather`

### Want more, just convince yourself
**Note:** First of all ensure a running clusterlogging instance. Then execute:
```
oc adm must-gather --image quay.io/periklis/cluster-logging-must-gather
```

/cc @jcantrill @ewolinetz @alanconway

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@ WORKDIR /go/src/github.com/openshift/cluster-logging-operator
 
 # COPY steps are in the reverse order of frequency of change
 COPY cmd ./cmd
+COPY must-gather ./must-gather
 COPY version ./version
 COPY scripts ./scripts
 COPY files ./files
@@ -15,9 +16,12 @@ COPY pkg ./pkg
 
 RUN make build
 
+FROM registry.svc.ci.openshift.org/ocp/4.6:cli as origincli
+
 FROM centos:centos7
 RUN INSTALL_PKGS=" \
       openssl \
+      rsync \
       " && \
     yum install -y $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
@@ -26,15 +30,19 @@ RUN INSTALL_PKGS=" \
     chmod og+w /tmp/ocp-clo
 COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/bin/cluster-logging-operator /usr/bin/
 COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/scripts/* /usr/bin/scripts/
+
 RUN mkdir -p /usr/share/logging/
 COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/files/ /usr/share/logging/
 
 COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/manifests /manifests
 RUN rm /manifests/art.yaml
 
+COPY --from=origincli /usr/bin/oc /usr/bin/
+COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/must-gather/* /usr/bin/
+
 # this is required because the operator invokes a script as `bash scripts/cert_generation.sh`
 WORKDIR /usr/bin
-ENTRYPOINT ["/usr/bin/cluster-logging-operator"]
+CMD ["/usr/bin/cluster-logging-operator"]
 LABEL io.k8s.display-name="OpenShift cluster-logging-operator" \
       io.k8s.description="This is a component of OpenShift Container Platform that manages the lifecycle of the Aggregated logging stack." \
       io.openshift.tags="openshift,logging,cluster-logging" \
--- a/must-gather/README.md
+++ b/must-gather/README.md
@@ -1,0 +1,168 @@
+cluster-logging must-gather
+=================
+
+`cluster-logging-must-gather` is a tool built on top of [OpenShift must-gather](https://github.com/openshift/must-gather)
+that expands its capabilities to gather Openshift Cluster Logging information.
+
+### Usage
+```sh
+oc adm must-gather --image=quay.io/openshift/origin-cluster-logging-operator -- /usr/bin/gather
+```
+
+The command above will create a local directory with a dump of the cluster-logging state.
+Note that this command will only get data related to the cluster-logging part of the OpenShift cluster.
+
+You will get a dump of:
+- The openshift-logging namespace and its children objects
+- The openshift-operators-redhat namespace and its children objects
+- The cluster-logging install objects
+- All cluster-logging CRD's definitions
+- All nodes objects
+- All persistent volumes objects
+- Custom logs, configurations and health status per component, i.e. collection, logStore, curation, visualization
+
+In order to get data about other parts of the cluster (not specific to cluster-logging) you should
+run `oc adm must-gather` (without passing a custom image). Run `oc adm must-gather -h` to see more options.
+
+Example must-gather for cluster-logging output:
+```
+├── cluster-logging
+│  ├── clo
+│  │  ├── cluster-logging-operator-74dd5994f-6ttgt
+│  │  ├── clusterlogforwarder_cr
+│  │  ├── cr
+│  │  ├── csv
+│  │  ├── deployment
+│  │  └── logforwarding_cr
+│  ├── collector
+│  │  ├── fluentd-2tr64
+│  ├── curator
+│  │  └── curator-1596028500-zkz4s
+│  ├── eo
+│  │  ├── csv
+│  │  ├── deployment
+│  │  └── elasticsearch-operator-7dc7d97b9d-jb4r4
+│  ├── es
+│  │  ├── cluster-elasticsearch
+│  │  │  ├── aliases
+│  │  │  ├── health
+│  │  │  ├── indices
+│  │  │  ├── latest_documents.json
+│  │  │  ├── nodes
+│  │  │  ├── nodes_stats.json
+│  │  │  └── thread_pool
+│  │  ├── cr
+│  │  ├── elasticsearch-cdm-lp8l38m0-1-794d6dd989-4jxms
+│  │  └── logs
+│  │     ├── elasticsearch-cdm-lp8l38m0-1-794d6dd989-4jxms
+│  ├── install
+│  │  ├── co_logs
+│  │  ├── install_plan
+│  │  ├── olmo_logs
+│  │  └── subscription
+│  └── kibana
+│     ├── cr
+│     ├── kibana-9d69668d4-2rkvz
+├── cluster-scoped-resources
+│  └── core
+│     ├── nodes
+│     │  ├── ip-10-0-146-180.eu-west-1.compute.internal.yaml
+│     └── persistentvolumes
+│        ├── pvc-0a8d65d9-54aa-4c44-9ecc-33d9381e41c1.yaml
+├── event-filter.html
+├── gather-debug.log
+└── namespaces
+   ├── openshift-logging
+   │  ├── apps
+   │  │  ├── daemonsets.yaml
+   │  │  ├── deployments.yaml
+   │  │  ├── replicasets.yaml
+   │  │  └── statefulsets.yaml
+   │  ├── batch
+   │  │  ├── cronjobs.yaml
+   │  │  └── jobs.yaml
+   │  ├── core
+   │  │  ├── configmaps.yaml
+   │  │  ├── endpoints.yaml
+   │  │  ├── events
+   │  │  │  ├── curator-1596021300-wn2ks.162634ebf0055a94.yaml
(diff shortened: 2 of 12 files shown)


must-gather/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM quay.io/openshift/origin-cli:latest
```
jcantrill, line 1: @bparees do you know if there is a way for us to pull `oc` into the CLO image and maybe provide it as part of CLO?  @periklis can we provide an alternate entrypoint with must gather

The goal here is to not need to productize yet another image

must-gather/README.md
```diff
@@ -0,0 +1,25 @@
...
+`cluster-logging-must-gather` is a tool built on top of [OpenShift must-gather](https://github.com/openshift/must-gather)
+that expands its capabilities to gather Openshift Cluster Logging information.
+
+### Usage
+```sh
+oc adm must-gather --image=quay.io/cluster-logging-dev/cluster-logging-must-gather
+```
+
+The command above will create a local directory with a dump of the cluster-logging state.
+Note that this command will only get data related to the cluster-logging part of the OpenShift cluster.
```
jcantrill, line 13: What does this mean?  you won't get say info about nodes?

must-gather/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+You will get a dump of:
```
jcantrill, line 15: Maybe "A not all-inclusive example of what is retrieved:"  Trying to not identify what is perceived as an exact  list

must-gather/Dockerfile
```diff
@@ -0,0 +1,7 @@
...
+
+# copy all collection scripts to /usr/bin
+COPY collection-scripts/* /usr/bin/
```
jcantrill, line 4: why specifically `collection-scripts` and not just `scripts`

must-gather/collection-scripts/common
```diff
@@ -0,0 +1,21 @@
...
+  local ns=${3:-$NAMESPACE}
+  local pattern=${4:-"Dockerfile-*logging*"}
+  echo ---- Env for $pod
+  containers=$(oc -n $ns get po $pod -o jsonpath='{.spec.containers[*].name}')
+  for container in $containers
+  do
+    dockerfile=$(oc -n $ns exec $pod -c $container -- find /root/buildinfo -name $pattern || :)
+    if [ -n "$dockerfile" ]
+    then
+      echo Dockerfile info: $dockerfile > $env_file
```
jcantrill, line 15: Is there another term we can use then 'Dockerfile' given we no longer use Docker.  Maybe its "Image info"

must-gather/collection-scripts/gather_collection_resources
```diff
@@ -0,0 +1,60 @@
...
+NAMESPACE="openshift-logging"
+
+CLO_COLLECTION_PATH="$BASE_COLLECTION_PATH/cluster-logging"
+collector_folder="$CLO_COLLECTION_PATH/collector"
+
+check_collector_connectivity() {
+  local pod=$1
+  echo "--Connectivity between $pod and elasticsearch" >> $collector_folder/$pod
+
+  es_host=$(oc -n $NAMESPACE  get pod $pod  -o jsonpath='{.spec.containers[0].env[?(@.name=="ES_HOST")].value}')
```
jcantrill, line 22: I don't believe this is relevent any longer in light of LF

must-gather/README.md
periklis: In general the must-gather plugin image replaces the default must-gather image. Nevertheless, you can ask `oc adm inspect` to get any standard k8s resource, which we do in the entrypoint `gather` script. Same for PVs, etc.

must-gather/Dockerfile
periklis: As in origin: https://github.com/openshift/must-gather/ I don't know if it is mandatory.

must-gather/Dockerfile
periklis: AFAICS, other operators just publish their must-gather images on developer repos on quay.io, e.g.:
https://github.com/openshift/ocs-operator/tree/master/must-gather#usage
https://github.com/openshift/ocs-operator/tree/master/must-gather#usage

You can do the same w/o productization, can't we?

must-gather/collection-scripts/gather_collection_resources
periklis: @ewolinetz WDYT? I transfered all these from logging-dump.

must-gather/Dockerfile
jcantrill: The discussion we had with thin the arch group was to include this as part of the CLO.  Suggestion was to base the CLO image on `quay.io/openshift/origin-cli:latest` and find a way to invoke an alternate entrypoint or use whatever convention that is provided for must gather.  Is there any reason NOT to do this? @alanconway

must-gather/Dockerfile
jcantrill: I don't have a firm preference here but it seems verbose

must-gather/Dockerfile
periklis: @jcantrill This is exactly what we have right now. Or do you see any difference.

ewolinetz: >  Suggestion was to base the CLO image on quay.io/openshift/origin-cli:latest

it is different than what it suggested here in that we wouldn't have a second image instead it would be invoked from the CLO image rather than this `must-gather` one

must-gather/collection-scripts/gather_collection_resources
ewolinetz: if we gather the LF CR then we should have this information (though it is technically still TP). logging-dump is still a little out of date with respect to our current deployment processes, i had updated it simply to gather additional information that we required for debugging -- no addressing of stale functionality

must-gather/collection-scripts/gather_elasticsearch_operator_resources
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
ewolinetz, line 1: this should likely live in the EO's own must-gather and CLO could call into that

must-gather/collection-scripts/gather_logstore_resources
```diff
@@ -0,0 +1,120 @@
+#!/bin/bash
```
ewolinetz, line 1: same comment as above

must-gather/collection-scripts/gather_visualization_resources
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
```
ewolinetz, line 1: same comment as above

must-gather/collection-scripts/gather_elasticsearch_operator_resources
jcantrill: While the separation would be nice, I don't see user's separating EO from CLO in this respect; it's just logging to them.  EO is not an independent product so we should just point user's to CLO and make a change in the future if EO ever truly is more independent.

ewolinetz: so what about the Jaeger case?

periklis: I favour for now a cluster-logging `must-gather` that encompasses the whole stack in one single command. This simplifies retiring logging-dump.sh for the start with more recognizable alternative. IMHO we can re-think what is a good must-gather for an independent elasticsearch-operator later. A little bit of overlap between both shouldn't hurt.

Regarding the jaeger case, I would say there should a jaeger must-gather for the tracing stack than encompasses elasticsearch resources the way they use them.

must-gather/collection-scripts/gather_logstore_resources
periklis: Resolving this comment in favour of eo-thread above.

must-gather/collection-scripts/gather_visualization_resources
periklis: Resolving this comment in favour of eo-thread above.

must-gather/collection-scripts/gather_collection_resources
periklis: I will keep this as is for now and we can wipe out stale things in a separate PR.

must-gather/Dockerfile
periklis: Will keep as is for now although more verbose to allow people recognize the structure and amend, e.g. third-party contributors.

must-gather/Dockerfile
periklis: I adapted the packaging to add everything in the single operator image. You can try it like this:
```
oc adm must-gather --image=quay.io/periklis/cluster-logging-operator:latest -- /usr/bin/gather
```
WDYT? (cc @jcantrill @ewolinetz)

Dockerfile
```diff
@@ -15,9 +16,12 @@ COPY pkg ./pkg
 
 RUN make build
 
+FROM quay.io/openshift/origin-cli:latest as origincli
```
periklis, line 19: @jcantrill / @ewolinetz Do you know if this needs an amendment in `ocp-build-data` for ART?

must-gather/Dockerfile
ewolinetz: my only concern with that is we would become an outlier on how to invoke `oc adm must-gather` vs how other teams would be called... 

unless there are other teams out there that would have do something similar, i would rather it be invoked like:
```
oc adm must-gather --image=quay.io/periklis/cluster-logging-operator:latest
```

and then we update how we invoke the normal operator process within our deployment (the entrypoint can be specified in that definition)

periklis: That's the point, you can not define the entrypoint via `oc adm must-gather`. You can only override the cmd via `-- /usr/bin/gather`. I believe this is ok, because user will just copy and paste the full oc-adm command from docs.

must-gather/collection-scripts/gather_elasticsearch_operator_resources
ewolinetz: thats fine, i personally don't think its a little overlap between what jaeger would want and CLO would want (regarding EO). if we think about it as "a customer has jaeger installed but has an issue with EO" i feel we would want to control what sort of information is gathered as part of troubleshooting it. yes, the jaeger team can always add more on top of that, but my original point was more about ensuring the things that are specific to debugging EO on its own should live in an EO provided `must-gather`:

* install plans
* CSVs
* CRs
* EO logs
* ES logs

must-gather/Dockerfile
jcantrill: I believe the fact it is an outlier is insignificant as long as we document it.  Documentation is significantly easier then having to setup the release train.  Also, users then have the scripts instead of having to go and find the image

must-gather/collection-scripts/gather_elasticsearch_operator_resources
jcantrill: Let's be clear, at the moment there are no EO-only customer's only.  We are not supporting an independent use of EO outside the logging and jaeger use case.  If we ever get to that point then we will make that change

Dockerfile
jcantrill: yes I believe it does

jcantrill: Is this how this works, FROM layers are munged together?

periklis: No FROM layers build different overlays. Directives in between are optional. Think of having multiple FROM directives on the top like branches that you want to merge by cherry-pick in a single final overlay.

PR #629, opened 2020-07-29
Updating CLO to use operator bundle

ewolinetz:
Generated using `make generate-bundle`

To address changes required as part of https://issues.redhat.com/browse/OLM-1810


ewolinetz:
@jcantrill i don't think we want to. the make target i have references the manifest dir for building the bundles and i'd like to wait and make sure this all gets fleshed out correctly before we switch to operator bundles entirely

PR #631, opened 2020-07-29
LOG-825: Added cluster wide tags

igor-karpukhin:
Added cluster-wide tags for LogForwarding

### DIFF
--- a/manifests/4.6/logforwardings.crd.yaml
+++ b/manifests/4.6/logforwardings.crd.yaml
@@ -72,6 +72,9 @@ spec:
                       - logs.app
                       - logs.infra
                       - logs.audit
+                  labels:
+                    type: map
+                    description: Pipeline labels in the key:value format
               required:
               - name
               - source
--- a/pkg/apis/logging/v1/cluster_log_forwarder_types.go
+++ b/pkg/apis/logging/v1/cluster_log_forwarder_types.go
@@ -188,6 +188,10 @@ type PipelineSpec struct {
 	// +required
 	InputRefs []string `json:"inputRefs"`
 
+	// Labels lists labels applied to this pipeline
+	//
+	// +optional
+	Labels map[string]string `json:"labels,omitempty"`
 	// Name is optional, but must be unique in the `pipelines` list if provided.
 	//
 	// +optional
--- a/pkg/generators/forwarding/fluentd/generators.go
+++ b/pkg/generators/forwarding/fluentd/generators.go
@@ -1,6 +1,7 @@
 package fluentd
 
 import (
+	"encoding/json"
 	"fmt"
 	"sort"
 	"text/template"
@@ -122,9 +123,9 @@ func (engine *ConfigGenerator) Generate(clfSpec *logging.ClusterLogForwarderSpec
 	}{
 		engine.includeLegacyForwardConfig,
 		engine.includeLegacySyslogConfig,
-		inputs.Has(string(logging.InputNameInfrastructure)),
-		inputs.Has(string(logging.InputNameApplication)),
-		inputs.Has(string(logging.InputNameAudit)),
+		inputs.Has(logging.InputNameInfrastructure),
+		inputs.Has(logging.InputNameApplication),
+		inputs.Has(logging.InputNameAudit),
 		sourceInputLabels,
 		sourceToPipelineLabels,
 		pipelineToOutputLabels,
@@ -173,7 +174,7 @@ func inputsToPipelines(pipelines []logging.PipelineSpec) logging.RouteMap {
 
 //generateSourceToPipelineLabels generates fluentd label stanzas to fan source types to multiple pipelines
 func (engine *ConfigGenerator) generateSourceToPipelineLabels(sourcesToPipelines logging.RouteMap) ([]string, error) {
-	configs := []string{}
+	var configs []string
 	for sourceType, pipelineNames := range sourcesToPipelines {
 		data := struct {
 			IncludeLegacySecureForward bool
@@ -183,7 +184,7 @@ func (engine *ConfigGenerator) generateSourceToPipelineLabels(sourcesToPipelines
 		}{
 			engine.includeLegacyForwardConfig,
 			engine.includeLegacySyslogConfig,
-			string(sourceType),
+			sourceType,
 			pipelineNames.List(),
 		}
 		result, err := engine.Execute("sourceToPipelineCopyTemplate", data)
@@ -196,14 +197,26 @@ func (engine *ConfigGenerator) generateSourceToPipelineLabels(sourcesToPipelines
 }
 
 func (engine *ConfigGenerator) generatePipelineToOutputLabels(pipelines []logging.PipelineSpec) ([]string, error) {
-	configs := []string{}
+	var configs []string
 	for _, pipeline := range pipelines {
+		var jsonLabels string
+
+		if pipeline.Labels != nil {
+			marshalledLabels, err := json.Marshal(pipeline.Labels)
+			if err != nil {
+				return nil, fmt.Errorf("unable to marshal pipeline labels: %v", err)
+			}
+			jsonLabels = string(marshalledLabels)
+		}
+
 		data := struct {
-			Name    string
-			Outputs []string
+			Name           string
+			Outputs        []string
+			PipelineLabels string
 		}{
 			pipeline.Name,
 			pipeline.OutputRefs,
+			jsonLabels,
 		}
 		result, err := engine.Execute("pipelineToOutputCopyTemplate", data)
 		if err != nil {
@@ -227,7 +240,7 @@ func (engine *ConfigGenerator) generatePipelineToOutputLabels(pipelines []loggin
 //  </match>
 // </label>
 func (engine *ConfigGenerator) generateOutputLabelBlocks(outputs []logging.OutputSpec, outputConf *logging.ForwarderSpec) ([]string, error) {
-	configs := []string{}
+	var configs []string
 	logger.Debugf("Evaluating %v forwarder logging...", len(outputs))
 	for _, output := range outputs {
 		logger.Debugf("Generate output type %v", output.Type)
--- a/pkg/generators/forwarding/fluentd/generators_suite_test.go
+++ b/pkg/generators/forwarding/fluentd/generators_suite_test.go
@@ -1,0 +1,13 @@
+package fluentd
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestGenerator(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Generating pipeline to output labels")
+}
--- a/pkg/generators/forwarding/fluentd/generators_test.go
+++ b/pkg/generators/forwarding/fluentd/generators_test.go
@@ -1,0 +1,168 @@
+package fluentd
+
+import (
+	"text/template"
+
+	. "github.com/openshift/cluster-logging-operator/test"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/pkg/generators"
+)
+
+var _ = Describe("Generating pipeline to output labels", func() {
+	var (
+		configGenerator *ConfigGenerator
+	)
+	BeforeEach(func() {
+		engn, err := generators.New("OutputLabelConf",
+			&template.FuncMap{
+				"labelName":           labelName,
+				"sourceTypelabelName": sourceTypeLabelName,
+			},
+			templateRegistry...)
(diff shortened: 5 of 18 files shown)


pkg/apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -188,6 +188,9 @@ type PipelineSpec struct {
 	// +required
 	InputRefs []string `json:"inputRefs"`
 
+	// Labels lists labels applied to this pipeline
+	// +optional
+	Labels map[string]string `json:"labels"`
```
jcantrill, line 193: I see this results in `labels: null`  @periklis  @alanconway  doesn't `json:"labels,omitempty"` make it go away during serialization if empty?

igor-karpukhin: Fixed

pkg/apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -188,6 +188,9 @@ type PipelineSpec struct {
 	// +required
 	InputRefs []string `json:"inputRefs"`
 
+	// Labels lists labels applied to this pipeline
+	// +optional
```
vimalk78, line 192: nit: add empty line here, same as for other fields

pkg/apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -188,6 +188,9 @@ type PipelineSpec struct {
...
+	Labels map[string]string `json:"labels,omitempty"`
```
vimalk78, line 193: is there a syntax or character set for these labels?

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -492,6 +492,14 @@ const sourceToPipelineCopyTemplate = `{{- define "sourceToPipelineCopyTemplate"
 
 const pipelineToOutputCopyTemplate = `{{- define "pipelineToOutputCopyTemplate" -}}
 <label {{labelName .Name}}>
+  {{ if .PipelineLabels }}
```
vimalk78, line 495: it is better to end the template code with an `-`, this doesnt add empty space if there are no labels. e.g.
` {{ if .PipelineLabels -}}`

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -492,6 +492,14 @@ const sourceToPipelineCopyTemplate = `{{- define "sourceToPipelineCopyTemplate"
...
+  <filter **>
+    @type record_transformer
+    <record>
+      openshift { "labels": {{.PipelineLabels}} }
+    </record>
+  </filter>
+  {{ end }}
```
vimalk78, line 502: same as above.
`{{ end -}}`

pkg/apis/logging/v1/cluster_log_forwarder_types.go
igor-karpukhin: Nope, it's just like labels for `pods`, a simple `key:value` pairs

pkg/generators/forwarding/fluentd/templates.go
igor-karpukhin: But it doesn't add an empty line even if there are labels. I see that this output:

pkg/generators/forwarding/fluentd/templates.go
igor-karpukhin: Fixed

pkg/generators/forwarding/fluentd/templates.go
igor-karpukhin: Fixed

pkg/apis/logging/v1/cluster_log_forwarder_types.go
igor-karpukhin: Fixed

igor-karpukhin:
@vimalk78 labels are added on top level like this: 
```
{
  "message" : "2020-07-30 08:04:21,765 - SVTLogger - INFO",
  "@timestamp" : "2020-07-30T08:04:21.996384+00:00",
  "level" : "unknown",
  "hostname" : "ip-10-0-153-186.us-east-2.compute.internal",

  "openshift" : {
    "labels" : {
      "label1" : "value1",
      "label2" : "value-2"
    }
  }

```

manifests/4.6/logforwardings.crd.yaml
```diff
@@ -72,6 +72,9 @@ spec:
                       - logs.app
                       - logs.infra
                       - logs.audit
+                  labels:
+                    type: map
+                    description: Pipeline labels in the same format as medata.labels
```
periklis, line 77: Does declaring labels having the same format as `metadata.labels` imply the same constraints? If yes, we should add some validation on this field (See k8s metadata labels contraints here: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set)

If not, just remove the reference `metadata.labels` and re-phase to `key value pairs`

pkg/generators/forwarding/fluentd/generators.go
```diff
@@ -196,14 +197,26 @@ func (engine *ConfigGenerator) generateSourceToPipelineLabels(sourcesToPipelines
 }
 
 func (engine *ConfigGenerator) generatePipelineToOutputLabels(pipelines []logging.PipelineSpec) ([]string, error) {
-	configs := []string{}
+	var configs []string
 	for _, pipeline := range pipelines {
+		var jsnLabels string
```
periklis, line 202: nit. jsn -> json

pkg/generators/forwarding/fluentd/generators.go
```diff
@@ -196,14 +197,26 @@ func (engine *ConfigGenerator) generateSourceToPipelineLabels(sourcesToPipelines
...
+
+		if pipeline.Labels != nil {
+			jl, err := json.Marshal(pipeline.Labels)
+			if err != nil {
+				return nil, err
```
periklis, line 207: Wrapping the error with a nice message `fmt.Errorf("failed marshaling pipelines labels: %s", err)` would make the error side easy to identify.

pkg/generators/forwarding/fluentd/generators_test.go
```diff
@@ -0,0 +1,235 @@
...
+	"reflect"
+	"strings"
+	"testing"
+	"text/template"
+
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/pkg/generators"
+)
+
+func TestConfigGenerator_generatePipelineToOutputLabels(t *testing.T) {
```
periklis, line 13: Golang tests don't use snake case. Propose to rename to: `TestGeneratePipelineToOutputLabels`. The test file `generators_test.go` is a good enough identifier w/o the need to duplicate information like in the `TestConfigGenerator` prefix.

pkg/generators/forwarding/fluentd/generators_test.go
```diff
@@ -0,0 +1,235 @@
+package fluentd
```
periklis, line 1: Please rewrite this test in ginkgo/gomega style as the whole package is using this style. It has the benefits that you can use the more advanced config parsing matcher `EqualTrimedLines`. It integrates the test in this package suite.

pkg/generators/forwarding/fluentd/generators_test.go
igor-karpukhin: https://golang.org/pkg/testing/
```
Multiple example functions for a package/type/function/method may be provided by appending a distinct suffix to the name. The suffix must start with a lower-case letter.

func Example_suffix() { ... }
func ExampleF_suffix() { ... }
func ExampleT_suffix() { ... }
func ExampleT_M_suffix() { ... }
```
So I don't think it breaks go test naming style. It is widely used, ex: https://github.com/hashicorp/consul/blob/master/api/agent_test.go#L170

pkg/generators/forwarding/fluentd/generators_test.go
igor-karpukhin: Fixed

pkg/generators/forwarding/fluentd/generators.go
igor-karpukhin: Fixed

pkg/generators/forwarding/fluentd/generators.go
igor-karpukhin: Fixed

manifests/4.6/logforwardings.crd.yaml
igor-karpukhin: Fixed

pkg/generators/forwarding/fluentd/generators_suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestGenerator(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Generating pipeline to output labels")
+}
```
periklis, line 13: AFAIK, this is obsolete, because a `fluentd_suite_test.go` already exists in this package.

pkg/generators/forwarding/fluentd/generators_test.go
```diff
@@ -0,0 +1,166 @@
...
+			{
+				InputRefs:  []string{logging.InputNameInfrastructure},
+				OutputRefs: []string{"infra-es"},
+				Labels:     map[string]string{"1": "2", "3": "4"},
+				Name:       "Pipeline-1",
+			},
+		}
+		got, err := configGenerator.generatePipelineToOutputLabels(pipelines)
+		Expect(err).To(BeNil())
+		Expect(got[0]).To(BeEquivalentTo(`<label @PIPELINE_1>
```
periklis, line 99: Consider using `Expect(results).To(EqualTrimLines(....))` from `"github.com/openshift/cluster-logging-operator/test"`. The diff is more concise and you can indent the expected literal string inside our code.

test/e2e/logforwarding/fluent/forward_to_fluent_test.go
```diff
@@ -177,6 +177,99 @@ var _ = Describe("[ClusterLogForwarder] Forwards logs", func() {
...
+			})
+
+			It("should send logs to the forward.Output logstore", func() {
+				name := fluentDeployment.GetName()
+				Expect(e2e.LogStores[name].HasInfraStructureLogs(helpers.DefaultWaitForLogsTimeout)).To(BeTrue(), "Expected to find stored infrastructure logs")
+				Expect(e2e.LogStores[name].HasApplicationLogs(helpers.DefaultWaitForLogsTimeout)).To(BeTrue(), "Expected to find stored application logs")
+				Expect(e2e.LogStores[name].HasAuditLogs(helpers.DefaultWaitForLogsTimeout)).To(BeTrue(), "Expected to find stored audit logs")
+			})
+
+			It("should contain all pipeline labels", func() {
```
periklis, line 260: Consider merging the two`It` blocks, because the `BeforeEach` will run for each `It` block again.

test/helpers/framework.go
```diff
@@ -64,6 +64,8 @@ type LogStore interface {
 
 	GrepLogs(expr string, timeToWait time.Duration) (string, error)
 
+	RetrieveLogs() (map[string]string, error)
+
```
periklis, line 68: Why do we need another log retrieval function on this interface? This blows up the interface. Consider using `ApplicationLogs` that returns a typed `logs` slice.

pkg/generators/forwarding/fluentd/generators_test.go
igor-karpukhin: The diff is more concise, I admit. But `EqualTrimLines` compares strings, not arrays of strings

periklis: Ah good to know :+1:

alanconway: EqualTrimLines internally converts the multi-line string to a slice of lines and does the diff on the slice. Gingko matchers take their values as interface{}, so we could easily change EqualTrimLines to work for both strings and slices of strings.

test/e2e/logforwarding/fluent/forward_to_fluent_test.go
igor-karpukhin: Fixed

pkg/generators/forwarding/fluentd/generators_test.go
igor-karpukhin: Fixed

pkg/generators/forwarding/fluentd/generators_suite_test.go
jcantrill: yes. i believe we only need one suite file

test/helpers/framework.go
jcantrill: Maybe a comment is in order here.  what does this provide that can not be fetched from one of the other methods?

PR #632, opened 2020-07-29
LOG-804: add metrics dashboard

blockloop:
This PR adds a new metrics dashboard called Openshift Logging

/cc @ewolinetz @lukas-vlcek 

image

### DIFF
--- a/pkg/k8shandler/dashboards.go
+++ b/pkg/k8shandler/dashboards.go
@@ -1,0 +1,32 @@
+package k8shandler
+
+import (
+	"path"
+
+	"github.com/openshift/cluster-logging-operator/pkg/utils"
+)
+
+const (
+	clusterLoggingDashboardFile = "dashboards/openshift-logging-dashboard.json"
+)
+
+// CreateOrUpdateDashboards reconciles metrics dashboards component for cluster logging
+func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateDashboards() (err error) {
+	spec := string(utils.GetFileContents(path.Join(utils.GetShareDir(), clusterLoggingDashboardFile)))
+
+	cm := NewConfigMap("grafana-dashboard-cluster-logging",
+		"openshift-config-managed",
+		map[string]string{
+			"openshift-logging.json": spec,
+		})
+	if cm.Labels == nil {
+		cm.Labels = map[string]string{}
+	}
+	cm.Labels["console.openshift.io/dashboard"] = "true"
+
+	if err := clusterRequest.CreateOrUpdateConfigMap(cm); err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/k8shandler/reconciler.go
+++ b/pkg/k8shandler/reconciler.go
@@ -54,6 +54,11 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
 		return fmt.Errorf("Unable to create or update collection for %q: %v", clusterLoggingRequest.cluster.Name, err)
 	}
 
+	// Reconcile Metrics Dashboards
+	if err = clusterLoggingRequest.CreateOrUpdateDashboards(); err != nil {
+		return fmt.Errorf("Unable to create or update metrics dashboards for %q: %w", clusterLoggingRequest.cluster.Name, err)
+	}
+
 	return nil
 }
 
--- a/files/dashboards/openshift-logging-dashboard.json
+++ b/files/dashboards/openshift-logging-dashboard.json
@@ -1,0 +1,1759 @@
+{
+  "__inputs": [],
+  "__requires": [ ],
+  "annotations": {
+    "list": [ ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "id": null,
+  "iteration": 1594321382217,
+  "links": [],
+  "panels": [
+    {
+      "collapsed": false,
+      "datasource": null,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 0
+      },
+      "id": 14,
+      "panels": [],
+      "title": "Overview",
+      "type": "row"
+    },
+    {
+      "cacheTimeout": null,
+      "colorBackground": true,
+      "colorValue": false,
+      "colors": [
+        "#37872D",
+        "#37872D",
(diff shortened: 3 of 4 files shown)


ewolinetz:
/approve
out of curiosity how did we generate the dashboard file?

ewolinetz:
is it possible to also attach a screenshot of the dashboard in the summary?

pkg/k8shandler/reconciler.go
```diff
@@ -44,6 +44,11 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
 		return fmt.Errorf("Unable to create or update visualization for %q: %v", clusterLoggingRequest.cluster.Name, err)
 	}
 
+	// Reconcile Dashboards
+	if err = clusterLoggingRequest.CreateOrUpdateVisualization(proxyConfig); err != nil {
+		return fmt.Errorf("Unable to create or update visualization for %q: %v", clusterLoggingRequest.cluster.Name, err)
+	}
+
```
ewolinetz, line 51: copy pasta?

pkg/k8shandler/dashboards.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"github.com/openshift/cluster-logging-operator/pkg/utils"
+)
+
+const (
+	clusterLoggingDashboardFile = "dashboards/openshift-logging-dashboard.json"
+)
+
+// CreateOrUpdateDashboards reconciles metrics dashboards component for cluster logging
+func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateDashboards() (err error) {
```
jcantrill, line 16: Let's use the word "ReconcileDashboards" since thats what we are doing

pkg/k8shandler/dashboards.go
```diff
@@ -0,0 +1,45 @@
...
+	if err = clusterRequest.createOrUpdateClusterLoggingDashboard(); err != nil {
+		return
+	}
+	return nil
+}
+
+// createOrUpdateClusterLoggingDashboard creates/updates the cluster logging dashboard ConfigMap
+func (clusterRequest *ClusterLoggingRequest) createOrUpdateClusterLoggingDashboard() error {
```
jcantrill, line 24: why is this split from the public method?

pkg/k8shandler/dashboards.go
```diff
@@ -0,0 +1,45 @@
...
+	b, err := ioutil.ReadFile(path.Join(utils.GetShareDir(), clusterLoggingDashboardFile))
```
jcantrill, line 25: where is the dashboard file?  there is a util function to grab files from <root>/files.  Ref what we do with fluent/run.sh

pkg/k8shandler/reconciler.go
blockloop: doh. Fixed

pkg/k8shandler/dashboards.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to read dashboard file: %w", err)
+	}
+
+	cm := NewConfigMap("grafana-dashboard-cluster-logging",
```
jcantrill, line 30: consider naming this just "cluster-logging-dashboard" or "dashboard-cluster-logging"

blockloop: All dashboards in that namespace are formatted like this `grafana-dashboard-<name>`

pkg/k8shandler/dashboards.go
blockloop: The util function searches a different dir than this ends up. I believe the util function you're talking about is `GetWorkingDirFileContents`? If so, that's what I tried originally and it searched the wrong directory.

pkg/k8shandler/dashboards.go
blockloop: I literally copied this file from visualization.go and modified it's contents.

pkg/k8shandler/dashboards.go
blockloop: The convention in reconciler.go is CreateOrUpdate<Resource> shouldn't we follow that convention? All of the other resources are CreateOrUpdate. I agree that Reconcile is probably better, but if I'm going to rename one I think we should rename them all.

pkg/k8shandler/dashboards.go
ewolinetz: https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/fluentd.go#L213

pkg/k8shandler/dashboards.go
ewolinetz: do we anticipate having additional dashboards?

blockloop:
> /approve
> out of curiosity how did we generate the dashboard file?

I deployed a new instance of grafana in-cluster. Then I created a dashboard and exported the JSON and made required changes for the dashboard. Once you have the exported JSON you put them in a configmap in the specified namespace.

periklis:
> > /approve
> > out of curiosity how did we generate the dashboard file?
> 
> I deployed a new instance of grafana in-cluster. Then I created a dashboard and exported the JSON and made required changes for the dashboard. Once you have the exported JSON you put them in a configmap in the specified namespace.

Does this also mean that we can re-import the JSON into grafana and make changes, e.g. fix bugs?

blockloop:
@periklis I share all of your concerns, however they are all problems with limitations on the console dashboards. For example, the cluster status 0=Green, 1=Yellow, 2=Red was mapped in the Grafana dashboard but it does not apply in the console dashboards. There is no way for us to accomplish this in the console dashboards today. 

As for the rest, the labels aren't simply mapped from Grafana and might require some fancy work if it's even possible. I'd like to revisit this later and figure out some better options. Those can be followups.

ewolinetz:
@periklis another thing to consider regarding this is it can be a follow up "bug fix" to address the dashboard

periklis:
@blockloop @ewolinetz ACK, I am fine with any follow-up approach you pick. The list above is more of the sort how to improve the whole dashboard. Worked for years in operations team and I myself swear to those things like unified units, legends, etc.

pkg/k8shandler/dashboards.go
blockloop: Not that I'm aware of, but potentially?

PR #635, opened 2020-07-30
Pin development tools with bingo

periklis:
### Description
This PR pins our development tools using [bingo](https://github.com/bwplotka/bingo). In detail:
- Each tool is built from a separate go modules file (e.g. `.bingo/operator-sdk.mod`).
- The tools are provided from a single `$GOBIN` directory on demand
- For each tool a specific make target and a shell variable is exported for downstream scripts. 

**Notes:** Requires Golang **1.14**, depends on #638, openshift/release#10631


.bingo/Variables.mk
```diff
@@ -0,0 +1,42 @@
...
+$(GOLANGCI_LINT): .bingo/golangci-lint.mod
+	@# Install binary/ries using Go 1.14+ build command. This is using bwplotka/bingo-controlled, separate go module with pinned dependencies.
+	@echo "(re)installing $(GOBIN)/golangci-lint-v1.27.0"
+	@cd .bingo && $(GO) build -mod=mod  -modfile=golangci-lint.mod -o=$(GOBIN)/golangci-lint-v1.27.0 "github.com/golangci/golangci-lint/cmd/golangci-lint"
+
+OPERATOR_SDK := $(GOBIN)/operator-sdk-v0.18.2
+$(OPERATOR_SDK): .bingo/operator-sdk.mod
+	@# Install binary/ries using Go 1.14+ build command. This is using bwplotka/bingo-controlled, separate go module with pinned dependencies.
+	@echo "(re)installing $(GOBIN)/operator-sdk-v0.18.2"
+	@cd .bingo && $(GO) build -mod=mod -modfile=operator-sdk.mod -o=$(GOBIN)/operator-sdk-v0.18.2 "github.com/operator-framework/operator-sdk/cmd/operator-sdk"
```
bwplotka, line 35: so I guess you manually changed it to `-mod=mod` right? What you changed here is not tool module vendoring II think

>  If invoked with -mod=mod, the go command loads modules from the module cache
even if there is a vendor directory present.

So, if you never installed this tool before on some CI with empty cache, those will be fetched I guess?

Is this option changes anything then?

alanconway:
/lgtm
/approve
Its a step in the right direction IMO, we can improve it as we go.

.bingo/Variables.mk
periklis: Indeed it is, because it tells go build to not look in my vendor dir and use the cache instead.

pkg/utils/utils.go
```diff
@@ -1,6 +1,7 @@
 package utils
 
 import (
+	// #nosec G501
```
alanconway, line 4: What does this mean?

alanconway:
/lgtm
Hah - got you with the enforced linter did I?

pkg/utils/utils.go
periklis: Disables rule #G501 from: https://github.com/securego/gosec#available-rules

alanconway:
/lgtm
Looks better every time ;)

PR #636, opened 2020-07-30
[wip] Refactor k8 types into packages

jcantrill:
Refactor code into small packages


pkg/k8shandler/factory/cronjob.go
```diff
@@ -0,0 +1,26 @@
+package factory
```
ewolinetz, line 1: this isn't really a factory in the normal paradigm, is it really necessary to break this out from the other `cronjob.go` file?

PR #637, opened 2020-07-31
[release-4.4] Bug 1852639: fixing kibana not rolling out with updated secret

ewolinetz:
Manual cherry-pick of https://github.com/openshift/elasticsearch-operator/pull/392

Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1852639


ewolinetz:
@bparees what should the approach be here for the bz dependency? this is already fixed in our later releases so it seems odd to open a bz against them...

bparees:
> @bparees what should the approach be here for the bz dependency? this is already fixed in our later releases so it seems odd to open a bz against them...

it's already fixed in 4.5 but there was no bug needed to fix it in 4.5?  (it got fixed as part of some non-bug PR?)

bparees:
The strict rule is to open newer release BZs anyway so someone can explicitly confirm the problem does not exist in the newer releases (the BZ will only be used to verify, not to deliver code)

ewolinetz:
@bparees i updated the bz this fixes to depend on the bz that the PR i'm cherrypicking from addressed... not sure why i didn't just think to do that before

PR #639, opened 2020-08-03
Trivial change to make linter happy.

alanconway:
Note: the "consider preallocating" linter warns if a slice variable is declared using `var x []foo`. If you know up front an upper bound for the size of the slice you can pre-allocate with `x := make([]foo, maxlen)` which is more efficient.  If you are deliberately starting with a 0 length slice use `x := []foo{}` which is equivalent to the `var` form but satisfies the linter.

/cc @jcantrill


alanconway:
@jcantrill This one is definitely a CI problem - the change is trivial lint fixing.

PR #645, opened 2020-08-04
Add CRD printer columns and categories

periklis:
### Description
This PR declares a grouping category namely `logging` for all custom resources managed by the cluster-logging-operator. In addition it enhances the `ClusterLogging` CR with inforative printer columns, e.g. management stated, etc.

### Example output
image

/cc @jcantrill @alanconway


jcantrill:
> ### Description

> 
> image
> 

I like where this is going but what is the value of identifying the component type or name for clusterlogging instance?  The management state makes sense but the other columns seem to provide limited value

periklis:
> > ### Description
> 
> > image
> 
> I like where this is going but what is the value of identifying the component type or name for clusterlogging instance? The management state makes sense but the other columns seem to provide limited value

Although we don't support multiple collectors/logstores/etc. yet or ever will, the columns reflect which type is in use currently. In sort being informative as `The ClusterLogging stack installed runs A, B, C, D`.

periklis:
@jcantrill @alanconway I removed the extra printer columns except for Management State. Can we move this PR forward?

bundle/manifests/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -221,6 +224,11 @@ spec:
                       items:
                         type: string
                       type: array
+                    labels:
```
alanconway, line 227: Looks like this crossed over from the message labeling PR?

bundle/manifests/logging.openshift.io_clusterloggings_crd.yaml
```diff
@@ -3,8 +3,14 @@ kind: CustomResourceDefinition
 metadata:
   name: clusterloggings.logging.openshift.io
 spec:
+  additionalPrinterColumns:
+  - JSONPath: .spec.managementState
```
alanconway, line 7: What is this?

manifests/4.6/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
 spec:
   group: logging.openshift.io
   names:
+    categories:
+    - logging
```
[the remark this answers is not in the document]
periklis: This is a category to get all CRs with a simple: `oc get logging`. See screenshot above.

bundle/manifests/logging.openshift.io_clusterlogforwarders_crd.yaml
periklis: Actually this is not intended to be part of this PR, but until the bundle generation is bound to CRD generation, we need to update the bundle with `make generate-bundle` manually.

bundle/manifests/logging.openshift.io_clusterloggings_crd.yaml
periklis: See screenshot above. It exposes the spec field as part of `oc get clusterlogging instance`. It enables to check the CRs state without doing an `oc describe clusterlogging instance`.

alanconway:
/hold
spinning on retest

periklis:
@alanconway Could reconsider adding the lgtm label. I had to rebase this PR on top of the cluster-scope for ClusterLogForwarder.

igor-karpukhin (approved):
/lgtm

PR #646, opened 2020-08-04
Bug 1866020: alert on error rate %

blockloop:
FluentD error alerts are now based on a % instead of a number to account
for high ingestion rates.

This PR also adds an ability for CLO to update alerting rules.

/cc @ewolinetz @periklis @lukas-vlcek

### DIFF
--- a/files/fluentd/fluentd_prometheus_alerts.yaml
+++ b/files/fluentd/fluentd_prometheus_alerts.yaml
@@ -31,13 +31,34 @@
     "labels":
       "service": "fluentd"
       "severity": "critical"
-  - "alert": "FluentdErrorsHigh"
-    "annotations":
-      "message": "In the last minute, {{ $value }} errors reported by fluentd {{ $labels.instance }}."
-      "summary": "Fluentd reports high number of errors"
-    "expr": |
-      sum by(instance, job) (rate(fluentd_output_status_num_errors[1m])) > 10
-    "for": "1m"
-    "labels":
-      "service": "fluentd"
-      "severity": "critical"
+
+  - alert: FluentDHighErrorRate
+    annotations:
+      message: |-
+        {{ $value }}% of records have resulted in an error by fluentd {{ $labels.instance }}.
+      summary: FluentD output errors are high
+    expr: |
+      100 * (
+        sum by(instance)(rate(fluentd_output_status_num_errors[2m]))
+      /
+        sum by(instance)(rate(fluentd_output_status_emit_records[2m]))
+      ) > 10
+    for: 15m
+    labels:
+      severity: warning
+
+  - alert: FluentDVeryHighErrorRate
+    annotations:
+      message: |-
+        {{ $value }}% of records have resulted in an error by fluentd {{ $labels.instance }}.
+      summary: FluentD output errors are very high
+    expr: |
+      100 * (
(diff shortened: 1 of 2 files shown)


files/fluentd/fluentd_prometheus_alerts.yaml
```diff
@@ -31,13 +31,26 @@
...
-      sum by(instance, job) (rate(fluentd_output_status_num_errors[1m])) > 10
-    "for": "1m"
-    "labels":
-      "service": "fluentd"
-      "severity": "critical"
+
+  - alert: FluentDHighErrorRate
+    annotations:
+      message: |-
+        FluentD has a high error rate
```
ewolinetz, line 38: so before we had more information in our messages, why do you feel its unnecessary?

files/fluentd/fluentd_prometheus_alerts.yaml
```diff
@@ -31,13 +31,26 @@
     "labels":
       "service": "fluentd"
       "severity": "critical"
-  - "alert": "FluentdErrorsHigh"
-    "annotations":
-      "message": "In the last minute, {{ $value }} errors reported by fluentd {{ $labels.instance }}."
-      "summary": "Fluentd reports high number of errors"
-    "expr": |
-      sum by(instance, job) (rate(fluentd_output_status_num_errors[1m])) > 10
```
ewolinetz, line 39: can you talk a little bit about the difference between the two expressions (this one vs your new one)?

what is `sum by(instance, job)` doing here? (it looks like then we see if we've had > 10 errors in a minute for a minute...)

and instead we are figuring out that if we have 25% of our fluentd records are errors over the course of 5 min for 15 minutes we will alert for high.

files/fluentd/fluentd_prometheus_alerts.yaml
blockloop: I can reuse the more verbose message. That part was unintentional.

ewolinetz: do we have the same sorting as we had before where we can be specific to a fluentd instance? (generally curious)

files/fluentd/fluentd_prometheus_alerts.yaml
blockloop: `sum by(instance, job)` will remove all other labels and sum the resulting datasets that are common. Before this alert was assuming that 10 errors per second was considered a problem. That's a fair belief if fluentd is only outputting 20 logs per second because that would mean that half of the outputs are errors. But the problem is if fluentd is outputting 5000 logs per second. At which point 10 errors per second only represents 0.2% of logs. Perhaps we can play with these percentages a bit and maybe decrease them to 10% or whatever we think might be a good number, however I don't believe that using error count without considering output count is valuable.

files/fluentd/fluentd_prometheus_alerts.yaml
blockloop: Yes, because the instance label is still included in the query (I've adjusted it again)

files/fluentd/fluentd_prometheus_alerts.yaml
lukas-vlcek: What value is stored in `$labels.instance` ? Will it help us identify specific node or pod_id?

PR #648, opened 2020-08-05
Bug 1871737:  Upgrade CRD apiVersion apiextensions.k8s.io from v1beta1 to v1

periklis:
### Description
This PR addresses a recommended upgrade of the `apiVersion` of cluster-logging-operator managed CRDs from `apiextensions.k8s.io/v1beta1` to `apiextensions.k8s.io/v1`. Former is deprecated in upcoming k8s release 1.19 and very likely removed in 1.20. In extent addressing this upgrade sooner than later frees the operator upgrade procedure from limbo to enable installation of version `N` on an OCP cluster version `N-1`, where the cluster version has dropped the deprecated versions.

/cc @jcantrill @ewolinetz

### DIFF
--- a/bundle/manifests/cluster-logging.v4.6.0.clusterserviceversion.yaml
+++ b/bundle/manifests/cluster-logging.v4.6.0.clusterserviceversion.yaml
@@ -250,6 +250,12 @@ spec:
           verbs:
           - "*"
         - apiGroups:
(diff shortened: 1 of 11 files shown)


jcantrill:
@alanconway is it safe to do this now or should we be waiting until the next release stream?

PR #649, opened 2020-08-05
Bug 1866531: [clf] Infrastructure logs are sent to application pipelines.

alanconway:
Fixed configuration template so that infrastructure logs won't "fall thru" to the @_APPLICATION label if there is no infrastructure source.

- Minor logging cleanp
- Avoid [] in pipeline names - translated to fluentd config names, legal but not wise.

/cc @ik


alanconway:
/retest
AWS quota error

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -291,24 +291,32 @@ const fluentConfTemplate = `{{- define "fluentConf" -}}
   </filter>
 
   # Relabel specific source tags to specific intermediary labels for copy processing
-{{ if .CollectInfraLogs }}
+  # Earlier matchers remove logs so they don't fall through to later ones.
```
vimalk78, line 294: the bug is that we read infra logs when only "application" is configured
this section of fluent.conf routes the logs to proper destination type, but still logs from infra space will be collected.

perhaps we should specifically exclude infra logs when "application" is configured by generating something like:
```
path "/var/log/containers/*.log"
exclude_path ["/var/log/containers/*_default_*", "/var/log/containers/*_kube-*_*", "/var/log/containers/*_openshift-*_*", "/var/log/containers/"]
```
and for "infrastructure"
```
path ["/var/log/containers/*_default_*", "/var/log/containers/*_kube-*_*", "/var/log/containers/*_openshift-*_*", "/var/log/containers/"]
```
and if both "appliation" and "infrastructure"
```
path "/var/log/containers/*.log"
```
(with appropriate check to exclude logging logs)

vimalk78: but if our approach is to capture ALL logs, and then decide where we want to send them (or not send them), then the solution is ok. but this conflicts with the solution for namespace filtering, which filters the logs at the log collection level.

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
```diff
@@ -383,7 +383,7 @@ var _ = Describe("[ClusterLogForwarder] Forwards logs", func() {
 							{
 								Name:       "test-app",
 								OutputRefs: []string{"syslogout"},
-								InputRefs:  []string{"application"},
+								InputRefs:  []string{"infrastructure"},
```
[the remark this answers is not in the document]
alanconway: Thanks I'll review the tests and fix this.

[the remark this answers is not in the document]
alanconway: Thanks, I'll review the tests and fix that.

pkg/generators/forwarding/fluentd/templates.go
alanconway: We were already scraping all the logs, these sections are simply re-labelling them to the right destinations. The problem was that if you don't ask for infra those, there was no infra match rule and they all fall thru to the applicaiton match rule.  My fix always creates all the sections but sends to 'null' instead of relabelling. It is no worse than the existing code and the same as far as namespaces are concerned.

I agree there is wasted work in scraping logs just to send them to 'null', we should review how we configure fluentd to address that. We can raise a new story for that.

Does that make sense or am I missing something?

vimalk78: yes, it makes sense.

/lgtm

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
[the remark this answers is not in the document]
vimalk78: @alanconway this is the reason for syslog test failures

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
```diff
@@ -523,7 +521,7 @@ var _ = Describe("[ClusterLogForwarder] Forwards logs", func() {
 					generatorPayload["facility_key"] = "local0"
 					generatorPayload["severity_key"] = "Informational"
 				})
-				It("should take from record", func() {
+				FIt("should take from record", func() {
```
vimalk78, line 524: left over from local testing.

pkg/generators/forwarding/fluentd/templates.go
mburke5678: @alanconway Does this PR make the note in the docs irrelevant?
_you must configure a pipeline and output for all three types of logs, application, infrastructure, and audit. If you do not specify a pipeline and output for a log type, those logs are not stored and will be lost._

Is @vimalk78 comment about needing pipelines for both infra and app required to get the container logs? Qialoing's  comment in the BZ  supports this: https://bugzilla.redhat.com/show_bug.cgi?id=1866531#c6

vimalk78: @mburke5678 , the PR #699 fixes comments from Qiaolong.

PR #655, opened 2020-08-07
LOG-831. Add scripts to run SVT workload

jcantrill:
This PR
* Adds a script to run SVT workload to verify log throughput performance
* Updates the makefile

This test will allow us to verify no regressions are introduces when  swapping around configuration or component deployment

Ref: https://issues.redhat.com/browse/LOG-831


jcantrill:
cc @alanconway @vimalk78

bparees:
non-shipping code changes, overriding bz

PR #658, opened 2020-08-10
Bug 1867687: Update audit file location for oauth-apiserver

jcantrill:
This PR:
*  Update audit file location for oauth-apiserver
* removes ref to env vars for the file location
* modifies run.sh to move the position file if it exists

Ref https://bugzilla.redhat.com/show_bug.cgi?id=1867687


jcantrill:
cc @alanconway

files/fluentd/run.sh
```diff
@@ -92,4 +92,9 @@ if type -p jemalloc-config > /dev/null 2>&1 ; then
     export LD_PRELOAD=$( jemalloc-config --libdir )/libjemalloc.so.$( jemalloc-config --revision )
     export LD_BIND_NOW=1 # workaround for https://bugzilla.redhat.com/show_bug.cgi?id=1544815
 fi
+if [ -f /var/log/openshift-apiserver/audit.log.pos ] ; then
+  #https://bugzilla.redhat.com/show_bug.cgi?id=1867687
+  mv /var/log/openshift-apiserver/audit.log.pos /var/log/oauth-apiserver.audit.log
```
alanconway, line 97: If we move the pos file, don't we have to move the logs as well? Or conversely if we don't move the logs, shouldn't fluetnd start fresh with the oath logs, which it hasn't seen before?

ewolinetz: +1 to the second part. not sure its necessary to move the `.pos` file because it would be based on the logs its reading. if anything leaving it would allow fluentd to finish reading any `audit.log` files that are left on the node

ewolinetz: looking at the fluentd config we'll be reading the old logs still in addition to the new log path. however if we don't move the `.pos` file fluentd we reread the old files and potentially give us duplicate log entries

ewolinetz:
:+1: from me

files/fluentd/run.sh
alanconway: Got it.

jcantrill:
@alanconway there is an open question that I can not answer.  Does oauth audit logs replace openshift api server audit logs or is it in addition to the logs we are collecting?  There is some discrepancy based on a ticket that @sichvoge pointed out

alanconway:
> @alanconway there is an open question that I can not answer. Does oauth audit logs replace openshift api server audit logs or is it in addition to the logs we are collecting? There is some discrepancy based on a ticket that @sichvoge pointed out

I believe they replace openshift API logs. The only reason to collect both would be if we need to bridge the gap to avoid dropping logs during the upgrade. I'm not too clear on the OCP upgrade process,if the OCP upgrade includes moving existing logs to the new location, then we don't need to watch both, but if it doesn't there might be unforwarded stuff in the old file.

JAORMX:
> @alanconway there is an open question that I can not answer. Does oauth audit logs replace openshift api server audit logs or is it in addition to the logs we are collecting? There is some discrepancy based on a ticket that @sichvoge pointed out

They don't replace the openshift API logs.

PR #659, opened 2020-08-12
Bug 1852341: fix legacy syslog k8s Metadata

blockloop:
The `use_record true` setting is trying to use the "level" from the k8s
audit log which [is Metadata](https://kubernetes.io/docs/tasks/debug-application-cluster/audit/#audit-policy). I believe we need to relabel k8s audit logs
"level" field to "k8s_audit_level" so that fluentd doesn't try to use it
as the severity and we can retain the original information.


jcantrill:
@ewolinetz i think this needs a template change as well .  @lukas-vlcek ?

lukas-vlcek:
If this "renames" `level` field to `k8s_audit_level` then please open a ticket against https://github.com/ViaQ/elasticsearch-templates as well and provide some more info. (Is it really rename/replace or this adds a new field and the `level` field is going to stay there?)

PR #664, opened 2020-08-13
Bug 1882524: fixing bundle image and updating makefile accordingly

ewolinetz:
Addresses and fixing bundle issues


bundle/manifests/cluster-logging.v4.6.0.clusterserviceversion.yaml
```diff
@@ -249,6 +249,12 @@ spec:
           - consoleexternalloglinks
           verbs:
           - "*"
```
alanconway, line 251: Is * necessary here or can we list the permissions needed?

bundle/manifests/cluster-logging.v4.6.0.clusterserviceversion.yaml
```diff
@@ -249,6 +249,12 @@ spec:
...
+        - apiGroups:
+          - ""
+          resources:
+          - configmaps
+          verbs:
+          - "*"
```
alanconway, line 257: Same

Makefile
```diff
@@ -128,11 +128,12 @@ test-cleanup:			# Only needed if e2e tests are interrupted.
...
 
-generate-bundle: $(OPM)
+MANIFEST_VERSION?="4.6"
+generate-bundle: regenerate $(OPM)
 	mkdir -p bundle; \
-	pushd bundle; \
-	$(OPM) alpha bundle generate --directory ../manifests/$(OCP_VERSION)/ --package cluster-logging-operator --channels $(OCP_VERSION) --output-dir .; \
-	popd
+	$(OPM) alpha bundle generate --directory manifests/${MANIFEST_VERSION} --package cluster-logging-operator --channels ${MANIFEST_VERSION} --default ${MANIFEST_VERSION} --output-dir bundle/; \
+	find bundle/manifests/ -type f ! \( -name "cluster-logging*" -o -name "*crd.yaml" \) -delete
```
alanconway, line 135: Why do we delete these files?

vimalk78, line 135: Please add bundle validation also in the same target, or in separate target
`$(OPERATOR_SDK) bundle validate --verbose ./bundle`

vimalk78 (commented):
is this bundle image only for upstream, or can we use this downstream also

bundle.Dockerfile
```diff
@@ -5,7 +5,7 @@ LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
 LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=cluster-logging-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=4.6
-LABEL operators.operatorframework.io.bundle.channel.default.v1=
+LABEL operators.operatorframework.io.bundle.channel.default.v1=4.6
 
-COPY manifests /manifests/
-COPY metadata /metadata/
+COPY bundle/manifests /manifests/
```
vimalk78, line 10: We can remove `bundle/manifest/kustomization.yaml` and `bundle/manifest/image-references` after the `COPY` step.
As these are not part of bundle, and cause validation failures.

bundle/manifests/cluster-logging.v4.6.0.clusterserviceversion.yaml
ewolinetz: that is something to fix in the `manifest` dir, the bundle simply pulls from there.

Makefile
ewolinetz: it causes errors when installing -- the catalog seems to validate what files are in the bundle, only csvs and crds are valid

bundle.Dockerfile
ewolinetz: they are removed with the `find` command in `generate-bundle`

ewolinetz:
@vimalk78 added bundle validation, PTAL

ewolinetz:
currently bundle validation fails, however that is due to the TP logforwarding still existing in the manifest file -- that should be cleaned up as a follow up PR since its outside the scope of what this PR is performing.
```make
DEBU[0000] Debug logging is set                         
INFO[0000] Found annotations file                        bundle-dir=bundle container-tool=docker
INFO[0000] Could not find optional dependencies file     bundle-dir=bundle container-tool=docker
ERRO[0000] Error: Value logging.openshift.io/v1alpha1, Kind=LogForwarding: CRD "logging.openshift.io/v1alpha1, Kind=LogForwarding" is present in bundle "clusterlogging.v4.6.0" but not defined in CSV 
```

bundle/manifests/logforwardings.crd.yaml
```diff
@@ -72,6 +72,9 @@ spec:
                       - logs.app
                       - logs.infra
                       - logs.audit
+                  labels:
```
vimalk78, line 75: we should remobe this CRD from bundle, since this is not owned in the CSV

ewolinetz:
this is only updating the Makefile and bundle directory which is not currently used by CI nor anything in the release process. can we get a bz exception on this please?

PR #666, opened 2020-08-17
Bug 1866681: Fixed missing cluster-wide tags

igor-karpukhin:
Fixed missing cluster-wide tags for LF pipelines


igor-karpukhin:
@alanconway I guess this pipeline reassemble function has to be rewritten so that all fields will be added automatically if they are passed the validation

PR #667, opened 2020-08-20
WIP LOG-679: POC for centralized normalizer

jcantrill:
This PR enables a dev preview POC for a centralized normalizer. By specifying a particular log forwarding topology, the CLO deploys a normalizer and collector.  As implemented:

* edgeNormalization - legacy that exists today where fluentd is deployed on the nodes performing both roles
* centralNormalization - devpreview that creates fluentbit collectors on the nodes writing to fluentd normalizers

Add the following to your clusterlogging instance:

```
kind: ClusterLogging
metadata:
  annotations:
    clusterlogging.openshift.io/enableDevPreviewTopology: "true"
    clusterlogging.openshift.io/logforwardingTopology: centralNormalization
```
**Note**:  This PR does not honor ClusterLogForwarding or the ClusterLogging collection spec.  Any desired changes to the configuration or deployment require the ClusterLogging instance to be placed into unmanaged state

cc @alanconway @sichvoge

### DIFF
--- a/pkg/k8shandler/logforwardingtopology/normalizer/collector_config.go
+++ b/pkg/k8shandler/logforwardingtopology/normalizer/collector_config.go
@@ -1,0 +1,83 @@
+package normalizer
+
+const (
+	fluentbitConf = `
+[SERVICE]
+    Log_Level info
+    Parsers_file /etc/fluent-bit/parsers.conf
+
+[INPUT]
+    Name tail
+    Path /var/log/containers/*_openshift*_*.log, /var/log/containers/*_kube*_*.log, /var/log/containers/*_default_*.log
+    Path_Key filename
+    Parser containerd
+    Exclude_Path /var/log/containers/collector*_openshift*_*.log, /var/log/containers/elasticsearch*_openshift*_*.log,/var/log/containers/kibana*_openshift*_*.log,/var/log/containers/normalizer*_openshift*_*.log
+    Tag kubernetes.*
+    DB /var/log/infra-containers.pos.db
+    Refresh_Interval 5
+
+[INPUT]
+    Name tail
+    Path /var/log/containers/*.log
+    Path_Key filename
+    Parser containerd
+    Exclude_Path /var/log/containers/*_openshift*_*.log, /var/log/containers/*_kube*_*.log, /var/log/containers/*_default_*.log
+    Tag kubernetes.*
+    DB /var/log/app-containers.pos.db
+    Refresh_Interval 5
+
+[INPUT]
+    Name tail
+    Path /var/log/audit/audit.log
+    Path_Key filename
+    Parser json
+    Tag linux-audit.log
+    DB /var/log/audit-linux.pos.db
+    Refresh_Interval 5
+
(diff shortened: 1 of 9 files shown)


pkg/k8shandler/logforwardingtopology/normalizer/collector_config.go
```diff
@@ -0,0 +1,83 @@
...
+    Log_Level info
+    Parsers_file /etc/fluent-bit/parsers.conf
+
+[INPUT]
+    Name tail
+    Path /var/log/containers/*_openshift*_*.log, /var/log/containers/*_kube*_*.log, /var/log/containers/*_default_*.log
+    Path_Key filename
+    Parser containerd
+    Exclude_Path /var/log/containers/collector*_openshift*_*.log, /var/log/containers/elasticsearch*_openshift*_*.log,/var/log/containers/kibana*_openshift*_*.log,/var/log/containers/normalizer*_openshift*_*.log
+    Tag kubernetes.*
```
sichvoge, line 15: Should we tag these as `infra`?

pkg/k8shandler/logforwardingtopology/normalizer/collector_config.go
```diff
@@ -0,0 +1,83 @@
...
+    DB /var/log/infra-containers.pos.db
+    Refresh_Interval 5
+
+[INPUT]
+    Name tail
+    Path /var/log/containers/*.log
+    Path_Key filename
+    Parser containerd
+    Exclude_Path /var/log/containers/*_openshift*_*.log, /var/log/containers/*_kube*_*.log, /var/log/containers/*_default_*.log
+    Tag kubernetes.*
```
sichvoge, line 25: Same as above but this time, should we tag these as `container` or `app`? Not sure if it is desired to have two different inputs but the tags are exactly the same. WDYT?

sichvoge:
A couple questions:

- Where would we assemble split lines by CRIO?
- Where would we assemble multiline log records?
- My concern, maybe it shouldn't be one, is that when you have multiple replicas of the Fluentd StatefulSet - can it happen that a split line could go to two different pods and therefore, would not be considered one line?

jcantrill:
> I made some basic attempts with fluentbit to reassemble using the multiline parser configurations thinking it should be more performant and make better use of memory. I was unsuccessful and the available examples are not very good. Additionally, I think the way CRIO splits and identifies split lines does not play well with the the patterns that fluentbit expects.

@sichvoge 
I am unable to see how multiline parser configs can be utilized in fluent bit to re-assemble split lines based on the crio format.  I don't believe fluentd concat plugin maps to it conceptionally.  I made a rudimentary attempt to write a lua script but there this mechanism assumes you are processing and returning a single record.  It supports globals but there is no mechanism AFAIK to cache records and then later dump them  into a processing pipeline; the API simply doesn't support it.  It seems we need to either:

* reassemble at the normalizer which introduces the challenges you already described
* introduce a fluentbit concat filter ourselves

jcantrill:
closing in favor of #762

PR #669, opened 2020-08-21
Bug 1868068: Evaluate daemonset init containers during reconciliation

jcantrill:
This PR:

* Evaluates daemonset init containers when reconciling the collector
* Refactors daemonset comparison to no longer mutate incoming daemonsets
* Renames the method to retrieve the TrustedCAHash to be a getter instead of getter/mutator

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1868068


alanconway:
> > /lgtm
> > Style note (no need to change) - when you have a series of if statements each of which returns, I find it neater to use a case statement. Not important enough to change the PR
> 
> Can you please provide an example as I'm curious how one might do that with a switch

Sorry, doesn't apply here. You can do a switch for simple boolean expressions, but you can't do an inline assignment in a case statement (I keep forgetting that) so it doesn't work for: err=x(); err != nil

PR #672, opened 2020-08-24
make tests less verbose

blockloop:
Removing the -v flag from go test prevents Go from printing any of the
t.Log messages unless the test fails.

/assign @jcantrill 
/cc @igor-karpukhin


ewolinetz:
manually adding `valid-bug` as this PR is only removing verbosity flags used during our tests. no code is being modified

PR #580, opened 2020-06-29
Bug 1870289: LOG-755: Make ClusterLogForwarder cluster scoped

alanconway:

PR #673, opened 2020-08-25
Bug 1870898: Revert WATCH_NAMESPACE to unbreak clusterlogging

jcantrill:
This PR reverts a change in https://github.com/openshift/cluster-logging-operator/pull/580 that breaks watching clusterlogging

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1870898

cc @ecordell please advise how we can watch target NS for one kind but watch global for a different cluster scoped resource


periklis (approved):
IMHO this is in-line with what I've suggested previously to watch out according to the operator-sdk docs:
https://sdk.operatorframework.io/docs/building-operators/golang/crds-scope/#crd-cluster-scoped-usage

(cc @alanconway )

igor-karpukhin (approved):
/lgtm

PR #674, opened 2020-08-25
Bug 1861224: Disable LogFowarding when legacy protocols in use

periklis:
### Description
This PR addresses a manual cherry-pick of #625 due to merge conflicts (see [here](https://github.com/openshift/cluster-logging-operator/pull/625#issuecomment-679119028)). In detail the PR backports the fix to allow using `LogForwarding` exclusively from legacy forward methods for syslog and fluentd-forward.

### Notes to the patch-manager
The merge-conflicts mentioned above manifest in the following features added to `release-4.5` and upwards and not present in `release-4.4`. The features are applied in the fluentd configuration generation and thus test development in upwards branches need to be slimmed down in downwards branches. Features in play (See expected fluentd configs `output_conf_legacy_test.go `):
- Filtering application logs per namespace
- Switch of the data model to static indices, e.g. `infra-write`
- Flatting kubernetes labels

### Disclaimer
This PR needs to include backporting the fix provided by #542 for >4.5. The reason we need this fix with the current PR is that we made LogForwarding exclusive from legacy methods. In turn `NUM_OUTPUTS` will always result to zero and needs a reset to one.

/cc @jcantrill @igor-karpukhin

### DIFF
--- a/files/fluentd/run.sh
+++ b/files/fluentd/run.sh
@@ -74,6 +74,10 @@ if [ ! -f "$FLUENT_CONF" ] ; then
     exit 1
 fi
 NUM_OUTPUTS=$(grep "path.*'$FILE_BUFFER_PATH" $FLUENT_CONF | wc -l)
+if [ $NUM_OUTPUTS -eq 0 ]; then
+    # Reset to default single output if log forwarding outputs all invalid
+    NUM_OUTPUTS=1
+fi
 
 # Get the available disk size.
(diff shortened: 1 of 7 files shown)


periklis:
@sdodson Can we pick this one soon please for the next 4.4 z-stream release?

periklis:
@derekwaynecarr Can we pick this for the next 4.5 z-stream? It eliminates some confusion in customer cases and makes troubleshooting less prone to false negatives.

PR #681, opened 2020-08-27
Fixed namespace filtering test case

vimalk78:
The test case was failing randomly because fluent-receiver receives logs generated in previous test cases. So added a check for logs generated from this test case.
Added cleanup function for fluent-receiver.

/cc @jcantrill @alanconway


jcantrill:
Merged to address test flakes

PR #683, opened 2020-08-29
LOG-914: CI test Improvements

vimalk78:
- Cleaned up fluend buffers after test case
 - Run cleanup.sh script to collect artifacts only if test case fails

fixes: https://issues.redhat.com/browse/LOG-914


jcantrill:
overriding BZ to merge since this is test improvement

PR #686, opened 2020-09-01
updating bundle labels to be used for nightly

jcantrill:
Updating bundle labels per:

* https://docs.engineering.redhat.com/display/CFC/Migration


jcantrill:
Overriding label to make nightly builds available

PR #688, opened 2020-09-03
LOG-747: Functional framework for fluentd testing

jcantrill:
This PR:

* introduces functional test framework for testing without operator dependency
* initially targeted to test fluent configurations
* initially introduces tests for fluent metrics

It should be possible to expand this framework to replace the current LF e2e tests with the expectation of writing from one container to the next

ref: https://issues.redhat.com/browse/LOG-747


test/helpers/functional/fluentd.go
```diff
@@ -0,0 +1,113 @@
...
+	logger.Infof("running command: %s", cmd)
+	return oc.Exec().Pod(f.Name).WithNamespace(f.Name).WithCmd(cmd[0], cmd[1:]...).RunFor(maxDuration)
+}
+
+//Deploy the objects needed to functional test
+func (f *FluentdFunctionalFramework) Deploy() error {
+	if f.Error = oc.Literal().From(fmt.Sprintf("oc create ns %s", f.Namespace)).Output(); f.Error != nil {
+		return f.Error
+	}
+	//generate empty config
```
vimalk78, line 66: empty config meaning?

test/builder/pod.go
```diff
@@ -0,0 +1,115 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+)
+
+type PodBuilder struct {
+	kubeclient *kubernetes.Clientset
+	obj        *corev1.Pod
+}
+
+func NewPodBuilder(namespace, name string) *PodBuilder {
```
vimalk78, line 17: Add method `AddSecurityContext` to let fluent pod run with `privileged: true`

test/helpers/functional/fluentd.go
```diff
@@ -0,0 +1,113 @@
...
+	if _, f.Error = builder.NewConfigMapBuilder(f.Namespace, certsName).
+		Add("tls.key", string(utils.GetWorkingDirFileContents("system.logging.fluentd.key"))).
+		Add("tls.crt", string(utils.GetWorkingDirFileContents("system.logging.fluentd.crt"))).
+		Create(); f.Error != nil {
+		return f.Error
+	}
+
+	podBuilder := builder.NewPodBuilder(f.Namespace, f.Name).
+		WithLabels(f.labels).
+		AddConfigMapVolume("config", f.Name).
```
vimalk78, line 91: Add Volume 
```
- hostPath:
    path: /var/log
    type: ""
  name: varlog
```
and volumeMount
```
volumeMounts:
- mountPath: /var/log
  name: varlog
```

internal/pkg/generator/forwarder/generator.go
```diff
@@ -0,0 +1,61 @@
...
+		return "", fmt.Errorf("Unable to create collector config generator: %v", err)
+	}
+
+	forwarder := &logging.ClusterLogForwarder{}
+	err = yaml.Unmarshal([]byte(clfYaml), forwarder)
+	if err != nil {
+		logger.Errorf("Error Unmarshalling %q: %v", clfYaml, err)
+		os.Exit(1)
+	}
+	logger.DebugObject("Unmarshalled %s", forwarder)
```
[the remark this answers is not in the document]
alanconway: @blockloop you did the logging cleanup on EO, is there a story up for doing the same to CLO? This issue should be fixed at the same time - get rid of DebugObject and replace it with the generic log statetment and test.YAMLString(forwarder) to write the object.

internal/cmd/forwarder-generator/main.go
```diff
@@ -1,21 +1,28 @@
...
+	"github.com/openshift/cluster-logging-operator/internal/pkg/generator/forwarder"
 	"github.com/openshift/cluster-logging-operator/pkg/logger"
 	"github.com/sirupsen/logrus"
 )
 
 // HACK - This command is for development use only
 func main() {
+
+	yamlFile := flag.String("file", "", "ClusterLogForwarder yaml file. - for stdin")
+	includeDefaultLogStore := flag.Bool("include-default-store", true, "Include the default storage when generating the config")
```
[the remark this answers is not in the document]
jcantrill: Maybe we dont.  It was added because the code this calls into has a boolean flag that represents this.  I did not follow the call but AFAIK this is the only way you get the default store.

alanconway (commented):
OK with general idea, but see alternative to builder approach at https://github.com/alanconway/cluster-logging-operator/blob/bz-1866531-forwarder/test/object.go#L63

internal/pkg/generator/forwarder/generator.go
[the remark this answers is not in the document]
jcantrill: > @blockloop you did the logging cleanup on EO, is there a story up for doing the same to CLO? This issue should be fixed at the same time - get rid of DebugObject and replace it with the generic log statetment and test.YAMLString(forwarder) to write the object.

@alanconway there is a card this spring that is assigned to @vparfonov to work with @blockloop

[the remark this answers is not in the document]
jcantrill: > How to see output from this function? i have never seen any output from this function in practice.
> Since this is a test, perhaps we change this function to generate outpout without enabling debug logs

you have to modify the loglevel like LOG_LEVEL=debug.  and it seems it only appears on test failure.  Also, not sure if the DEBUG log works correctly for cases like "trace"

Makefile
```diff
@@ -138,6 +138,14 @@ undeploy-elasticsearch-operator:
 deploy-example: deploy
 	oc create -n $(NAMESPACE) -f hack/cr.yaml
 
+test-functional:
+	@export LOG_LEVEL=info; \
```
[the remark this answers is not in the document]
jcantrill: > change to `@export LOG_LEVEL=${LOG_LEVEL:-info}` to control log level

Thoughts on how these will work together? https://github.com/openshift/cluster-logging-operator/pull/693/files#diff-b67911656ef5d18c4ae36cb6741b7965R4

pkg/certificates/certificates.go
```diff
@@ -0,0 +1,24 @@
...
+
+	"github.com/openshift/cluster-logging-operator/pkg/logger"
+)
+
+func GenerateCertificates(namespace, scriptsDir, logStoreName, workDir string) (err error) {
+	script := fmt.Sprintf("%s/cert_generation.sh", scriptsDir)
+	return RunCertificatesScript(namespace, logStoreName, workDir, script)
+}
+
+func RunCertificatesScript(namespace, logStoreName, workDir, script string) (err error) {
```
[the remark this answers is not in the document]
jcantrill: > Why do we need to run a script to gen certs? In the interest of reducing our bash madness, can we do this with https://golang.org/pkg/crypto/x509/ ? Just asking, I haven't thought it thru.

We can, we have an open JIRA to replace the script (which is log time legacy) with go code but it hasn't been a priority since it worked and is tech debt

test/builder/configmap.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"k8s.io/client-go/kubernetes"
+)
+
+type ConfigMapBuilder struct {
+	kubeclient *kubernetes.Clientset
+	obj        *corev1.ConfigMap
+}
+
+func NewConfigMapBuilder(namespace, name string) *ConfigMapBuilder {
```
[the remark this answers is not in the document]
jcantrill: We should have a conversation regarding the difference in approaches.  I'm proposing a ["fluent"](https://en.wikipedia.org/wiki/Fluent_interface) pattern in the interest of making the tests more expressive and less terse.  The point is to hide the "building" and object "creation" on purpose.  It's not exactly BDD (e.g. Given/When/Then) but does provide a mechanism ideally to replace xUnit style testing

test/functional/metrics/fluentd_test.go
```diff
@@ -0,0 +1,56 @@
...
+	)
+
+	BeforeEach(func() {
+
+		framework = functional.NewFluentdFunctionalFramework().WithLogForwarder(clusterLogForwarder)
+		Expect(framework.Deploy()).To(BeNil())
+	})
+
+	AfterEach(func() {
+		oc.Literal().From(fmt.Sprintf("oc delete ns %s --force --grace-period=0 --ignore-not-found", framework.Namespace)).Output()
```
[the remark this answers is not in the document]
jcantrill: I'm adding a if check but not certain of the value

test/builder/pod.go
jcantrill: Well add as needed but wasn't required for this test. Not certain we will need in with functional tests in general

test/helpers/functional/fluentd.go
jcantrill: The point is we shouldnt need to mount any node level directories for the test.  This would allow us to directly inject messages into the container and remove the setup without having artifacts on the host

vimalk78: How do we plan to inject logs into fluentd's pipeline? i.e. what will be the `<source>` ?

test/functional/metrics/fluentd_test.go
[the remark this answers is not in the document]
alanconway: If you don't care about a return value the way to get past lint is:

    _ = oc.Literal()...

That tells the go compiler and the linter that you know there's a return value and you're ignoring it on purpose.

alanconway:
I have an overlapping approach for e2e tests, but I think we should merge both and then we'll be able to discuss and pick and choose the best bits of each approach into a unified set of testing packages.

Makefile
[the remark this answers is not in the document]
jcantrill: exported var at top of file

pkg/certificates/certificates.go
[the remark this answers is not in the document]
jcantrill: https://issues.redhat.com/browse/LOG-422

test/helpers/functional/fluentd.go
jcantrill: we can still define a source to the same path but jump onto the container via exec and echo something in the required format or something like that , maybe a shared init container

test/builder/configmap.go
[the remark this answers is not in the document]
alanconway: I like the fluent style in a from-scratch API. My concerns in this case are:
1. The wrappers are more code to write where bugs can hide. to be consistent we have to add more wrappers every time we need to use a new k8s type.
2.

jcantrill:
@alanconway updated to use the new client and framework but couldn't bring myself to completely remove the builders as building up pods is very terse without it.  please review

test/helpers/functional/fluentd.go
alanconway: Just FYI: The image has fluent-cat so you could use that to inject messages directly into a `forward` source. You could also echo/cat stuff into a file that fluent is tailing as you point out. Check test/helpers/cmd package has a Reader type that can read from pod files, you could create equivalent Writer types to write to pod files or via fluent-cat direct to fluentd.

alanconway:
/lgtm

@jcantrill I can see the value in the builders for complicated sub-structs. I already had to provide some helper functions for the ClusterLogging setup, and a ClusterLoggingBuilder would be a better solution to that.

I have some small changes to suggest but it'll be easier if you merge as-is and then one of us can take care of further changes.

Suggested Changes:

1. Move the builders to the runtime package for use by e2e and functional tests.
2. Keep the existing runtime NewX methods so people can mix the approaches (see below).
3. Drop the client member from builders

2 reasons for change 3 - 

First: there are other possible clients, so better to keep the in-memory construction separate from the choice of client. Not a big change: instead of `podBuilder.Create()` it becomes `client.Create(podBuilder.Pod())`

Second: makes the PodBuilder usable as a wrapper of an existing Pod, so you can say things like:
```
pod := runtime.NewPod(namespace, name) // Or create a Pod some other way...
runtime.PodBuilder{pod}.AddVolumeMount(...)...
ExpectOK(c.Create(pod))
```
That way you can mix the explicit + builder approaches as best suits the circumstances. For simpler cases the direct approach might be clearer, for more complicated cases the Builder avoids a lot of laborious repetition of struct details.

Thoughts?

alanconway (approved):

jcantrill:
> Suggested Changes:
> 
> 1. Move the builders to the runtime package for use by e2e and functional tests.
Moved
> 2. Keep the existing runtime NewX methods so people can mix the approaches (see below).
> 3. Drop the client field from builders

Done

alanconway (approved):
I like it. Can you just export the obj field so you can both wrap the raw struct and extract it again? e.g.
```
type PodBuilder struct { Pod *corev1.Pod }

builder.AddFoo().AddBar()
myPod := builder.Pod
```
Otherwise LGTM.

test/functional/framework.go
```diff
@@ -0,0 +1,138 @@
...
+	test      *client.Test
+	pod       *corev1.Pod
+}
+
+func init() {
+	maxDuration, _ = time.ParseDuration("2m")
+}
+
+func NewFluentdFunctionalFramework() *FluentdFunctionalFramework {
+	opt := log.WithVerbosity(uint8(9))
```
eranra, line 44: @jcantrill Just this have to be fixed or can be set by external ENV variable ?

jcantrill: this could be evaluatedl like https://github.com/openshift/cluster-logging-operator/pull/751 but we don't have  a single place to set this.  you will only see the logs if the test fails though i would prefer it to not be fixed like this

jcantrill: Updated to depend on env var

jcantrill:
> I like it. Can you just export the obj field so you can both wrap the raw struct and extract it again? e.g.
> 
> ```
> type PodBuilder struct { Pod *corev1.Pod }
> 
> builder.AddFoo().AddBar()
> myPod := builder.Pod
> ```
> 
> Otherwise LGTM.

modified

test/functional/framework.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	return framework
+}
+
+func (f *FluentdFunctionalFramework) Cleanup() {
+	f.test.Close()
+}
+
+func (f *FluentdFunctionalFramework) RunCommand(cmdString string) (string, error) {
+	cmd := strings.Split(cmdString, " ")
```
eranra, line 76: @jcantrill this is not doing good split for the command line parameters ... 
Can you change to import::

shellquote "github.com/kballard/go-shellquote"

cmd, _ := shellquote.Split(cmdString)

This splits the string much better and allows the exec command to really act like in shell

alanconway: Actually I'd prefer RunCommand(args...string) so that you have explicit control of the arguments in Go. Shell quoting and splitting on whitespace are both accident prone when you combine them with Go's string literal rules. e.g. what does this do: RunCommand("echo \"\\\"foo bar\\\"\"")?

PR #692, opened 2020-09-10
make e2e framework watches more resilient

jcantrill:
This PR rebuilds the e2d test framework kubeclient everytime to avoid issues in watches waiting for deployments


jcantrill:
overriding valid label because this adds ci stability

PR #693, opened 2020-09-10
Bug 1877968: Reduce the chattiness of status updates and reconciliation

jcantrill:
This PR:

* Modifies status creation to reduce the status updates and noise in the logs

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1877968

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 # Define the target to run if make is called with no arguments.
 default: check
 
+LOG_LEVEL?=fatal
 export KUBECONFIG?=$(HOME)/.kube/config
 
 export GOBIN=$(CURDIR)/bin
@@ -140,7 +141,7 @@ deploy-example: deploy
 
 test-unit:
 	LOGGING_SHARE_DIR=$(CURDIR)/files \
-	LOG_LEVEL=fatal \
+	LOG_LEVEL=$(LOG_LEVEL) \
 	go test -cover -race ./pkg/...
 
 test-cluster:
--- a/pkg/k8shandler/collection_test.go
+++ b/pkg/k8shandler/collection_test.go
@@ -163,3 +163,60 @@ var _ = Describe("Reconciling", func() {
 		})
 	})
 })
+
+var _ = Describe("compareFluentdCollectorStatus", func() {
+	defer GinkgoRecover()
+	var (
+		lhs loggingv1.FluentdCollectorStatus
+		rhs loggingv1.FluentdCollectorStatus
+	)
+
+	BeforeEach(func() {
+		lhs = loggingv1.FluentdCollectorStatus{
+			DaemonSet:  constants.FluentdName,
+			Conditions: map[string]loggingv1.ClusterConditions{},
+			Nodes:      map[string]string{},
+			Pods:       map[loggingv1.PodStateType][]string{},
+		}
+		rhs = loggingv1.FluentdCollectorStatus{
+			DaemonSet:  constants.FluentdName,
+			Conditions: map[string]loggingv1.ClusterConditions{},
+			Nodes:      map[string]string{},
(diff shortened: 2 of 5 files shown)


pkg/k8shandler/collection_test.go
```diff
@@ -7,6 +7,7 @@ import (
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	configv1 "github.com/openshift/api/config/v1"
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
 	loggingv1 "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
```
periklis, line 11: Duplicate import, consider keeping the k8s-convention compliant `loggingv1`

periklis (commented):
One small import path issue, otherwide :lgtm:

pkg/k8shandler/forwarding.go
```diff
@@ -134,10 +134,10 @@ func (clusterRequest *ClusterLoggingRequest) NormalizeForwarder() (*logging.Clus
 	} else {
 		if len(unready)+len(degraded) > 0 {
 			status.Conditions.SetCondition(condDegraded(logging.ReasonInvalid, "degraded pipelines: invalid %v, degraded %v", unready, degraded))
-			logger.Infof("ClusterLogForwarder degraded")
+			logger.Debug("ClusterLogForwarder degraded")
```
ewolinetz, line 137: since this ends up in the cluster conditions does having this in the logs buy you guys anything?

pkg/k8shandler/forwarding.go
```diff
@@ -134,10 +134,10 @@ func (clusterRequest *ClusterLoggingRequest) NormalizeForwarder() (*logging.Clus
...
 		}
 		status.Conditions.SetCondition(condReady)
-		logger.Infof("ClusterLogForwarder is ready")
+		logger.Debug("ClusterLogForwarder is ready")
```
ewolinetz, line 140: same ask here

pkg/k8shandler/forwarding.go
jcantrill: This leads to my general question from slack regarding status and logging any of these messages.  What value does this provide if the result is reflected in a CR status?

jcantrill: removing

pkg/k8shandler/forwarding.go
jcantrill: removing

pkg/k8shandler/status.go
```diff
@@ -166,6 +167,7 @@ func translatePodMap(podStateMap elasticsearch.PodStateMap) logging.PodStateMap
 
 func getPodState(podStateMap elasticsearch.PodStateMap, podStateType elasticsearch.PodStateType) []string {
 	if v, ok := podStateMap[podStateType]; ok {
+		sort.Strings(v)
```
ewolinetz, line 170: since its not explicit in the summary -- is a core issue that the names of the fluentd pods, while remaining the same, if they are populated in a different order will be recognized as being different and therefore cause a status update?

Makefile
```diff
@@ -1,6 +1,7 @@
 # Define the target to run if make is called with no arguments.
 default: check
 
+LOG_LEVEL?=fatal
```
alanconway, line 4: `export LOG_LEVEL?=fatal` so it applies to anything that make runs, then no need to repeat it in individual targets.

alanconway (approved):

pkg/k8shandler/status.go
jcantrill: just seeing this, but yes, that is what i found

PR #695, opened 2020-09-11
Fixing test flakes

vimalk78:
1. Changed `cat /shared/consumed.logs` to `tail -n 5000 /shared/consumed.logs` to limit the logs are fetched from pod
2. fixed closure in wait.Poll to return nil if the command in closure gets an error
3. fixed location of `must-gather.log`
4. fixed log generator namespace already exists error


jcantrill:
overriding bugzilla label to address CI and testing issues

PR #696, opened 2020-09-14
Bug 1880012: Move custom Ginkgo matchers to matchers package.

alanconway:
Move all our custom matchers into the 'matchers' package.

Ginko convention uses anonymous imports, but that's a bad idea for anything
else. Therefore Gingko extensions must be in a separate package from other code.

/cc @jcantrill
/assign @jcantrill


jcantrill (commented):
/lgtm

PR #700, opened 2020-09-17
Add checking go version before run command

vparfonov:


scripts/check-go-version.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -e
+MINIMAL_VERSION=1.13
```
periklis, line 4: Minimal version is 1.14 for this repository, because we use the `-modfile` argument for `go build`.

scripts/check-go-version.sh
```diff
@@ -0,0 +1,12 @@
...
+VERSION=$( go version | grep -o '1\.[0-9|\.]*')
+
+function version { echo "$@" | awk -F. '{ printf("%d%03d%03d\n", $1,$2,$3); }'; }
+
+if [ $(version $VERSION) -lt $(version $MINIMAL_VERSION) ]; then
```
periklis, line 9: nit. I suggest adding double quotes around command expansion. Version include dots and they may add to expansion friction ;)

Makefile
```diff
@@ -204,3 +205,6 @@ cluster-logging-operator-install:
 # uninstalls the cluster-logging operator
 cluster-logging-operator-uninstall:
 	olm_deploy/scripts/operator-uninstall.sh
+
+check-go-version:
```
periklis, line 209: I suggest to add this command to the `.PHONY` targets as it is not relevant for file/dirname expansion by make.

scripts/check-go-version.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/sh
+
+set -e
```
periklis, line 3: I suggest to expand the flags to `set -euo pipefail`. It ensure exiting immediately also on right-most pipe fails and unset variables/parameters.

periklis (commented):
LGTM besides suggested improvements.

scripts/check-go-version.sh
alanconway: Suggest the script take the version directly from go.mod to avoid duplication: 
```
MINVERSION=$(awk '/^go/ {print $2}' go.mod)
```

alanconway (changes requested):
I thought the presence of the mod file would enforce the go version?

From a make perspective this is not great. The .PHONY check-go-version target is always be out of date, so it will for every target that depends on it to run regardless of whether it is really out of date. Suggest instead to check it once when the Makefile loads:

```
ifneq (0,$(shell check-version.sh; echo $$? ))
$(error "bad version")
endif
```

periklis:
> I thought the presence of the mod file would enforce the go version?
> 
> From a make perspective this is not great. The .PHONY check-go-version target is always be out of date, so it will for every target that depends on it to run regardless of whether it is really out of date. Suggest instead to check it once when the Makefile loads:
> 
> ```
> ifneq (0,$(shell check-version.sh; echo $$? ))
> $(error "bad version")
> endif
> ```

The [go directive](https://golang.org/ref/mod#go-mod-file-go) in `go.mod` is only responsible to tell the compiler which language features are available when building the present module. In other words this means only builds will break if you say build with an 1.14 and include 1.15 features. But I believe this is a rare case since most people use their IDEs, editors w/ LSP so far and all these tools compile your code on the fly for error checking.

scripts/check-go-version.sh
vparfonov: done

Makefile
vparfonov: done

alanconway:
I still don't understand what this check is for. If your go compiler is inadequate, the build will fail. 
The mod file enforces a standard notion of what "adequate" means.
What do we gain by adding a check of our own that does almost-but-not-exactly the same thing?

vparfonov:
going to close this PR, looks like checking useless

PR #709, opened 2020-09-24
Bug 1877968: Remove status update message

jcantrill:
This PR removes verbose status update messages.

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1877968


periklis (approved):
Amen

igor-karpukhin (approved):
/lgtm. We could also change the level of this message from `Info` to `Debug`

PR #712, opened 2020-09-25
Bug 1882524: Fixed 4.6 manifests folder and bundle validation

vimalk78:
The CLO operator bundle container contained patch files. This leads to failures to push the Errata Advisory to stage.

 - removed logforwardings.crd.yaml from bundle/manifest which is not present in CSV
 - moved crd patches from manifests/4.6 (which is used by ART) to manifest/patches, and modified crd generation scripts to use these patch files


jcantrill (commented):
/approve

periklis (changes requested):
Why doing this copy/remove of the patches and not a simple remove from the bundle dir like the logforwarding CRD?

igor-karpukhin (approved):
/lgtm

PR #720, opened 2020-10-01
Bug 1884275: Revert CLF cluster scope

alanconway:
Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1884275

/assign @jcantrill
/cc @periklis

### DIFF
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -39,8 +39,8 @@ To run CLO as a local process:
 * `make debug`: Run CLO in the `dlv` debugger.
 
 To run the e2e tests:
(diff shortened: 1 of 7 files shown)


jcantrill:
@alanconway looks good but are there also changes needed to the "must-gather" bits now the resource is namespace?

alanconway:
> @alanconway looks good but are there also changes needed to the "must-gather" bits now the resource is namespace?

Nope, I checked that and it looks like I never updated the must-gather scripts when I made the scope change in the first place :blush:

alanconway:
/retest

e2e tests are passing on my local CRC cluster, suspect timeouts due to CI overload

PR #725, opened 2020-10-04
New packages for e2e tests with sample test.

alanconway:
TLDNR; Look at [example of new forwarding test ](https://github.com/alanconway/cluster-logging-operator/blob/new-e2e-tests/test/e2e/logforwarding/fluent/fluent_test.go#L1)

Goals:
- Faster tests: run cluster actions concurrently, report slow actions.
  - New fluentd forwarding tests < 2 min, old tests ~ 5 min.
- Only pre-test requirement is `make deploy`
- Don't fail if resources are left behind by other tests, delete and recreate.
- Passing tests delete their resources, failing tests leave them for debugging.
- Labels for easy cleanup: `oc delete all -A -l test-client`

Package summary:

test
- UniqueName: strong crypto random IDs to avoid collisions.
- CmdReader: tail output of `oc exec` commands - read instead of polling.
- FailGroup: paralellize Ginkgo assertions safely.
- UnlessFailed: clean-up only on pass, leave evidence for debugging on fail.

test/runtime
- generic manipulation of objects via the runtime.Object interface.
- easy constructors for corev1 and loggingv1 objects.
- Decode objects from YAML (also test.MustUnmarshal to fill in parts of objects.)

test/client:
- Generic client for any object struct type using runtime.Object interface.
- Client.Watch: watch for changes instead of polling.
- Recreate: Delete+create - no more "already exists" failures.
- Process-wide singleton client (multiple client creation is slow)
- Auto-label resources for easy clean-up: `oc delete all -A -l test-client`
- CreateUniqueNS: unique namespace per test

test/client/fluentd:
- Flexible fluentd receiver, any number of listening ports.
- Tail output files instead of polling
- TLS secured sources.

test/certificate:
- Simplified certificate management for tests.

test/e2e/logforwarding/fluentnew:
- Sample test using the above (will eventually replace existing fluent test)
- Validates log data, doesn't just check for existence.
- Paralellizes operations where possible using FailGroup.


test/cmdreader_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/openshift/cluster-logging-operator/test"
+	. "github.com/openshift/cluster-logging-operator/test/matchers"
+)
+
+var _ = Describe("CmdReader", func() {
+	FIt("reads lines", func() {
```
vimalk78, line 14: focussed test

alanconway:
/hold
for comments

test/client/client_suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestClient(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Client Suite")
```
periklis, line 12: IIRC correct we configure the junit reporter for E2E tests. I suggest to consider the similar ginkgo register routines in our E2E test suite.

test/client/client_test.go
```diff
@@ -0,0 +1,93 @@
...
+
+	BeforeEach(func() {
+		c = Get()
+		ns = c.CreateUniqueNS()
+		test.Infof("test namespace: %v", ns.Name)
+		data = map[string]string{"a": "b"}
+	})
+
+	AfterEach(func() {
+		test.UnlessFailed(func() { ExpectOK(c.Remove(ns)) })
```
periklis, line 29: Does this imply that we keep the namespace if a failure occurs? Is this required for a test must-gather later?

test/client/client_test.go
```diff
@@ -0,0 +1,93 @@
...
+	It("creates object with data and automatic labels", func() {
+		cm := object.NewConfigMap(ns.Name, "foo", data)
+		ExpectOK(c.Create(cm))
+		cm2 := object.NewConfigMap(ns.Name, "foo", nil)
+		ExpectOK(c.Get(cm2))
+		Expect(cm2.Data).To(Equal(data))
+		Expect(cm2.Labels).To(HaveKeyWithValue(LabelKey, LabelValue))
+	})
+
+	Context("deletes", func() {
```
periklis, line 41: Typo? Shouldn't this be `recreates`?

test/client/client_test.go
```diff
@@ -0,0 +1,93 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/openshift/cluster-logging-operator/test"
+	. "github.com/openshift/cluster-logging-operator/test/client"
+	. "github.com/openshift/cluster-logging-operator/test/matchers"
+	"github.com/openshift/cluster-logging-operator/test/object"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/watch"
+)
+
+var _ = Describe("Client", func() {
```
periklis, line 14: Isn't `defer ginkgo.GinkgoRecover()` required for a testcase that makes use of go routines?

test/client/fluentd/fluentd_suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestFluentd(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Fluentd Suite")
```
periklis, line 12: IIRC correct we configure the junit reporter for E2E tests. I suggest to consider the similar ginkgo register routines in our E2E test suite.

test/client/fluentd/receiver_test.go
```diff
@@ -0,0 +1,49 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/cluster-logging-operator/test"
+	"github.com/openshift/cluster-logging-operator/test/client"
+	. "github.com/openshift/cluster-logging-operator/test/client/fluentd"
+	. "github.com/openshift/cluster-logging-operator/test/matchers"
+	"github.com/openshift/cluster-logging-operator/test/object"
+)
+
+var _ = Describe("Receiver", func() {
```
periklis, line 16: Isn't defer ginkgo.GinkgoRecover() required for a testcase that makes use of go routines?

test/client/fluentd/receiver_test.go
```diff
@@ -0,0 +1,49 @@
...
+	FIt("receives and saves data", func() {
```
periklis, line 17: The whole suite is currently focus on this single test case!!!

test/cmdreader.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+	"os/exec"
+	"time"
+)
+
+const (
+	// DefaultTimeout for operations that are expected to succeed.
+	DefaultTimeout = 5 * time.Minute
+	// ShortTimeout for operations that are *expected* to time out.
+	ShortTimeout = 10 * time.Second
```
periklis, line 15: I suggest to rename `DefaultTimeout` to `DefaultSuccessTimeout` and `ShortTimeout` to `DefaultFailureTimeout`

test/cmdreader_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"os/exec"
+	"time"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	. "github.com/openshift/cluster-logging-operator/test"
+	. "github.com/openshift/cluster-logging-operator/test/matchers"
+)
+
+var _ = Describe("CmdReader", func() {
```
periklis, line 13: Same here missing `ginkgo.GinkgoRecover()`?

test/e2e/logforwarding/fluentnew/fluentnew_suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestPipelines(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "SingleInput Suite")
```
periklis, line 12: Same here. Maybe missing the junit reporter.

test/e2e/logforwarding/fluentnew/fluentnew_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	loggingv1 "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/test"
+	"github.com/openshift/cluster-logging-operator/test/client"
+	"github.com/openshift/cluster-logging-operator/test/client/fluentd"
+	. "github.com/openshift/cluster-logging-operator/test/matchers"
+	"github.com/openshift/cluster-logging-operator/test/object"
+	"github.com/openshift/cluster-logging-operator/test/object/clusterlogging"
```
periklis, line 15: `object` is quite a generic pkg name. Why not calling it `crds`?

test/e2e/logforwarding/fluentnew/fluentnew_test.go
```diff
@@ -0,0 +1,129 @@
...
+)
+
+var cl = clusterlogging.New()
+
+var _ = BeforeSuite(func() { ExpectOK(client.Get().Recreate(cl)) })
+
+var _ = AfterSuite(func() { _ = client.Get().Remove(cl) })
+
+var _ = Describe("[ClusterLogForwarder]", func() {
+	var (
```
periklis, line 26: Same here missing ginkgo.GinkgoRecover() because we use go routines?

test/client/client_suite_test.go
alanconway: Yep. The J2EE reporter and the collection of debugging artifacts should also move to Go in a later PR.

test/client/client_test.go
alanconway: It does, that's something to be considered. When I'm running tests one-by-one, it's much easier if the actual namespace is still around. When we're running on CI I'm not sure if that's helpful or not. We might want to configure this based on whether we're in CI or not, or maybe it's just OK to leak these resources since they don't interfere with other tests and can be cleaned up with an `oc delete -l test-client` We'll see how it works in practice.

test/client/client_test.go
alanconway: yep

test/client/fluentd/fluentd_suite_test.go
alanconway: +1

test/cmdreader.go
alanconway: I like it.

alanconway: I despise the very notion of a "DefaultFailureTimeout" but I haven't figured out how to eliminate it yet. It's "wishful thinking" testing - "well I didn't look very hard and I didn't find anything, so it must not be there."

test/client/client_test.go
alanconway: It's required in the goroutine. The FailGroup type takes care of this for you so you can safely say:
```
g := test.FailGroup
g.Go(func() { Expect(blah).To(BeBoring()); }
g.Go(func() { Expect(geewhiz).To(BeGreatlyExciting()); }
g.Wait()
```
If you start goroutines by some other method then you would need the defer ginkgo stuff.

test/e2e/logforwarding/fluentnew/fluentnew_test.go
alanconway: Because corev1.Pod etc. are not CRDs, and because the central type in the API is called runtime.Object. I'm open to suggestions though, I had trouble deciding on this this. It's inconsistent as it stands because I gave clusterlogging a package of its own but no the other types.

My first take was:

```
package test
type Object struct{ o runtime.Object }
func (o Object) NamespacedName() ....
```

So you can say:

```
var o runtime.Object
obj := test.Object(o)
m := obj.Meta()
nn := obj.NamespacedName()
```

I was pretty pleased with myself, but in practice I almost always ended up only needing to call one method so it was usually:

```
nn := test.Object(o).NamespacedName()
```

in which case a simple package function is, well, simpler. I thought I might be getting fancy for no real gain (I do that sometimes) What do you think? I could be persuaded to go back to that, or if you have something simpler in mind...

periklis: I suggest to improve identifying speed and lessen cognitive bias between apimachinery are code, why not replicate the package name as `runtime`. That buys us:
```
github.com/openshift/cluster-logging-operator/test/runtime/clusterlogging
```
which is easier to relate to
```
"k8s.io/apimachinery/pkg/runtime"
```
WDYT?

alanconway: +1 I like it

test/cmdreader_test.go
alanconway: Fixed

test/e2e/logforwarding/fluentnew/fluentnew_test.go
alanconway: Fixed

periklis (approved):
Looks good to me so far. I will leave it to others to review and apply the labels.

Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 AS builder
+FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 AS builder
```
jcantrill, line 1: Should probably do this rebased from https://github.com/openshift/cluster-logging-operator/pull/736

README.md
```diff
@@ -2,36 +2,32 @@
...
 
-The CLO can also *forward* logs over multiple protocols, to multiple types of store, on- or off-cluster
+The CLO can also *forward* logs over multiple protocols, to multiple types of log stores, on- or off-cluster
 
-The CLO (Cluster Logging Operator) *owns* the following APIs:
+The CLO *owns* the following APIs:
 
 * ClusterLogging: Top level control of cluster-wide logging resources
 * ClusterLogForwarder: Configure forwarding of logs to external sources
 * Collector: configure global collector resource use.
```
jcantrill, line 25: Remove since this no longer exists and was devpreview api

README.md
```diff
@@ -2,36 +2,32 @@
...
 
-To install a released version of cluster logging see the [Openshift Documentation](https://docs.openshift.com/)
+To install a released version of cluster logging see the [Openshift Documentation](https://docs.openshift.com/), (e.g., [OCP v4.5](https://docs.openshift.com/container-platform/4.5/logging/cluster-logging-deploying.html))
```
jcantrill, line 27: @mburke5678 is there a live like to 4.6?

bundle/manifests/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -13,7 +13,7 @@ spec:
     shortNames:
     - clf
     singular: clusterlogforwarder
-  scope: Cluster
+  scope: Namespaced
```
jcantrill, line 16: Isn't this part of a PR already merged?

cmd/manager/main.go
```diff
@@ -47,46 +43,27 @@ var (
...
-	//
-	// The logger instantiated here can be changed to any logger
-	// implementing the logr.Logger interface. This logger will
-	// be propagated through the whole operator, generating
-	// uniform and structured logs.
-	logf.SetLogger(zap.Logger())
-
-	if logLevel := os.Getenv("LOG_LEVEL"); logLevel != "" {
-		level, err := logrus.ParseLevel(logLevel)
+	logLevel, present := os.LookupEnv("LOG_LEVEL")
```
jcantrill, line 48: Already part of the log PR refactoring which is merged?

docs/HACKING.md
```diff
@@ -49,7 +47,7 @@ NOTE: e2e tests require a clean cluster starting point, `make undeploy-all` will
 There are two types of test:
 
 * *Unit tests:* run in seconds, no cluster needed, verify internal behavior - behind the API. New code *should* be reasonably well covered by unit tests but we don't have a formal requirement.
-* *E2E tests:* run in minutes to hour(s), requires a cluster. There are several ways to run these:
+* *E2E tests:* run in minutes to hour(s), require a cluster. There are several ways to run these:
```
jcantrill, line 50: its not 'requires'?

hack/get-openshift-client.sh
```diff
@@ -8,6 +8,10 @@ DOWNLOAD_URL=$(curl -s -L  https://openshift-release.svc.ci.openshift.org/api/v1
 
 NAME="openshift-client-linux-$VERSION.tar.gz"
 mkdir -p bin
+if [ -f "bin/$NAME" ]; then
```
jcantrill, line 11: merged PR?

jcantrill (changes requested):

alanconway:
@jcantrill are you ready to LGTM this?

Dockerfile
alanconway: Done

alanconway: Done.

README.md
alanconway: This file wasn't part of the PR, but fixed anyway/

README.md
alanconway: Fixed - removed the version specific reference, people can find the latest from the home page.

cmd/manager/main.go
alanconway: Didn't touch this file. After rebase it matches latest master.

bundle/manifests/logging.openshift.io_clusterlogforwarders_crd.yaml
alanconway: Not sure what changes got picked up here. I didn't touch that file in this PR. Have a look at the rebase.

hack/get-openshift-client.sh
alanconway: Again, didn't touch - check rebased PR

alanconway:
@jcantrill figured out the diff confusion, this was still targetted at master-post-release. Switched to master should make more sense now.

docs/HACKING.md
alanconway: Merge confusion. Check again.

test/helpers/fluentd/secret.tmp
```diff
@@ -0,0 +1,7 @@
+{
```
jcantrill, line 1: should this be deleted?

test/runtime/logger.go
```diff
@@ -0,0 +1,23 @@
...
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+// NewLogger creates a pod that will print `count` lines to stdout, waiting for
+// `delay` between each line.  Lines are of the form "<timestamp> [n] `message`"
+// where n is the number of lines output so far. Once done printing the pod will
+// be idle but will not exit until deleted.
+func NewLogger(namespace, name string, count int, delay time.Duration, message string) *corev1.Pod {
```
jcantrill, line 14: Can we call this LogGenerator or the like and rename the file.  I think "Logger" i think of logging my error messages or the link

jcantrill (changes requested):

test/client/client.go
```diff
@@ -0,0 +1,213 @@
+package client
```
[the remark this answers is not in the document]
alanconway: It's a self-contained unit of functionality that can be re-used in other circumstances so I think it makes sense for it to be a package.

test/client/client.go
```diff
@@ -0,0 +1,213 @@
...
+		if cfg, err = config.GetConfig(); err != nil {
+			return nil, err
+		}
+	}
+	c := &Client{
+		cfg:     cfg,
+		ctx:     context.Background(),
+		rests:   map[schema.GroupVersion]rest.Interface{},
+		timeout: timeout,
+		labels:  make(map[string]string, len(labels)),
```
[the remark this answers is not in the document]
alanconway: Nope, go map is pointer-like (as are slice, array and chan) so a copy of a map points to the same set of key/value pairs as the original.

test/client/client.go
```diff
@@ -0,0 +1,213 @@
...
+	timeout time.Duration
+	labels  map[string]string
+}
+
+// New client.
+//
+// Operations will fail if they take longer than timeout.
+// The labels will be applied to each object created by the client.
+func New(cfg *rest.Config, timeout time.Duration, labels map[string]string) (*Client, error) {
+	var err error
```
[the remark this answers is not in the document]
alanconway: Nope, return values must all be named or all be nameless.

test/client/client.go
```diff
@@ -0,0 +1,213 @@
...
+}
+
+// Remove is like Delete but ignores NotFound errors.
+func (c *Client) Remove(o runtime.Object) error {
+	return crclient.IgnoreNotFound(c.Delete(o))
+}
+
+// Recreate first calls Remove, then Create.
+//
+// NOTE: Won't work for a namespace as they do not delete immediately.
```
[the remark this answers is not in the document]
alanconway: I came across this https://github.com/kubernetes/kubernetes/issues/42594#issuecomment-285801993 which says "The current state is that kubectl delete always waits (for ex: it waits for pods to be deleted when running kubectl delete rs or kubectl delete deployment). Namespace deletion and pod graceful deletion are exceptions." However that's not a very definitive reference. I'll beef up Recreate so it does a Watch() for the delete event.

test/helpers/fluentd/secret.tmp
alanconway: Yes, done.

test/runtime/logger.go
alanconway: Done.

test/e2e/logforwarding/fluent/fluent_secure_test.go
```diff
@@ -0,0 +1,58 @@
...
+			"tls.key":       clfCert.PrivateKeyPEM(),
+			"ca-bundle.crt": ca.CertificatePEM(),
+			"ca.key":        ca.PrivateKeyPEM(),
+			"shared_key":    []byte("top-secret"),
+		})
+		t.group.Go(func() {
+			ExpectOK(t.Recreate(secret)) // Create secret before CLF.
+			ExpectOK(t.Recreate(clf))
+		})
+		expectLines(t, "tls", 1, secureLoggerMessage, "viaq_index_name", "app-write")
```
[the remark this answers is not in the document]
alanconway: Fair point, I'll make it a method on Source.

test/const.go
```diff
@@ -0,0 +1,27 @@
...
+	// Ideally tests for non-happening should have a positive wait with
+	// DefaultSuccessTimeout for something that we know *does* happen, and that we
+	// know never happens *before* the thing we don't expect to happen.  Then look
+	// for positive proof that the thing didn't happen. A negative timeout result
+	// never proves that the thing won't happen in 1 more millisecond.
+	//
+	DefaultFailureTimeout = 1 * time.Second
+
+	InstanceName               = "instance"
+	OpenshiftLoggingNS         = "openshift-logging"
```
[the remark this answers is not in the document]
alanconway: :) circular dependency is exactly what happend. Since package test doesn't have any dependencies, I can re-define helpers/constants.go in terms of the test/const.go so there's only one definition.

test/cmdreader.go
```diff
@@ -0,0 +1,74 @@
+package test
```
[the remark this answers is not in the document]
alanconway: Done.
Can you summarize what you think belongs in package test vs. in a package under helpers? I can drop in a README for future devs adding new stuff.

test/certificate/certificate.go
```diff
@@ -0,0 +1,157 @@
+package certificate
```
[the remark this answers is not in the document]
alanconway: helpers it is.

alanconway:
@jcantrill I think I've addressed all your comments. LGTM?

test/client/client.go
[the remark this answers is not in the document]
jcantrill: I wasn't suggesting it should not be a package but that we may have other clients that could logically be in their own packages:
```
test/client/openshift/client.go
test/client/podman/client.go
```

test/client/test.go
```diff
@@ -0,0 +1,32 @@
+package client
```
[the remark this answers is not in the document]
jcantrill: Still open question

test/e2e/logforwarding/fluent/fluent_test.go
```diff
@@ -0,0 +1,154 @@
...
+	var t *Test
+	BeforeEach(func() { t = NewTest(message) })
+	AfterEach(func() { t.Close() })
+
+	Context("with app/infra/audit receiver", func() {
+		BeforeEach(func() {
+			t.receiver.AddSource("application", "forward", 24224)
+			t.receiver.AddSource("infrastructure", "forward", 24225)
+			t.receiver.AddSource("audit", "forward", 24226)
+			t.group.Go(func() { ExpectOK(t.receiver.Create(t.Client)) })
```
jcantrill, line 65: As a team, we had a conversation of terse vs explicit and frankly I find this syntax non-intuitive. Is there something we can do here by name that can make it more obvious what it does?  I don't understand why its structured this way at all

test/e2e/logforwarding/fluent/fluent_test.go
```diff
@@ -0,0 +1,154 @@
...
+		})
+
+		It("forwards application logs only", func() {
+			clf := runtime.NewClusterLogForwarder()
+			addPipeline(clf, t.receiver.Sources["application"])
+			t.group.Go(func() { ExpectOK(t.Recreate(clf)) })
+			// Use ctx to let empty checks run until we get the expected lines.
+			ctx, cancel := context.WithCancel(context.Background())
+			t.group.Go(func() {
+				defer cancel() // Cancel goroutines verifying empty when we get our logs.
```
jcantrill, line 75: Is this something that can be automatically done for us via a helper method?

test/e2e/logforwarding/fluent/fluent_test.go
```diff
@@ -0,0 +1,154 @@
...
+				r := t.Reader("application")
```
jcantrill, line 76: Maybe more verbose of what I'm getting?   "ReaderForSourceType(" ??

test/e2e/logforwarding/fluent/fluent_test.go
alanconway: How would you like it structured? Here are some other options
```
// Use struct field names to be more explicit
receiver.AddSource("name", Source{Port: 24224, Type: "forward"})
// Provide all sources as map on creation, no Add function
receiver := NewReceiver(map[string]*Source{ "name": Source{Port: 24224, Type: "forward"}, .... }
```
I used "AddSource/Source" because that's what fluentd calls it, personally I would tend to use "Listen/Listener" to follow the go net library. I guess it depends if we want to follow fluent or Go conventions here.

test/e2e/logforwarding/fluent/fluent_test.go
alanconway: We could mke the WithCancel line shorter with a terse function name, but I think that's a bad case of choosing terse over explicit. 

Possibly you could persuade me to take a different approach entirely - the idea here is that we are expecting logs from some sources, and not from others. Therefore we wait (forever) on the sources where we expect nothing, and we cancel those waits when we've got everything we expected from the other sources.

A simpler alternative would be to wait with a fixed timeout where we expect nothing, then no need for cancellation. I don't like fixed timeouts because they are always too long or too short, and that's one of the reasons our tests currently spend a lot of time sleeping for nothing. Using the wait time for expected results seems like a nice natural timeout based real conditions when the test runs.

test/e2e/logforwarding/fluent/fluent_test.go
alanconway: Done.

test/e2e/logforwarding/fluent/fluent_test.go
alanconway: Updated using explicit struct fields, I like this better than the original. Whaddya think @jcantrill ?

jcantrill: The source was less of a concern to me then `t.group.Go(func() { ExpectOK(t.receiver.Create(t.Client)) })` which doesn't read well to be or make it obvious whats going on.   I'm coming from the perspective of what is it I need to do when I add a test and why am I doing it

test/e2e/logforwarding/fluent/fluent_test.go
jcantrill: I'm not saying we should not do this or this is not correct.  I'm trying to emphasize the "why"?  Why am I doing this and if I'm writing a test is it going to be obvious that I forgot to add it.  Can the framework/test design make it more obvious what to  do and why by either:
* making a name that makes it explicit
* Adding a function that "does it for me"

test/e2e/logforwarding/fluent/fluent_test.go
alanconway: This line is starting a goroutine to create a recever using t.Client.
go func() {...}() is how you start a goroutine, so 
g.Go(func() {...}) starts a gouroutine managed by the group g. This is syntactically pretty close, and the same pattern as the errgroup package which is a proposed Go extension https://godoc.org/golang.org/x/sync/errgroup.
ExpectOK() is shorthand for Expect(x).To(Succeed()) which *really* annoys me for unnecessry typing in Ginkgo. 
Which part is problematic?

test/e2e/logforwarding/fluent/fluent_test.go
alanconway: The goroutine that starts at line 75 is going to block until it reads some expected log output (or the read times out).
Following this (not shown above) are two goroutines that will block reading on files that shoud *NOT* get any data. I pass this context to those Read() methods. If Read()s succeed that's an error - I wasn't expecting data there.
Once the first groutine succeeds, it cancels the others, otherwise the test will be stuck for some overly-long timeout waiting for data that won't come.
I'm not sure how to express that more clearly - context.Context is the standard way to cancel goroutines if their work is no longer needed. "cancel()" is the name the context library uses.

PR #731, opened 2020-10-08
Minor fix to the scripts under ./hack 

huikang:
- save some time during the build process
- avoid 404 if the file has already downloaded

### DIFF
--- a/hack/generate-crd.sh
+++ b/hack/generate-crd.sh
@@ -30,7 +30,7 @@ cp manifests/patches/${KUSTOMIZATIONS_FILE} ${MANIFESTS_DIR}
 echo "---------------------------------------------------------------"
 echo "Kustomize: Patch CRDs for singeltons"
 echo "---------------------------------------------------------------"
-oc kustomize "${MANIFESTS_DIR}" | \
+bin/oc kustomize "${MANIFESTS_DIR}" | \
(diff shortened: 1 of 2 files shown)


periklis (approved):
/ok-to-test

periklis (approved):
/lgtm

jcantrill (commented):
/approve

hack/generate-crd.sh
```diff
@@ -30,7 +30,7 @@ cp manifests/patches/${KUSTOMIZATIONS_FILE} ${MANIFESTS_DIR}
 echo "---------------------------------------------------------------"
 echo "Kustomize: Patch CRDs for singeltons"
 echo "---------------------------------------------------------------"
-oc kustomize "${MANIFESTS_DIR}" | \
+bin/oc kustomize "${MANIFESTS_DIR}" | \
```
vimalk78, line 33: sorry for the late comment, but this change fixes the `oc` to be in `bin/` folder 
the make target `openshift-client`
```
openshift-client:    
    @type -p oc > /dev/null || bash hack/get-openshift-client.sh 
```
checks the presence  of `oc` anywhere in path.

May be we can add a print statement which suggests user to add oc to path before starting the download, rather than starting download silently in background

PR #733, opened 2020-10-08
Patch Dockerfile to use CentOS images.

alanconway:
The default Dockerfile is using RHEL8 repositories that are not accessible to all.
This will be fixed, but meantime, this enables optional built with CentOS packages:

    PATCH=Dockerfile.centos.patch make ...

/assign @jcantrill
/cc @periklis

### DIFF
--- a/Dockerfile.centos.patch
+++ b/Dockerfile.centos.patch
@@ -1,0 +1,27 @@
+--- Dockerfile	2020-10-16 10:26:19.124105649 -0400
++++ Dockerfile.centos	2020-10-16 10:28:25.713003840 -0400
+@@ -1,4 +1,4 @@
+-FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.7 AS builder
++FROM registry.svc.ci.openshift.org/openshift/release:golang-1.14 AS builder
+ WORKDIR /go/src/github.com/openshift/cluster-logging-operator
+ 
+ # COPY steps are in the reverse order of frequency of change
+@@ -17,9 +17,8 @@
+ 
+ RUN make build
+ 
+-FROM registry.svc.ci.openshift.org/ocp/4.7:cli AS origincli
++FROM centos:centos7 AS centos
+ 
+-FROM registry.svc.ci.openshift.org/ocp/4.7:base
+ RUN INSTALL_PKGS=" \
+       openssl \
+       rsync \
+@@ -38,7 +37,6 @@
+ COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/manifests /manifests
+ RUN rm /manifests/art.yaml
+ 
+-COPY --from=origincli /usr/bin/oc /usr/bin/
+ COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/must-gather/collection-scripts/* /usr/bin/
+ 
+ # this is required because the operator invokes a script as `bash scripts/cert_generation.sh`
--- a/Makefile
+++ b/Makefile
@@ -86,10 +86,12 @@ clean:
 	@rm -rf bin tmp _output
 	go clean -cache -testcache ./...
 
-image:
+PATCH?=Dockerfile.patch
+Dockerfile.local: Dockerfile $(PATCH)
+	patch -o Dockerfile.local Dockerfile $(PATCH)
+
+image: Dockerfile.local
 	@if [ $${SKIP_BUILD:-false} = false ] ; then \
-		cp Dockerfile Dockerfile.local ; \
-		patch Dockerfile.local Dockerfile.patch ; \
 		podman build -t $(IMAGE_TAG) . -f Dockerfile.local; \
 	fi
 
@@ -154,7 +156,7 @@ test-cluster:
 MANIFEST_VERSION?="4.6"
 generate-bundle: regenerate $(OPM)
 	MANIFEST_VERSION=${MANIFEST_VERSION} hack/generate-bundle.sh
-	
+
 .PHONY: generate-bundle
 
 # NOTE: This is the CI e2e entry point.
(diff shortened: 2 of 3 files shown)


k-keiichi-rh:
Thank you for your work.
I could build cluster-logging-operator with CentOS packages using "PATCH=Dockerfile.centos.patch make deploy-image".

Dockerfile
```diff
@@ -18,17 +18,12 @@ COPY pkg ./pkg
...
-      openssl \
-      rsync \
-      " && \
-    yum install -y $INSTALL_PKGS && \
-    rpm -V $INSTALL_PKGS && \
-    yum clean all && \
-    mkdir /tmp/ocp-clo && \
-    chmod og+w /tmp/ocp-clo
+
+RUN dnf -v -y install openssl
```
periklis, line 23: I suggest to keep `rsync` because it is a mandatory utility for must-gather. A similar dnf command to `yum clean all` should be also kept here for minimizing container size.

Dockerfile.centos.patch
```diff
@@ -0,0 +1,31 @@
+diff --git a/Dockerfile b/Dockerfile
+index 23b9fce7..a480fe28 100644
+--- a/Dockerfile
++++ b/Dockerfile
+@@ -1,4 +1,4 @@
+-FROM registry.svc.ci.openshift.org/ocp/builder:rhel-8-golang-1.15-openshift-4.6 AS builder
++FROM registry.svc.ci.openshift.org/openshift/release:golang-1.13 AS builder
```
periklis, line 7: The minimum golang version here should be `1.14` for this patch file. We use couple of 1.14-only features in our code base AFAIK.

periklis (changes requested):
I suggest to keep the RHEL-based approach as the default builder for the team and make the Centos 7 a viable optional target for our public contributors. WDYT?

Dockerfile
```diff
@@ -18,17 +18,12 @@ COPY pkg ./pkg
 RUN make build
 
 FROM registry.svc.ci.openshift.org/ocp/4.6:cli as origincli
-
 FROM registry.svc.ci.openshift.org/ocp/4.6:base
-RUN INSTALL_PKGS=" \
```
jcantrill, line 23: do we know the base image includes dnf vs yum?  also minor point but this `rpm -V $INSTALL_PKGS` ensures the package actually installs and that we don't get false success.  If there are recommendations to an alternative please add

Dockerfile.centos.patch
```diff
@@ -0,0 +1,31 @@
+diff --git a/Dockerfile b/Dockerfile
```
jcantrill, line 1: Why patch the dockerfile and not simply provide a `Dockerfile.local` or the like since `Dockerfile` is the source of truth for the downstream for the time being

jcantrill (changes requested):

Dockerfile
alanconway: It's already present in the base image, so installing it just causes "already installed" noise.
However I'll put it back in case we ever use a base image without it. +1 to the dnf clean.

Dockerfile.centos.patch
alanconway: +1. Also I am keeping RHEL as the default, the HACKING.md explains how to use the centos patch.

Dockerfile
alanconway: I've backed out my changes to the original Dockerfile, you and Peri spotted too many flaws!

yum verifies packages after download (and fails if there's a problem) so the rpm -V is not necessary but not a big deal.

Dockerfile.centos.patch
alanconway: That's exactly why. Dockerfile is the source of truth, we create Dockerfile.local using a patch. The default for local builds is Dockerfile.patch as before, this change allows you to select an alternate patch for a centos base with `PATCH=Dockerfile.centos.patch make ...`

alanconway:
@periklis @jcantrill Fixed all your comments, can I get an LGTM?

periklis (approved):
/lgtm

Dockerfile
jcantrill: This was not to verify the packages but to verify that every package I requested was actually found and installed.  I have seen builds complete the `yum install` but actually the run of the image failed because yum didnt find a package and therefore did not install it.  This was the point of this addition.  If yum has some other mechanism to verify that would be preferred IMO

Dockerfile.centos.patch
```diff
@@ -0,0 +1,27 @@
...
+ 
+-FROM registry.svc.ci.openshift.org/ocp/4.7:base
+ RUN INSTALL_PKGS=" \
+       openssl \
+       rsync \
+@@ -38,7 +37,6 @@
+ COPY --from=builder /go/src/github.com/openshift/cluster-logging-operator/manifests /manifests
+ RUN rm /manifests/art.yaml
+ 
+-COPY --from=origincli /usr/bin/oc /usr/bin/
```
jcantrill, line 24: this is not the same based on the change:
* former provided `oc`
* later provides must-gather scripts

Would not we need both?  Maybe not needed if we are only doing devel

jcantrill (changes requested):

alanconway:
@jcantrill @periklis Can I get a new LGTM? I removed some un-necessary changes.

Dockerfile.centos.patch
periklis: I suggest we may need a compromise here, because we won't get both Centos-based development and oc. We should clearly state for external devs that centos based development has the limitation of not being able to using the final image for must-gather executions

periklis (approved):

Dockerfile.centos.patch
alanconway: OC is not available without the origincli images, which are not public. However our own makefile will download it on demand so I think this works OK for developers - it's never going to be used for CI.

periklis (approved):
/lgtm

PR #739, opened 2020-10-13
LOG-945: Remove Kibana5 cleanup code

jcantrill:
This PR:

* Removes the Kibana cleanup code that was required from 4.4 to 4.5 when Kibana responsibility was moved to EO

Note:  User's are blocked from upgrading directly to 4.6 and later rendering this no longer required

Ref: https://issues.redhat.com/browse/LOG-945


alanconway (commented):
/lgtm

PR #741, opened 2020-10-14
LOG-949: Always reconcile services to desired state

jcantrill:
This PR addresses tech debt to:

* Constantly reconcile services to the state desired by the operator
* Move service constructor into an appropriately named factory package
* Add service comparator into the utils pkge

https://issues.redhat.com/browse/LOG-949


igor-karpukhin (approved):
/lgtm

PR #743, opened 2020-10-15
Bug 1888420: Add collector source input metrics

jcantrill:
This PR add input source metrics for fluent collector so we may start understand what's happening at the collection piont

* adds `cluster_logging_collector_input_record_total{tag:, hostname:}`
* adds `cluster_logging_collector_input_record_bytes{tag:, hostname:}` this is an approximation

Ref: https://bugzilla.redhat.com/show_bug.cgi?id=1888420

image

cc @blockloop

### DIFF
--- a/pkg/generators/forwarding/fluentd/fluent_conf_test.go
+++ b/pkg/generators/forwarding/fluentd/fluent_conf_test.go
@@ -154,7 +154,7 @@ var _ = Describe("Generating fluentd config", func() {
     rotate_wait 5
     tag kubernetes.*
     read_from_head "true"
-    @label @CONCAT
+    @label @MEASURE
     <parse>
       @type multi_format
       <pattern>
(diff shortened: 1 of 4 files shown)


pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -1088,7 +1186,56 @@ var _ = Describe("Generating fluentd config", func() {
                 time_format %Y-%m-%dT%H:%M:%S.%N%z
               </parse>
             </source>
-
+			<label @MEASURE>
```
blockloop, line 1189: Indention seems off

jcantrill: updated... but this particular test has some formatting issues

pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -170,7 +170,56 @@ var _ = Describe("Generating fluentd config", func() {
       </pattern>
     </parse>
   </source>
-  
+  <label @MEASURE>
+	<filter **>
+		@type record_transformer
+		enable_ruby
+		<record>
+		msg_size ${record.to_s.length}
```
syedriko, line 178: indentation

PR #744, opened 2020-10-19
Fix Makefile removing 'test-cleanup' as dependency for 'run'

eranra:
The Makefile 'run' step is dependent on 'test-cleanup' that do not exist in the Makefile. 

This PR is removing that dependency


periklis (approved):
/lgtm

PR #745, opened 2020-10-19
Reduce resource requests for AWS example to default provisioning

eranra:
When deploying default AWS cluster, provisioned CPU resources are limited and without specific resource limitations Elastic-search and Fluentd container do not deploy. 

This PR is reducing the resources required to lower level so that ES and Fluentd  will be deployed

### DIFF
--- a/hack/cr-aws.yaml
+++ b/hack/cr-aws.yaml
@@ -11,9 +11,11 @@ spec:
       storage: {}
       redundancyPolicy: "ZeroRedundancy"
       resources:
-        limits:
-          cpu: 500m
-          memory: 4Gi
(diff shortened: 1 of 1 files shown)


jcantrill:
Until operators are able to dynamically adjust resources to account for the needs of logging, we reject this change.  It looks as though the motivation for this PR is to land the logging stack without regards to what is required to support log collection for a cluster.  Elasticsearch and fluent are both resource intensive applications even on clusters that are not heavily loaded.  Until the operators can manage resources on their own it is preferable to require a known default versus assuming the cluster will scale as needed.  Granted documentation could be better.

PR #750, opened 2020-10-20
make run does not deploy-eo

jcantrill:
This PR updates the Makefile to:

* decouple make run  from automatically deploying EO (doesn't handle already deployed operator nor should it need to)
* remove a non-existent dependent target
* fix CRD to rely on `oc` in the path

cc @alanconway

### DIFF
--- a/hack/generate-crd.sh
+++ b/hack/generate-crd.sh
@@ -30,7 +30,7 @@ cp manifests/patches/${KUSTOMIZATIONS_FILE} ${MANIFESTS_DIR}
 echo "---------------------------------------------------------------"
 echo "Kustomize: Patch CRDs for singeltons"
 echo "---------------------------------------------------------------"
-bin/oc kustomize "${MANIFESTS_DIR}" | \
+oc kustomize "${MANIFESTS_DIR}" | \
     awk -v clf="${MANIFESTS_DIR}/${CLF_CRD_FILE}" \
         -v clo="${MANIFESTS_DIR}/${CLO_CRD_FILE}"\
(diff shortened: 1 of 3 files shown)


hack/generate-crd.sh
```diff
@@ -30,7 +30,7 @@ cp manifests/patches/${KUSTOMIZATIONS_FILE} ${MANIFESTS_DIR}
 echo "---------------------------------------------------------------"
 echo "Kustomize: Patch CRDs for singeltons"
 echo "---------------------------------------------------------------"
-bin/oc kustomize "${MANIFESTS_DIR}" | \
+oc kustomize "${MANIFESTS_DIR}" | \
```
periklis, line 33: It assumes you have oc installed on your PATH. IIRC, we have a make target [openshift-client](https://github.com/openshift/cluster-logging-operator/blob/master/Makefile#L57) that should be  a pre-requisite tarket of crd generation ensuring that you have an oc client in your repo's temporary bin folder.

periklis (changes requested):

hack/generate-crd.sh
alanconway: The Makefile puts bin/oc in front of PATH so this will work fine if run via the Makefile. Hard-coding bin/oc is wrong since we don't download oc if it is already in the path. For the full belt-and-suspenders we could make this:

    PATH=./bin:$PATH oc kustomize ... 

We should probably fix up the oc download to be more consistent with what you did with bingo - always download a known version. But that's a different PR...

alanconway (commented):
/lgtm

PR #752, opened 2020-10-20
Bug 1879150: Updating how we scale down ES nodes in CR to preserve the number of es node definitions

ewolinetz:
Addresses https://bugzilla.redhat.com/show_bug.cgi?id=1879150

This change works to preserve the number of es nodes we create in the `es/elasticsearch` object managed by CLO to prevent EO flagging the change as invalid due to nodes missing.

This allows us to scale from `4 -> 3` nodes (and in theory `n -> 0` however EO will prevent this since the cluster would not have the required number of data or master nodes though to clear a cluster one should delete the elasticsearch CR)

### DIFF
--- a/pkg/k8shandler/logstore.go
+++ b/pkg/k8shandler/logstore.go
@@ -18,7 +18,8 @@ import (
 )
 
 const (
-	elasticsearchResourceName = "elasticsearch"
+	elasticsearchResourceName       = "elasticsearch"
+	maximumElasticsearchMasterCount = int32(3)
 )
 
 func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateLogStore() (err error) {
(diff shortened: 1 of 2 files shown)


jcantrill (commented):
/lgtm

pkg/k8shandler/logstore.go
```diff
@@ -199,34 +214,36 @@ func newElasticsearchCR(cluster *logging.ClusterLogging, elasticsearchName strin
...
-			Roles:     []elasticsearch.ElasticsearchNodeRole{"client", "data", "master"},
-			NodeCount: 3,
-			Storage:   logStoreSpec.ElasticsearchSpec.Storage,
+	// if we had more than 1 es node before, we also want to enter this condition
+	if logStoreSpec.NodeCount > maximumElasticsearchMasterCount || len(existing.Spec.Nodes) > 1 {
+
+		// we need to check this because if we scaled down we can enter this block
+		if logStoreSpec.NodeCount > maximumElasticsearchMasterCount {
+			esNodes[0].NodeCount = maximumElasticsearchMasterCount
 		}
```
blockloop, line 232: This is a bit confusing to me. If we scaled down then how can logStoreSpec.NodeCount be greater than maximumElasticsearchMasterCount?

blockloop: I guess if someone had more master nodes before we introduced the new max?

ewolinetz: this cover the case where we scale up from 3, so where only the first part of that condition is true

PR #756, opened 2020-10-22
Enabled more linters.

alanconway:
Enabled most of the available linters in golangci_lint.
Made minor code fixes to conform to the linters. No functional changes.

This PR has multiple commits to show what changes were for what linters.
If all are approved I will squash before merge.

/hold
/cc @periklis
/cc @jcantrill

### DIFF
--- a/golangci.yaml
+++ b/golangci.yaml
@@ -1,12 +1,34 @@
-linters:
-  disable:
-    - megacheck
+ linters:
+  disable-all: true
   enable:
-    - gosimple
-    - staticcheck
-    - unused
-    - interfacer
+    - bodyclose
+    - deadcode
+    - depguard
+    - dogsled
     - errcheck
-    - prealloc
+    - gocyclo
+    - gofmt
+    - goprintffuncname
     - gosec
+    - gosimple
+    - govet
+    - ineffassign
+    - nakedret
+    - nolintlint
+    - scopelint
+    - staticcheck
+    - structcheck
+    - typecheck
+    - unconvert
+    - varcheck
+    - gocritic
+    - unused
+    - misspell
 
+# Consider enabling these, some cleanup required:
+# gomnd: An analyzer to detect magic numbers. [fast: true, auto-fix: false]
+# stylecheck: Stylecheck is a replacement for golint [fast: true, auto-fix: false]
+# whitespace: Tool for detection of leading and trailing whitespace [fast: true, auto-fix: true]
+# wsl: Whitespace Linter - Forces you to use empty lines! [fast: true, auto-fix: false]
+# dupl: Tool for code clone detection [fast: true, auto-fix: false]
+# goimports: Goimports does everything that gofmt does. Additionally it checks unused imports [fast: true, auto-fix: true]
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,7 @@ lint: $(GOLANGCI_LINT)
 
 fmt:
 	@echo gofmt		# Show progress, real gofmt line is too long
-	@gofmt -s -l -w $(shell find pkg cmd -name '*.go')
+	find pkg cmd test -name '*.go' | xargs -r gofmt -s -l -w
 
 # Do all code/CRD generation at once, with timestamp file to check out-of-date.
 GEN_TIMESTAMP=.zz_generate_timestamp
--- a/internal/cmd/forwarder-generator/main.go
+++ b/internal/cmd/forwarder-generator/main.go
@@ -7,10 +7,10 @@ import (
 
 	yaml "gopkg.in/yaml.v2"
 
+	"github.com/ViaQ/logerr/log"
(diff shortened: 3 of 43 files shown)


golangci.yaml
```diff
@@ -1,12 +1,35 @@
...
+  disable-all: true
   enable:
+    - bodyclose
+    - deadcode
+    - depguard
+    - dogsled
+    - errcheck
+    - gocyclo
+    - gofmt
+    - goimports
```
periklis, line 11: `goimports` assumes that `make fmt` is using `goimports`.

golangci.yaml
```diff
@@ -1,12 +1,35 @@
...
+    - goprintffuncname
+    - gosec
     - gosimple
+    - govet
+    - ineffassign
+    - interfacer
```
periklis, line 17: `interfacer ` is deprecated from its maintainer being nasty making bad suggestions.

golangci.yaml
```diff
@@ -1,12 +1,35 @@
...
+    - nakedret
+    - nolintlint
+    - rowserrcheck
```
periklis, line 20: `rowserrcheck` is not applicable in this repo, as we don't make use of golang's `sql` package.

golangci.yaml
```diff
@@ -1,12 +1,35 @@
...
+    - scopelint
```
periklis, line 21: `scopelint` is superseded by `exportloopref`.

golangci.yaml
```diff
@@ -1,12 +1,35 @@
...
     - staticcheck
+    - structcheck
+    - typecheck
+    - unconvert
+    - varcheck
+    - gocritic
```
periklis, line 27: `gocritic ` is highly opinionated, but let's see if we get used to it.

golangci.yaml
```diff
@@ -1,12 +1,35 @@
...
     - unused
-    - interfacer
-    - errcheck
-    - prealloc
-    - gosec
+    - misspell
+    - dupl
```
periklis, line 30: `dupl` is nice but may be a source of headaches when using `client-go` structs or api-machinery's `metav1` package.

golangci.yaml
```diff
@@ -1,12 +1,35 @@
...
 
+# Consider enabling these, some cleanup required:
+#    - stylecheck  # Replacement for golint, valid but numerous.
+#    - gomnd       # Avoid magic numbers.
+#    - whitespace  # Consistent use of whitespace.
```
periklis, line 35: I suggest to add `wsl` to the list above instead of `whitespace` only.

periklis (changes requested):
In general LGTM, only a couple of points to consider:
- Linting for `goimports` should also reflect that we use `goimports` in `make fmt`
- A couple of suggested linters are obsolete because of harmful suggestions or superseded by others.
- I suggest to add `wsl` to the standard repertoire for increased code readbility.

golangci.yaml
alanconway: +1

golangci.yaml
alanconway: +1

golangci.yaml
alanconway: +1

golangci.yaml
alanconway: That one was interesting - I was fully expecting to eliminate it for noise, but actually out our code was already almost 100% compliant! It seems like we share its opinions :) We can remove if it turns out to be a pain.

golangci.yaml
alanconway: exportloopref isn't supported by golangci-lint 1.27 and I am having trouble upgrading it. Can we put this in with scoprelint and upgrade/replace in a subsequent merge?

golangci.yaml
alanconway: +1

golangci.yaml
alanconway: Moved it to the wish list as it is reporting some very valid cut/paste abuse in the tests and I'd rather not take on any more in this merge. Would like to enable later.

golangci.yaml
alanconway: I'm having trouble installing goimports with bingo. I've put it on the commented wish list for now. I'd like to come back to it but I want to get this merge in ASAP as it touches a lot of files and will become a merge-out nightmare quickly.

alanconway:
@periklis agreed on all points. Have addressed some and left others for a future merge. This merge touches a lot of files and will get out of hand quickly if it stays out too long. LGTM?

golangci.yaml
periklis: I am ok with doing this later.

periklis (approved):
LGTM

alanconway:
> LGTM

I'll be needing that in /lgtm form if you please :)

/unhold

periklis (approved):
/lgtm

PR #758, opened 2020-10-23
create hack/create-aws-cluster.sh

blockloop:
This script is a wrapper for openshift-install that I created for myself that
has proven useful for testing. As much as I hate the term 'sane defaults' I
think this is probably a 'sane defaults' version for what we typically use. 
This should give us a simple script to create a cluster without hassling and an
easy entry point for new members of the team and/or new contributors.

/cc @openshift/openshift-team-logging 
/assign @alanconway

### DIFF
--- a/hack/create-aws-cluster.sh
+++ b/hack/create-aws-cluster.sh
@@ -1,0 +1,182 @@
+#!/usr/bin/env bash
+#
+# Author: Brett Jones <blockloop>
+# Purpose: Provide a simple/default set of instructions for easily creating an
+# OpenShift cluster for testing purposes
+
+set -ueo pipefail
+
+ME=$(basename "$0")
+KERBEROS_USERNAME=${KERBEROS_USERNAME:-$(whoami)}
+CLUSTER_DIR=${CLUSTER_DIR:-${HOME}/.openshift/cluster}
+QUIET=${QUIET:-0}
+FORCE=${FORCE:-0}
+SSH_KEY=
+
+usage() {
+	cat <<-EOF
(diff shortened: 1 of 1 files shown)


alanconway (approved):
Suggest rename to `create-aws-cluster`. CRC is a viable alternative and there might be others in future, so we may have other helper scripts at some point.
With that change I'll LGTM.

hack/create-cluster.sh
```diff
@@ -0,0 +1,182 @@
...
+	  networkType: OpenShiftSDN
+	  serviceNetwork:
+	  - 172.30.0.0/16
+	fips: false
+	platform:
+	  aws:
+	    region: us-west-2
+	publish: External
+	pullSecret: |-
+	  $(tr -d '[:space:]' < ~/.docker/config.json )
```
vimalk78, line 132: this should have openshift-install pull secret which is downloaded along with installer. should that pull secret be saved in `~/.docker/config.json ` ?

blockloop: Yes. As part of the google doc pull-secrets section describes.

blockloop:
> Suggest rename to `create-aws-cluster`. CRC is a viable alternative and there might be others in future, so we may have other helper scripts at some point.
> With that change I'll LGTM.

done @alanconway

blockloop:
I actually think this script can be modified slightly to change the type of cluster based on an argument/flag provided since a majority of the YAML stays the same. Either way I'll leave it create-aws-cluster for now and iterate later.

periklis (approved):
/lgtm

PR #763, opened 2020-10-29
Fixes in the must-gather scripts: syntax and curator diagnostics

syedriko:
### Description

- Fixed script syntax
```
./gather_collection_resources: line 66: syntax error near unexpected token `done'
./gather_collection_resources: line 66: `done'
```

- Added installation of the xz package to the Dockerfile
```
nice: 'xz': No such file or directory
```

- Fixed curator diagnostics

/cc @periklis 
/assign @jcantrill

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,6 +23,7 @@ FROM registry.svc.ci.openshift.org/ocp/4.7:base
 RUN INSTALL_PKGS=" \
       openssl \
       rsync \
+      xz \
       " && \
     yum install -y $INSTALL_PKGS && \
     rpm -V $INSTALL_PKGS && \
--- a/must-gather/collection-scripts/common
+++ b/must-gather/collection-scripts/common
@@ -9,7 +9,7 @@ get_env() {
   containers=$(oc -n $ns get po $pod -o jsonpath='{.spec.containers[*].name}')
   for container in $containers
   do
-    dockerfile=$(oc -n $ns exec $pod -c $container -- find /root/buildinfo -name $pattern || :)
+    dockerfile=$(oc -n $ns exec $pod -c $container -- find /root/buildinfo -name $pattern)
     if [ -n "$dockerfile" ]
     then
       echo Image info: $dockerfile > $env_file
--- a/must-gather/collection-scripts/gather_cluster_logging_operator_resources
+++ b/must-gather/collection-scripts/gather_cluster_logging_operator_resources
@@ -10,7 +10,7 @@ if [ "${BASE_COLLECTION_PATH}" = "" ]; then
     BASE_COLLECTION_PATH=$(pwd)
 fi
 
-NAMESPACE="openshift-logging"
+NAMESPACE=openshift-logging
 
 CLO_COLLECTION_PATH="$BASE_COLLECTION_PATH/cluster-logging"
 clo_folder="$CLO_COLLECTION_PATH/clo"
@@ -18,7 +18,7 @@ clo_folder="$CLO_COLLECTION_PATH/clo"
 echo "Gathering data for cluster-logging-operator"
 mkdir -p "$clo_folder"
 
-pods=$(oc -n $NAMESPACE get pods -l name=cluster-logging-operator -o jsonpath={.items[*].metadata.name})
+pods=$(oc -n $NAMESPACE get pods -l name=cluster-logging-operator -o jsonpath='{.items[*].metadata.name}')
 for pod in $pods
 do
     get_env $pod $clo_folder $NAMESPACE "Dockerfile-*operator*"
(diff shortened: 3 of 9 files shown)


syedriko:
@periklis Two questions for you:
- do we need the || : pattern anywhere in the must-gather scripts?
- should we be calling check_curator_connectivity from https://github.com/openshift/cluster-logging-operator/blob/ff68206e4fe49ddb01d639df7322eaf07ad83815/must-gather/collection-scripts/gather_curation_resources#L34? check_curator_connectivity refers to $pod that only makes sense inside the loop.

periklis (commented):
Is this complete? Otherwise LGTM

syedriko:
> Is this complete? Otherwise LGTM

I had a couple more questions we could address on this branch^^^

periklis:
> @periklis Two questions for you:
> 
>     * do we need the || : pattern anywhere in the must-gather scripts?

I believe this is only a leftover from the old `logging-dump.sh` script. I kept the code as is when I transferred it  here.

>     * should we be calling check_curator_connectivity from https://github.com/openshift/cluster-logging-operator/blob/ff68206e4fe49ddb01d639df7322eaf07ad83815/must-gather/collection-scripts/gather_curation_resources#L34
>       ? check_curator_connectivity refers to $pod that only makes sense inside the loop.

That was maybe always wrong. But @ewolinetz / @jcantrill may know best.

syedriko:
cc @jcantrill @ewolinetz

periklis (approved):

Dockerfile
```diff
@@ -23,6 +23,7 @@ FROM registry.svc.ci.openshift.org/ocp/4.7:base
 RUN INSTALL_PKGS=" \
       openssl \
       rsync \
+      xz \
```
jcantrill, line 26: Do we know if this is available downstream? Consider asking ART

must-gather/collection-scripts/common
```diff
@@ -9,7 +9,7 @@ get_env() {
   containers=$(oc -n $ns get po $pod -o jsonpath='{.spec.containers[*].name}')
   for container in $containers
   do
-    dockerfile=$(oc -n $ns exec $pod -c $container -- find /root/buildinfo -name $pattern || :)
+    dockerfile=$(oc -n $ns exec $pod -c $container -- find /root/buildinfo -name $pattern)
```
jcantrill, line 12: Does this not break us on failure of this line?

must-gather/collection-scripts/gather_cluster_logging_operator_resources
```diff
@@ -29,16 +29,16 @@ oc -n $NAMESPACE get deployment cluster-logging-operator -o yaml > $clo_folder/d
...
-oc -n ${NAMESPACE} get secrets -o yaml > ${clo_folder}/secrets.yaml 2>&1 || :
-oc -n ${NAMESPACE} get cronjobs -o yaml > ${clo_folder}/cronjobs.yaml 2>&1 || :
-oc -n ${NAMESPACE} get deployments -o wide > ${clo_folder}/deployments.txt 2>&1 ||    :
-oc -n ${NAMESPACE} get ds -o wide > ${clo_folder}/daemonsets.txt 2>&1 ||    :
-oc -n ${NAMESPACE} get pods -o wide > ${clo_folder}/pods.txt 2>&1 || :
-oc -n ${NAMESPACE} extract secret/elasticsearch --to=${clo_folder} ||:
-oc -n ${NAMESPACE} extract configmap/fluentd --to=${clo_folder} ||:
-oc -n ${NAMESPACE} extract configmap/secure-forward --to=${clo_folder} ||:
-oc -n ${NAMESPACE} extract secret/secure-forward --to=${clo_folder} ||:
+oc -n $NAMESPACE get logforwarding instance -o yaml > "${clo_folder}/logforwarding_cr"
```
jcantrill, line 32: Does this not break us on failure of this line?

must-gather/collection-scripts/common
syedriko: There's no set -e in these scripts. if the above line fails, dockerfile will be empty and the following if will skip over it.

must-gather/collection-scripts/gather_cluster_logging_operator_resources
syedriko: There's no set -e. This line does fail if there's no LF instance, but the script keeps on going.

Dockerfile
syedriko: It is. The downstream base image openshift-ose-base:v4.6.0.20201030.115241 is based on UBI 8.2, which in turn does provide the xz package according to https://catalog.redhat.com/software/containers/ubi8/ubi-minimal/5c359a62bed8bd75a2c3fba8?container-tabs=packages .

PR #765, opened 2020-10-29
Bug 1888958: Store secrets in one place and utilize mutex

jcantrill:
### Description
This PR updates the certificate generation code to:
* Add info message why certs are being (re)generated
* Adds a mutex to the cert code to ensure changes are made in a single threaded fashion
* Moves stashing of generated certs to the "master-certs" secret

/assign @syedriko @alanconway 

### Links
* https://bugzilla.redhat.com/show_bug.cgi?id=1888958

### DIFF
--- a/pkg/k8shandler/certificates.go
+++ b/pkg/k8shandler/certificates.go
@@ -2,164 +2,98 @@ package k8shandler
 
 import (
 	"fmt"
-	"os/exec"
+	"io/ioutil"
+	"path"
+	"sync"
 
 	"github.com/ViaQ/logerr/log"
 	"github.com/openshift/cluster-logging-operator/pkg/certificates"
 	"github.com/openshift/cluster-logging-operator/pkg/constants"
 	"github.com/openshift/cluster-logging-operator/pkg/utils"
 	"k8s.io/apimachinery/pkg/api/errors"
-
-	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-var deprecatedKeys = sets.NewString("app-ca", "app-key", "app-cert", "infra-ca", "infra-key", "infra-cert")
+var mutex sync.Mutex
 
-// golang doesn't allow for const maps
-var secretCertificates = map[string]map[string]string{
-	"master-certs": {
-		"masterca":  "ca.crt",
-		"masterkey": "ca.key",
-	},
-	"elasticsearch": {
-		"elasticsearch.key": "elasticsearch.key",
-		"elasticsearch.crt": "elasticsearch.crt",
-		"logging-es.key":    "logging-es.key",
-		"logging-es.crt":    "logging-es.crt",
-		"admin-key":         "system.admin.key",
-		"admin-cert":        "system.admin.crt",
-		"admin-ca":          "ca.crt",
-	},
-	"kibana": {
-		"ca":   "ca.crt",
-		"key":  "system.logging.kibana.key",
-		"cert": "system.logging.kibana.crt",
-	},
-	"kibana-proxy": {
-		"server-key":     "kibana-internal.key",
-		"server-cert":    "kibana-internal.crt",
-		"session-secret": "kibana-session-secret",
-	},
-	"curator": {
-		"ca":       "ca.crt",
-		"key":      "system.logging.curator.key",
-		"cert":     "system.logging.curator.crt",
-		"ops-ca":   "ca.crt",
-		"ops-key":  "system.logging.curator.key",
-		"ops-cert": "system.logging.curator.crt",
-	},
-	"fluentd": {
-		"ca-bundle.crt": "ca.crt",
-		"tls.key":       "system.logging.fluentd.key",
-		"tls.crt":       "system.logging.fluentd.crt",
-	},
+//Syncronize blocks single threads access using the certificate mutex
+func Syncronize(action func() error) error {
+	mutex.Lock()
+	defer mutex.Unlock()
+	return action()
 }
 
-func (clusterRequest *ClusterLoggingRequest) extractSecretToFile(secretName string, key string, toFile string) (err error) {
-	secret, err := clusterRequest.GetSecret(secretName)
+func (clusterRequest *ClusterLoggingRequest) extractMasterCerts() (err error) {
+	secret, err := clusterRequest.GetSecret(constants.MasterCASecretName)
 	if err != nil {
 		if errors.IsNotFound(err) {
+			return nil
+		}
+		return fmt.Errorf("Unable to get secret %s: %v", constants.MasterCASecretName, err)
+	}
+	workDir := utils.GetWorkingDir()
+	for name, value := range secret.Data {
+		if err != utils.WriteToWorkingDirFile(path.Join(workDir, name), value) {
 			return err
 		}
-		return fmt.Errorf("Unable to extract secret %s to file: %v", secretName, err)
 	}
 
-	value, ok := secret.Data[key]
-
-	// check to see if the map value exists
-	if !ok {
-		if deprecatedKeys.Has(key) {
-			log.Info("No secret data found. Please be aware but likely not an issue for deprecated keys", "key", key, "deprecatedKeys", deprecatedKeys.List())
-		} else {
-			log.Info("No secret data found", key)
-
-		}
-		return nil
-	}
-
-	return utils.WriteToWorkingDirFile(toFile, value)
+	return nil
 }
 
 func (clusterRequest *ClusterLoggingRequest) writeSecret() (err error) {
 
+	secrets, err := loadFilesFromWorkingDir()
+	if err != nil {
+		return err
+	}
 	secret := NewSecret(
 		constants.MasterCASecretName,
 		clusterRequest.Cluster.Namespace,
-		map[string][]byte{
-			"masterca":  utils.GetWorkingDirFileContents("ca.crt"),
-			"masterkey": utils.GetWorkingDirFileContents("ca.key"),
-		})
-
+		secrets,
+	)
 	utils.AddOwnerRefToObject(secret, utils.AsOwner(clusterRequest.Cluster))
 
-	err = clusterRequest.CreateOrUpdateSecret(secret)
+	return clusterRequest.CreateOrUpdateSecret(secret)
+}
+
+func loadFilesFromWorkingDir() (map[string][]byte, error) {
+	workDir := utils.GetWorkingDir()
+	files, err := ioutil.ReadDir(workDir)
 	if err != nil {
-		return
+		return nil, err
 	}
-
-	return nil
-}
-
-func (clusterRequest *ClusterLoggingRequest) readSecrets() (err error) {
-
-	for secretName, certMap := range secretCertificates {
-		if err = clusterRequest.extractCertificates(secretName, certMap); err != nil {
-			return
+	results := map[string][]byte{}
+	for _, f := range files {
+		content := utils.GetFileContents(path.Join(workDir, f.Name()))
+		if content != nil {
+			results[f.Name()] = content
+		} else {
+			log.V(0).Info("The content is nil for certificate file", "file", f.Name())
 		}
 	}
-
-	return nil
-}
-
-func (clusterRequest *ClusterLoggingRequest) extractCertificates(secretName string, certs map[string]string) (err error) {
-
-	for secretKey, certPath := range certs {
-		if err = clusterRequest.extractSecretToFile(secretName, secretKey, certPath); err != nil {
-			if errors.IsNotFound(err) {
-				return nil
-			}
-			return
-		}
-	}
-
-	return nil
+	return results, nil
 }
 
 //CreateOrUpdateCertificates for a cluster logging instance
 func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCertificates() (err error) {
+	return Syncronize(func() error {
+		if err = clusterRequest.extractMasterCerts(); err != nil {
+			fmt.Printf("Error %v", err)
+			return err
+		}
 
-	// Pull master signing cert out from secret in logging.Spec.SecretName
-	if err = clusterRequest.readSecrets(); err != nil {
-		return
-	}
+		scriptsDir := utils.GetScriptsDir()
+		updated := false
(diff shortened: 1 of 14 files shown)


scripts/cert_generation.sh
```diff
@@ -265,3 +299,13 @@ generate_certs 'system.admin'
 generate_certs 'kibana-internal' "$(generate_extensions false false kibana)"
 generate_certs 'elasticsearch' "$(generate_extensions true true $LOG_STORE{,-cluster}{,.${NAMESPACE}.svc}{,.cluster.local})"
 generate_certs 'logging-es' "$(generate_extensions false true $LOG_STORE{,.${NAMESPACE}.svc}{,.cluster.local})"
+
+if [ ! -f "${WORKING_DIR}/kibana-session-secret" ] || [ "$(du -b "${WORKING_DIR}/kibana-session-secret" | cut -f1)" = 0 ] ; then
```
syedriko, line 303:
```
[ ! -f "${WORKING_DIR}/kibana-session-secret" ] || [ "$(du -b "${WORKING_DIR}/kibana-session-secret" | cut -f1)" = 0 ] 
``` 
is better expressed as
```
[ ! -s "${WORKING_DIR}/kibana-session-secret" ]
```

scripts/cert_generation.sh
```diff
@@ -265,3 +299,13 @@ generate_certs 'system.admin'
...
+  info "Generating kibana session secret"
+  python -c "import string,random;print(''.join(random.choices(string.ascii_letters + string.digits, k=32)))" >  "${WORKING_DIR}/kibana-session-secret"
+fi
+
+if [ -f ${WORKING_DIR}/action.log ] ; then
+  cat ${WORKING_DIR}/action.log
+  rm ${WORKING_DIR}/action.log
+fi
```
syedriko, line 311: Missing newline here

pkg/k8shandler/certificates.go
```diff
@@ -2,138 +2,81 @@ package k8shandler
...
-		map[string][]byte{
-			"masterca":  utils.GetWorkingDirFileContents("ca.crt"),
-			"masterkey": utils.GetWorkingDirFileContents("ca.key"),
-		})
-
+		secrets,
+	)
 	utils.AddOwnerRefToObject(secret, utils.AsOwner(clusterRequest.Cluster))
 
 	err = clusterRequest.CreateOrUpdateSecret(secret)
```
syedriko, line 49: this can be just
```
return clusterRequest.CreateOrUpdateSecret(secret)
```

pkg/k8shandler/certificates.go
```diff
@@ -2,138 +2,81 @@ package k8shandler
...
-	return nil
+	return results, nil
 }
 
 //CreateOrUpdateCertificates for a cluster logging instance
 func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCertificates() (err error) {
-
-	// Pull master signing cert out from secret in logging.Spec.SecretName
-	if err = clusterRequest.readSecrets(); err != nil {
+	mutex.Lock()
```
syedriko, line 71: It is not quite enough to lock this mutex here, we also need to lock it around accesses to the cert files, in 
```
[localhost cluster-logging-operator]$ ack GetWorkingDirFileContents pkg
pkg/k8shandler/logstore.go
144:		"elasticsearch.key": utils.GetWorkingDirFileContents("elasticsearch.key"),
145:		"elasticsearch.crt": utils.GetWorkingDirFileContents("elasticsearch.crt"),
146:		"logging-es.key":    utils.GetWorkingDirFileContents("logging-es.key"),
147:		"logging-es.crt":    utils.GetWorkingDirFileContents("logging-es.crt"),
148:		"admin-key":         utils.GetWorkingDirFileContents("system.admin.key"),
149:		"admin-cert":        utils.GetWorkingDirFileContents("system.admin.crt"),
150:		"admin-ca":          utils.GetWorkingDirFileContents("ca.crt"),

pkg/k8shandler/curation.go
172:			"ca":       utils.GetWorkingDirFileContents("ca.crt"),
173:			"key":      utils.GetWorkingDirFileContents("system.logging.curator.key"),
174:			"cert":     utils.GetWorkingDirFileContents("system.logging.curator.crt"),
175:			"ops-ca":   utils.GetWorkingDirFileContents("ca.crt"),
176:			"ops-key":  utils.GetWorkingDirFileContents("system.logging.curator.key"),
177:			"ops-cert": utils.GetWorkingDirFileContents("system.logging.curator.crt"),

pkg/k8shandler/visualization.go
146:			"ca":   utils.GetWorkingDirFileContents("ca.crt"),
147:			"key":  utils.GetWorkingDirFileContents("system.logging.kibana.key"),
148:			"cert": utils.GetWorkingDirFileContents("system.logging.kibana.crt"),
160:	sessionSecret = utils.GetWorkingDirFileContents("kibana-session-secret")
170:			"server-key":     utils.GetWorkingDirFileContents("kibana-internal.key"),
171:			"server-cert":    utils.GetWorkingDirFileContents("kibana-internal.crt"),

pkg/k8shandler/fluentd.go
287:			"ca-bundle.crt": utils.GetWorkingDirFileContents("ca.crt"),
288:			"tls.key":       utils.GetWorkingDirFileContents("system.logging.fluentd.key"),
289:			"tls.crt":       utils.GetWorkingDirFileContents("system.logging.fluentd.crt"),
```

pkg/k8shandler/certificates.go
```diff
@@ -2,138 +2,81 @@ package k8shandler
...
+	defer mutex.Unlock()
+	if err = clusterRequest.extractMasterCerts(); err != nil {
+		fmt.Printf("Error %v", err)
 		return
 	}
 
 	scriptsDir := utils.GetScriptsDir()
-	if err = GenerateCertificates(clusterRequest.Cluster.Namespace, scriptsDir, "elasticsearch", utils.DefaultWorkingDir); err != nil {
+	if err = GenerateCertificates(clusterRequest.Cluster.Namespace, scriptsDir, "elasticsearch", utils.GetWorkingDir()); err != nil {
```
syedriko, line 79: If the script could indicate by its return value that it didn't change anything, we could skip writing the secret and cut down on API server traffic quite a bit.

pkg/k8shandler/certificates.go
```diff
@@ -2,138 +2,81 @@ package k8shandler
...
-		"tls.key":       "system.logging.fluentd.key",
-		"tls.crt":       "system.logging.fluentd.crt",
-	},
-}
+var mutex sync.Mutex
 
-func (clusterRequest *ClusterLoggingRequest) extractSecretToFile(secretName string, key string, toFile string) (err error) {
-	secret, err := clusterRequest.GetSecret(secretName)
+func (clusterRequest *ClusterLoggingRequest) extractMasterCerts() (err error) {
+	secret, err := clusterRequest.GetSecret(constants.MasterCASecretName)
```
syedriko, line 19: When the new code runs for the first time after an upgrade, it'll dump files "masterca" and "masterkey" on disk and keep them circulating. Just might account for that and skip them altogether.

pkg/k8shandler/certificates.go
syedriko, line 49: this can be just
```
return clusterRequest.CreateOrUpdateSecret(secret)
```

pkg/k8shandler/certificates.go
```diff
@@ -2,138 +2,81 @@ package k8shandler
...
+	return results, nil
 }
 
 //CreateOrUpdateCertificates for a cluster logging instance
 func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCertificates() (err error) {
-
-	// Pull master signing cert out from secret in logging.Spec.SecretName
-	if err = clusterRequest.readSecrets(); err != nil {
+	mutex.Lock()
+	defer mutex.Unlock()
```
syedriko, line 72: Checking for 0-length files while populating per-component secrets was one of the bullet points, too.

syedriko:
@jcantrill Added error checking to reading of ES certs:
https://github.com/syedriko/cluster-logging-operator/blob/bz_1888958/pkg/k8shandler/logstore.go#L155

scripts/cert_generation.sh
jcantrill: Added... but why does it matter if there is a newline after `fi`

syedriko: It makes GitHub's diff happy :)

pkg/k8shandler/certificates.go
jcantrill: updated as we discussed

pkg/k8shandler/certificates.go
jcantrill: i believe the changes to the script resolve this

pkg/k8shandler/certificates.go
```diff
@@ -2,163 +2,111 @@ package k8shandler
...
-	for secretKey, certPath := range certs {
-		if err = clusterRequest.extractSecretToFile(secretName, secretKey, certPath); err != nil {
-			if errors.IsNotFound(err) {
-				return nil
-			}
-			return
-		}
+	results := map[string][]byte{}
+	for _, f := range files {
+		results[f.Name()] = utils.GetFileContents(path.Join(workDir, f.Name()))
```
syedriko, line 67: We could use a check for GetFileContents() failing here.

pkg/k8shandler/certificates.go
```diff
@@ -2,163 +2,111 @@ package k8shandler
...
 	log.V(3).Info("Running script", "script", script, "workDir", workDir, "namespace", namespace, "logStoreName", logStoreName)
 	cmd := exec.Command(script, workDir, namespace, logStoreName)
-	result, err := cmd.Output()
-
-	if log.V(3).Enabled() {
-		log.V(3).Info("cert_generation output", "output", string(result))
-	}
-	if err != nil {
-		log.V(2).Error(err, "Error")
+	out, err := cmd.Output()
```
syedriko, line 105: Should we check for err != nil and bail right here?

pkg/k8shandler/curation.go
```diff
@@ -164,18 +164,23 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateCuratorConfigMap() er
 }
 
 func (clusterRequest *ClusterLoggingRequest) createOrUpdateCuratorSecret() error {
-
-	curatorSecret := NewSecret(
-		"curator",
-		clusterRequest.Cluster.Namespace,
-		map[string][]byte{
+	var secrets = map[string][]byte{}
```
syedriko, line 167: There's no need to allocate a map here
```
var secrets map[string][]byte
```
should do it.

pkg/k8shandler/curation.go
```diff
@@ -164,18 +164,23 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateCuratorConfigMap() er
...
+	Syncronize(func() error {
+		secrets = map[string][]byte{
```
syedriko, line 169: We're not checking for possible errors from GetWorkingDirFileContents()

pkg/k8shandler/certificates.go
alanconway: This is ok as an interim fix but in the longer term a more maintainable approach is  goroutines+channels (or Communicating Sequential Processes as they were called in the 1960s, sometimes a good idea has to wait for the world to catch up) I can do a short talk about it if that's of interest. The best one-liner from the Go folks is:

    "share memory by communicating, don't communicate by sharing memory".

In summary: don't build mutex walls around thread-safe data structures, only to realize - too late - how threads will end up contending for them. Instead design your process as a set of simple loops with local, contention-free, data; then have them communicate their shared concerns over channels.

jcantrill: @syedriko and I discussed the proposed implementation solely as a fix to the current problem with the understanding we will better design cert generation and syncing when moving to replace the script logic with golang lib

pkg/k8shandler/certificates.go
jcantrill: The called function logs an error and returns nil.  Adding check for nil and log message. Skipping adding the file if the content is nil

pkg/k8shandler/curation.go
jcantrill: I think this is OK not to check missing here.  We need a bunch of secrets and we should fail and reconcile again.  The cert gen code should regen if something is missing.  If it does not that is a bug in the regen code IMO

syedriko: But if we don't check we don't fail and stick empty buffers into secrets.

jcantrill: What is the action we can take in this situation?  What do we believe will change during a subsequent reconciliation that would resolve the issue?  The best we can do is really generate a log message and/or update status.  This particular file is probably moot as we want to remove curation from CLO but  if we are to do anything then it should be applied to the other places where we take similar action

syedriko: We can abort, refuse to update the secret and potentially make our situation worse. GetWorkingDirFileContents() is not very nuanced in how it reports errors, but there can be a genuine I/O failure that will make GetWorkingDirFileContents() return nil while there's a good cert in the secret as well as on disk. By proceeding we make things worse by wiping out a good cert from a secret. We will recover in time, but this is a disruption we can avoid.

scripts/cert_generation.sh
```diff
@@ -265,3 +299,18 @@ generate_certs 'system.admin'
 generate_certs 'kibana-internal' "$(generate_extensions false false kibana)"
 generate_certs 'elasticsearch' "$(generate_extensions true true $LOG_STORE{,-cluster}{,.${NAMESPACE}.svc}{,.cluster.local})"
 generate_certs 'logging-es' "$(generate_extensions false true $LOG_STORE{,.${NAMESPACE}.svc}{,.cluster.local})"
+
+if [ ! -s "${WORKING_DIR}/kibana-session-secret" ] ; then
+  info "Generating kibana session secret"
+  result=$(python -c "import string,random;print(''.join(random.choice(string.ascii_letters + string.digits) for i in range(32)))")
```
syedriko, line 305: This achieves almost the same end, with a shorter alphabet of letters, without the Python dependency
```
[root@e33867c18033 cluster-logging-operator]# dd if=/dev/urandom count=1 ibs=16 status=none | hexdump -e '"%02X"'
4306BCE91BC0C99B7AB482ECC3E73047
```

jcantrill: updated to remove python dep

PR #767, opened 2020-11-01
Improvements to test support packages

alanconway:
Clearer object IDs.
Begin/end logging of client operations (at V(7),V(8)) with timing.  
Cleaned up dead code.
Environment variables SUCCESS_TIMEOUT FAILURE_TIMEOUT to control test timeouts.

/cc jcantrill


jcantrill (commented):
/lgtm

PR #768, opened 2020-11-02
Add license header for *.go files

vparfonov:

### Description
Add license header for *.go files

/cc 
/assign 

/cherry-pick


jcantrill:
It would be awsome to have either a "pre-check" or linter that would wither fix files that don't have the license or fail the build

vparfonov:
> It would be awsome to have either a "pre-check" or linter that would wither fix files that don't have the license or fail the build

will do

blockloop:
The Apache 2.0 license does not require the header to exist on every source file, but "recommends" it. I think if we're not legally required to do this then we should avoid having to deal with it altogether.

PR #770, opened 2020-11-04
Fix 'fmt' for Mac OS users

vparfonov:

### Description

`-r` (`--no-run-if-empty`)  disables execution if output is empty,
remove '-r' from xargs options, output will never be empty

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s): N/A
- Bugzilla: N/A
- Github issue: N/A
- JIRA: N/A
- Enhancement proposal: N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,7 @@ lint: $(GOLANGCI_LINT)
 
 fmt:
 	@echo gofmt		# Show progress, real gofmt line is too long
-	find pkg cmd test -name '*.go' | xargs -r gofmt -s -l -w
+	find pkg cmd test -name '*.go' | xargs gofmt -s -l -w
 
 # Do all code/CRD generation at once, with timestamp file to check out-of-date.
 GEN_TIMESTAMP=.zz_generate_timestamp


Makefile
```diff
@@ -1,5 +1,6 @@
 # Define the target to run if make is called with no arguments.
 default: check
+UNAME_S := $(shell uname -s)
```
periklis, line 3: This variable implies the `OS`, so let's call it so. Also a good practice is to lowercase everything `shell uname -s | tr '[A-Z]' '[a-z]'`

Makefile
```diff
@@ -96,7 +97,13 @@ lint: $(GOLANGCI_LINT)
 
 fmt:
 	@echo gofmt		# Show progress, real gofmt line is too long
-	find pkg cmd test -name '*.go' | xargs -r gofmt -s -l -w
+	@if [ $(UNAME_S) == 'Linux' ] ; then \
+		find pkg cmd test -name '*.go' | xargs -r gofmt -s -l -w; \
+	fi
+	#BSD (MacOS) xargs does not have option -r (--no-run-if-empty)
+	@if [ $(UNAME_S) = 'Darwin' ] ; then \
+		find pkg cmd test -name '*.go' | xargs gofmt -s -l -w; \ #BSD xargs does not have this behavior, so there's no need to disable it.
```
periklis, line 105: If the whole fix is about `xargs`, could you check if you could use the one from brew?

periklis (changes requested):

Makefile
vparfonov: I think is nothing to fix in `xargs` side, just different behavior

Makefile
```diff
@@ -96,7 +97,13 @@ lint: $(GOLANGCI_LINT)
...
 fmt:
 	@echo gofmt		# Show progress, real gofmt line is too long
-	find pkg cmd test -name '*.go' | xargs -r gofmt -s -l -w
+	@if [ $(OS_NAME) == 'Linux' ] ; then \
+		find pkg cmd test -name '*.go' | xargs -r gofmt -s -l -w; \
+	fi
+	# BSD (MacOS) xargs does not have option -r (--no-run-if-empty)
+	@if [ $(OS_NAME) = 'Darwin' ] ; then \
+		find pkg cmd test -name '*.go' | xargs gofmt -s -l -w; \
+	fi
```
periklis, line 106: The checks should use now lowercase, too, e.g. `linux`.

periklis (changes requested):

Makefile
periklis: If you use the xargs from homebrew aka from GNU, you should get the same behavior right?

Makefile
vparfonov: done

Makefile
```diff
@@ -96,7 +97,13 @@ lint: $(GOLANGCI_LINT)
 
 fmt:
 	@echo gofmt		# Show progress, real gofmt line is too long
-	find pkg cmd test -name '*.go' | xargs -r gofmt -s -l -w
+	@if [ $(OS_NAME) == 'linux' ] ; then \
+		find pkg cmd test -name '*.go' | xargs -r gofmt -s -l -w; \
+	fi
+	# BSD (MacOS) xargs does not have option -r (--no-run-if-empty)
+	@if [ $(OS_NAME) = 'darwin' ] ; then \
+		find pkg cmd test -name '*.go' | xargs gofmt -s -l -w; \
```
alanconway, line 105: Actually we can just delete the -r, it'll never be empty (or if it is, somebody's deleted the entire code-base, which probably deserves to be an error!) I was coming from writing a different script where the -r was needed and I put it by reflex.

alanconway (changes requested):

Makefile
vparfonov: thx

periklis (commented):
LGTM, please squash the commits.

PR #773, opened 2020-11-05
Remove need for LOGGING_SHARE_DIR in tests.

alanconway:
Use the current working directory to locate /cluster-logging-operator/files.
Drop the env var LOGGING_SHARE_DIR.
Allows all tests to run directly with `go test`, with no extra environment settings.

/cc @blockloop
/assign @alanconway

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 # Define the target to run if make is called with no arguments.
 default: check
 
-export LOG_LEVEL?=9
 export KUBECONFIG?=$(HOME)/.kube/config
 
 export GOBIN=$(CURDIR)/bin
@@ -64,7 +63,7 @@ build-debug:
 
 # Run the CLO locally - see HACKING.md
 RUN_CMD?=go run
-run: 
+run:
 	@ls $(MANIFESTS)/*crd.yaml | xargs -n1 oc apply -f
 	@mkdir -p $(CURDIR)/tmp
 	CURATOR_IMAGE=quay.io/openshift/origin-logging-curator:latest \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -1,7 +1,6 @@
 # Define the target to run if make is called with no arguments.
 default: check
 
-export LOG_LEVEL?=fatal
```
jcantrill, line 4: I recently added a change to set this to "9" in either another PR or commit... as this is useful IMO. maybe we should consider lowing the level and still make it overridable

pkg/utils/utils.go
```diff
@@ -181,12 +182,21 @@ func GetFileContents(filePath string) []byte {
...
 
+const defaultShareDir = "/usr/share/logging"
+
 func GetShareDir() string {
-	shareDir := os.Getenv("LOGGING_SHARE_DIR")
-	if shareDir == "" {
-		return "/usr/share/logging"
+	// shareDir is <repo-root>/files - try to find from working dir.
+	const sep = string(os.PathSeparator)
+	const repoRoot = sep + "cluster-logging-operator" + sep
```
jcantrill, line 190: When is this ever true?  I recall even the built image writes to something like `/go/src/openshift/....`

Makefile
alanconway: We don't need to set it in Makefile for it to be overridable, whoever runs make can set in their env. I suggest we use the default by default (i.e. don't set it), and if someone wants noisy tests they can `LOG_LEVEL=9 make check` or whatever.

pkg/utils/utils.go
alanconway: Dunno, I was copying the old behaviour which uses this as the final fallback. If it's never used then we should perhaps make it an an error/panic if we get this far.

blockloop: You should use fmt.Sprintf instead string + string + string

alanconway: Can I have an LGTM?

blockloop (approved):
/lgtm

PR #781, opened 2020-11-10
Updated the CentOS Dockerfile patch

syedriko:
### Description
Updated the Dockerfile patch based on CentOS for upstream builds to align better with the downstream - Go 1.15 and CentOS 8.

/cc @jcantrill 
/assign @alanconway


jcantrill (commented):
/approve
/lgtm

PR #751, opened 2020-10-20
Update log lib and Fix LOG_LEVEL and documentation to reflect verbosity as an int

jcantrill:
This PR:

* Fix LOG_LEVEL and documentation to reflect verbosity as an int
* Updates the dependent log adapter library

cc @vparfonov @blockloop

PR #782, opened 2020-11-10
LOG-994: functional benchmarker

jcantrill:
This PR provides an internal command for benchmarking collector configurations

cc @alanconway 

ref: https://issues.redhat.com/browse/LOG-994

```
./bin/functional-benchmarker --sample
Sample:
{
  "docker": {
    "container_id": "4215905f05d3f8133fd9185755ead3d7e5bbb782da7ddf7272ae5944e6db019e"
  },
  "kubernetes": {
    "container_name": "fluentd",
    "container_image": "quay.io/openshift/origin-logging-fluentd:latest",
    "container_image_id": "quay.io/openshift/origin-logging-fluentd@sha256:cefe1014ef0c531b720abdad1d85ebca1ad29b0fa5c8225c80d3f60cd5a8eb4d",
    "pod_name": "functional",
    "pod_id": "3f964382-04f3-4fbc-ab91-02484e9e3b09",
    "host": "ip-10-0-130-38.us-east-2.compute.internal",
    "labels": null,
    "flat_labels": [
      "test-client=true",
      "testname=functional",
      "testtype=functional"
    ],
    "master_url": "https://kubernetes.default.svc",
    "namespace_name": "testhack-1709398484b528",
    "namespace_id": "01f08ae0-37b6-4f7d-b950-fe141160460f"
  },
  "message": "6urHqtq5i0Jc8JQrjQ7eBM7XXf4nThz3GMi2gULgq3VIshmSr8YREbfde7LlUNicjsgRuAZbxa16oFYEMQY1mK5KCzJ6mTPEBo22bEjlynp29PkAyEEQwPYde9HXQNzoSQESXlp4PiI77rOAqxnt7FhmFlmmUqH4hW0KgsQRRFF6Ib9jfs6vyOEpJ6Ul4H5k3sEFzhu7OYiU2Hcjuh8aHjX0YMvxZAsaA6nJNcET0CPyk2znT4bKeS2oqj2Ww87WDTzqvqWZpXgE6tM8r4jNNCcAZf63k1aw4FHsaEXlsMmEQmmE2DKGJIdxNoZcdRdRQtjeeJY7LdFUl1FWeanMtzPMDHUfDuL2Ar8JfsgAGwYnU2D49q4ZLQpXBM6VteCrJMkq9MIYwXGNJLp6o6068hEoJd69HoSx00hOjStt8QNo62Qx55D8GlutYIrPsZfo8liBRCDJGUeDkmkG6rqTFNX7nN2n4R9bavD2BKywHL3dFPHrVv8dmEAUJS4XDzpRrDoY6TOQMq7AqFheF6lP9nWYaEKYWyIyTjVTvLzT4OJfDQXjH4sC9mYTnCnTPrsAirRbAyQHg5fnzgtMamWKWsavJEYpFh9lY3b3umhYSmJpIG5CAjfQojK4Xc4DXG8rawv6Y2pwQdubJwNj0ltuxXidOVBBQLiXlj9D5j1XVnAvobfDoqHxYhae0BrWu7Sfeas5vVw251U0zScW6RLtcEjRkt8Oy8i7N1f2zGlvI5G0xJr3K3UMwcoiBA0EiMc91Am1vhWEV3kCRlJMmYMZJzlBhd2Pthf616sBf4TRDUT3MV3TvXlvxLvEL7SMpYsLYyrZTpSoKeCFfNVQfcGOmnRFbl7mg5JiMMefoLx51sRmGjFvMV57JHTvY5JVDkT7Eqwalau0jKEU2RgD71snKWFeuMSYzBb0AxLL56DsFUx8C8xrVwAVd2OCUim3MB0d4mdPsZFqvXKocuOYLDeK6KqzolEjlmB4ZfxOzErN7WZQQbMWoSxXoUszmKp9R59y",
  "level": "info",
  "hostname": "ip-10-0-130-38.us-east-2.compute.internal",
  "pipeline_metadata": {
    "collector": {
      "ipaddr4": "10.128.2.10",
      "ipaddr6": "",
      "inputname": "fluent-plugin-systemd",
      "name": "fluentd",
      "received_at": "2020-11-17T22:09:53.054485+00:00",
      "version": "1.7.4 1.6.0"
    }
  },
  "@timestamp": "2020-11-17T22:09:48.284723+00:00",
  "viaq_index_name": "app-write",
  "viaq_msg_id": "MDJkMjJjYTItZDJmNi00Y2U0LTk5NmUtYjg2Yjg3Njg3NzRi",
  "openshift": {
    "labels": {}
  },
  "epoc_in": 1605650988.284723,
  "epoc_out": 1605650996.6136672
}
  Total Msg: 10000
Size(bytes): 1024
 Elapsed(s): 29.67829061s
    Mean(s): 3.679317
     Min(s): 0.226056
     Max(s): 8.329003
  Median(s): 3.427630
 Mean Bloat: 2.179606

```


jcantrill:
needs rebase off https://github.com/openshift/cluster-logging-operator/pull/751

internal/cmd/functional-benchmarker/main.go
```diff
@@ -0,0 +1,128 @@
...
+	fmt.Printf("Size(bytes): %d\n", *msgSize)
+	fmt.Printf(" Elapsed(s): %s\n", endTime.Sub(startTime))
+	fmt.Printf("    Mean(s): %f\n", mean(jsonlogs))
+	fmt.Printf("     Min(s): %f\n", min(jsonlogs))
+	fmt.Printf("     Max(s): %f\n", max(jsonlogs))
+	fmt.Printf("  Median(s): %f\n", median(jsonlogs))
+	fmt.Printf(" Mean Bloat: %f\n", meanBloat(jsonlogs))
+}
+
+func meanBloat(logs helpers.Logs) float64 {
```
syedriko, line 93: All the statistics functions blow up with empty `logs`, let's add a check for that before we do anything else
```
if len(logs) == 0 {
    return 0
}
```

internal/cmd/functional-benchmarker/main.go
```diff
@@ -0,0 +1,128 @@
...
+func min(logs helpers.Logs) float64 {
+	diffs := sortLogsByTimeDiff(logs)
+	return diffs[0]
+}
+func max(logs helpers.Logs) float64 {
+	diffs := sortLogsByTimeDiff(logs)
+	return diffs[len(diffs)-1]
+}
+func sortLogsByTimeDiff(logs helpers.Logs) []float64 {
+	diffs := []float64{}
```
syedriko, line 122: `diffs := make([]float64, len(logs))`

internal/cmd/functional-benchmarker/main.go
```diff
@@ -0,0 +1,128 @@
...
+	for _, e := range logs {
+		diffs = append(diffs, e.Difference())
```
syedriko, line 124:
```
for i, e := range logs {
		diffs[i] = e.Difference()
```
This will make it O(n) as opposed to O(n^2)

internal/cmd/functional-benchmarker/main.go
jcantrill: to improve efficiency?

syedriko: Yep. It goes together with the next comment

internal/cmd/functional-benchmarker/main.go
```diff
@@ -0,0 +1,131 @@
...
+	msgSize := flag.Uint64("size", 1024, "The message size in bytes")
+	verbosity := flag.Int("verbosity", 0, "")
+	doCleanup := flag.Bool("docleanup", true, "set to false to preserve the namespace")
+	sample := flag.Bool("sample", false, "set to true to dump a sample message")
+
+	flag.Parse()
+
+	log.MustInit("functional-benchmark")
+	log.SetLogLevel(*verbosity)
+	log.V(1).Info("Args: %v", os.Args)
```
syedriko, line 35: Logger.Info() works a bit differently (https://github.com/go-logr/logr/blob/master/logr.go). This needs to be something like
```
log.V(1).Info("Starting functional benchmarker", "command line", os.Args)
```
which will log
```
{"command line":["./bin/functional-benchmarker","-verbosity=9"],"component":"functional-benchmark","message":"Starting functional benchmarker","ts":"2020-12-02T19:44:00.546444871Z"}
```

The way it is now logs
```
{"component":"functional-benchmark","message":"Args: %v","ts":"2020-12-02T20:14:02.68607955Z"}
```

internal/cmd/functional-benchmarker/main.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+	jsonlogs, err := helpers.ParseLogs(fmt.Sprintf("[%s]", strings.Join(logs, ",")))
+	if err != nil {
+		log.Error(err, "Error parsing logs")
+		os.Exit(1)
+	}
+	if *sample {
+		fmt.Printf("Sample:\n%s\n", test.JSONString(jsonlogs[0]))
+	}
+	fmt.Printf("  Total Msg: %d\n", *totalMessages)
```
syedriko, line 83: Here's a more efficient implementation of the statistics functions. 
- sorting is only done once
- `for`/range is replaced by a simple `for` to avoid copying large AllLog structs from the `logs` slice into the local `e`
- the two `mean` functions refactored the best we can without generics
```
	timeDiffs := sortLogsByTimeDiff(jsonlogs)
	fmt.Printf("  Total Msg: %d\n", *totalMessages)
	fmt.Printf("Size(bytes): %d\n", *msgSize)
	fmt.Printf(" Elapsed(s): %s\n", endTime.Sub(startTime))
	fmt.Printf("    Mean(s): %f\n", mean(jsonlogs))
	fmt.Printf("     Min(s): %f\n", min(timeDiffs))
	fmt.Printf("     Max(s): %f\n", max(timeDiffs))
	fmt.Printf("  Median(s): %f\n", median(timeDiffs))
	fmt.Printf(" Mean Bloat: %f\n", meanBloat(jsonlogs))
}

func meanBloat(logs helpers.Logs) float64 {
	return genericMean(logs, (*helpers.AllLog).Bloat)
}

func mean(logs helpers.Logs) float64 {
	return genericMean(logs, (*helpers.AllLog).Difference)
}

func genericMean(logs helpers.Logs, f func(l *helpers.AllLog) float64) float64 {
	if len(logs) == 0 {
		return 0
	}
	var total float64
	for i := 0; i < len(logs); i++ {
		total += f(&logs[i])
	}
	return total / float64(len(logs))
}

func median(diffs []float64) float64 {
	if len(diffs) == 0 {
		return 0
	}
	return diffs[(len(diffs)/2)+1]
}

func min(diffs []float64) float64 {
	if len(diffs) == 0 {
		return 0
	}
	return diffs[0]
}

func max(diffs []float64) float64 {
	if len(diffs) == 0 {
		return 0
	}
	return diffs[len(diffs)-1]
}

func sortLogsByTimeDiff(logs helpers.Logs) []float64 {
	diffs := make([]float64, len(logs))
	for i := 0; i < len(logs); i++ {
		diffs[i] = logs[i].Difference()
	}
	sort.Slice(diffs, func(i, j int) bool { return diffs[i] < diffs[j] })
	return diffs
}
```

syedriko: Submitted a PR for this: https://github.com/jcantrill/cluster-logging-operator/pull/3

jcantrill: updated

alanconway (changes requested):
We need a few more stats:
- percentiles: median, 99%, 99.99%, 99.999%
- deviation
- stats for latency - can use timestamp + current time as estimate. Need to make sure clocks are reasonably synced.

Percentiles are more useful than the mean. All the standard math around means and std-deviations (T tests, etc.) only work well if the data is normally distributed, which this data almost certainly won't be. They can be very misleading on non-normal data sets.
 
Best way to calculate stats is to do a fast dump of all the data points with current time to a file during the benchmark, then do the stats when the benchmark is over. Doing stats during the benchmark slows you down and can skew the results.

Here's a benchmarking tool from a previous project, take a quick look at the options and output: https://github.com/ssorj/quiver

We absolutely need stats for latency. You can easily generate throughtput numbers in a short-duration benchmark that are completely unsustainable in production because the benchmark can stuff buffers faster than they drain, and then drain them at leisure. A production system at the same througput would blow out buffer limits, lose data and work up large latencies. By measuring latency and throughput together you can get realistic results even in short-duration benchmarks - a given throughput is only sustainable if the latencies stay within bounds during the benchmark.

jcantrill:
> We need a few more stats:
> 
>     * percentiles: median, 99%, 99.99%, 99.999%
> 
>     * deviation
> 
>     * stats for latency - can use timestamp + current time as estimate. Need to make sure clocks are reasonably synced.

Can we add these as needed?

> 
> Percentiles are more useful than the mean. All the standard math around means and std-deviations (T tests, etc.) only work well if the data is normally distributed, which this data almost certainly won't be. They can be very misleading on non-normal data sets.
> 
> Best way to calculate stats is to do a fast dump of all the data points with current time to a file during the benchmark, then do the stats when the benchmark is over. Doing stats during the benchmark slows you down and can skew the results.

Stats and benchmarking are only done at the end of shipping messages through the tool

> We absolutely need stats for latency. You can easily generate throughtput numbers in a short-duration benchmark that are completely unsustainable in production because the benchmark can stuff buffers faster than they drain, and then drain them at leisure. A production system at the same througput would blow out buffer limits, lose data and work up large latencies. By measuring latency and throughput together you can get realistic results even in short-duration benchmarks - a given throughput is only sustainable if the latencies stay within bounds during the benchmark.

My vision for this tool is only for a rough understanding of efficiency of our processing pipelines.  It is not connected to an actual cluster in anyway so I'm not sure these suggestions are relevant. It doesn't take into account context switching and other things that may occur when collecting from multiple containers.  

We need to either:
* Modify the tool to do e2e benchmarking
* Utilize a different tool that may be more appropriate.

jcantrill:
Updated to latest code.  @eranra any idea why this fails the functional test but other jobs on master succeed ?  Seems like the strict validation should be in both places

eranra:
@jcantrill please apply, this will resolve and the functional test will pass
[pr782.diff.txt](https://github.com/openshift/cluster-logging-operator/files/5702980/pr782.diff.txt)

Also more information and slack thread https://coreos.slack.com/archives/GGUR75P60/p1608113567088100?thread_ts=1607964932.496700&cid=GGUR75P60

alanconway:
@jcantrill agreed we can extend later with more stats.

> My vision for this tool is only for a rough understanding of efficiency of our processing pipelines. 

Agreed, but you still need to evaluate both latency and throughput together - one without the other is meaningless. I made this mistake in a previous project and regretted it.

1. You can get dream latencies at 1msg/sec but that's meaningless. Latency will go up when you push the system, buffers fill, CPUs get busy etc. So it's meaningless to say "latency is 100ms", you have to say "latency is <100ms at up to 1Mb/sec throughput".

2. You can get unrealistic throughputs in short run tests if you don't watch the latency (bit me hard with Qpid) Basically your sender fills a growing queue/buffer and your receiver pulls the data off at leisure. Both sender and receiver are running at full capacity so it looks ok, but the queue/buffer is growing continuously so it's not sustainable. You can catch this by looking at latency - on a "good" run latency should be fairly stable, on a "bad" run you'll see be a big spread between the min and max latency as entries spend longer and longer in the growing buffer. You can also catch this by doing very long running tests, but it's valuable to be able to get reliable data quickly during development.

test/client/test.go
```diff
@@ -38,3 +40,24 @@ func (t *Test) Close() {
 		fmt.Printf("============\n\n")
 	}
 }
+
+//HackTest wraps the singleton test client for use with hack testing
```
alanconway, line 44: Call this "NamespaceClient" rather than "HackTest"? Its a client with a namespace that isn't related to any particular test.

test/functional/cluster_log_forwarder.go
```diff
@@ -5,7 +5,7 @@ import (
 )
 
 const (
-	forwardOutputName   = "fluentForward"
+	ForwardOutputName   = "fluentForward"
```
alanconway, line 8: I think there's already a constant for this in apis/logging.

alanconway (changes requested):
The binary forwarder-generator should not be checked-in.

alanconway (commented):
/lgtm

jcantrill:
Manually merging as all tests are passing and merge bot seems to be ignoring me

PR #784, opened 2020-11-11
Functional test for application logforwarding format

eranra:
### Description

This PR adds a new functional test that verifies the formatting of fluentd logforwarding functionality. If injects known log line as input to first fluentd instance and reads the log line back from secondary fluentd post logforwarding formating.

/cc @jcantrill
/assign @jcantrill

### Links
- JIRA:
https://issues.redhat.com/browse/LOG-975


test/functional/logforwarding/suite_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	"github.com/onsi/ginkgo/reporters"
+	. "github.com/onsi/gomega"
+)
+
+func TestFunctionMetrics(t *testing.T) {
```
jcantrill, line 11: TestLogForwarding?

test/functional/framework.go
```diff
@@ -176,8 +176,7 @@ func (f *FluentdFunctionalFramework) addOutputContainers(b *runtime.PodBuilder,
 	return nil
 }
 
-func (f *FluentdFunctionalFramework) WritesApplicationLogs(numOfLogs int) error {
-	msg := "2020-11-04T18:13:59.061892999+00:00 stdout F Functional test message $n"
```
jcantrill, line 180: We may need an alternate function.  I think there is value in:
* writing just N number of messages regardless of format
* writing an explicit message so you can verify an explicit output
* taking a message and having the generator maybe add the current time, whether it is split, or not, etc

eranra: Fixed

test/functional/logforwarding/message_format_test.go
```diff
@@ -0,0 +1,52 @@
...
+		applicationLogLine := fmt.Sprintf("%s stdout F %s $n", timeStamp, message)
+		Expect(framework.WritesMessageToApplicationLogs(applicationLogLine,1)).To(BeNil())
+		// Read line from Log Forward output
+		raw, err := framework.ReadApplicationLogsFrom("fluentforward")
+
+		Expect(err).To(BeNil(), "Expected no errors reading the logs")
+		logs, err := helpers.ParseLogs(raw)
+		Expect(err).To(BeNil(), "Expected no errors parsing the logs")
+
+		Expect(logs[0].Timestamp).Should(Equal(timeStamp), fmt.Sprintf("Expected correct Timestamp in %#v", logs[0]))
```
jcantrill, line 47: I think we should reconcider how we verify.  The intent is to verify the output format.  The result could have missing fields we expect or extra we do not.  Maybe we should deserialize into generic map and compare against that.  Or maybe we should be doing a raw string comparison if we can ensure the output is repeatable

eranra: @jcantrill I created code that uses template on the fields to compare input and output ... the only thing that you need to do now as to test is either (1) specify concreate values to test against or (2) start a string with pre so it will check for pre-fix or (3) write '*' in a field and than we test that there is 'something' in the output field ... 

This seems very flexible + the code for the test is really short now ... just one line that tests against the input with template fields.

test/functional/logforwarding/message_format_test.go
```diff
@@ -0,0 +1,61 @@
...
+				NamespaceName:"*",
+				NamespaceID: "*",
+				OrphanedNamespace: "*"},
+		}
+
+		// Write log line as input to fluentd
+		applicationLogLine := fmt.Sprintf("%s stdout F %s $n", testFormatTemplate.Timestamp, testFormatTemplate.Message)
+		Expect(framework.WritesMessageToApplicationLogs(applicationLogLine,1)).To(BeNil())
+		// Read line from Log Forward output
+		raw, err := framework.ReadApplicationLogsFrom("fluentforward")
```
jcantrill, line 52: While testing my functional benchmarker I had issues with this method returning correctly because of the timestamp was in the past.  I would verify this actually works for  you in that you may need to pass a "Now" timestamp in order for fluent to not buffer the logs elsewhere

test/functional/logforwarding/message_format_test.go
```diff
@@ -0,0 +1,61 @@
...
+		Expect(err).To(BeNil(), "Expected no errors reading the logs")
+		// Parse log line
+		logs, err := helpers.ParseLogs(raw)
+		Expect(err).To(BeNil(), "Expected no errors parsing the logs")
+		// Compare to expected template
+		outputTestLog := logs[0]
+		Expect(outputTestLog.CompareLog(testFormatTemplate)).To(BeTrue(), "Expected output to fit expected format")
```
jcantrill, line 59: Consider converting this into a matcher like https://github.com/openshift/cluster-logging-operator/blob/master/test/matchers/envvar.go#L1

test/helpers/types.go
```diff
@@ -99,3 +102,102 @@ func (l logs) NonEmpty() bool {
...
+		if strings.HasPrefix(value.(string)[4:], templateValue.(string)[4:])  {
+			logger.V(3).Info("CompareLogLogic: Fit prefix ", "name", name, "value", value)
+			return true
+		}
+	}
+
+	return false
+}
+
+func (output Log) CompareLog(input Log) bool {
```
jcantrill, line 168: We should add tests to exercise this code

test/helpers/types.go
```diff
@@ -99,3 +102,102 @@ func (l logs) NonEmpty() bool {
...
+		return true
+	}
+	switch reflect.TypeOf(i).Kind() {
+	case reflect.Ptr, reflect.Map, reflect.Array, reflect.Chan, reflect.Slice:
+		return reflect.ValueOf(i).IsNil()
+	}
+	return false
+}
+
+func DeepFields(iface interface{}, namePrefix string) ([]reflect.Value, []string) {
```
jcantrill, line 117: We should add tests to exercise this code

test/helpers/types.go
jcantrill, line 168: I think turning this into a "matcher" and having some sort of output that explicitly identifies the diffs could be really useful

test/functional/logforwarding/message_format_test.go
eranra: @jcantrill In this test since we are always writing a single log line and reading back a single line there is no issue. This should be extended for more advanced tests like the benchmarking but not as part of this PR. I ran this large amount of times and this was always working for me.

test/functional/logforwarding/message_format_test.go
eranra: Done

test/helpers/types.go
eranra: Done

eranra: Done

test/helpers/types.go
eranra: Done

test/matchers/log_format.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+func DeepFields(iface interface{}, namePrefix string) ([]reflect.Value, []string) {
+	fields := make([]reflect.Value, 0)
+	names :=  make([]string,0)
+
+	ifv := reflect.ValueOf(iface)
+	ift := reflect.TypeOf(iface)
+
+	for i := 0; i < ift.NumField(); i++ {
```
jcantrill, line 61: can you not use a range here:
```
for i, name := range ift.NumField
```

test/matchers/log_format.go
```diff
@@ -0,0 +1,139 @@
...
+func compareLogLogic (name string, templateValue interface{}, value interface{}) bool {
+	if templateValue == value { // Same value is ok
+		logger.V(3).Info("CompareLogLogic: Same value for", "name", name, "value", value)
+		return true
+	}
+	if templateValue == "*" && !isNil(value) { // Any value, not Nil is ok if template value is "*"
+		logger.V(3).Info("CompareLogLogic: Any value for * ", "name", name, "value", value)
+		return true
+	}
+	if strings.HasPrefix(templateValue.(string), "Pre-") { // Any prefix value is ok if template value starts with "Pre-"
```
jcantrill, line 95: Would we be better served and more flexible to utilize regex and maybe do that kind of comparison if the string begins/ends with a slash like `/.foo.*/`

test/matchers/log_format.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	logger.V(3).Info("CompareLogLogic: Mismatch", "name", "templateValue", templateValue, "value", value)
+	return false
+}
+
+func CompareLog(template Log, log Log) (bool, error) {
+	templateFieldValues, templateFieldNames := DeepFields(template,"")
+	logFieldValues, logFieldNames := DeepFields(log,"")
+	for i := 0; i < len(logFieldValues); i++ {
```
jcantrill, line 109: why not range where you can get the number and entry during interation?

test/matchers/log_format.go
```diff
@@ -0,0 +1,139 @@
...
+		logFieldValue := logFieldValues[i].Interface()
+		logFieldName := logFieldNames[i]
+		foundMatchFields := false
+		for j := 0; j < len(templateFieldValues); j++ {
```
jcantrill, line 113: same question here

test/matchers/log_format.go
eranra: Nope ...  NumField returns a struct type's field count. (will not work as a range)

test/matchers/log_format.go
eranra: @jcantrill  ... If this is ok for first version of this PR I want to stay with simple "Pre".  I totally agree that adding regex will help but for the first set of tests prefix is what we need and adding regex will complicate the tests because escape characters do not pass k8s checks and than there will be need to split the field from the template in the test, so making this a bit more complicated.

+ adding more code for regex ... so if this make sense let's keep this as-is for first version and once we will advance we will work on that area. I have more ideas for advanced checks but let's not add this at this cycle

test/matchers/log_format.go
eranra: fixed

test/matchers/log_format.go
eranra: fixed

test/matchers/log_format.go
eranra: Update: I added support for regex and removed "Pre-".  Changed the test code to use.

jcantrill (commented):
/approve
@alanconway please weigh in on the approach here

alanconway (approved):
There are some overlaps with the approach I introduced in https://github.com/openshift/cluster-logging-operator/blob/master/test/e2e/logforwarding/fluent/fluent_test.go but the test is useful so lets get it in now and deal with test uniformity and reducing overlap later - we already have to do that anyway for existing tests so we just add this one to the bag.

/lgtm

jcantrill (commented):
/approve

test/helpers/types/types.go
```diff
@@ -272,11 +277,26 @@ type AllLog struct {
 	Annotations              Annotations      `json:"annotations"`
 	K8SAuditLevel            string           `json:"k8s_audit_level"`
 	OpenshiftAuditLevel      string           `json:"openshift_audit_level"`
+	OpenshiftLabels          OpenshiftMeta    `json:"openshift"`
+}
+
+func StrictlyParseLogs(in string, logs interface{}) error {
+	logger.V(3).Info("ParseLogs", "content", in)
```
vimalk78, line 284: i will suggest not to print the logs here. sometimes the volume of logs can be quite high, and it fills up the CI console. Then it becomes difficult to find the error from CI logs.

suggest to print the logs only if the parsing fails. OR just print something like "logs parsing failed, error=%v. Please look in the collected test artifacts"

eranra: @vimalk78 I thought about that but decided that this is not creating so many logs ... on the flip side this is really good way to understand how things are working and how the test is comparing existing with required fields ... so decided to keep that in (for now) ... I am planning to add in the next PR some more tests (audit etc...) that use this functionality and want to follow up on the process. Once we will feel more confident that this is really comparing well and doing it's work ... totally agree we can remove this print message.

PR #785, opened 2020-11-12
Functional tests for Logforwarding audit formats

eranra:
### Description

This PR extends log forwarding format testing to include also audit log tests 

/cc @jcantrill 
/assign @jcantrill 

### Links
- JIRA:
  https://issues.redhat.com/browse/LOG-975

### DIFF
--- a/test/functional/framework.go
+++ b/test/functional/framework.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math/rand"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
@@ -22,6 +23,24 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
+
+const (
+	applicationLog = "application"
+	auditLog       = "audit"
+	k8sAuditLog    = "k8s"
+)
+
+var fluentdLogPath = map[string]string{
+	applicationLog: "/var/log/containers",
+	auditLog:       "/var/log/audit",
+	k8sAuditLog:    "/var/log/kube-apiserver",
+}
+
+var outputLogFile = map[string]string{
+	applicationLog: "/tmp/app-logs",
+	auditLog:       "/tmp/audit-logs",
+	k8sAuditLog:    "/tmp/audit-logs",
+}
 
 var (
 	maxDuration          time.Duration
@@ -179,7 +198,7 @@ func (f *FluentdFunctionalFramework) DeployWithVisitor(visit runtime.PodBuilderV
 	}
 	log.V(2).Info("waiting for fluentd to be ready")
 	err = wait.PollImmediate(time.Second*2, time.Second*30, func() (bool, error) {
-		output, err := oc.Literal().From(fmt.Sprintf("oc logs -n %s pod/%s %s", f.test.NS.Name, f.Name, constants.FluentdName )).Run()
+		output, err := oc.Literal().From(fmt.Sprintf("oc logs -n %s pod/%s %s", f.test.NS.Name, f.Name, constants.FluentdName)).Run()
 		if err != nil {
 			return false, nil
 		}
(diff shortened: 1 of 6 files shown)


alanconway (approved):
/approve

test/functional/framework.go
```diff
@@ -169,6 +189,26 @@ func (f *FluentdFunctionalFramework) Deploy() (err error) {
 	if err != nil {
 		return fmt.Errorf("service could not be started")
 	}
+
+	log.V(2).Info("waiting for fluentd to be ready")
+	err = wait.PollImmediate(time.Second*2, time.Second*30, func() (bool, error) {
```
alanconway, line 194: test/client/Client.Watch is a more efficient way to wait for pods but there is a lot of code still using polling so I wouldn't hold up this merge just for that. At some point we'll need to make a sweep of the tests to bring them up to date

alanconway (commented):
/lgtm

test/functional/output_forward.go
```diff
@@ -27,10 +34,31 @@ const unsecureFluentConf = `
 		@type json
 	</format>
 </match>
-<match **>
-	@type stdout
+
+<filter linux-audit.log**>
```
jcantrill, line 38: Why are we doing any special parsing/processing the message in the output container?  One point of doing functional tests is to verify we are generating our normalization and collection pipeline correctly.  Needing to parse the output shipped from the main fluent container would leave me to believe:
* We are not normalizing something correctly
* We are unnecessarily post-processing something for the test

eranra: @jcantrill apparently the forwarded time_key format as-is for "linux-audit.log" do not work with @type file to write logs to file ... so you **can't** write the output format of  "linux-audit.log" to file ... so ... I spend almost two working days trying to understand what is going on here. I learned that you are using proprietary RH parser to parse the logs and do not use the required "%Y-%m-%dT%H:%M:%S.%N%z" (somewhere in the ViaQ github project) and to overcome that I had to change the format reading time from @timestamp and writing to time_key is the correct format ...  I did not want (or know if this make sense) to change the parsing code ...

eranra: @jcantrill @alanconway  I invested more time today to make this PR even more robust and make the tests easier to create ... look on the templates ... they are very easy to create right now .... 

The big challenge for me is that I do not know what are the correct formats ... so with this PR we have all the building blocks ... but now we need to feed this will some more complicated formats to check ... and make sure that all output fields are as expected. I can just see what is the output and compare but I am not sure that this is correct output

eranra: @alanconway look also on prev. comment

alanconway: We really need to review & lay out the data model properly, it's on my TODO list. 

Meantime how about we reduce the expectations of the test to just items we are confident about, and extend it later with more details.

My concern with the template appproach is that it works for equality/regexp tests on string values, but it's no use for complicated non-string values like time.Time for instance. However I think it's worth merging this and improving later, since the test already provides some value.

/lgtm

alanconway: +1, I say again, lets merge it and improve it as we go.

alanconway: /lgtm

test/functional/framework.go
```diff
@@ -23,6 +24,24 @@ import (
...
+const (
+	applicationLog = "application"
+	auditLog       = "audit"
+	k8sAuditLog    = "k8s"
+)
+
+var fluentdLogPath = map[string]string{
+	applicationLog: "/var/log/containers",
+	auditLog:       "/var/log/audit",
+	k8sAuditLog:    "/var/log/kube-apiserver",
```
jcantrill, line 36: I don't know it matters for this test but i think you are missing the openshift apiserver audit logs here

test/functional/logforwarding/message_format_test.go
```diff
@@ -4,14 +4,30 @@ import (
 	"fmt"
```
jcantrill, line 4: Please move this test to `test/functional/normalization`

test/functional/logforwarding/message_format_test.go
```diff
@@ -22,48 +38,100 @@ var _ = Describe("[LogForwarding] Functional tests for message format", func() {
...
+		msg := fmt.Sprintf("audit(%.3f:24287)", float64(nanoTime.UnixNano())/float64(time.Second))
+		auditLogLine := fmt.Sprintf("type=%s msg=%s:", msgType, msg)
+		// Template expected as output Log
+		var outputLogTemplate = types.LinuxAuditLog{
+			Message:       auditLogLine,
+			ViaqIndexName: "audit-write",
+			AuditLinux: types.AuditLinux{
+				Type:     msgType,
+				RecordID: "*"},
+			Timestamp:        nanoTime,
```
jcantrill, line 95: This is a nit but something I was thinking about when you mentioned it.  We do desire the time to be the correct format but it doesn't matter if it is "time"; it should be a string.  JSON only supports the basic types anyway.  This may be but a detail.

jcantrill:
/hold cancel
retest

jcantrill:
Applied valid bz label as this is test code improvements

PR #786, opened 2020-11-12
Fix cmd.Reader to include stderr text in returned errors.

alanconway:
/assign @jcantrill
/cc @syedriko

### DIFF
--- a/test/helpers/cmd/reader.go
+++ b/test/helpers/cmd/reader.go
@@ -2,12 +2,14 @@ package cmd
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"io"
 	"os/exec"
 	"regexp"
+	"strings"
 
 	"github.com/openshift/cluster-logging-operator/test"
 	"github.com/openshift/cluster-logging-operator/test/runtime"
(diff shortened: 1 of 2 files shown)


test/helpers/cmd/reader.go
```diff
@@ -142,3 +146,28 @@ func TailReader(pod *corev1.Pod, path string) (*Reader, error) {
...
+	}
+	return err
+}
+
+const stderrLimit = 1024
+
+type stderrBuffer struct{ b bytes.Buffer }
+
+func (b *stderrBuffer) Write(data []byte) (int, error) {
+	max := stderrLimit - b.b.Len()
```
syedriko, line 165: Rename max -> canAdd or remainingRoom?

test/helpers/cmd/reader.go
```diff
@@ -142,3 +146,28 @@ func TailReader(pod *corev1.Pod, path string) (*Reader, error) {
...
+	if max < 0 {
+		max = 0
```
syedriko, line 167: Would it make sense to just return here instead of going with a degenerate case of max == 0?

syedriko (approved):

PR #788, opened 2020-11-13
gather artifacts for e2e setup

jcantrill:
### Description
This PR adds artifact gathering for e2e setup to try and understand e2e CI falures

### DIFF
--- a/hack/test-e2e-olm.sh
+++ b/hack/test-e2e-olm.sh
@@ -3,6 +3,28 @@
 current_dir=$(dirname "${BASH_SOURCE[0]}" )
 source "${current_dir}/lib/init.sh"
 source "${current_dir}/lib/util/logs.sh"
+source "${current_dir}/testing-olm/utils"
+
+get_setup_artifacts=true
+CLUSTER_LOGGING_OPERATOR_NAMESPACE=${CLUSTER_LOGGING_OPERATOR_NAMESPACE:-openshift-logging}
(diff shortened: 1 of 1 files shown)


hack/test-e2e-olm.sh
```diff
@@ -3,6 +3,28 @@
...
+test_artifactdir="${ARTIFACT_DIR}/$(basename ${BASH_SOURCE[0]})"
+if [ ! -d $test_artifactdir ] ; then
+  mkdir -p $test_artifactdir
+fi
+cleanup(){
+  local return_code="$?"
+  
+  set +e
+  if [ "$return_code" != "0" -a $get_setup_artifacts ] ; then 
+    oc get all -n openshift-operators-redhat > $testartifactdir/openshift-operators-redhat_all.txt
```
huikang, line 20: How about adding a variable `ELASTICSEARCH_OPERATOR_NAMESPACE` like line 9 with the default value `openshift-operators-redhat`?

jcantrill: all other tests assume it lives here.  do we have need to parameterize it?

huikang: Sorry, I only looked at the change at line 9 and did not check the rest of the file. If so, it I agreed that there is no need to parameterize to be consistent.

alanconway (approved):
/lgtm

PR #790, opened 2020-11-15
Bug 1891886: Support Adding log source info to log message

vimalk78:
### Description
 - A fix for bug [1891886](https://bugzilla.redhat.com/show_bug.cgi?id=1891886).
 - Added a field in LF CRD `AddLogSource`, which when enabled, adds namespace_name, pod_name, container_name to the `message` field of the record.
 - This can be further used with `PayloadKey` set to `"message"` to achieve the same behavior as older syslog plugin.
 - The solution is to add a `record_modifier` section in the fluent config, which copies the `namespace_name` , `pod_name` , `container_name` from the `kubernetes` section, and copies it into `message`
  - also fixed two lint issues in funtional tests

#### Syslog Payload before this change:
```
<15>1 2020-11-15T17:06:14+00:00 fluentd-9hkb4 mytag - - -  {"msgcontent"=>"My life is my message", "timestamp"=>"2020-11-15 17:06:09", "tag_key"=>"rec_tag", "index"=>56}
```

#### Syslog Payload After this change:
```
<15>1 2020-11-16T10:49:37+00:00 crc-j55b9-master-0 mytag - - -  namespace_name=clo-test-6327,pod_name=log-generator-ff9746c49-qxm7l,container_name=log-generator,message={"msgcontent"=>"My life is my message", "timestamp"=>"2020-11-16 10:49:36", "tag_key"=>"rec_tag", "index"=>76}
```

#### Example LF CR :
```
spec:    
  outputs:    
  - name: syslogout    
    syslog:    
      addLogSource: true    
      facility: user    
      payloadKey: message    
      rfc: RFC3164    
      severity: debug    
      tag: mytag    
    type: syslog    
    url: tls://syslog-receiver.openshift-logging.svc:24224    
  pipelines:    
  - inputRefs:    
    - application    
    name: test-app    
    outputRefs:    
    - syslogout    

```
*Note* : This behavior works for both RFC3164 and RFC5424

/cc @alanconway @jcantrill 
/assign @alanconway 

### Links
- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1824003#c26 (actual syslog BZ which holds the conversation about the issue faced by customer. This solution is inspired from this BZ, especially comment 26)
- Bugzille: https://bugzilla.redhat.com/show_bug.cgi?id=1891886

### DIFF
--- a/pkg/generators/forwarding/fluentd/templates.go
+++ b/pkg/generators/forwarding/fluentd/templates.go
@@ -628,6 +628,21 @@ const outputLabelConfJsonParseNoretryTemplate = `{{- define "outputLabelConfJson
 	merge_json_log false
 	replace_json_log true
   </filter>
+{{ if .Target.Syslog.AddLogSource }}
+  <filter **>
+	@type record_modifier
+	<record>
+	  kubernetes_info ${if record.has_key?('kubernetes'); record['kubernetes']; else {}; end}
+	  namespace_info  ${if record['kubernetes_info'] != nil && record['kubernetes_info'] != {}; "namespace_name=" + record['kubernetes_info']['namespace_name']; else nil; end}
+	  pod_info        ${if record['kubernetes_info'] != nil && record['kubernetes_info'] != {}; "pod_name=" + record['kubernetes_info']['pod_name']; else nil; end}
+	  container_info  ${if record['kubernetes_info'] != nil && record['kubernetes_info'] != {}; "container_name=" + record['kubernetes_info']['container_name']; else nil; end}
+	  msg_key         ${if record.has_key?('message') && record['message'] != nil; record['message']; else nil; end}
+      msg_info        ${if record['msg_key'] != nil && record['msg_key'].is_a?(Hash); require 'json'; "message="+record['message'].to_json; elsif record['msg_key'] != nil; "message="+record['message']; else nil; end}
+      message         ${if record['msg_key'] != nil && record['kubernetes_info'] != nil && record['kubernetes_info'] != {}; record['namespace_info'] + ", " + record['container_info'] + ", " + record['pod_info'] + ", " + record['msg_info']; else record['message']; end}
+	</record>
+	remove_keys kubernetes_info, namespace_info, pod_info, container_info, msg_key, msg_info
+  </filter>
+{{end -}}
   <match **>
     @type copy
 {{include .StoreTemplate . "" | indent 4}}
@@ -799,6 +814,7 @@ const storeSyslogTemplate = `{{- define "storeSyslog" -}}
 	{{end -}}
 	protocol {{.Protocol}}
 	packet_size 4096
+	hostname "#{ENV['NODE_NAME']}"
 {{ if .Target.Secret -}}
   tls true
   ca_file '{{ .SecretPath "ca-bundle.crt"}}'
--- a/manifests/4.7/logging.openshift.io_clusterlogforwarders_crd.yaml
+++ b/manifests/4.7/logging.openshift.io_clusterlogforwarders_crd.yaml
@@ -99,6 +99,9 @@ spec:
                     syslog:
                       description: Syslog provides optional extra properties for output type `syslog`
                       properties:
+                        addLogSource:
+                          description: AddLogSource adds log's source information to the log message If the logs are collected from a process; namespace_name, pod_name, container_name is added to the log
+                          type: boolean
                         appName:
                           description: "AppName is APP-NAME part of the syslog-msg header \n AppName needs to be specified if using rfc5424"
                           type: string
--- a/manifests/5.0/logging.openshift.io_clusterlogforwarders_crd.yaml
+++ b/manifests/5.0/logging.openshift.io_clusterlogforwarders_crd.yaml
@@ -99,6 +99,9 @@ spec:
                     syslog:
                       description: Syslog provides optional extra properties for output type `syslog`
                       properties:
+                        addLogSource:
+                          description: AddLogSource adds log's source information to the log message If the logs are collected from a process; namespace_name, pod_name, container_name is added to the log
+                          type: boolean
                         appName:
                           description: "AppName is APP-NAME part of the syslog-msg header \n AppName needs to be specified if using rfc5424"
                           type: string
--- a/pkg/apis/logging/v1/output_types.go
+++ b/pkg/apis/logging/v1/output_types.go
@@ -65,6 +65,12 @@ type Syslog struct {
 	// +optional
 	PayloadKey string `json:"payloadKey,omitempty"`
 
+	// AddLogSource adds log's source information to the log message
+	// If the logs are collected from a process; namespace_name, pod_name, container_name is added to the log
+	//
+	// +optional
+	AddLogSource bool `json:"addLogSource,omitempty"`
+
 	// Rfc specifies the rfc to be used for sending syslog
 	//
 	// Rfc values can be one of:
--- a/pkg/generators/forwarding/fluentd/output_conf_syslog_test.go
+++ b/pkg/generators/forwarding/fluentd/output_conf_syslog_test.go
@@ -143,6 +143,7 @@ var _ = Describe("Generating external syslog server output store config blocks",
         severity debug
     	protocol tcp
     	packet_size 4096
+		hostname "#{ENV['NODE_NAME']}"
 		timeout 60
 		timeout_exception true
 	    keep_alive true
@@ -187,6 +188,7 @@ var _ = Describe("Generating external syslog server output store config blocks",
         severity debug
     	protocol udp
     	packet_size 4096
+		hostname "#{ENV['NODE_NAME']}"
         <buffer >
         @type file
         path '/var/lib/fluentd/syslog_receiver'
@@ -225,6 +227,7 @@ var _ = Describe("Generating external syslog server output store config blocks",
         severity debug
     	protocol tcp
     	packet_size 4096
+		hostname "#{ENV['NODE_NAME']}"
         tls true
         ca_file '/var/run/ocp-collector/secrets/some-secret/ca-bundle.crt'
         verify_mode true
@@ -272,6 +275,7 @@ var _ = Describe("Generating external syslog server output store config blocks",
         severity debug
     	protocol udp
     	packet_size 4096
+		hostname "#{ENV['NODE_NAME']}"
         tls true
         ca_file '/var/run/ocp-collector/secrets/some-secret/ca-bundle.crt'
         verify_mode true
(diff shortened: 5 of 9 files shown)


alanconway (commented):
This fix needs to be in context of larger concerns:

- "Source on/off" is not enough. We will need a coherent API for filtering _all_ metadata fields, so we need to avoid ad-hoc extensions.

- This is not just a syslog issue, it affects all output types.

- Our data model needs to be sanitized. It is already bulky and complex, we need to think carefully before adding redundant fields.

The original data model is at https://github.com/openshift/origin-aggregated-logging/blob/master/fluentd/configs.d/openshift/filter-viaq-data-model.conf

alanconway:
If you remove the changes to the ClusterLogForwarderSpec, so the API changes are all in SyslogOutput then this is OK. I don't like the comma-separated-prefix message format, but we have customers who have adapted to it. I'll put together a proposal for more generic metadata manipulation, we can live with the overlap. I've posted some questions on the BZ, but I guess we will need a backwards-compatibe story regardless.

vimalk78:
@alanconway , there are no changes to `ClusterLogForwarderSpec` , only in the syslog output type.
Earlier i started with a generic solution to 'merge' fields, but discarded it as being too generic , and scope being much larger than what customers needed.

```
// MergeFields allows to pick fields from logs (e.g. log metadata) and add them into other field (e.g. message)
//
// Example spec:
//  mergeFields:
//   - name: container
//     fromPath: docker.kubernetes.container_name
//     toPath: message
//   - fromPath: docker.kubernetes.namespace_name
//     toPath: message
//   - fromPath: docker.kubernetes.pod_name
//     toPath: message
//
//
// original log mesage:
//  docker:{
//   kubernetes:{
//    "container_name"=>"applicaion-container",
//    "namespace_name"=>"application-namespace",
//    "pod_name"=>"application-pod",
//
//  }
//  message: "this is a log line"
//
// merged log message:
//  docker:{
//   kubernetes:{
//    "container_name"=>"applicaion-container",
//    "namespace_name"=>"application-namespace",
//    "pod_name"=>"application-pod",
//
//  }
//  message: "namespace_name=\"application-namespace\", pod_name=\"application-pod\", container=\"application-container\",message=\"this is a log line\""
type MergeFields struct {
      // From specifies the jsonpath of the field whose value needs to be merged
      From string `json:"fromPath"`
      // To specifies the jsonpath of the field to which the value needs to be merged into
      To string `json:"toPath"`
      // Key specifies the key under which the value needs to be merged. if omitted, last part of `From` is used as Key
      // +optional
      Key string `json:"key,omitempty"`
}
```

alanconway:
> @alanconway , there are no changes to `ClusterLogForwarderSpec` , only in the syslog output type.
> Earlier i started with a generic solution to 'merge' fields, but discarded it as being too generic , and scope being much larger than what customers needed.

+1, this is the right fix for the customer as they need the exact old behavior (got a reply to that effect in the BZ) which is incompatible with a clean JSON model at the output.

I do want to add your generic merge notion, but building on  https://github.com/openshift/enhancements/pull/518 (would appreciate your review there) - your mergeFields would become a new field for the filter type, merging meta-data fields into the `structured` payload rather than into the top level.



>  - 
> ```
> // MergeFields allows to pick fields from logs (e.g. log metadata) and add them into other field (e.g. message)
> //
> // Example spec:
> //  mergeFields:
> //   - name: container
> //     fromPath: docker.kubernetes.container_name
> //     toPath: message
> //   - fromPath: docker.kubernetes.namespace_name
> //     toPath: message
> //   - fromPath: docker.kubernetes.pod_name
> //     toPath: message
> //
> //
> // original log mesage:
> //  docker:{
> //   kubernetes:{
> //    "container_name"=>"applicaion-container",
> //    "namespace_name"=>"application-namespace",
> //    "pod_name"=>"application-pod",
> //
> //  }
> //  message: "this is a log line"
> //
> // merged log message:
> //  docker:{
> //   kubernetes:{
> //    "container_name"=>"applicaion-container",
> //    "namespace_name"=>"application-namespace",
> //    "pod_name"=>"application-pod",
> //
> //  }
> //  message: "namespace_name=\"application-namespace\", pod_name=\"application-pod\", container=\"application-container\",message=\"this is a log line\""
> type MergeFields struct {
>       // From specifies the jsonpath of the field whose value needs to be merged
>       From string `json:"fromPath"`
>       // To specifies the jsonpath of the field to which the value needs to be merged into
>       To string `json:"toPath"`
>       // Key specifies the key under which the value needs to be merged. if omitted, last part of `From` is used as Key
>       // +optional
>       Key string `json:"key,omitempty"`
> }
> ```

vimalk78:
> +1, this is the right fix for the customer as they need the exact old behavior (got a reply to that effect in the BZ) which is incompatible with a clean JSON model at the output.

i think then we can merge this PR, right ?

> I do want to add your generic merge notion, but building on [openshift/enhancements#518](https://github.com/openshift/enhancements/pull/518) (would appreciate your review there) - your mergeFields would become a new field for the filter type, merging meta-data fields into the `structured` payload rather than into the top level.

will review the enhancement proposal

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -622,6 +622,19 @@ const outputLabelConfJsonParseNoretryTemplate = `{{- define "outputLabelConfJson
 	merge_json_log false
 	replace_json_log true
   </filter>
+{{ if .Target.Syslog.AddLogSource }}
+  <filter **>
+	@type record_modifier
+	<record>
+	  namespace_info      ${if record.has_key?('kubernetes'); "namespace_name=" + record['kubernetes']['namespace_name']; else nil; end}
```
k-keiichi-rh, line 629: I tried this patch and checked if we can send audit logs via syslog forwarding.
But it failed to send audit logs with the following error:
```
2020-12-04 18:23:57 +0000 [warn]: dump an error event: error_class=TypeError error="no implicit conversion of nil into String" location="/usr/local/share/gems/gems/fluent-plugin-record-modifier-2.1.0/lib/fluent/plugin/filter_record_modifier.rb:174:in `+'" tag="k8s-audit.log" time=2020-12-04 18:23:57.473698000 +0000 record={"kind"=>"Event", "apiVersion"=>"audit.k8s.io/v1", "level"=>"info", "auditID"=>"45543604-2e70-44c6-ab25-6b670eac13d5", "stage"=>"ResponseComplete", "requestURI"=>"/apis/authentication.k8s.io/v1/tokenreviews", "verb"=>"create", "user"=>{"username"=>"system:serviceaccount:openshift-multus:multus", "uid"=>"a48aa3fc-eacb-4cec-934a-31a65955d6dd", "groups"=>["system:serviceaccounts", "system:serviceaccounts:openshift-multus", "system:authenticated"]}, "sourceIPs"=>["192.168.200.11"], "userAgent"=>"kube-rbac-proxy/v0.0.0 (linux/amd64) kubernetes/$Format", "objectRef"=>{"resource"=>"tokenreviews", "apiGroup"=>"authentication.k8s.io", "apiVersion"=>"v1"}, "responseStatus"=>{"code"=>201}, "requestReceivedTimestamp"=>"2020-12-04T18:23:57.473698Z", "stageTimestamp"=>"2020-12-04T18:23:57.479767Z", "annotations"=>{"authentication.k8s.io/legacy-token"=>"system:serviceaccount:openshift-multus:multus", "authorization.k8s.io/decision"=>"allow", "authorization.k8s.io/reason"=>"RBAC: allowed by ClusterRoleBinding \"multus-admission-controller-webhook\" of ClusterRole \"multus-admission-controller-webhook\" to ServiceAccount \"multus/openshift-multus\""}, "k8s_audit_level"=>"Metadata", "message"=>nil, "hostname"=>"master2.hawk2p7-1.cluster.sub.nec.test", "pipeline_metadata"=>{"collector"=>{"ipaddr4"=>"192.168.200.12", "inputname"=>"fluent-plugin-systemd", "name"=>"fluentd", "received_at"=>"2020-12-04T18:23:57.481102+00:00", "version"=>"1.7.4 1.6.0"}}, "@timestamp"=>"2020-12-04T18:23:57.473698+00:00", "viaq_index_name"=>"audit-write", "viaq_msg_id"=>"MDZmMDI3ZTItNDY3MC00ZTMwLWE3YTktNzhkNWNhZWQ2YmU0", "kubernetes"=>{}}
```
As for Audit log, the "record.has_key?('kubernetes');" doesn't work well.
Because audit record also has "kubernetes" key in its record although the value is "{}".

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -622,6 +622,19 @@ const outputLabelConfJsonParseNoretryTemplate = `{{- define "outputLabelConfJson
...
+	  pod_info            ${if record.has_key?('kubernetes'); "pod_name=" + record['kubernetes']['pod_name']; else nil; end}
+	  container_info      ${if record.has_key?('kubernetes'); "container_name=" + record['kubernetes']['container_name']; else nil; end}
+	  message_info        ${if record['message'].is_a?(Hash); require 'json'; "message=#{record['message'].to_json}" ; else "message=" + record['message']; end}
+	  message             ${if record.has_key?('kubernetes'); record['namespace_info'] + "," + record['pod_info'] + "," + record['container_info'] + "," + record['message_info']; else record['message']; end}
```
k-keiichi-rh, line 633: In the legacy way, the format was "namespace_name=xxx, container_name=xxx, pod_name=xxx, message=xxx".
https://github.com/openshift/origin-aggregated-logging/blob/ee4935b6e1a4507a3714452e99acf0c945c2568c/fluentd/out_syslog.rb#L138

Could you please change the format to the following to keep the format compatibility(a space is needed after ",", the order is namespace=>container=>pod) between the new way and the legacy way:
```
message             ${if record.has_key?('kubernetes');  record['namespace_info'] + ", " + record['container_info'] + ", " + record['pod_info'] + ", "  + record['message_info']; else record['message']; end}
```

k-keiichi-rh (commented):
> +1, this is the right fix for the customer as they need the exact old behavior (got a reply to that effect in the BZ) which is incompatible with a clean JSON model at the output.

Thank you for your help and support.

> #### Syslog Payload After this change:
> ```
> <15>1 2020-11-16T10:49:37+00:00 crc-j55b9-master-0 mytag - - -  namespace_name=clo-test-6327,pod_name=log-generator-ff9746c49-qxm7l,container_name=log-generator,message={"msgcontent"=>"My life is my message", "timestamp"=>"2020-11-16 10:49:36", "tag_key"=>"rec_tag", "index"=>76}
> ```

There are two log formats:

- RFC3164 : "<#{pri}>#{generate_timestamp_rfc3164} #{@hostname} #{@tag}: #{@content}"
- RFC5424 : "<#{pri}>1 #{generate_timestamp_rfc5424} #{@hostname} #{@appname} #{@procid} #{@msgid} #{structured_data} #{@content}"

It looks like it's designed to change the format by setting the "rfc" field in "syslog" config.
 https://github.com/openshift/origin-aggregated-logging/blob/ee4935b6e1a4507a3714452e99acf0c945c2568c/fluentd/lib/syslog_protocol/lib/syslog_protocol/packet.rb#L23
> #### Example LF CR :
> ```
> spec:    
>   outputs:    
>   - name: syslogout    
>     syslog:    
>       addLogSource: true    
>       facility: user    
>       payloadKey: message    
>       rfc: RFC3164    

However, the log format is RFC5424's one even if we set "rfc: RFC3164".
So I think this behavior is unexpected one.
In the legacy way, the format was RFC3164's one. We want to keep same format same as the legacy one.

k-keiichi-rh:
> However, the log format is RFC5424's one even if we set "rfc: RFC3164".
> So I think this behavior is unexpected one.

I am sorry for the confusion.
This syslog forwarding with "rfc: RFC3164" works fine as we expected. I confirmed it sends log with "<#{pri}>#{generate_timestamp_rfc3164} #{@hostname} #{@tag}: #{@content}" via  remote_syslog_sender.

And I found out that the unexpected behavior I reported is caused by incorrect configuration of rsyslogd.
When I tested with the following rsyslogd configuration,  the log with RFC3164 format is transferred to the rsyslogd server.
```
$ cat /etc/rsyslog.d/ocp.conf 
$template RFC3164fmt,"<%PRI%>%TIMESTAMP% %HOSTNAME% %syslogtag%%msg%\n"
local0.info /var/log/ocp_lf/openshift_infra.log;RFC3164fmt
& stop
local1.info /var/log/ocp_lf/openshift_audit.log;RFC3164fmt
& stop
local2.info /var/log/ocp_lf/openshift_app.log;RFC3164fmt
& stop
```

pkg/generators/forwarding/fluentd/templates.go
k-keiichi-rh: Thank you for your work.

> Could you please change the format to the following to keep the format compatibility(a space is needed after ",", the order is namespace=>container=>pod) between the new way and the legacy way:
> 
> ```
> message             ${if record.has_key?('kubernetes');  record['namespace_info'] + ", " + record['container_info'] + ", " + record['pod_info'] + ", "  + record['message_info']; else record['message']; end}
> ```

Is this change acceptable for you? If not, is there any reason?

alanconway (commented):
/lgtm

jcantrill (commented):
/hold cancel
/lgtm

PR #794, opened 2020-11-16
Attempt a fix to functional test flake

vimalk78:
Attempt a fix to functional test flake
 - wait for fluentd service to be ready before curl-ing it
 - fixed "odd number of arguments" to structured logging in oc/runner
### Description
The metrics functional test fails because `curl` to the metrics endpoint fails. So wait for the service endpoint to have IPs before attempting to curl it.

/cc @jcantrill 
/assign @alanconway

### DIFF
--- a/test/functional/framework.go
+++ b/test/functional/framework.go
@@ -154,6 +154,21 @@ func (f *FluentdFunctionalFramework) Deploy() (err error) {
 	if err = f.test.Client.Get(f.pod); err != nil {
 		return err
 	}
+	log.V(2).Info("waiting for service endpoints to be ready")
+	err = wait.PollImmediate(time.Second*2, time.Second*10, func() (bool, error) {
+		ips, err := oc.Get().WithNamespace(f.test.NS.Name).Resource("endpoints", f.Name).OutputJsonpath("{.subsets[*].addresses[*].ip}").Run()
+		if err != nil {
+			return false, nil
+		}
+		// if there are IPs in the service endpoint, the service is available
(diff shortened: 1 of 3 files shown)


test/functional/framework.go
```diff
@@ -154,6 +154,24 @@ func (f *FluentdFunctionalFramework) Deploy() (err error) {
 	if err = f.test.Client.Get(f.pod); err != nil {
 		return err
 	}
+	log.V(2).Info("waiting for service endpoints to be ready")
+	serviceUp := false
+	count := 0
+	for count < 10 && !serviceUp {
```
jcantrill, line 160: Wouldnt this be better resolved with the Poll functions we've used elsewhere

test/functional/framework.go
```diff
@@ -154,6 +154,24 @@ func (f *FluentdFunctionalFramework) Deploy() (err error) {
...
+		ips, err := oc.Get().WithNamespace(f.test.NS.Name).Resource("endpoints", f.Name).OutputJsonpath("{.subsets[*].addresses[*].ip}").Run()
```
jcantrill, line 161: can we `oc wait` for a service?

vimalk78: no we cant. service doesnt have a conditions in their status
https://github.com/kubernetes/kubernetes/issues/80828

test/functional/framework.go
vimalk78: updated to use `wait.PollImmediate`

jcantrill (commented):
/approve
/lgtm

PR #797, opened 2020-11-18
Fix `make check`

eranra:
### Description
Adding environment variable GODEBUG to make file with value x509ignoreCN=0 so that `/home/eranra/go/src/github.com/openshift/cluster-logging-operator/pkg/k8shandler/certificates_test.go` will work

/cc @jcantrill 
/assign @jcantrill


jcantrill (commented):
/lgtm

PR #801, opened 2020-11-18
Bug 1894640: lower default max_retry_wait

jcantrill:
# Description
This PR lowers the default max_retry_wait to resolve cases where it appears fluent is functional but not pushing messages. The current retry algo is an exponential backoff that will cause fluent to not retry try soon enough

ref: https://bugzilla.redhat.com/show_bug.cgi?id=1894640

/assign @vimalk78 @alanconway

backport of https://github.com/openshift/cluster-logging-operator/pull/799


alanconway (commented):
/lgtm

PR #811, opened 2020-11-23
Bug 1891738: Make operator not depend on elasticsearch-operator

jcantrill:
### Description
This PR modifies the reconciliation loop to:
* skip reconciliation of storage, visualization and curation if the managed store is not defined

*Note*: while testing it was observed
* storage et al will fail (as expected) to deploy if the EO is not deployed
* It is possible to see the following error message in the logs if you deploy eo, logging, and then remove eo, logging, and delete the CRD
```
E1123 19:13:13.053559       1 reflector.go:178] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:224: Failed to list *v1.Elasticsearch: the server could not find the requested resource (get elasticsearches.logging.openshift.io)
```
Deleting the CRD is a manual step which is unlikely to be done unless a user explicitly does so

/cc @igor-karpukhin 
/assign @alanconway @vimalk78 

### Links
* https://bugzilla.redhat.com/show_bug.cgi?id=1891738

### DIFF
--- a/pkg/k8shandler/clusterloggingrequest.go
+++ b/pkg/k8shandler/clusterloggingrequest.go
@@ -23,6 +23,10 @@ type ClusterLoggingRequest struct {
 	ForwarderSpec logging.ClusterLogForwarderSpec
 }
 
+func (clusterRequest *ClusterLoggingRequest) IncludesManagedStorage() bool {
+	return clusterRequest.Cluster != nil && clusterRequest.Cluster.Spec.LogStore != nil
+}
+
 // TODO: determine if this is even necessary
 func (clusterRequest *ClusterLoggingRequest) isManaged() bool {
 	return clusterRequest.Cluster.Spec.ManagementState == logging.ManagementStateManaged
--- a/pkg/k8shandler/clusterloggingrequest_test.go
+++ b/pkg/k8shandler/clusterloggingrequest_test.go
@@ -1,0 +1,40 @@
+package k8shandler
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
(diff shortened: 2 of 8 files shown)


jcantrill:
cc @mburke5678

pkg/k8shandler/clusterloggingrequest.go
```diff
@@ -23,6 +23,10 @@ type ClusterLoggingRequest struct {
 	ForwarderSpec logging.ClusterLogForwarderSpec
 }
 
+func (clusterRequest *ClusterLoggingRequest) IncludesManagedStorage() bool {
+	return clusterRequest.Cluster != nil && clusterRequest.Cluster.Spec.LogStore != nil
+}
+
 // TODO: determine if this is even necessary
```
alanconway, line 30: I think we can erase this TODO?

pkg/k8shandler/clusterloggingrequest_test.go
```diff
@@ -0,0 +1,40 @@
...
+	)
+	BeforeEach(func() {
+		clr = &ClusterLoggingRequest{
+			Cluster: &logging.ClusterLogging{
+				Spec: logging.ClusterLoggingSpec{},
+			},
+		}
+
+	})
+	Describe("#IncludesManagedStorage", func() {
```
alanconway, line 22: Does '#' mean something or is it a typo?

alanconway (commented):
/lgtm

pkg/k8shandler/clusterloggingrequest_test.go
jcantrill: It's only a means to chain the describes together so its more obvious we are referring to a given type and method `ClusterLoggingRequest#IncludesManagesStorage`

PR #812, opened 2020-11-23
Fix Makefile for Go 1.15 legacy certificate incompatibility.

alanconway:
Building and running tests with Go 1.15 causes certificate parsing errors.  The
issue is described https://golang.org/doc/go1.15#commonname The fix in Makefile
is: export GODEBUG=x509ignoreCN=0

We should sanitize our certs to avoid the deprecated practice but for now we can
disable the new check in Go.

Also fixed minor gofmt/lint issues in test/functional/framework.go

/approve @jcantril
/cc @ewolinetz


alanconway:
Fixed by another PR

PR #813, opened 2020-11-24
Bug 1901096: Set forward heartbeat to none

jcantrill:
### Description
This PR
*  sets the fluentd forward heartbeat to "none" to avoid dns resolution issues from stopping the pod from starting
* Expands the functional framework to add visitors in order to test this issue
```
2020-11-23 20:26:38 +0000 [error]: unexpected error error_class=SocketError error="getaddrinfo: Name or service not known"
  2020-11-23 20:26:38 +0000 [error]: /usr/local/share/gems/gems/fluentd-1.7.4/lib/fluent/plugin/out_forward.rb:680:in `getaddrinfo'
  2020-11-23 20:26:38 +0000 [error]: /usr/local/share/gems/gems/fluentd-1.7.4/lib/fluent/plugin/out_forward.rb:680:in `resolve_dns!'
  2020-11-23 20:26:38 +0000 [error]: /usr/local/share/gems/gems/fluentd-1.7.4/lib/fluent/plugin/out_forward.rb:666:in `resolved_host'
  2020-11-23 20:26:38 +0000 [error]: /usr/local/share/gems/gems/fluentd-1.7.4/lib/fluent/plugin/out_forward.rb:516:in `validate_host_resolution!'
  2020-11-23 20:26:38 +0000 [error]: /usr/local/share/gems/gems/fluentd-1.7.4/lib/fluent/plugin/out_forward.rb:237:in `block in configure'
  2020-11-23 20:26:38 +0000 [error]: /usr/local/share/gems/gems/fluentd-1.7.4/lib/fluent/plugin/out_forward.rb:227:in `each'
  2020-11-23 20:26:38 +0000 [error]: /usr/local/share/gems/gems/fluentd-1.7.4/lib/fluent/plugin/out_forward.rb:227:in `configure'
```

/cc @vimalk78 @igor-karpukhin 
/assign @alanconway 

### Links
* https://bugzilla.redhat.com/show_bug.cgi?id=1901096

### DIFF
--- a/pkg/generators/forwarding/fluentd/fluent_conf_test.go
+++ b/pkg/generators/forwarding/fluentd/fluent_conf_test.go
@@ -1014,20 +1014,21 @@ var _ = Describe("Generating fluentd config", func() {
 				<match **>
 					# https://docs.fluentd.org/v1.0/articles/in_forward
 				@type forward
+				heartbeat_type none
 
 				<buffer>
 					@type file
 					path '/var/lib/fluentd/secureforward_receiver'
 					queued_chunks_limit_size "#{ENV['BUFFER_QUEUE_LIMIT'] || '1024' }"
-          total_limit_size "#{ENV['TOTAL_LIMIT_SIZE'] ||  8589934592 }" #8G
-          chunk_limit_size "#{ENV['BUFFER_SIZE_LIMIT'] || '1m'}"
-          flush_mode interval
+					total_limit_size "#{ENV['TOTAL_LIMIT_SIZE'] ||  8589934592 }" #8G
+					chunk_limit_size "#{ENV['BUFFER_SIZE_LIMIT'] || '1m'}"
+					flush_mode interval
 					flush_interval 5s
 					flush_at_shutdown true
 					flush_thread_count 2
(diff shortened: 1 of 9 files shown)


alanconway (changes requested):
Code looks OK but I don't understand the problem you're fixing. What do heartbeats have to do with DNS resolution?

jcantrill:
> Code looks OK but I don't understand the problem you're fixing. What do heartbeats have to do with DNS resolution?

The plugin generates the referenced error while trying to ensure there is a connection with destination by way of a "heartbeat".  This will cause the pod initialization to fail completely because it is unable to resolve the destination

alanconway:
> > Code looks OK but I don't understand the problem you're fixing. What do heartbeats have to do with DNS resolution?
> 
> The plugin generates the referenced error while trying to ensure there is a connection with destination by way of a "heartbeat". This will cause the pod initialization to fail completely because it is unable to resolve the destination

Right, but won't we it fail anyway it tries to connect?
I'm OK with this for now, I think we need to review how we handle & report errors from fluentd but that's another task.
/lgtm

PR #814, opened 2020-11-24
Bug 1901256: Make operator not depend on elasticsearch-operator

jcantrill:
### Description
This PR modifies the reconciliation loop to:
* skip reconciliation of storage, visualization and curation if the managed store is not defined

*Note*: while testing it was observed
* storage et al will fail (as expected) to deploy if the EO is not deployed
* It is possible to see the following error message in the logs if you deploy eo, logging, and then remove eo, logging, and delete the CRD
```
E1123 19:13:13.053559       1 reflector.go:178] sigs.k8s.io/controller-runtime/pkg/cache/internal/informers_map.go:224: Failed to list *v1.Elasticsearch: the server could not find the requested resource (get elasticsearches.logging.openshift.io)
```
Deleting the CRD is a manual step which is unlikely to be done unless a user explicitly does so

/cc @igor-karpukhin 
/assign @alanconway @vimalk78 

### Links
* https://bugzilla.redhat.com/show_bug.cgi?id=1901256
* 4.6 Backport of #811

### DIFF
--- a/hack/logforwarder.yaml
+++ b/hack/logforwarder.yaml
@@ -1,0 +1,12 @@
+apiVersion: "logging.openshift.io/v1"
+kind: ClusterLogForwarder
+metadata:
+  name: instance 
+  namespace: openshift-logging 
+spec:
+  pipelines:
+   - name: application-logs 
+     inputRefs: 
+     - application
+     outputRefs:
+     - default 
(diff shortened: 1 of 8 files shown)


igor-karpukhin (approved):
/lgtm

PR #820, opened 2020-12-01
Add testcases to test-unit Makefile target

vimalk78:
### Description
  - these testcases were already present but missing from test-unit make target
  - moved unit tests that require a cluster to functional tests
  - made oc helper tests to use functional test namespace generation
  - `oc.Literal().From()` supports varargs, so no need of separate `fmt.Sprintf()`
  - fixed race condition error in oc/runner.go

/cc @igor-karpukhin @syedriko 
/assign @jcantrill


jcantrill (commented):
/approve
/lgtm

jcantrill (commented):
/lgtm

Makefile
```diff
@@ -156,7 +156,7 @@ test-unit:
 	CURATOR_IMAGE=quay.io/openshift/origin-logging-curator:latest \
 	FLUENTD_IMAGE=$(FLUENTD_IMAGE) \
 	go test -cover -race ./pkg/...
-	go test -cover -race ./test/helpers/
+	go test -cover -race ./test/helpers/...
```
jcantrill, line 159: @vimalk78 thinking a bout this, does it make more sense to leave these helper tests in there original packages and just move the test execution to be part of functional?  Maybe we make `test-unit` only compile them for sanity but not run?

vimalk78: we could say that the tests in `./test/helpers` are only for helpers required in `e2e` or `functional` tests. the `make` default target already compiles the e2e tests for sanity, we could compile these also there.

jcantrill (commented):
/lgtm

Makefile
vimalk78: updated PR as per suggestion

jcantrill:
adding label 'valid bug' since this is to improve test code base

vimalk78:
need lgtm label

igor-karpukhin (approved):
/lgtm

PR #826, opened 2020-12-03
Modify internal/forwarder-generator to create default conf

jcantrill:
### Description
This PR:
* Modifies the internal bin/forwarder-generator to produce a default conf with no input
* Include legacy forward config

/cc @vimalk78 
/assign @alanconway


alanconway (commented):
/accept
/lgtm

PR #827, opened 2020-12-04
Bug 1901424: Skip setting kibana replicas to 1 with zero es node

huikang:
- test cases updated

### Description
This PR adds an additional condition check of the elasticsearch nodecount before setting the kibana replica to 1.

/cc @lukas-vlcek 
/assign @jcantrill

### Links
- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1901424


jcantrill (commented):
/approve
/lgtm

PR #830, opened 2020-12-06
Check that elastic search containers are ready

eranra:
### Description
This PR fix waitForElasticsearchPods function to make sure that the function really waits for "ready state" containers in elastic-search pods

/cc @jcantrill 
/assign @jcantrill 

### Links
https://coreos.slack.com/archives/GGUR75P60/p1607246317123600?thread_ts=1607243962.121800&cid=GGUR75P60

### DIFF
--- a/test/helpers/framework.go
+++ b/test/helpers/framework.go
@@ -119,7 +119,7 @@ func (tc *E2ETestFramework) DeployLogGeneratorWithNamespace(namespace string) er
 		Containers: []corev1.Container{container},
 	}
 	deployment := k8shandler.NewDeployment("log-generator", namespace, "log-generator", "test", podSpec)
-	clolog.Info("Deploying %q to namespace: %q...", deployment.Name, deployment.Namespace)
+	clolog.Info("Deploying LogGenerator to namespace", "deployment name", deployment.Name, "namespace", deployment.Namespace)
 	deployment, err := tc.KubeClient.AppsV1().Deployments(namespace).Create(context.TODO(), deployment, opts)
 	if err != nil {
 		return err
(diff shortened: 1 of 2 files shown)


test/helpers/framework.go
```diff
@@ -260,19 +260,25 @@ func (tc *E2ETestFramework) waitForElasticsearchPods(retryInterval, timeout time
 			clolog.Error(err, "Error listing elasticsearch pods")
 			return false, nil
 		}
-		if len(pods.Items) == 0 {
+		numberOfPods := len(pods.Items)
```
vimalk78, line 263: This is probably not the expected number of pods. We deploy ES cluster by creating an instance of Elasticsearch CR. better to compare with the length of `elasticsearch.spec.nodes.nodeCount`

but i think in the e2e test we create only one ES

eranra: Hi @vimalk78 thanks for this comment ... agree ... but ... I see that this is a fixed/ hardcoded value of 1 

https://github.com/openshift/cluster-logging-operator/blob/master/test/helpers/elasticsearch.go#L194

The CR is not saved to any variable, so to know nodeCount there is a need to read it back from k8s (by CR name?). 

https://github.com/openshift/cluster-logging-operator/blob/master/test/helpers/elasticsearch.go#L256

Taking into account the rest of the waiting functions (look on fluentd that do not check the count as expected) this enhancement can be postponed to following PR

eranra: @vimalk78 I see that this should theoretically be the value in tc.ClusterLogging.Spec.LogStore.NodeCount but as said above in some cases this is saved and in some other it is not ... so can't count on that
		
`

test/helpers/framework.go
```diff
@@ -260,19 +260,25 @@ func (tc *E2ETestFramework) waitForElasticsearchPods(retryInterval, timeout time
...
 			for _, status := range pod.Status.ContainerStatuses {
 				clolog.V(3).Info("Checking status of", "PodName", pod.Name, "ContainerID", status.ContainerID, "status", status.Ready)
-				if !status.Ready {
-					return false, nil
+				if status.Ready {
+					containersReadyCount ++
 				}
 			}
 		}
+		if containersReadyCount >= numberOfPods {
```
jcantrill, line 277: These numbers are not related other then there are N containers for every pod defined by ES.  I think we should rethink this check and we do a similar check in the functional framework that may be reusable here or able to be moved into a utility function

eranra: @jcantrill @vimalk78 

Please advice what to compare as this issue is one of the big killers for e2e tests. If you think we should look for ES CR and read from that ... this is doable but let me know if you want to go that path ?

eranra: Updated logic to make sure we have at least one container ready and no containers waiting ... this makes sure that the CR was read by the controller and it started to create containers and that there are no waiting containers ...

@vimalk78 @jcantrill please approve

jcantrill (commented):
/approve

jcantrill:
/lgtm

Applying valid label as this is test code

PR #831, opened 2020-12-07
WIP: Re-implementation of the certificate generating script in pure Go

syedriko:
### Description
Generate certificates using a Go library instead of shelling out to a script to make cert generation more robust and efficient

/cc @ewolinetz @alanconway 
/assign @jcantrill

### Links
- JIRA: https://issues.redhat.com/browse/LOG-422


pkg/k8shandler/certs.go
```diff
@@ -422,15 +426,40 @@ func (clusterRequest *ClusterLoggingRequest) populateCA() error {
...
 	if ret.privKey, err = rsa.GenerateKey(rand.Reader, rsaKeyLength); err != nil {
 		return
 	}
 	pubKeySHA1 := sha1.Sum(x509.MarshalPKCS1PublicKey(&ret.privKey.PublicKey))
+	var serial *big.Int
+	if serial, err = clusterRequest.incCertSerial(); err != nil {
+		return
+	}
 	cloCAMutex.Lock()
 	defer cloCAMutex.Unlock()
```
alanconway, line 460: Can you isolate the shared state that needs locking in a smaller type to make it clear why we need this lock?
I noticed is the global `extensions` map is shared state, if that's the only thing, then you could replace it with a sync.Map and drop the lock.

alanconway (changes requested):

alanconway (approved):
Approve the direction, but looks like this is unfinished.

alanconway:
@syedriko Is this dead or are you planning to finish it?

syedriko:
The plan is to get back to this, but move the code to the Elastic operator, per https://github.com/openshift/enhancements/pull/576.

alanconway:
/close
see previous comment

PR #838, opened 2020-12-13
fluentd instance not getting created in CRC error fix

sasagarw:
### Description

CRC does not start monitoring by default, so this was causing fluentd instance to not start on CRC.
Not creating prometheus rules should not block deployment, but only mark the operator as "degraded"

This PR:
- Removes the error caused by Fluentd Prometheus Rule.
- The fluentd instance will be up and running successfully.

/cc @vimalk78  
/assign

### DIFF
--- a/pkg/k8shandler/collection.go
+++ b/pkg/k8shandler/collection.go
@@ -72,7 +72,7 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection(proxyConfi
 		}
 
 		if err = clusterRequest.createOrUpdateFluentdPrometheusRule(); err != nil {
-			return
+			log.Error(err, "unable to create or update fluentd prometheus rule")
 		}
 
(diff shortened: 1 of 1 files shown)


vimalk78 (commented):
how is the operator to be marked as degraded? will OLM detect it? or we need to update the status as such?

jcantrill:
We should update status

pkg/k8shandler/collection.go
```diff
@@ -72,7 +72,7 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection(proxyConfi
 		}
 
 		if err = clusterRequest.createOrUpdateFluentdPrometheusRule(); err != nil {
-			return
+			log.Error(err, "unable to create or update fluentd prometheus rule")
```
alanconway, line 75: This is a serious condition we should not simply drop it into the logs and carry on. It needs to be notified on status conditions and/or alerted. I will LGTM this since it's better than nothing, but we need to come back to this kind of error mis-handling and clean it up.

/lgtm

PR #839, opened 2020-12-15
LOG-1002: Add cloudwatch logforwarding type

jcantrill:
### Description
This PR add logforwarding to Cloudwatch capabilities

/cc @igor-karpukhin 
/assign @alanconway 

### Links
https://issues.redhat.com/browse/LOG-1002

/hold


alanconway (approved):
/lgtm

We may want to modify the naming and API but we should merge this as a first cut and work from there.

jcantrill:
> We may want to modify the naming and API but we should merge this as a first cut and work from there.

@alanconway since the API is firm here, I hesitate to merge because then that IS the API.  I'm still working to create TP branches where we could merge this and call it beta.  That's ultimately what I would most desire to deliver for ROSA unless we are confident in the api

alanconway:
/lgtm cancel

Here's the API enhancement proposal: https://github.com/openshift/enhancements/pull/570
Not a big change, has some small extra bits.

jcantrill:
@alanconway review comments please

alanconway:
LGTM but do we want to start the tech-preview branch for this?

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -590,6 +591,50 @@ const pipelineToOutputCopyTemplate = `{{- define "pipelineToOutputCopyTemplate"
...
+    @type cloudwatch_logs
+    auto_create_stream true
+    region {{ .Target.Cloudwatch.Region }}
+    log_group_name_key cw_group_name
+	log_stream_name_key cw_stream_name
+    remove_log_stream_name_key true
+    remove_log_group_name_key true
+    auto_create_stream true
+    concurrency 2
+    aws_key_id "#{open('{{ .SecretPath "aws_access_key_id"}}','r') do |f|f.read end}"
```
alanconway, line 630: Only set this if the key is found, otherwise we'll have a new  crash loop when its missing. Here's what I did for the general TLS stuff:

```
{{- with $path := .SecretPathIfFound "tls.key"}}
tls_client_private_key_path "{{$path}}"
{{- end}}
```

Also if these keys are required, they should be checked as part of VerifyOutputSecret, but only for cloudwatch outputs.

pkg/k8shandler/clusterloggingrequest.go
```diff
@@ -25,6 +25,14 @@ type ClusterLoggingRequest struct {
 
 	// OutputSecrets are retrieved during validation and used for generation.
 	OutputSecrets map[string]*corev1.Secret
+
+	//CLFVerifier is a collection of functions to control verification
+	//of ClusterLogForwarding
+	CLFVerifier ClusterLogForwarderVerifier
```
alanconway, line 31: Not sure this is the best approach. We need to customize the verify behavior per-output, not for the whole forwarder,

Perhaps a map of map[string]VerifyFunc = { "cloudwatch": ...; "kafka": ... }, so Normalize can call the right verify func for each output type.

alanconway (changes requested):
Looks good, minor changes requested.

pkg/generators/forwarding/fluentd/fluent_conf.go
```diff
@@ -74,6 +74,29 @@ func (conf *outputLabelConf) Port() string {
...
+			return "${record['kubernetes']['namespace_id']}"
+		default:
+			return logging.InputNameApplication
+		}
+	}
+	return ""
+}
+func (conf *outputLabelConf) LogGroupPrefix() string {
+	if conf.Target.Type == logging.OutputTypeCloudwatch {
+		prefix := conf.Target.Cloudwatch.GroupPrefix
```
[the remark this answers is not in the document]
jcantrill: I don't believe I can do that without losing the ability to treat the following as different  cases:
* nil = use default
* non-nil and not empty = use value

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -590,6 +591,50 @@ const pipelineToOutputCopyTemplate = `{{- define "pipelineToOutputCopyTemplate"
 </label>
 {{- end}}`
 
+const outputLabelConfCloudwatch = `{{- define "outputLabelConfCloudwatch" -}}
```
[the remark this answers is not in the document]
jcantrill: Added to GA ticket for completion https://issues.redhat.com/browse/LOG-1102

pkg/generators/forwarding/fluentd/templates.go
jcantrill: These are absolutely required for the plugin to work.  Will the generator code prior to the template successfully skip output generation for missing secrets?

pkg/k8shandler/clusterloggingrequest.go
jcantrill: This was added to allow the internal config generator bin to work without being connected to the cluster and really is not intended to be used otherwise.  We can repurpose it if needed but ATM I don't seed the need.

alanconway: OK, understood. However we still need different secret verification for cloudwatch (needs "aws_...") vs. other TLS outputs. We could have a global map of verify functions keyed by output type, and have output types with special needs put a function in the map that the validation code will call if not nil. What do you think?

jcantrill: sure.  if that is needed seems reasonable

pkg/generators/forwarding/fluentd/templates.go
jcantrill: added

pkg/k8shandler/clusterloggingrequest.go
alanconway: Not needed for this PR. This is OK

alanconway (commented):
/lgtm

PR #841, opened 2020-12-16
Functional tests for LogForwarding EventRouter messages

vparfonov:

### Description
Functional tests for Logforwarding EventRouter messages

/cc @jcantrill 
/assign @jcantrill

### Links

- JIRA:https://issues.redhat.com/browse/LOG-1008


igor-karpukhin (approved):
/lgtm

test/functional/normalization/eventrouter_test.go
```diff
@@ -0,0 +1,107 @@
...
+			FromInput(logging.InputNameApplication).
+			ToFluentForwardOutput()
+		Expect(framework.Deploy()).To(BeNil())
+	})
+	AfterEach(func() {
+		framework.Cleanup()
+	})
+
+	It("Should parse eventrouter `ADDED` message and check values", func() {
+		pod := types.MockPod()
```
jcantrill, line 35: There is code here that is duplicated and can be merged into either `var` or BeforeEach as well as utility functions (maybe for the common expectations).

test/helpers/types/types.go
```diff
@@ -54,6 +59,14 @@ type Collector struct {
 	Version    string    `json:"version"`
 }
 
+// EventData encodes an eventrouter event and previous event, with a verb for
+// whether the event is created or updated.
+type EventData struct {
```
jcantrill, line 64: Do we need to define "Event" here in lieu of something that already formally exists in Kubernetes?

test/helpers/types/types.go
```diff
@@ -336,3 +349,40 @@ func (l Logs) NonEmpty() bool {
...
+			Name:      podName,
+			Namespace: namespace,
+			UID:       types.UID(uuidStr),
+		},
+		Spec: v1.PodSpec{},
+	}
+	return &fakePod
+}
+
+func MockEvent(ref *v1.ObjectReference, eventType, reason, message string) *v1.Event {
```
jcantrill, line 371: Can we use naming conventions for constructors here?   maybe "NewMockEvent" or "NewTestEvent"

test/functional/normalization/eventrouter_test.go
```diff
@@ -0,0 +1,105 @@
...
+		event := eventData.Event
+		Expect(event).ToNot(BeNil())
+		Expect(event.Type).To(BeEquivalentTo(corev1.EventTypeNormal))
+		Expect(event.Message).To(BeEquivalentTo(message))
+		Expect(event.Reason).To(BeEquivalentTo(reason))
+		Expect(event.InvolvedObject.Name).To(BeEquivalentTo(podName))
+		Expect(event.InvolvedObject.Namespace).To(BeEquivalentTo(ns))
+	})
+
+	It("Should parse eventrouter `UPDATED` message and check values", func() {
```
jcantrill, line 67: @alanconway @eranra What are your thoughts on writing this test using the template style that @eranra introduced for audit logs?  Is there reason we should be validating values more strictly or is it sufficient to know we transformed an "event" log to include the desired fields?

test/helpers/types/types.go
vparfonov: "EventData" is EventRouter citizen, with Kubernetes Event inside

test/helpers/types/types.go
vparfonov: Ok

test/functional/normalization/eventrouter_test.go
alanconway: I think it is sufficient to check the expected change, though I wouldn't object to more validation.
As long as we have some tests that do full validation, I don't think every test needs to - unless testing something that is likely to invalidate unknown parts of the record.

eranra: @jcantrill the purpose of that templates idea was exactly cases like this one that there are multiple `expect` statements and instead of that code using one nice and more understandable template. I get @alanconway point that it is not critical ... I can help to create if this is needed,.

BTW; It is consistent in the code (tests, also prod) that there are multiple ways/frameworks for same functionalities ... I see that as very sick area of the code base and that consolidating that into more agreed and coherent one-way of doing things will accelerate development and increase correctness and debuggability. So for example in the templates case ... I advice either to start using that in all new tests (like this one) or revert to multiple `Expect` statements ... having both for long time in the code without pure direction does not feel good to me.

vparfonov: I can rewrite with `FitLogFormatTemplate` function, but I think it will be complicated for code reading because need to know how actually function works. BTW `FitLogFormatTemplate` also can provide errors. But I agree that need to do in one way

jcantrill: @vparfonov please rewrite the test to use the template validation so we have consistency

vparfonov: ok

test/functional/normalization/eventrouter_test.go
```diff
@@ -0,0 +1,128 @@
...
+	AfterEach(func() {
+		framework.Cleanup()
+	})
+
+	It("Should parse EventRouter `ADDED` message and check values", func() {
+		podRef, err := reference.GetReference(scheme.Scheme, pod)
+		Expect(err).To(BeNil())
+		newEvent := types.NewMockEvent(podRef,
+			corev1.EventTypeNormal,
+			string(utils.GetRandomWord(8)),
```
eranra, line 64: I propose to change to some static string like "reason" it will be easier to debug and understand than using random letters

test/functional/normalization/eventrouter_test.go
```diff
@@ -0,0 +1,128 @@
...
+			string(utils.GetRandomWord(32)))
```
eranra, line 65: same ^^^

test/functional/normalization/eventrouter_test.go
```diff
@@ -0,0 +1,128 @@
...
+	"github.com/openshift/cluster-logging-operator/test/helpers/types"
+	"github.com/openshift/cluster-logging-operator/test/matchers"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/reference"
+	"strings"
+	"time"
+)
+
+const timestamp string = "1985-10-21T09:00:00.00000+00:00"
```
eranra, line 20: You can move this under `Describe`, no need for this to be global variable

test/functional/normalization/eventrouter_test.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	templateForAnyCollector = types.PipelineMetadata{Collector: types.Collector{
+		Ipaddr4:    "*",
+		Inputname:  "*",
+		Name:       "*",
+		Version:    "*",
+		ReceivedAt: time.Time{},
+	},
+	}
+	nanoTime, _ = time.Parse(time.RFC3339Nano, timestamp)
```
eranra, line 38: Exactly like `timestamp` this can move into the Describe context (you will need to pass `nanoTime`  to the template creator but this is ok )

test/functional/normalization/eventrouter_test.go
```diff
@@ -0,0 +1,128 @@
...
+		msg := strings.ReplaceAll(fmt.Sprintf("%s stdout F %s", timestamp, jsonStr), "\"", "\\\"")
+		err = framework.WriteMessagesToApplicationLog(msg, 1)
+		Expect(err).To(BeNil())
+
+		raw, err := framework.ReadApplicationLogsFrom("fluentforward")
+		Expect(err).To(BeNil(), "Expected no errors reading the logs")
+		var logs []types.EventRouterLog
+		err = types.StrictlyParseLogs(raw, &logs)
+		Expect(err).To(BeNil(), "Expected no errors parsing the logs")
+		var outputLogTemplate = OutPutTemplate(jsonStr)
```
eranra, line 78: Consider change to variable to `expectedLogTemplate` and function name to `ExpectedLogTemplateBuilder`

In anyevent `OutPut` is one word in english ... so use `Output` and not `OutPut ` :-)

eranra (commented):
Most of the comments are minor ... I tested the code with debugger and it works fine and validates all relevant fields ... so this is almost ready. Good work 👍

test/functional/normalization/eventrouter_test.go
vparfonov: fixed

test/functional/normalization/eventrouter_test.go
vparfonov: fixed

test/functional/normalization/eventrouter_test.go
vparfonov: fixed

test/functional/normalization/eventrouter_test.go
vparfonov: fixed

test/functional/normalization/eventrouter_test.go
```diff
@@ -0,0 +1,118 @@
...
+			ToFluentForwardOutput()
+		Expect(framework.Deploy()).To(BeNil())
+		pod = types.NewMockPod()
+	})
+	AfterEach(func() {
+		framework.Cleanup()
+	})
+
+	It("Should parse EventRouter `ADDED` message and check values", func() {
+		podRef, err := reference.GetReference(scheme.Scheme, pod)
```
jcantrill, line 59: There is duplicate code here from each "It" that can extracted into a function to condense the work.  You might even consider adding it in `var` as a named function of some kind.  Given the similarity for all of these I wonder if you can do something like:

```
Describe("my test", func(){
  for _, verb := range []string{"ADDED","UPDATED"} {
    It(fmt.Sprintf("should parse %s", verb), func(){
       //testcode
    })
  }
})
```

PR #847, opened 2021-01-07
Bug 1905910: Correctly extract master-certs to workind directory

jcantrill:
### Description
This PR
* Corrects the cert extraction logic to extract all certs from the master-cert secret
* Reformats the cert gen script output to be parsable to allow explicit understandying of why certs were regenerated
* Adds a test to expose the issue

/cc @anpingli 
/assign @ewolinetz @vimalk78 

### Links
* https://bugzilla.redhat.com/show_bug.cgi?id=1905910

### DIFF
--- a/pkg/certificates/certificates.go
+++ b/pkg/certificates/certificates.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"os/exec"
 
+	"sigs.k8s.io/yaml"
+
 	"github.com/ViaQ/logerr/log"
 )
 
@@ -21,8 +23,25 @@ func RunCertificatesScript(namespace, logStoreName, workDir, script string) (err
 	log.V(3).Info("Cert generation", "out", result, "err", err)
 	if result != "" {
 		updated = true
-		log.Info("cert_generation output", "output", result)
+		dumpLogs(result)
 	}
 	log.V(3).Info("Returning", "err", err, "updated", updated)
 	return err, updated
+}
+
+func dumpLogs(raw string) {
+	genLogs := []map[string]string{}
+	err := yaml.Unmarshal([]byte(raw), &genLogs)
+	if err == nil {
+		for _, eventlog := range genLogs {
+			keyvalues := []interface{}{}
+			for k, v := range eventlog {
+				keyvalues = append(keyvalues, k, v)
+			}
+			log.Info("cert_generation output", keyvalues...)
+		}
+		return
+	}
+	log.Error(err, "Unable to unmarshal cert_generation to structured output")
+	log.Info("cert_generation output", "output", raw)
 }
--- a/scripts/cert_generation.sh
+++ b/scripts/cert_generation.sh
@@ -10,34 +10,58 @@ REGENERATE_NEEDED=0
 
 function info() {
   local msg=$1
-  echo "[INFO] ${msg}" >> ${WORKING_DIR}/action.log
+  local type=${2:-}
+  local reason=${3:-}
+  local out="\"msg\":\"$msg\""
+  if [ -n "$type" ] ; then
+    out="$out, \"type\":\"$type\""
+  fi
+  if [ -n "$reason" ] ; then
+    out="$out, \"reason\":\"$reason\""
+  fi
+  if [ -f ${WORKING_DIR}/action.log ] && [ "$(stat -c%s ${WORKING_DIR}/action.log)" != "0" ] ; then
+    echo "," >> ${WORKING_DIR}/action.log
+  fi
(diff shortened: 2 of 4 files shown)


pkg/certificates/certificates.go
```diff
@@ -21,8 +23,25 @@ func RunCertificatesScript(namespace, logStoreName, workDir, script string) (err
...
+			keyvalues := []interface{}{}
+			for k, v := range eventlog {
+				keyvalues = append(keyvalues, k, v)
+			}
+			log.Info("cert_generation output", keyvalues...)
+		}
+		return
+	}
+	log.Error(err, "Unable to unmarshal cert_generation to structured output")
+	log.Info("cert_generation output", "output", raw)
```
ewolinetz, line 46: you have both an error and info level here? why not move the `"output", raw` into the Error log?

scripts/cert_generation.sh
```diff
@@ -306,6 +336,7 @@ if [ ! -s "${WORKING_DIR}/kibana-session-secret" ] ; then
 fi
 
 if [ -f ${WORKING_DIR}/action.log ] ; then
-  cat ${WORKING_DIR}/action.log
+  sed -i ':a;N;$!ba;s/\n//g' ${WORKING_DIR}/action.log
```
ewolinetz, line 339: can you add a comment to what this sed line is doing?

scripts/cert_generation.sh
```diff
@@ -10,34 +10,58 @@ REGENERATE_NEEDED=0
...
     echo 00 > ${WORKING_DIR}/ca.serial.txt
   fi
 }
 
 function generate_signing_ca() {
   local regen=false
-  if [ ! -f ${WORKING_DIR}/ca.crt ] || [ "$(file -b ${WORKING_DIR}/ca.crt)" != "PEM certificate" ] ; then
-    info "Missing file or invalid type: ${WORKING_DIR}/ca.crt"
+  local fileExists=$(test -f ${WORKING_DIR}/ca.crt;echo $?)
+  if [ "$fileExists" == "1" ] ; then
```
ewolinetz, line 44: given we never use `fileExists` ever again after this, can we just keep it in the `if` block? needing to do an `echo $?` and then check that value manually seems like an anti-pattern here

scripts/cert_generation.sh
```diff
@@ -10,34 +10,58 @@ REGENERATE_NEEDED=0
...
+    info "${WORKING_DIR}/ca.crt" "Regenerate" "FileMissing"
+    regen=true
+  fi
+  local filetype=$(file -b ${WORKING_DIR}/ca.crt)
```
ewolinetz, line 48: same here

scripts/cert_generation.sh
```diff
@@ -10,34 +10,58 @@ REGENERATE_NEEDED=0
...
+  if [ "$fileExists" == "0" ] && [ "$filetype" != "PEM certificate" ] ; then
+    info "${WORKING_DIR}/ca.crt '$filetype' != 'PEM certificate'" "Regenerate" "InvalidFileType"
     regen=true
   fi
-  if [ ! -f ${WORKING_DIR}/ca.key ] || [ "$(file -b ${WORKING_DIR}/ca.key)" != "ASCII text" ] ; then
-    info "Missing file or invalid type: ${WORKING_DIR}/ca.key"
+  fileExists=$(test -f ${WORKING_DIR}/ca.key;echo $?)
```
ewolinetz, line 53: same here

pkg/certificates/certificates.go
jcantrill: This is intentional.  The PR adds the ability to have structured logs from the script which should help clarify why cert regen is happening.  What i wanted here was to preserve the previous output in the event we could not parse the output to provide some indicator that regen occurred.  So the logs here are:  Error and tell me why; give me the original output too.

scripts/cert_generation.sh
jcantrill: we use this in line 49.  I purposely added this distinction in order to get the singular reason for regen; missing versus filetype.  It otherwise logged both which did not pinpoint why

scripts/cert_generation.sh
jcantrill: separated "filetype" in order to be able to add the outcome to a log message to be returned to the caller.  all of this will be much better when we move to golang based cert cen

scripts/cert_generation.sh
jcantrill: yes.  this is condensing the output and removing newlines.  Not sure if its needed in order to properly consume on the golang side...

PR #851, opened 2021-01-11
Verify labels attached to logs on rsyslog server

sasagarw:
### Description

This PR:
- Adds an e2e tests to verify that the labels are attached to the logs on rsyslog server for rfc5424 and rfc3164.

/cc @anpingli  
/assign @vimalk78 @jcantrill  

/cherry-pick

### DIFF
--- a/test/e2e/logforwarding/syslog/forward_to_syslog_test.go
+++ b/test/e2e/logforwarding/syslog/forward_to_syslog_test.go
@@ -23,7 +23,8 @@ import (
 
 const (
 	// grep "pod_name"=>"<generator-pod-name>" <syslog-log-filename>
-	rsyslogFormatStr = `grep %s %%s| grep pod_name | tail -n 1 | awk -F' ' '{print %s}'`
+	rsyslogFormatStr       = `grep %s %%s| grep pod_name | tail -n 1 | awk -F' ' '{print %s}'`
+	rsyslogLabelsFormatStr = `grep %s %%s| grep pod_name | tail -n 1 | sed -n -e 's/^.*openshift://p'`
 )
 
 var _ = Describe("[ClusterLogForwarder] Forwards logs", func() {
@@ -41,6 +42,7 @@ var _ = Describe("[ClusterLogForwarder] Forwards logs", func() {
 		grepprocid       string
 		grepmsgid        string
 		greptag          string
+		grepLabels       string
 		logGenPod        string
 		logGenNS         string
 	)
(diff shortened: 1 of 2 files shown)


test/e2e/logforwarding/syslog/forward_to_syslog_test.go
```diff
@@ -139,6 +143,17 @@ var _ = Describe("[ClusterLogForwarder] Forwards logs", func() {
...
 				expectedProcID := forwarder.Spec.Outputs[0].Syslog.ProcID
 				Expect(logStore.GrepLogs(grepprocid, helpers.DefaultWaitForLogsTimeout)).To(Equal(expectedProcID), "Expected: "+expectedProcID)
+				grepMsgContent := fmt.Sprintf(`grep %s %%s| grep pod_name | tail -n 1 | sed -n -e 's/^.*openshift://p'`, logGenPod)
+				str, err := logStore.GrepLogs(grepMsgContent, helpers.DefaultWaitForLogsTimeout)
+				Expect(err).To(BeNil())
+				str = strings.ReplaceAll(str, "=>", ":")
+				msg := map[string]map[string]string{}
+				err = json.Unmarshal([]byte(str), &msg)
+				Expect(err).To(BeNil())
+				expectedLabelSecure := forwarder.Spec.Pipelines[0].Labels["secure"]
```
jcantrill, line 153: This is part of the setup which you already know that could be defined initially as a `const`.  Doing so earlier in the test will allow you to remove these lines making the test more clear.  Comment applies to all these labels

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
```diff
@@ -139,6 +143,17 @@ var _ = Describe("[ClusterLogForwarder] Forwards logs", func() {
 				Expect(logStore.GrepLogs(grepmsgid, helpers.DefaultWaitForLogsTimeout)).To(Equal(expectedMsgID), "Expected: "+expectedMsgID)
 				expectedProcID := forwarder.Spec.Outputs[0].Syslog.ProcID
 				Expect(logStore.GrepLogs(grepprocid, helpers.DefaultWaitForLogsTimeout)).To(Equal(expectedProcID), "Expected: "+expectedProcID)
+				grepMsgContent := fmt.Sprintf(`grep %s %%s| grep pod_name | tail -n 1 | sed -n -e 's/^.*openshift://p'`, logGenPod)
```
jcantrill, line 146: please implement the needed method in `syslog.go`:
```
func (syslog *syslogReceiverLogStore) HasApplicationLogs(timeToWait time.Duration) (bool, error) {
	return false, fmt.Errorf("Not implemented")
}
```

jcantrill (commented):
/approve
/hold

cc @vimalk78

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
sasagarw: Why we need to implement this method? The logs generated by log-generator pod while testing is shipped to infra.log. There is no application logs generated during the process.

test/e2e/logforwarding/syslog/forward_to_syslog_test.go
sasagarw: Do you mean to define all the "expected" variables as constants above?

jcantrill: reduce code duplication as much as we can where it makes sense

jcantrill:
@vimalk78 @sasagarw is this still a PR we want to pursue?

PR #860, opened 2021-01-14
Bug 1904071: Fluentd chunkLimitSize wrong values echoed

sasagarw:
### Description

This PR
- echoes correct chunkLimitSize which is set in ClusterLogging CR
- adds environment variable on fluentd pod

/cc @vimalk78  
/assign @jcantrill  

/cherry-pick 

### Links

- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1904071


pkg/k8shandler/fluentd.go
```diff
@@ -361,6 +361,19 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, proxyConfig *configv1.Pr
 		{Name: "POD_IP", ValueFrom: &v1.EnvVarSource{FieldRef: &v1.ObjectFieldSelector{APIVersion: "v1", FieldPath: "status.podIP"}}},
 	}
 
+	forwarderSpec := logging.ForwarderSpec{}
+	if cluster.Spec.Forwarder != nil {
+		forwarderSpec = *cluster.Spec.Forwarder
+		if forwarderSpec.Fluentd.Buffer != nil {
+			if forwarderSpec.Fluentd.Buffer.ChunkLimitSize != "" {
+				fluentdContainer.Env = append(fluentdContainer.Env, v1.EnvVar{Name: "BUFFER_SIZE_LIMIT", Value: string(forwarderSpec.Fluentd.Buffer.ChunkLimitSize)})
```
jcantrill, line 369: I don't believe this is the correct solution.  These values should be calculated as part of the run.sh

sasagarw: I thought in that way too then we should ssh into the fluentd pod and grep these values. Is this what you feel is more correct way?
BUFFER_SIZE_LIMIT=$(cat fluent.conf | grep chunk_limit_size | awk 'NR==1{print $2}' | sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc)

sasagarw: Other than this we don't have access to values because it's not present in fluentd pod environment

PR #863, opened 2021-01-15
Bug 1904071: Fluentd chunkLimitSize wrong values echoed

sasagarw:
### Description

This PR
- echoes correct chunkLimitSize which is set in ClusterLogging CR
- adds environment variable on fluentd pod
- adds a util for serializing/deserializing values such as - 750k, 8m, 1g

/cc @vimalk78 
/assign @jcantrill 

### Links
- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1904071


files/fluentd/run.sh
```diff
@@ -77,7 +77,13 @@ export TOTAL_LIMIT_SIZE
 # Calculate the max number of queued chunks given the size of each chunk
 # and the max allowed space per buffer
 ##
-BUFFER_SIZE_LIMIT=$(echo ${BUFFER_SIZE_LIMIT:-8m} |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc)
+BUFFER_SIZE_RE='^[0-9]+[kKmMgG]?$'
```
jcantrill, line 80: I'm missing the reason we need to execute a regex against this value.  The sed will do nothing if there is no match leaving you with the raw value.  Also we can default missing values with straight bash substitutions

sasagarw: Because in fluent.conf file, if we havn't provided any value in ClusterForwarder CR then chunk_limit_size will be set to "#{ENV['BUFFER_SIZE_LIMIT'] || '8m'}". So if this is set then will go to else and see if environment variable is set. If that's also not set then default value of "8m" will be set as chunk_limit_size

alanconway: It's pretty horrible to be doing this in bash at all. Why aren't we doing this in Go using the standard https://godoc.org/k8s.io/apimachinery/pkg/api/resource Quantity to serialize/deserialize?
@sasagarw I understand you're following the existing pattern, but maybe it's time to do some clean up.

pkg/utils/resource/quantity.go
```diff
@@ -0,0 +1,767 @@
...
+			break Suffix
+		}
+	}
+	// we encountered a non decimal in the Suffix loop, but the last character
+	// was not a valid exponent
+	err = ErrFormatWrong
+	return
+}
+
+// ParseQuantity turns str into a Quantity, or returns an error.
```
alanconway, line 263: Don't copy the k8s resource code here. We just need to add the modified parse function, we can import the k8s resource package:

alanconway:
```
package scratch

import (	
	 "k8s.io/apimachinery/pkg/api/resource"
)

func ParseQuantity(str string) (resource.Quantity, error) { ... }

````
Then we can use it as `q, err := utils.ParseQuantity(x)` to get support for our string format, but leave maintenance of the Quantity code to the k8s project.

**Review by alanconway — CHANGES_REQUESTED** [review]

**Review by alanconway — CHANGES_REQUESTED** [review]
Looks good. Please add unit tests to the utils and k8shandler packages to validate that it parses a few values correctly and sets fluent environment as expected. After that I will approve.

File: pkg/utils/quantity.go [hunk 3]
```diff
@@ -0,0 +1,90 @@
+package utils
```
jcantrill, line 1: Is this  a copy of another file elsewhere? Are there unit tests we can bring in too?

pkg/k8shandler/fluentd.go
```diff
@@ -361,6 +362,21 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, proxyConfig *configv1.Pr
...
+	if cluster.Spec.Forwarder != nil {
+		if cluster.Spec.Forwarder.Fluentd.Buffer != nil {
+			if cluster.Spec.Forwarder.Fluentd.Buffer.ChunkLimitSize != "" {
+				if chunkLimitSize, err := utils.ParseQuantity(string(cluster.Spec.Forwarder.Fluentd.Buffer.ChunkLimitSize)); err == nil {
+					fluentdContainer.Env = append(fluentdContainer.Env, v1.EnvVar{Name: "BUFFER_SIZE_LIMIT", Value: strconv.FormatInt(chunkLimitSize.Value(), 10)})
+				}
+			}
+			if cluster.Spec.Forwarder.Fluentd.Buffer.TotalLimitSize != "" {
+				if totalLimitSize, err := utils.ParseQuantity(string(cluster.Spec.Forwarder.Fluentd.Buffer.TotalLimitSize)); err == nil {
+					fluentdContainer.Env = append(fluentdContainer.Env, v1.EnvVar{Name: "TOTAL_LIMIT_SIZE", Value: strconv.FormatInt(totalLimitSize.Value(), 10)})
```
jcantrill, line 374: It looks like the run script doesn't take this into account.  It will require adjustments to skip calculations of the value if it is provided by the operator.  

We may also want to add some sort of warning message in the run script about exceeding disk value.  @alanconway we calculate this value now to avoid using all the disk which this change and allowed configuration risks re-introducing

pkg/utils/quantity.go
alanconway: It's based on the k8s resources package, just providing a parser that accepts our suffixes as well as the k8s standard ones for backwards compat. Does need some unit tests.

pkg/k8shandler/fluentd.go
alanconway: @sasagarw does this value need to be modified in this PR?

sasagarw: @alanconway not compulsorily required in this PR. We can skip for now. I will just work with fluentd chunk_limit_size value.

alanconway (approved):
/lgtm

PR #864, opened 2021-01-15
LOG-784: Rework checking for 'managementState' to make deployment working if it set

vparfonov:

### Description
Rework checking for 'managementState' to make deployment working if it set to "unmanagement" or not set at all

/cc 
/assign 

/cherry-pick 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-784

### DIFF
--- a/hack/cr-unmanaged.yaml
+++ b/hack/cr-unmanaged.yaml
@@ -3,6 +3,7 @@ kind: "ClusterLogging"
 metadata:
   name: "instance"
 spec:
+  managementState: "Unmanaged"
   logStore:
     type: "elasticsearch"
     elasticsearch:
--- a/hack/testing-olm/assertions
+++ b/hack/testing-olm/assertions
@@ -16,3 +16,14 @@ assert_kibana_instance_exists() {
 	# verify kibana crd
 	os::cmd::try_until_success "oc -n $NAMESPACE get kibana kibana" ${TIMEOUT_MIN}
 }
+
+assert_resources_does_not_exist(){
+	os::cmd::try_until_failure "oc -n $NAMESPACE get cronjob curator"
+	os::cmd::try_until_failure "oc -n $NAMESPACE get ds fluentd"
+	os::cmd::try_until_failure "oc -n $NAMESPACE get elasticsearch elasticsearch"
+
+}
+
+assert_kibana_instance_does_not_exists() {
+	os::cmd::try_until_failure "oc -n $NAMESPACE get kibana kibana"
+}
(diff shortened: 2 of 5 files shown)


pkg/k8shandler/logstore.go
```diff
@@ -288,9 +288,9 @@ func (clusterRequest *ClusterLoggingRequest) removeElasticsearchCR(elasticsearch
 
 func (clusterRequest *ClusterLoggingRequest) createOrUpdateElasticsearchCR() (err error) {
 
-	if !clusterRequest.isManaged() {
-		return nil
-	}
+	//if !clusterRequest.isManaged() {
```
jcantrill, line 291: This makes sense, we shouldn't need this guard anymore

pkg/k8shandler/logstore.go
```diff
@@ -311,11 +311,15 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateElasticsearchCR() (er
 		return fmt.Errorf("Failure creating Elasticsearch CR: %v", err)
 	}
 
-	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
-		return clusterRequest.updateElasticsearchCRIfRequired(esCR)
-	})
-
-	return retryErr
+	if clusterRequest.isManaged() {
```
jcantrill, line 314: This change should not be necessary

pkg/controller/clusterlogging/clusterlogging_controller.go
```diff
@@ -82,9 +82,9 @@ func (r *ReconcileClusterLogging) Reconcile(request reconcile.Request) (reconcil
 		return reconcile.Result{}, err
 	}
 
-	if instance.Spec.ManagementState == loggingv1.ManagementStateUnmanaged {
-		return reconcile.Result{}, nil
-	}
+	//if instance.Spec.ManagementState == loggingv1.ManagementStateUnmanaged {
```
jcantrill, line 85: I would expect a check like this and something similar may also be needed in the other controllers.  Essentially:
* Managed or blank or nil = "Managed" - continue
* Unmanaged = "Unmanaged" - do nothing and return early

pkg/k8shandler/logstore.go
syedriko: We shouldn't be leaving this commented out. If we don't need it - remove it, there's revision history in git we can go back to.

pkg/controller/clusterlogging/clusterlogging_controller.go
vparfonov: In case we do: `Unmanaged = "Unmanaged" - do nothing and return early`.
[k8shandler.Reconcile()](https://github.com/openshift/cluster-logging-operator/blob/master/pkg/controller/clusterlogging/clusterlogging_controller.go#L89) will never be called even for create (this is how it works now in master) and instances will not be created

jcantrill: yes.  if you are initially creating an "unmanaged" clusterlogging instance it is reasonable we do nothing; that is what you told us

periklis (commented):
Looks good to me, leaving for @jcantrill to put his stamp

jcantrill (commented):
/approve
/lgtm

PR #871, opened 2021-01-29
LOG-1034: Allow selectively running e2e tests for EO

jcantrill:
### Description
This PR modifies Makefile and e2e scripts to allow selectively running tests for cases like the EO that only needs to be gated on the managed ES test

### Links
https://issues.redhat.com/browse/LOG-1034

cc @ewolinetz

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,8 @@ export NAMESPACE?=openshift-logging
 
 FLUENTD_IMAGE?=quay.io/openshift/origin-logging-fluentd:latest
 REPLICAS?=0
+export E2E_TEST_INCLUDES?=
+export CLF_TEST_INCLUDES?=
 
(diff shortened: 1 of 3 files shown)


ewolinetz:
this seems overly complicated... so we'll still have to update both repos?
why not just have a second ci target specific for log forwarding?

jcantrill:
> this seems overly complicated... so we'll still have to update all repos?
It's a minor change to existing jobs and the only changes are to the CI definitions themselves for EO and elasticsearch-proxy. The changes are minimal

> why not just have a second ci target specific for log forwarding?
From CLO perspective,  it makes sense for all e2e tests to be run together in lieu of having to add and maintain another job.  This change is not invasive and accomplishes the task of allowing the ability to only run specific tests.

jcantrill:
cc @vimalk78 @alanconway

igor-karpukhin (approved):
/lgtm

jcantrill:
Adding bz/valid label to unblock testing for EO.  This change only affects test code

PR #884, opened 2021-02-05
Master into postrelease

jcantrill:
### Description

Update master-post-release to latest master

cc @vimalk78


alanconway:
Do we need postrelease any more? If we use a release-5.0 branch for the release shut-down, then we can keep using master for 5.1 development.

jcantrill:
> Do we need postrelease any more? If we use a release-5.0 branch for the release shut-down, then we can keep using master for 5.1 development.

@alanconway we likely do not, but we do need at least one last merge of master into this and this back to master as this branch has some changes we landed for 5.1

jcantrill:
overriding test as its a failure in eventrouter that is fixed on master

jcantrill:
manually merging to get back into master

PR #890, opened 2021-02-12
LOG-1125: Set component NO_PROXY,HTTP_PROXY from environment variables

jcantrill:
### Description
This PR address the issue by:
* relying on OLM to set proxy vars on the operator which can then be propagated to the operands
* Removes watching the proxy config object directly

/cc @ewolinetz  @vimalk78 

### Links
* https://bugzilla.redhat.com/show_bug.cgi?id=1915448
* https://issues.redhat.com/browse/LOG-1125


syedriko:
@jcantrill Can we come up with a better title? This one doesn't make sense.

jcantrill:
> @jcantrill Can we come up with a better title? This one doesn't make sense.

better?

pkg/utils/utils.go
```diff
@@ -418,40 +417,15 @@ func EnvVarResourceFieldSelectorEqual(resource1, resource2 v1.ResourceFieldSelec
...
-			Value: proxyConfig.Status.NoProxy,
-		},
-			v1.EnvVar{
-				Name:  "no_proxy",
-				Value: proxyConfig.Status.NoProxy,
+	for _, envvar := range []string{"HTTPS_PROXY", "https_proxy", "HTTP_PROXY", "http_proxy", "NO_PROXY", "no_proxy"} {
+		if value := os.Getenv(envvar); value != "" {
+			envVars = append(envVars, v1.EnvVar{
+				Name:  envvar,
+				Value: value,
```
ewolinetz, line 426: just to clarify (and maybe you want to add a comment so its clear why this is working) we're checking for these env vars on the CLO because OLM will add it to the operator deployment, right?

ewolinetz:
looks straight forward to me, we may want to add a quick comment so this doesn't seem like complete magic why it works the way it does

pkg/utils/utils.go
jcantrill: > just to clarify (and maybe you want to add a comment so its clear why this is working) we're checking for these env vars on the CLO because OLM will add it to the operator deployment, right?

This was the message as I understood from OLM team, OLM will inject those vars into the operator deployments

PR #894, opened 2021-02-16
LOG-1143: fix using spec'd secret in config instead of output name

jcantrill:
### Description
fix using the CLF spec'd secret instead of assuming its the same as the output name

/cc @igor-karpukhin @vimalk78 
/assign @alanconway 

### Links
https://issues.redhat.com/browse/LOG-1143


alanconway (approved):
/lgtm
/approve

PR #906, opened 2021-02-19
LOG-962: Add test to verify forward to Logstash

jcantrill:
### Description
This PR:

* adds test to verify forwarding to logstash over forward protocol

### Links
* https://issues.redhat.com/browse/LOG-962

cc @sichvoge


sichvoge:
That logstash configuration should work but not entirely since the time format would be wrong. As we discussed in the ticket, you need to configure `nanosecond_precision => true` so that you workaround the protocol version issue. See https://issues.redhat.com/browse/LOG-962?focusedCommentId=15414355&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-15414355

igor-karpukhin (approved):
/lgtm

PR #918, opened 2021-02-25
Changed beta1 to preview, regenerated bundle

vimalk78:
### Description
Changed beta1 to preview
Fixed OCP version in bundle to `v4.7`
Fixed channel to `tech-preview`
Regenerated bundle

/assign @jcantrill


Makefile
```diff
@@ -166,9 +166,10 @@ test-unit:
 test-cluster:
 	go test  -cover -race ./test/... -- -root=$(CURDIR)
 
-MANIFEST_VERSION?="5.1.beta1"
+OPENSHIFT_VERSIONS?="v4.7"
+MANIFEST_VERSION?="5.1.preview"
```
alanconway, line 170: Should be "preview.1" in case we find problems and need a second preview version.

alanconway (changes requested):

Makefile
vimalk78: The builds added to a version will have release numbers. We can attach new builds to the same version if we find problems.
e.g. `registry-proxy.engineering.redhat.com/rh-osbs/openshift-logging-cluster-logging-rhel8-operator:v5.1.0.preview-1` 
next build will be `registry-proxy.engineering.redhat.com/rh-osbs/openshift-logging-cluster-logging-rhel8-operator:v5.1.0.preview-2` etc

Adding `5.1.preview.2` would mean a totally different version.

vimalk78: Also i am not sure if `5.1.preview.1` is allowed as a CSV version

alanconway: There's no big difference between `v5.1.0.preview-2` and `v5.1.0.preview.2` - they are both _different versions_ from `.preview-1` or `.preview.1`. There's no distinction between "totally different" or "partly different" -  the meaning of version fields is outside the scope of  OLM and CSV, which follow the https://semver.org rules. 
Both `.preview-1` and `.preview.1` are valid semantic versions, the difference is sort order. A dot separated field with only digits is sorted numerically, any other field is sorted alphanumerically. 
So `preview.9` < `preview.10` which is what we want (if we ever get to 10 previews!) 
but `preview-10` < `preview-9` which will confuse  OLM upgrades.

alanconway (approved):
/lgtm

PR #932, opened 2021-03-03
WIP:: LOG-1157 Add initial json log format test

eranra:
### Description
Initial test for json fomat
Implementation of the JSON forwarding proposal
https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/forwarding-json-structured-logs.md

/cc @vimalk78 
/assign @jcantrill 

### Links
https://issues.redhat.com/browse/LOG-1157

PR #934, opened 2021-03-05
LOG-1157: Implement Json Forwarding: Parse Json Message

vimalk78:
### Description
This PR is to implement the Json Forwarding Proposal https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/forwarding-json-structured-logs.md

 - Changed the ClusterLogForwarder API
 - Generated fluent.conf for doing json parsing based on fluentd parse plugin https://docs.fluentd.org/filter/parser
 - Add functional tests
 - rebased on latest master
 - changed functional benchmarker to use separate config for fluentdForward which adds `epoc_in/epoc_out` which interferes in testing normal functional test cases

/cc @jcantrill @eranra 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-1157
- Enhancement proposal: https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/forwarding-json-structured-logs.md

### DIFF
--- a/manifests/5.1/logging.openshift.io_clusterlogforwarders_crd.yaml
+++ b/manifests/5.1/logging.openshift.io_clusterlogforwarders_crd.yaml
@@ -91,6 +91,13 @@ spec:
                   description: Output defines a destination for log messages.
                   properties:
                     elasticsearch:
+                      properties:
+                        indexKey:
+                          description: IndexKey specifies the metadata key to be used as name of elasticsearch index It takes precedence over IndexName
+                          type: string
+                        indexName:
+                          description: IndexName specifies the name of elasticsearch schema
+                          type: string
                       type: object
                     fluentdForward:
                       type: object
@@ -246,6 +253,11 @@ spec:
                       items:
                         type: string
                       type: array
+                    parse:
+                      description: "Parse enables parsing of log entries into structured logs \n Logs are parsed according to parse value, only `json` is supported as of now."
+                      enum:
+                      - json
+                      type: string
                   required:
                   - inputRefs
                   - outputRefs
--- a/pkg/apis/logging/v1/cluster_log_forwarder_types.go
+++ b/pkg/apis/logging/v1/cluster_log_forwarder_types.go
@@ -197,6 +197,14 @@ type PipelineSpec struct {
 	//
 	// +optional
 	Name string `json:"name,omitempty"`
+
+	// Parse enables parsing of log entries into structured logs
+	//
+	// Logs are parsed according to parse value, only `json` is supported as of now.
+	//
+	// +kubebuilder:validation:Enum:=json
+	// +optional
+	Parse string `json:"parse,omitempty"`
 }
 
 // ClusterLogForwarderList is a list of ClusterLogForwarders
--- a/pkg/apis/logging/v1/output_types.go
+++ b/pkg/apis/logging/v1/output_types.go
@@ -124,4 +124,15 @@ type Kafka struct {
 // Placeholders for configuration of other types
 
 type FluentdForward struct{}
-type Elasticsearch struct{}
+type Elasticsearch struct {
+	// IndexKey specifies the metadata key to be used as name of elasticsearch index
+	// It takes precedence over IndexName
+	//
+	// +optional
+	IndexKey string `json:"indexKey,omitempty"`
+
+	// IndexName specifies the name of elasticsearch schema
+	//
+	// +optional
+	IndexName string `json:"indexName,omitempty"`
+}
--- a/pkg/certificates/certificates.go
+++ b/pkg/certificates/certificates.go
@@ -7,6 +7,7 @@ import (
 	"sigs.k8s.io/yaml"
 
 	"github.com/ViaQ/logerr/log"
+	"github.com/openshift/cluster-logging-operator/pkg/utils"
 )
 
 func GenerateCertificates(namespace, scriptsDir, logStoreName, workDir string) (err error, updated bool) {
@@ -20,6 +21,8 @@ func RunCertificatesScript(namespace, logStoreName, workDir, script string) (err
 	cmd := exec.Command(script, workDir, namespace, logStoreName)
 	out, err := cmd.Output()
 	result := string(out)
+	// get error string from certificate generation script
+	err = utils.WrapError(err)
 	log.V(3).Info("Cert generation", "out", result, "err", err)
 	if result != "" {
 		updated = true
--- a/pkg/generators/forwarding/fluentd/fluent_conf_test.go
+++ b/pkg/generators/forwarding/fluentd/fluent_conf_test.go
@@ -1949,6 +1949,59 @@ var _ = Describe("Generating fluentd config", func() {
 		})
 		Expect(sources.List()).To(ContainElements("application", "audit"))
 	})
+
+	Describe("Json Parsing", func() {
+
+		fw := &logging.ClusterLogForwarderSpec{
+			Outputs: []logging.OutputSpec{
+				{
+					Type: logging.OutputTypeElasticsearch,
+					Name: "apps-es-1",
+					URL:  "https://es.svc.messaging.cluster.local:9654",
+				},
+			},
+			Inputs: []logging.InputSpec{
+				{
+					Name:        "myInput",
+					Application: &logging.Application{},
+				},
+			},
+			Pipelines: []logging.PipelineSpec{
+				{
+					Name:       "apps-pipeline",
+					InputRefs:  []string{"myInput"},
+					OutputRefs: []string{"apps-es-1"},
+					Parse:      "json",
+				},
+			},
+		}
+		It("should generate config for Json parsing", func() {
+			spec, err := generator.generatePipelineToOutputLabels(fw.Pipelines)
+			Expect(err).To(BeNil())
+			Expect(spec[0]).To(EqualTrimLines(`
+  <label @APPS_PIPELINE>
+    <filter **>
+      @type parser
+	  key_name message
+	  reserve_data yes
+	  hash_value_field structured
+	  <parse>
+	    @type json
+		json_parser oj
+	  </parse>
+	</filter>
+    <match **>
+      @type copy
+      <store>
+        @type relabel
+        @label @APPS_ES_1
+      </store>
+    </match>
+  </label>`))
+		})
+
+	})
+
 	var _ = DescribeTable("Verify generated fluentd.conf for valid forwarder specs",
 		func(yamlSpec, wantFluentdConf string) {
 			var spec logging.ClusterLogForwarderSpec
--- a/pkg/generators/forwarding/fluentd/templates.go
+++ b/pkg/generators/forwarding/fluentd/templates.go
@@ -546,7 +546,7 @@ const sourceToPipelineCopyTemplate = `{{- define "sourceToPipelineCopyTemplate" 
 <label {{sourceTypelabelName .Source}}>
   <match **>
     @type copy
-{{ range $index, $pipelineLabel := .PipelineNames }}
+{{- range $index, $pipelineLabel := .PipelineNames }}
     <store>
       @type relabel
       @label {{labelName $pipelineLabel}}
@@ -566,9 +566,21 @@ const pipelineToOutputCopyTemplate = `{{- define "pipelineToOutputCopyTemplate" 
     </record>
   </filter>
   {{ end -}}
+  {{ if (eq .Parse "json") -}}
+  <filter **>
+    @type parser
+    key_name message
+    reserve_data yes
+    hash_value_field structured
+    <parse>
+      @type json
+	  json_parser oj
+    </parse>
+  </filter>
+  {{ end -}}
   <match **>
     @type copy
-{{ range $index, $target := .Outputs }}
+{{- range $index, $target := .Outputs }}
     <store>
       @type relabel
       @label {{labelName $target}}
--- a/test/functional/normalization/json_parsing_test.go
+++ b/test/functional/normalization/json_parsing_test.go
@@ -1,0 +1,222 @@
+package normalization
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+
+	"github.com/ViaQ/logerr/log"
+	"github.com/google/go-cmp/cmp"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/test/functional"
+	"github.com/openshift/cluster-logging-operator/test/helpers/types"
+)
+
+const (
+	Json = `
+    {
+      "a": "Alpha",
+      "b": true,
+      "c": 12345,
+      "d": [
+      true,
+      [
+        false,
+        [
+        -123456789,
+        null
+        ],
+        3.9676,
+        [
+        "Something else.",
+        false
+        ],
+        null
+      ]
+      ],
+      "e": {
+      "zero": null,
+      "one": 1,
+      "two": 2,
+      "three": [
+        3
+      ],
+      "four": [
+        0,
+        1,
+        2,
+        3,
+        4
+      ]
+      },
+      "f": null,
+      "h": {
+        "a": {
+          "b": {
+            "c": {
+              "d": {
+                "e": {
+                  "f": {
+                    "g": null
+                     }
+                   }
+                 }
+               }
+             }
+           }
+         },
+      "i": [
+            [
+             [
+              [
+               [
+                [
+                 [
+                  null
+                 ]
+                ]
+               ]
+              ]
+             ]
+            ]
+          ]
+    }
+    `
+)
+
+var _ = Describe("[LogForwarding] Json log parsing", func() {
+	var (
+		framework *functional.FluentdFunctionalFramework
+		clfb      *functional.ClusterLogForwarderBuilder
+		expected  map[string]interface{}
+		empty     map[string]interface{}
+	)
+	BeforeEach(func() {
+		_ = json.Unmarshal([]byte(Json), &expected)
+		empty = map[string]interface{}{}
+		framework = functional.NewFluentdFunctionalFramework()
+		clfb = functional.NewClusterLogForwarderBuilder(framework.Forwarder).
+			FromInput(logging.InputNameApplication).
+			ToFluentForwardOutput()
+	})
+	AfterEach(func() {
+		framework.Cleanup()
+	})
+
+	It("should parse json message into structured", func() {
+		clfb.Forwarder.Spec.Pipelines[0].Parse = "json"
+		Expect(framework.Deploy()).To(BeNil())
+
+		// Log message data
+		jsonMsg := strings.ReplaceAll(Json, "\n", "")
+		expectedMessage := jsonMsg
+		jsonMsg = strings.ReplaceAll(jsonMsg, "\"", "\\\"")
+		timestamp := "2020-11-04T18:13:59.061892+00:00"
+
+		// Write log line as input to fluentd
+		applicationLogLine := fmt.Sprintf("%s stdout F %s", timestamp, jsonMsg)
+		Expect(framework.WriteMessagesToApplicationLog(applicationLogLine, 10)).To(BeNil())
+
+		// Read line from Log Forward output
+		raw, err := framework.ReadApplicationLogsFrom(logging.OutputTypeFluentdForward)
+		Expect(err).To(BeNil(), "Expected no errors reading the logs")
+
+		// Parse log line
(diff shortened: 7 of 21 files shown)


manifests/5.1/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -173,6 +173,18 @@ spec:
...
                         type: string
                       type: array
+                    parse:
+                      default: json
+                      description: "Parse enables parsing of the log entries into structured logs \n Logs are parsed according to parse value, only `json` is supported as of now."
+                      enum:
+                      - json
+                      type: string
+                    schemaKey:
+                      description: SchemaKey specifies the metadata key to be used as name of log schema It takes precedence over SchemaName
```
jcantrill, line 183: @alanconway i missed this detail in the proposal, but why should `schemaKey` take precendence?

pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -1949,6 +1949,82 @@ var _ = Describe("Generating fluentd config", func() {
...
+    </match>
+  </label>`))
+		})
+
+		It("should generate schema from schemaName", func() {
+			fw.Pipelines[0].SchemaName = "myapp-logs"
+			spec, err := generator.generatePipelineToOutputLabels(fw.Pipelines)
+			Expect(err).To(BeNil())
+			Expect(spec[0]).To(EqualTrimLines(`
+  <label @APPS_PIPELINE>
```
jcantrill, line 2007: Shouldnt there be a `record_modifier` here to add in the schema name?

vimalk78: The PR is WIP, as of now it is only for API changes and generating parser code.
I will delete this test

manifests/5.1/logging.openshift.io_clusterlogforwarders_crd.yaml
vimalk78: https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/forwarding-json-structured-logs.md#clusterlogforwarder-configuration

manifests/5.1/logging.openshift.io_clusterlogforwarders_crd.yaml
```diff
@@ -173,6 +173,18 @@ spec:
                       items:
                         type: string
                       type: array
+                    parse:
+                      default: json
+                      description: "Parse enables parsing of the log entries into structured logs \n Logs are parsed according to parse value, only `json` is supported as of now."
```
eranra, line 178: s/"of the log"/"of log"

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -566,9 +566,20 @@ const pipelineToOutputCopyTemplate = `{{- define "pipelineToOutputCopyTemplate"
...
+  {{ if (eq .Parse "json") -}}
+  <filter **>
+    @type parser
+    key_name message
+    reserve_data yes
+    hash_value_field structured
+    <parse>
+      @type json
+    </parse>
+  <filter>
```
eranra, line 578: @vimalk78 We talked on using `record_modifier` to allow us more flexability

```
 <filter **>
    @type record_modifier
    <record>
      structured      ${require 'json'; JSON.parse(record['message'])}
    </record>
  </filter>
```

vimalk78: I used `parse` plugin as suggested by @jcantrill , for two reasons
- It is a generic plugin which can support multiple parsers, and we also want such extensibility in our API
- Originally we either wanted a `message` OR `structured` field after parsing, which needed logic to remove the other. Now requirement is relaxed, so the scope of potential plugins is also increased.

eranra: @vimalk78 agree ... when we did the inital testing we looked at those two options ... one thing I wonder is performance diffrences ... JSON is known to te expensive to parse ... so if we know that one of the options is much more efficent (again, I don't know) ... maybe it make sense to use the less resource consuming option.

I get the point on future flexability ... agree.

If I remeber correcly there is another small diffrence ... with `parser` we get an empty `structured` field if there is nothing to parse while with `record_modifier` if there is no parsing we have no struct at all ... both are ok, but I think that for customers the second option might be prefered

manifests/5.1/logging.openshift.io_clusterlogforwarders_crd.yaml
eranra: @vimalk78 I am trying to understand the logic Schema starting with understanding `SchemaKey` ... let's say that I have 2 apps names `app1`, `app2`  I want `app1` json parsing to go to index `app1` and I want `app2` json parsing to go to index `app2`

I need to define two `schemaKey' values app1 and app2 right ???

 schemaKey: kubernetes.labels.app1
 schemaKey: kubernetes.labels.app2

pkg/generators/forwarding/fluentd/templates.go
vimalk78: `record_modifier` always adds a `structured` field irrespective of content in `message` field. 

```
structured      ${require 'json'; JSON.parse(record['message'])}
```
The above code will add a field with whatever is the result of `JSON.parse`. It could even be an error message, exception etc, it need not be an empty struct. We will need to handle error conditions more carefully.

I havent checked the performance.`parse` plugin may be slower than `JSON.parse` but has more features.

manifests/5.1/logging.openshift.io_clusterlogforwarders_crd.yaml
eranra: @vimalk78 so ... if I enable "json parsing" and there is a valid json for a log line ... are we going to send the message also to the regular index and not just to the "by schema" index ?

eranra: s/"schema It takes"/"schema. It takes"

pkg/generators/forwarding/fluentd/templates.go
eranra: @vimalk78 ok, parser it will be :-)  ... BTW: I also saw another remark from @jcantrill  in the same spirit https://github.com/openshift/cluster-logging-operator/pull/932/files#r587764712

manifests/5.1/logging.openshift.io_clusterlogforwarders_crd.yaml
sichvoge: No, you are sending the entire log record (which includes the `message` field) into a particular index. By using our Elasticsearch instance, this means into the "schema index".

test/functional/logforwarding/json-parsing.go
```diff
@@ -0,0 +1,106 @@
...
+		jsonMsg = strings.ReplaceAll(jsonMsg, "\n", "")
+		expectedMessage := jsonMsg
+		jsonMsg = strings.ReplaceAll(jsonMsg, "\"", "\\\"")
+		timestamp := "2020-11-04T18:13:59.061892+00:00"
+
+		// Write log line as input to fluentd
+		applicationLogLine := fmt.Sprintf("%s stdout F %s", timestamp, jsonMsg)
+		Expect(framework.WriteMessagesToApplicationLog(applicationLogLine, 10)).To(BeNil())
+		// Read line from Log Forward output
+		raw, err := framework.ReadApplicationLogsFrom("fluentforward")
```
jcantrill, line 76: use logging.outputtypeforward

test/functional/logforwarding/json-parsing.go
```diff
@@ -0,0 +1,106 @@
...
+		Expect(err).To(BeNil(), "Expected no errors reading the logs")
+		// Parse log line
+		var logs []types.ApplicationLog
+		err = types.StrictlyParseLogs(raw, &logs)
+		Expect(logs[0].Structured).To(Equal(expected), "received structured not matching")
+		Expect(logs[0].Message).To(Equal(expectedMessage), "received message not matching")
+		Expect(err).To(BeNil(), "Expected no errors parsing the logs")
```
jcantrill, line 83: shouldnt we check this first?

test/functional/logforwarding/json-parsing.go
```diff
@@ -0,0 +1,106 @@
...
+		// Write log line as input to fluentd
+		applicationLogLine := fmt.Sprintf("%s stdout F %s", timestamp, jsonMsg)
+		Expect(framework.WriteMessagesToApplicationLog(applicationLogLine, 10)).To(BeNil())
+		// Read line from Log Forward output
+		raw, err := framework.ReadApplicationLogsFrom("fluentforward")
+		Expect(err).To(BeNil(), "Expected no errors reading the logs")
+		// Parse log line
+		var logs []types.ApplicationLog
+		err = types.StrictlyParseLogs(raw, &logs)
+		Expect(logs[0].Structured).To(Equal(expected), "received structured not matching")
```
jcantrill, line 81: maybe update the message failure to be more explicit  telling us the fact the parsed message was

test/functional/logforwarding/json-parsing.go
```diff
@@ -0,0 +1,106 @@
...
+		Expect(logs[0].Message).To(Equal(expectedMessage), "received message not matching")
```
jcantrill, line 82: @alanconway please confirm.  i think the expectation is that "message" is empty:

```
For each log entry; if parse: json is set and the entry is valid JSON, the output record will have a structured field. Otherwise the output record will have a message field.
```

vimalk78: As per conversation last week, this requirement is relaxed. Now `message` will remain as is. `structured` will be populated after parsing.

test/functional/logforwarding/json-parsing.go
vimalk78: changed

test/functional/logforwarding/json-parsing.go
vimalk78: will change after master-post-release merges to master

test/functional/logforwarding/json-parsing.go
```diff
@@ -0,0 +1,106 @@
+package logforwarding
```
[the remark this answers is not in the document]
vimalk78: moved file to `normalization`

test/functional/logforwarding/json-parsing.go
vimalk78: changed

test/functional/logforwarding/json-parsing.go
vimalk78: done

pkg/apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -197,6 +197,26 @@ type PipelineSpec struct {
 	//
 	// +optional
 	Name string `json:"name,omitempty"`
+
+	// Parse enables parsing of log entries into structured logs
+	//
+	// Logs are parsed according to parse value, only `json` is supported as of now.
+	//
+	// +kubebuilder:validation:Enum:=json
+	// +kubebuilder:default:=json
```
alanconway, line 206: No default, it is optional but if present the only allowed value is "json"
```suggestion
```

test/functional/normalization/json_parsing_test.go
```diff
@@ -0,0 +1,177 @@
...
+                 ]
+                ]
+               ]
+              ]
+             ]
+            ]
+          ]
+    }
+    `
+)
```
alanconway, line 85: Wooohooo, now *thats* a test case! :+1:

test/functional/normalization/json_parsing_test.go
```diff
@@ -0,0 +1,177 @@
...
+		Expect(err).To(BeNil(), "Expected no errors reading the logs")
+
+		// Parse log line
+		var logs []types.ApplicationLog
+		err = types.StrictlyParseLogs(raw, &logs)
+		Expect(err).To(BeNil(), "Expected no errors parsing the logs")
+		Expect(logs[0].Structured).To(Equal(map[string]interface{}{}), "received non-empty json parsed structure")
+		Expect(logs[0].Message).To(Equal(message), "received message not matching")
+	})
+	XIt("should not parse invalid json message into structured", func() {
```
alanconway, line 152: Why disabled?

test/functional/normalization/json_parsing_test.go
```diff
@@ -0,0 +1,177 @@
...
+		Expect(err).To(BeNil(), "Expected no errors reading the logs")
+
+		// Parse log line
+		var logs []types.ApplicationLog
+		err = types.StrictlyParseLogs(raw, &logs)
+		Expect(err).To(BeNil(), "Expected no errors parsing the logs")
+		Expect(logs[0].Structured).To(Equal(map[string]interface{}{}), "received structured not matching")
+		Expect(logs[0].Message).To(Equal(message), "received message not matching")
+	})
+
```
alanconway, line 176: Maybe add one case to make sure that we don't parse valid json if Parse is not set, i.e. we default to original behavior.

test/helpers/types/types.go
```diff
@@ -343,7 +337,25 @@ func StrictlyParseLogs(in string, logs interface{}) error {
 
 type Logs []AllLog
 
-func (t *AllLog) ElapsedEpoc() float64 {
+func ToJsonLogs(logs []string) string {
```
alanconway, line 340: Replace with existing `test.JSONLine()`, does the same thing.

alanconway (commented):
We talked about  moving the schema stuff to the Elasticsearch output, my bad - I haven't updated the proposal. I will do that today, should be only a minor refactoring to adapt the PR. Otherwise looks good, a few minor suggestions.

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -566,9 +566,20 @@ const pipelineToOutputCopyTemplate = `{{- define "pipelineToOutputCopyTemplate"
     </record>
   </filter>
   {{ end -}}
+  {{ if (eq .Parse "json") -}}
+  <filter **>
+    @type parser
+    key_name message
+    reserve_data yes
```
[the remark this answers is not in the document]
vimalk78: me and @eranra  discussed with @alanconway, and conclusion was

- parse the content from message field into json and save in a new field named  structured
- if the content is non-json can keep the structured field as empty

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -566,9 +566,20 @@ const pipelineToOutputCopyTemplate = `{{- define "pipelineToOutputCopyTemplate"
     </record>
   </filter>
   {{ end -}}
+  {{ if (eq .Parse "json") -}}
+  <filter **>
```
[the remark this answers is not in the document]
vimalk78: Agreed. we dont have an `@error` block in our conf, so couldnt test it.

pkg/apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -197,6 +197,26 @@ type PipelineSpec struct {
...
+	// +optional
+	Parse string `json:"parse,omitempty"`
+
+	// SchemaKey specifies the metadata key to be used as name of log schema.
```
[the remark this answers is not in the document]
vimalk78: i will delete these.

pkg/certificates/certificates.go
```diff
@@ -20,6 +20,10 @@ func RunCertificatesScript(namespace, logStoreName, workDir, script string) (err
 	cmd := exec.Command(script, workDir, namespace, logStoreName)
 	out, err := cmd.Output()
 	result := string(out)
+	// get error string from certificate generation script
+	if ee, ok := err.(*exec.ExitError); ok {
+		err = fmt.Errorf("%s", string(ee.Stderr))
+	}
```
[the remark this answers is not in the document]
vimalk78: thanks. i didnt know about this function.
Without this wrapping, we get a dump of `[]byte` and gets encoded into base64 by the ViaQ/logger.

pkg/apis/logging/v1/cluster_log_forwarder_types.go
vimalk78: changed

test/functional/normalization/json_parsing_test.go
vimalk78: copied it from https://github.com/anilmaurya/fast_jsonparser/blob/master/benchmark/json_benchmark.rb#L3

test/functional/normalization/json_parsing_test.go
vimalk78: This is disabled because the parser returns `nil` instead of empty `structured` field

test/functional/normalization/json_parsing_test.go
vimalk78: Added

test/functional/normalization/json_parsing_test.go
vimalk78: we need to fix the behavior for invalid json parsing. the choice of parser gives different results.
There are 3 json parsers available https://docs.fluentd.org/parser/json#json_parser

- oj
- yajl
- json

With each parser we get different output of we parse this invalid json `{"key":"v}`

- oj parses into `"structured":"key"`
- yajl parses into `"structured":"key"`
- json parses into `"structured":{}`

each has different performance characteristics perhaps https://github.com/anilmaurya/fast_jsonparser/blob/master/benchmark/json_benchmark.rb#L38-L41

Cc: @alanconway @jcantrill @eranra

test/helpers/types/types.go
vimalk78: this didnt give the same result. so didnt change

test/functional/normalization/json_parsing_test.go
jcantrill: oj is the default right?  we should see if we can configure it to be strict https://github.com/ohler55/oj/blob/develop/pages/Modes.md  and then follow-up to maybe choose a faster parser

jcantrill (commented):
/approve

test/helpers/types/types.go
alanconway: Of course - line breaks. Sorry.

pkg/generators/forwarding/fluentd/templates.go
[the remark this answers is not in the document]
alanconway: There is an RFE for that, I'll create an epic with more details and a note to revisit the JSON code when it is implemented.

test/functional/normalization/json_parsing_test.go
alanconway: Looks great but we do need to handle invalid JSON. I guess we need to dig into the oj modes.  The json parser produces an acceptable result `"structured":{}` but it's worth some effort to get the faster parser as this is critical path for JSON customers. We don't have much room to further annoy them :)

vimalk78: The Oj options are set here
https://github.com/fluent/fluentd/blob/v1.7.4/lib/fluent/env.rb#L24
```
  DEFAULT_OJ_OPTIONS = {bigdecimal_load: :float, mode: :compat, use_to_json: true}
```
other options in the `env.rb` can be overridden via environment variables except this one

vimalk78: even in master branch it is same

vimalk78: We also get logs
```
2021-03-16 08:49:45 +0000 [info]: Oj is not installed, and failing back to Yajl for json parser
```
Thats why in the avove comment we see `Oj` and `Yajl` giving same result

vimalk78: The fluentd documentation gives a hint
https://docs.fluentd.org/quickstart/faq#fluentd-warns-oj-is-not-installed-and-falling-back-to-yajl-for-json-parser

pkg/apis/logging/v1/output_types.go
```diff
@@ -124,4 +124,15 @@ type Kafka struct {
 // Placeholders for configuration of other types
 
 type FluentdForward struct{}
-type Elasticsearch struct{}
+type Elasticsearch struct {
+	// IndexKey specifies the metadata key to be used as name of elasticsearch index
+	// It takes precedence over IndexName
+	//
+	// +optional
+	IndexKey string `json:"indexKey,omitempty"`
```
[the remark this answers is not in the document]
vimalk78: 👍🏽

test/functional/normalization/json_parsing_test.go
vimalk78: I have added the `Oj` json parser to fluentd image. With `oj` parser we get empty `structured` field for invalid json.

vimalk78:
I have added the `Oj` json parser to fluentd image. With oj parser we get empty `structured` field for invalid json.

PR #938, opened 2021-03-08
Update container registry to registry.ci.openshift.org

jcantrill:
### Description
This PR updates the container registry per dev email regarding the decommissioning of the other registry


jcantrill:
added valid bz label as these are changes for development

PR #940, opened 2021-03-09
LOG-1187: Set component NO_PROXY,HTTP_PROXY from environment variables

jcantrill:
### Description
This PR address the issue by:

relying on OLM to set proxy vars on the operator which can then be propagated to the operands
Removes watching the proxy config object directly

### Links
backport of: https://github.com/openshift/cluster-logging-operator/pull/890
https://issues.redhat.com/browse/LOG-1187


vimalk78 (approved):
/lgtm

PR #942, opened 2021-03-12
BUG 1918920: keep the ES cluster status if the cluster is to be redep…

huikang:
…loyed

### Description
When a cluster is set to be redeployed due to a secret change with the changes to the ES setting (e.g., CPU requirement in the logstore), current implementation will recreate the es type which overwrites the certRedeploy status. This PR fixes the issue by keeping the status if the cluster is set to be redeployed.

/cc @ewolinetz 
/assign @jcantrill 

/cherry-pick release-5.0

### Links
- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1918920
- JIRA: https://issues.redhat.com/browse/LOG-1205

### DIFF
--- a/pkg/k8shandler/logstore.go
+++ b/pkg/k8shandler/logstore.go
@@ -257,7 +257,7 @@ func (cr *ClusterLoggingRequest) newElasticsearchCR(elasticsearchName string, ex
 	var es *elasticsearch.Elasticsearch
 	if existing != nil && existing.Name != "" {
 		// using existing es if a valid one exists
-		es = existing
+		es = existing.DeepCopy()
 	} else {
 		es = cr.emptyElasticsearchCR(elasticsearchName)
 	}
--- a/pkg/k8shandler/logstore.go
+++ b/pkg/k8shandler/logstore.go
@@ -225,6 +225,7 @@ func (cr *ClusterLoggingRequest) newElasticsearchCR(elasticsearchName string, ex
 	// build Nodes
 	esNodes = append(esNodes, esNode)
 
+	keepEsStatus := false
 	// if we had more than 1 es node before, we also want to enter this condition
 	if logStoreSpec.NodeCount > maximumElasticsearchMasterCount || len(existing.Spec.Nodes) > 1 {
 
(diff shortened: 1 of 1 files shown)


pkg/k8shandler/logstore.go
```diff
@@ -246,6 +247,10 @@ func (cr *ClusterLoggingRequest) newElasticsearchCR(elasticsearchName string, ex
 
 		esNodes = append(esNodes, dataNode)
 
+		// if the cluster is set to be redeployed, keep the status
+		if len(existing.Status.Nodes) > 0 && existing.Status.Nodes[0].UpgradeStatus.ScheduledForCertRedeploy == v1.ConditionTrue {
```
jcantrill, line 251: why keep a specific status and not the entirety?

jcantrill:
@huikang please:
* copy this issue into jira for verification by QE against 5.1
* clone the 5.1 issue for verification against 5.0
* re-target bz to 4.6.z and ref the JIRA issues

pkg/k8shandler/logstore.go
ewolinetz: is there a way we can ensure that CLO doesn't touch the status field for the es CR at all?
ideally it would only update the spec for an update

huikang: If so, we probably need only to update the Spec of the existing CR, instead of creating a new one.

huikang:
@jcantrill , done. Created two jira issues
- 5.1: https://issues.redhat.com/browse/LOG-1205
- 5.0 https://issues.redhat.com/browse/LOG-1206
- bz is targeted to 4.6.z

pkg/k8shandler/logstore.go
jcantrill: why do we need to conditionally keep the status instead of just always keeping it?

pkg/k8shandler/logstore.go
```diff
@@ -255,7 +254,13 @@ func (cr *ClusterLoggingRequest) newElasticsearchCR(elasticsearchName string, ex
...
 	indexManagementSpec := indexmanagement.NewSpec(logStoreSpec.RetentionPolicy)
 
-	es := cr.emptyElasticsearchCR(elasticsearchName)
+	var es *elasticsearch.Elasticsearch
+	if existing != nil && existing.Name != "" {
+		// using existing es if a valid one exists
+		es = existing.DeepCopy()
+	} else {
+		es = cr.emptyElasticsearchCR(elasticsearchName)
+	}
```
ewolinetz, line 263: i'm not following why this is the fix. this makes it seem like something in our create/update logic is broken and this is a bandaid. 

Maybe I'm not understanding the root issue though?

jcantrill: I'm not following either.  Typically the pattern is:
* new proto
* Create(proto)
* if fail and already exists:
* get instance
* update instance to equal proto (or the parts as needed

Red-GV:
@jcantrill I think this PR can be closed in favor of https://github.com/openshift/elasticsearch-operator/pull/695 to fix this issue.

PR #944, opened 2021-03-15
Bug 1932443: Fix missing metrics port in manifests

vparfonov:

### Description
Expose ports for `http` and `cr` metrics in the ClusterServiceVersion deployment manifest for the cluster-logging-operator.
Fix generate bundle command
Generate bundle.

/cc @periklis 
/assign @jcantrill 

/cherry-pick 

### Links

- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1932443

### DIFF
--- a/manifests/4.6/cluster-logging.v4.6.0.clusterserviceversion.yaml
+++ b/manifests/4.6/cluster-logging.v4.6.0.clusterserviceversion.yaml
@@ -315,6 +315,11 @@ spec:
                 imagePullPolicy: IfNotPresent
                 command:
                 - cluster-logging-operator
+                ports:
+                  - containerPort: 8383
+                    name: http-metrics
+                  - containerPort: 8686
+                    name: cr-metrics
                 env:
                   - name: WATCH_NAMESPACE
                     valueFrom:
--- a/bundle/manifests/cluster-logging.v4.6.0.clusterserviceversion.yaml
+++ b/bundle/manifests/cluster-logging.v4.6.0.clusterserviceversion.yaml
@@ -159,9 +159,9 @@ spec:
   - type: SingleNamespace
     supported: true
   - type: MultiNamespace
-    supported: true
+    supported: false
   - type: AllNamespaces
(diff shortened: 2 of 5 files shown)


periklis (changes requested):
Needs a run `make bundle` and include the bundle copy of the CSV.

vparfonov:
> Needs a run `make bundle` and include the bundle copy of the CSV.
@periklis `make bundle` target not found for CLO

vparfonov:
> > Needs a run `make bundle` and include the bundle copy of the CSV.
> > @periklis `make bundle` target not found for CLO

Looks like need to run `make generate-bundle`

manifests/4.6/cluster-logging.v4.6.0.clusterserviceversion.yaml
```diff
@@ -315,6 +315,11 @@ spec:
                 imagePullPolicy: IfNotPresent
                 command:
                 - cluster-logging-operator
+                ports:
+                  - containerPort: 8383
+                    name: http-metrics
+                  - containerPort: 8686
+                    name: cr-metrics
```
periklis, line 322: Have you tested this on a test cluster? Do the SVCs use the ports by name or by port-number directly? If by name, do the names match between here and SVCs?

periklis (commented):
In general looks good to me. I suggest someone from log core takes also a look as the extra changes in the bundle directory indicate that the bundle was not sync'ed for a long time. I am not sure if the old state was delivered to customer clusters already. If so we would break compatibility with changes like CLF from cluster-scoped to namespace-scoped.

manifests/4.6/cluster-logging.v4.6.0.clusterserviceversion.yaml
vparfonov: Yes, i tested it. All deployed normally. Service use port-number directly.

periklis (approved):
/lgtm

PR #946, opened 2021-03-16
fix internal generator tool and always test

jcantrill:
### Description
This PR:
* fixes the internal clf generator to properly deserialize yaml and respect camelcase
* adds Makefile target to compile and test when doing unit tests to ensure it at least runs without error

cc @vimalk78 @alanconway @igor-karpukhin

### DIFF
--- a/internal/cmd/forwarder-generator/main.go
+++ b/internal/cmd/forwarder-generator/main.go
@@ -40,18 +40,17 @@ func main() {
 		os.Exit(1)
 	}
 	log.V(1).Info("Args: %v", os.Args)
-
 	var reader func() ([]byte, error)
 	switch *yamlFile {
 	case "-":
-		reader = func() ([]byte, error) { return ioutil.ReadFile(*yamlFile) }
-	case "":
-		reader = func() ([]byte, error) { return []byte{}, nil }
-	default:
 		reader = func() ([]byte, error) {
 			stdin := bufio.NewReader(os.Stdin)
 			return ioutil.ReadAll(stdin)
 		}
+	case "":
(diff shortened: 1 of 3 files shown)


internal/cmd/forwarder-generator/main.go
```diff
@@ -40,18 +40,17 @@ func main() {
...
 	case "-":
-		reader = func() ([]byte, error) { return ioutil.ReadFile(*yamlFile) }
-	case "":
-		reader = func() ([]byte, error) { return []byte{}, nil }
-	default:
 		reader = func() ([]byte, error) {
 			stdin := bufio.NewReader(os.Stdin)
 			return ioutil.ReadAll(stdin)
 		}
+	case "":
```
vimalk78, line 50: i had the same changes in my syslog-functional-tests PR. but i never get to finish it

internal/pkg/generator/forwarder/generator.go
```diff
@@ -3,7 +3,7 @@ package forwarder
 import (
 	"fmt"
 
-	yaml "gopkg.in/yaml.v2"
+	yaml "sigs.k8s.io/yaml"
```
vimalk78, line 6: any test case to check the effect of this change?

jcantrill: This will test it https://github.com/openshift/cluster-logging-operator/pull/946/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R161

igor-karpukhin (approved):
/lgtm

PR #952, opened 2021-03-19
Cleanup of the eventrouter e2e test

syedriko:
### Description
- Fixed the eventrouter e2e test failure due to bad image name when running in dev env
- Removed noise from the output of the eventrouter e2e test. The events for debugging can be found in the artifacts dir.
- Simplified Logging version detection

/cc @jcantrill 
/assign @alanconway


jcantrill (commented):
/approve

PR #953, opened 2021-03-19
Cleaned up pattern matching for selective running of logforwarding e2e tests

syedriko:
### Description
Cleaned up pattern matching for the logforwarding e2e tests.
    The CLF_TEST_INCLUDES env variable takes a PCRE that directories in test/e2e/logforwarding are matched against. Examples:
    - CLF_TEST_INCLUDES=syslog matches syslog and sysloglegacy
    - CLF_TEST_INCLUDES=syslog|fluent matches syslog, sysloglegacy, fluent and fluentlegacy
    - CLF_TEST_INCLUDES=^syslog$ matches just syslog

/cc @jcantrill 
/assign @alanconway

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -177,7 +177,7 @@ generate-bundle: regenerate $(OPM)
 
 # NOTE: This is the CI e2e entry point.
 test-e2e-olm: $(JUNITREPORT)
-	INCLUDES=$(E2E_TEST_INCLUDES) CLF_INCLUDES=$(CLF_TEST_INCLUDES) hack/test-e2e-olm.sh
+	INCLUDES="$(E2E_TEST_INCLUDES)" CLF_INCLUDES="$(CLF_TEST_INCLUDES)" hack/test-e2e-olm.sh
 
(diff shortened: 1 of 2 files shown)


jcantrill:
How are these more unique or better?

syedriko:
> How are these more unique or better?

Now there's no way to run just the syslog test or the fluent test. You say `export CLF_TEST_INCLUDES=syslog` and it runs both syslog and sysloglegacy. Same thing for fluent, both fluent and fluentlegacy run.

alanconway:
Can we make whatever is doing the substring match do a full-string match instead?  E.g. if it's `go test` that takes a regexp so you could change `go test -run 'blah'` to `go test -run '^blah$'`

Changing all the directory names makes merging test changes tricky.

syedriko:
Let me see what I can do...

alanconway:
/lgtm

I'd like to get all these tests runnable directlly by `go test` and get rid of all the scripting stuff, but that's another story.

PR #954, opened 2021-03-19
Adding container chart LOG-1165-Dashboard

ajaygupta978:
### Description
This PR address creating total bytes logged by go-watcher and total bytes collected by fluentd. 

/cc 
/assign 
Screenshot 2021-03-19 at 6 40 25 PM

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-1165
- Enhancement proposal:

### DIFF
--- a/files/dashboards/openshift-logging-dashboard.json
+++ b/files/dashboards/openshift-logging-dashboard.json
@@ -1708,6 +1708,92 @@
         "align": false,
         "alignLevel": null
       }
+    },
+    {
+      "aliasColors": {},
+      "bars": false,
+      "dashLength": 10,
+      "dashes": false,
+      "datasource": null,
+      "fill": 1,
+      "fillGradient": 0,
+      "gridPos": {
+        "h": 7,
+        "w": 8,
+        "x": 16,
+        "y": 49
+      },
+      "hiddenSeries": false,
+      "id": 41,
+      "legend": {
+        "avg": false,
+        "current": false,
+        "max": false,
+        "min": false,
+        "show": true,
+        "total": false,
+        "values": false
+      },
+      "lines": true,
+      "linewidth": 1,
+      "nullPointMode": "null",
+      "percentage": false,
+      "pluginVersion": "7.1.1",
+      "pointradius": 2,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "topk(10, round(rate(fluentd_input_status_total_bytes_logged[5m])))",
+          "interval": "",
+          "legendFormat": "{{path}}",
+          "refId": "A"
+        }
+      ],
+      "thresholds": [],
+      "timeFrom": null,
+      "timeRegions": [],
+      "timeShift": null,
+      "title": "Logs emitted (top 10 containers) - Bytes/Second",
+      "tooltip": {
+        "shared": true,
+        "sort": 0,
+        "value_type": "individual"
+      },
+      "type": "graph",
+      "xaxis": {
+        "buckets": null,
+        "mode": "time",
+        "name": null,
+        "show": true,
+        "values": []
+      },
+      "yaxes": [
+        {
+          "format": "short",
+          "label": null,
+          "logBase": 1,
+          "max": null,
+          "min": null,
+          "show": true
+        },
+        {
+          "format": "short",
+          "label": null,
+          "logBase": 1,
+          "max": null,
+          "min": null,
+          "show": true
+        }
+      ],
+      "yaxis": {
+        "align": false,
+        "alignLevel": null
+      }
     }
   ],
   "refresh": "30s",


jcantrill:
@ajaygupta978 how does the dashboard actually get deployed?  should there be some additional change to CLO code to consume it?

files/dashboards/openshift-logging-dashboard.json
```diff
@@ -1708,18 +1798,233 @@
...
+      "pointradius": 2,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "topk(10, fluentd_input_status_total_bytes_logged) by (path)",
```
blockloop, line 1861: This is going to show more than 5 items because it maps the top 5 items at each
interval instead of across the whole graph. To accomplish what you're trying to
do you need to use the following query:

```
query_result(topk by(path)(5, avg_over_time(fluentd_input_status_total_bytes_logged[${__range_s}s])))
```

blockloop:
> @ajaygupta978 how does the dashboard actually get deployed? should there be some additional change to CLO code to consume it?

@jcantrill That JSON file is created as a configmap that's read by the
monitoring operator automatically. See [pkg/k8shandler/dashboards.go](pkg/k8shandler/dashboards.go)

files/dashboards/openshift-logging-dashboard.json
ajaygupta978: @blockloop 
1)  `query_result(topk by(path)(10, avg_over_time(fluentd_input_status_total_bytes_logged[${__range_s}s])`  throwing error in while `(topk by(path)(10, avg_over_time(fluentd_input_status_total_bytes_logged[${__range_s}s])` works. 

2) I see several values in the Graph as shown in attached screenshot.  Is this query shows top 10 containers in the given range(time-interval) ? So in the timeline one will see 10 containers every range ? is this expected ? 
charts_range_topk

ajaygupta978: @jcantrill  Currently I am facing above issue so seeking help from Brett.

ajaygupta978:
@blockloop I think instead of `avg_over_time` we should use `increase`.  For example  if we are at current time-instant `T` then avg_over_time(fluentd_input_status_total_bytes_logged[5m])  will consider all data points from [T, T-5m]. If scrape interval is 30s then we will get 10 such data points and avg will be calculated.  Every 30s we will be calculating  `avg_over_time` using past 5m data points. Based on plot interval we will have many such `avg_over_time` data points which will be plotted in graph.  
Since `fluentd_input_status_total_bytes_logged` metric is counter so avg_over_time will always be increasing and will not give results we want. 
increase() function calculates difference between two time instance so in our case it will show  multiple data points where each data point shows how many bytes collected in last [T, T-5m].
In the attached screen-shot Avg_over_time showing increasing graph because metric is counter. In increase screenshot values are constant because simulated container pushing logs with constant rate(10 logs per second). 

**Avg_over_time Screenshot**
Avg_over_time

**Increase screenshot**
increase


Also I think  there is no mechanism to provide time-interval  by user so we have to keep it to some pre-defined constant say 1m,5m or some other value.

ajaygupta978:
@blockloop I have update my PR. We had discussed to show table view but I think till the time we don't have support of $range timeinterval chart will be static and show only last 15 mins. Instead I think we should show graph which can show values at different interval. It might not show more than top 10 in the time-range but still its better than static table view. 
I have attached latest screenshot also with changes. 

top10

blockloop:
I think the table view is probably the best view, however, the lack of a time 
window selection makes the table view show only the most recent time
range. With this limitation, I would lean toward having two table views - 
one for the last 15m and one for the last 1h. The line graph doesn't make
much sense to me, but I'll leave that up to you and @jcantrill

ajaygupta978:
@jcantrill could you please provide your suggestion ?

files/dashboards/openshift-logging-dashboard.json
```diff
@@ -7,8 +7,8 @@
   "editable": true,
   "gnetId": null,
   "graphTooltip": 0,
-  "id": null,
-  "iteration": 1594321382217,
+  "id": 1,
+  "iteration": 1616139020699,
```
eranra, line 11: not sure why id and iteration was changed .... you might want to get those back to original if not needed

files/dashboards/openshift-logging-dashboard.json
```diff
@@ -35,6 +35,12 @@
         "#d44a3a"
       ],
       "datasource": "$datasource",
+      "fieldConfig": {
+        "defaults": {
+          "custom": {}
+        },
+        "overrides": []
+      },
```
eranra, line 43: You can remove those empty lines if no need to change them ... not sure why they are added

files/dashboards/openshift-logging-dashboard.json
```diff
@@ -1678,7 +1769,6 @@
         "value_type": "individual"
       },
       "type": "graph",
-      "span": 2,
```
eranra, line 1681: All changesup until here ... seems not really needed please get them back to original

files/dashboards/openshift-logging-dashboard.json
```diff
@@ -1708,18 +1798,233 @@
...
+          "interval": "",
+          "legendFormat": "{{path}}",
+          "refId": "A"
+        }
+      ],
+      "thresholds": [],
+      "timeFrom": null,
+      "timeRegions": [],
+      "timeShift": null,
+      "title": "Fluentd Input Total Bytes Logged in last 15 minutes",
```
eranra, line 1871: Remove the word `Fluentd` ... the metric represents bytes generated by containers so not related to fluentd.

Whould advice to replace title with: "Logs emitted (top 10 containers)  - Bytes/Second"

@jcantrill  ^^^ == ?

files/dashboards/openshift-logging-dashboard.json
```diff
@@ -1708,18 +1798,233 @@
...
+      "pointradius": 2,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "topk(10, increase(fluentd_input_status_total_bytes_logged[15m]) > 0)",
```
eranra, line 1861: @ajaygupta978 I know that we agreed that rate and increase are similar but I think that changing to rate function will make things more understood ... also ant particular reason for 15m ... maybe reduce that number so that the metric will update more quickly ... how about 5 minutes ?

files/dashboards/openshift-logging-dashboard.json
```diff
@@ -1708,18 +1798,233 @@
...
+      "pointradius": 2,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "topk(10, increase(fluentd_input_status_total_bytes_collected[15m]) > 0)",
```
eranra, line 1969: same

files/dashboards/openshift-logging-dashboard.json
eranra: @ajaygupta978 in any event it will be good to normalize the information presented to users so that it is per seconds and not related to the period of metric collection (i.e. 15m or any other value)

files/dashboards/openshift-logging-dashboard.json
```diff
@@ -1708,18 +1798,233 @@
...
+          "interval": "",
+          "legendFormat": "{{path}}",
+          "refId": "A"
+    }
+  ],
+      "thresholds": [],
+      "timeFrom": null,
+      "timeRegions": [],
+      "timeShift": null,
+      "title": "Fluentd Input Total Bytes Collected in last 15 minutes",
```
eranra, line 1979: @ajaygupta978 maybe we do not need this one ... if you add log loss ... I think that customers will want to see just log loss .... it is not interesting to see top 10 collected logs .... this is just my thinking ... 

@jcantrill === ?

In any event if we decide to keep that, I would advice to change into::

"Logs collected by cluster-logging (top 10 containers) - Bytes/Second"

files/dashboards/openshift-logging-dashboard.json
```diff
@@ -1755,5 +2061,5 @@
   "timezone": "",
   "title": "Openshift Logging",
   "uid": "seyA3sMMk",
-  "version": 1
-}
+  "version": 2
+}
```
eranra, line 2065: @ajaygupta978 more lines to get back to original here in the end :-)

ajaygupta978: Done.

files/dashboards/openshift-logging-dashboard.json
ajaygupta978: Yeah I agree. I just remove this.

files/dashboards/openshift-logging-dashboard.json
ajaygupta978: done

files/dashboards/openshift-logging-dashboard.json
ajaygupta978: Yeah agree even I noticed in the code other charts are using 5m and rate function make more sense because its normalises.

files/dashboards/openshift-logging-dashboard.json
ajaygupta978: done

files/dashboards/openshift-logging-dashboard.json
ajaygupta978: yeah I reverted them.

files/dashboards/openshift-logging-dashboard.json
ajaygupta978: These are coming from Granafa but now I removed them.

files/dashboards/openshift-logging-dashboard.json
ajaygupta978: reverted.

ajaygupta978:
@blockloop @jcantrill @alanconway 
@eranra  and I had discussion about log-loss and based on that discussion we think `rate` function is more appropriate than `increase` because rate produced normalised values which is emitted bytes/sec while increase shows logs emitted in last X mins so it dependent on X and not normalized.  Users should be able to calculated total logs emitted for given time-period using value provided by chart. 
Here are two charts attached. In case of High-stress we observed  2.77 MB/sec and low-stress 208k/sec which is consistent with ground truth.
`High-stress container`
High_stress_container
`Low-stress containter`
Low_stress_containter

alanconway:
How does this PR relate to https://github.com/openshift/cluster-logging-operator/pull/971 ?

ajaygupta978:
> How does this PR relate to #971 ?

This PR is for top 10 charts for log-emitted while #971  is for showing log-loss.

alanconway:
Agree re rate() function, as I read it is specifically aimed at X-per-second situations like we have here.
Need to label the lines in the graph somehow: I guess the top line is total bytes-loged by the top 10 containers, and the bottom line is the total loss?

alanconway:
Feature idea  (not for this PR, maybe later) when someone is looking at one of these graphs with named logs underneath, would it be possible to enable a click from an individual log name to a graph showing the logged/collected/lost for just that log?

alanconway:
Need to label the lines in the graph somehow: they don't correspond to the colours of the logs below. Looks like the top line is a total of all 10, what is the bottom line? Would be good to have human-readable text as well as the LogQL query.
 
Can we graph the bytes-logged for each container in colour as well as of the totals?

ajaygupta978:
> Feature idea (not for this PR, maybe later) when someone is looking at one of these graphs with named logs underneath, would it be possible to enable a click from an individual log name to a graph showing the logged/collected/lost for just that log?

I really doubt we can add any interactivity apart from current tool tip. OCP dashboard are mostly static graphs but will check with monitoring team.

ajaygupta978:
> Need to label the lines in the graph somehow: they don't correspond to the colours of the logs below. Looks like the top line is a total of all 10, what is the bottom line? Would be good to have human-readable text as well as the LogQL query.

There are two containers in top line. These are heavy stress containers and emitting  at same rate so they are merged into single line. If you see `High-stress container` screenshot it shows green tool tip which corresponds to 2nd container green label in the bottom so colors are consistent. Bottom line shows 8 low-stress containers and since they also generate at same rate so all of them are merged. In the 2nd screenshot we can see dark brown tool tip which should match low-stress container which is either 9th/10th container and missed because lack of space in screen shot. 

> Can we graph the bytes-logged for each container in colour as well as of the totals?
bytes-logged is covered in the chart for top 10 containers.  For total bytes we need to create another chart.May be we can create another PR for it if we agree to create it.
@eranra

alanconway:
> > Need to label the lines in the graph somehow: they don't correspond to the colours of the logs below. Looks like the top line is a total of all 10, what is the bottom line? Would be good to have human-readable text as well as the LogQL query.
> 
> There are two containers in top line. These are heavy stress containers and emitting at same rate so they are merged into single line. If you see `High-stress container` screenshot it shows green tool tip which corresponds to 2nd container green label ..

@ajaygupta978 OK, you are already doing what I wanted - it just isn't clear on the image. I guess with the live console you can zoom or use tool tips make it clearer.

/lgtm

jcantrill:
@ajaygupta978 please squash into single commit

PR #955, opened 2021-03-19
LOG-1123: fix routing of app logs per namespace

jcantrill:
### Description
This PR
* fixes routing of application messages by namespace so they end up at the desired outputs

/cc @alanconway @vimalk78 

/cherrypick release-5.0

### Links
https://issues.redhat.com/browse/LOG-1223

### DIFF
--- a/pkg/generators/forwarding/fluentd/fluent_conf_test.go
+++ b/pkg/generators/forwarding/fluentd/fluent_conf_test.go
@@ -71,13 +71,21 @@ var _ = Describe("Generating fluentd config", func() {
 		}
 	})
 
-	It("should generate fluent config for sending given namespaces logs only to output", func() {
+	It("should generate fluent config for sending specific namespace logs to separate pipelines", func() {
 		forwarder = &logging.ClusterLogForwarderSpec{
 			Outputs: []logging.OutputSpec{
 				{
 					Type: logging.OutputTypeElasticsearch,
 					Name: "apps-es-1",
 					URL:  "https://es.svc.messaging.cluster.local:9654",
+					Secret: &logging.OutputSecretSpec{
+						Name: "my-es-secret",
+					},
+				},
+				{
+					Type: logging.OutputTypeElasticsearch,
+					Name: "apps-es-2",
+					URL:  "https://es2.svc.messaging.cluster.local:9654",
 					Secret: &logging.OutputSecretSpec{
 						Name: "my-es-secret",
 					},
@@ -90,12 +98,28 @@ var _ = Describe("Generating fluentd config", func() {
 						Namespaces: []string{"project1-namespace", "project2-namespace"},
 					},
 				},
+				{
+					Name: "myInput2",
+					Application: &logging.Application{
+						Namespaces: []string{"dev-apple", "project2-namespace"},
+					},
+				},
 			},
 			Pipelines: []logging.PipelineSpec{
+				{
+					Name:       "my-default-pipeline",
+					InputRefs:  []string{"application"},
+					OutputRefs: []string{"apps-es-2"},
+				},
 				{
 					Name:       "apps-pipeline",
 					InputRefs:  []string{"myInput"},
 					OutputRefs: []string{"apps-es-1", "apps-es-2"},
+				},
+				{
+					Name:       "apps-pipeline2",
+					InputRefs:  []string{"myInput2"},
+					OutputRefs: []string{"apps-es-2"},
 				},
 			},
 		}
@@ -463,12 +487,9 @@ var _ = Describe("Generating fluentd config", func() {
     <match **_default_** **_kube-*_** **_openshift-*_** **_openshift_** journal.** system.var.log**>
       @type null
     </match>
-    <match kubernetes.**_project1-namespace_** kubernetes.**_project2-namespace_** >
+    <match kubernetes.**>
       @type relabel
       @label @_APPLICATION
-    </match>
-    <match kubernetes.** >
-      @type null
     </match>
     <match linux-audit.log** k8s-audit.log** openshift-audit.log**>
       @type null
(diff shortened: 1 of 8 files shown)


k-keiichi-rh:
Hello, I also tried to fix Bug 1925361 by using label router plugin in the following patch.
https://github.com/openshift/cluster-logging-operator/pull/865/commits/5c691df350cb1112dfbbc60702260943373440b6

After merging this PR, I will drop the above patch and rebase your commit.

pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -482,31 +503,69 @@ var _ = Describe("Generating fluentd config", func() {
...
+      </store>
+    </match>
     <match **>
       @type copy
-
       <store>
-        @type relabel
-        @label @APPS_PIPELINE
+	    @type relabel
+	    @label @MY_DEFAULT_PIPELINE
```
vimalk78, line 535: does this `<match>` order against namespace tag means `MY_DEFAULT_PIPELINE` receives all namespaces **except** `project1-namespace, project2-namespace, dev-apple` namespace? is that the intention?

https://docs.fluentd.org/configuration/config-file#note-on-match-order

Do we ensure a wider match appears lower in the order?

vimalk78: I built forwarder generator from this PR and generated fluent.conf from below yaml. I think this yaml was part of a BZ about NS forwarding
```
kind: ClusterLogForwarder
metadata:
  name: instance
  namespace: openshift-logging
spec:
  inputs:
    - application:
        namespaces:
         - test
      name: specific-app-logs
  outputs:
   - name: rsyslog-ex
     type: syslog
     syslog:
       facility: local0
       rfc: RFC3164
       payloadKey: message
       severity: informational
     url: 'tcp://10.10.93.88:514'
  pipelines:
    - inputrefs:
        - application
        - audit
        - infrastructure
      name: application-audit-infrastructure-logs
      outputrefs:
        - default
    - inputrefs:
        - specific-app-logs
      name: forward-to-external
      outputrefs:
        - rsyslog-ex
```
It generates cofig with wider match on top. So perhaps we cannot guarantee match order in generated conf.


```
# Relabel specific sources (e.g. logs.apps) to multiple pipelines
<label @_APPLICATION>
  <match **> 
    @type copy
    <store>
      @type relabel
      @label @APPLICATION_AUDIT_INFRASTRUCTURE_LOGS
    </store>
  </match>
  <match kubernetes.**_test_**>      #<<<< this will never match
    @type copy
    <store>
      @type relabel
      @label @FORWARD_TO_EXTERNAL
    </store>
  </match>
</label>
<label @_AUDIT>
  <match **>
    @type copy
    <store>
      @type relabel
      @label @APPLICATION_AUDIT_INFRASTRUCTURE_LOGS
    </store>
  </match>
</label>
<label @_INFRASTRUCTURE>
  <match **>
    @type copy
    <store>
      @type relabel
      @label @APPLICATION_AUDIT_INFRASTRUCTURE_LOGS
    </store>
  </match>
</label>
```

jcantrill: I will review your findings.  Each match block should be 1:1 to a namespace since we know where the namepace falls within the tag.  The blocks should be ordered such that ns matches occur before a catch all which is `**`.  If it does not do this then that is a bug.

jcantrill: I see the issue.  This is because we have the concept of `application` which means all, and a subset of application logs which come from a defined pipeline.  It has to know how to deal with both

jcantrill: actually, the test case looks like it covers this scenario, so need to investigate

alanconway: I think you need to add MY_DEFAULT_ROUTE to every copy stanza, because it receives from every namespace. Then the ** catch-all will add any namespaces that were not explicitly matched.

Is this really easier or more efficient than using https://github.com/banzaicloud/fluent-plugin-label-router? It does exactly what we want - define a set of routes, guarded by a set of label/namepsace matching rules.

alanconway (changes requested):
Looks OK except for the copy issue discussed above that needs fixing.

I'm ok with releasing this fix for the immediate problem, but I think this will be more maintainable if we replace it with the label router plugin instead. We can consider that when the label selector PR is ready since it also needs that plugin.

pkg/generators/forwarding/fluentd/fluent_conf_test.go
jcantrill: I have no familiarity with this plugin.  I was trying to solve a bz that has been outstanding since essentially GA.  I'm not sure what you are recommended here.  The only issue I can see was the "anything left over" should be caught by "**" has to go last, which I resolved.  The pipeline via labels is basically:  INGRESS->_TYPE ->PIPELINE->OUTPUT where application logs entering _TYPE are specifically copied to multiple PIPELINES as defined by the input namespacing.

jcantrill:
> may be not in scope of this PR but can we use `*` to match more than one namespace
> I used a yaml which specifies namespace with `*` in it.
> 
> ```
> spec:
>   inputs:
>     - application:
>         namespaces:
>          - "test*abc"
>       name: specific-app-logs
> ```
> 
> and forwarder generator generated conf
> 
> ```
>   <match kubernetes.**_test*abc_**>
>     @type copy
>     <store>
>       @type relabel
>       @label @FORWARD_TO_EXTERNAL
>     </store>
>   </match>
> ```
I'm not certain what you are hinting at here.  It looks like we are still ok even in this situation and that this is a desirable side-affect

pkg/generators/forwarding/fluentd/fluent_conf_test.go
k-keiichi-rh: This may be off-topic, but please let me explain what I tried to do in https://github.com/openshift/cluster-logging-operator/pull/865.

As @alanconway said, label router plugin would help us to maintain a set of label/namepsace matching rules.

In https://github.com/openshift/cluster-logging-operator/pull/865, I tried to introduce Input Selector Block, which defines a set of routes by using label router plugin.
This block will re-label a subset of application logs based on namespaces as well as their Kubernetes metadata.
We will be able to centrally control  label/namespace matching rules by label router plugin. 
### Example LF CR:
```
spec:
  inputs:
   - name: app-product-logs
     application:
       namespaces:
         - test1
       selector
         matchLabels: 
           environment: product
   - name: app-dev-logs
     application:
       namespaces:
         - test2
       selector
         matchLabels: 
           environment: dev
  outputs:
   - name: syslog1
     ...
   - name: syslog2
     ...
  pipelines:
   - name: forward-product-logs-to-syslog1
     inputRefs:
     - app-product-logs
     outputRefs:
     - syslog1
   - name: forward-dev-logs-to-syslog2
     inputRefs:
     - app-dev-logs
     outputRefs:
     - syslog2
   - name: all-default
     inputRefs:
     - audit
     - infrastructure
     - application
     outputRefs:
     - default
```
### fluent.conf after applying Example LF CR:
```
<label @_APPLICATION>
  <match **>
    @type copy
    <store>
      @type relabel
      @label @ALL_DEFAULT
    </store>
    <store>
      @type relabel
      @label @_APPLICATION_BASED_ON_INPUT_SELECTOR
    </store>
  </match>
</label>
<label @_APPLICATION_BASED_ON_INPUT_SELECTOR>
  <match **>
    @type label_router       # this will re-label application logs as defined by the input namespacing instead of copying to multiple PIPELINES
    <route>
      @label @FORWARD_PRODUCT_LOGS_TO_SYSLOG1
      <match>
        labels environment:product
        namespaces test1
      </match>
    </route>
    <route>
      @label @FORWARD_DEV_LOGS_TO_SYSLOG2
      <match>
        labels environment:dev
        namespaces test2
      </match>
    </route>
  </match>
</label>

<label @_AUDIT>
  <match **>
    @type copy
    <store>
      @type relabel
      @label @ALL_DEFAULT
    </store>
  </match>
</label>

<label @_INFRASTRUCTURE>
  <match **>
    @type copy
    <store>
      @type relabel
      @label @ALL_DEFAULT
    </store>
  </match>
</label>
```

vimalk78:
> 
> I'm not certain what you are hinting at here. It looks like we are still ok even in this situation and that this is a desirable side-affect

I meant we are not defining what user can input as namespace, and we should define its schema. Is it a namespace name as defined by k8s? or any arbitrary string

jcantrill:
> > I'm not certain what you are hinting at here. It looks like we are still ok even in this situation and that this is a desirable side-affect
> 
> I meant we are not defining what user can input as namespace, and we should define its schema. Is it a namespace name as defined by k8s? or any arbitrary string

This sounds more like we need to add some validation into the CSV to retrict the values of namespace which as you noted is separate from this PR.  Strict validation as a namespace may not be desirable as we could leverage the use of wildcards for users.  We do require, however, something that is a valid flluent 'match' if we are going to drop it straight into the config.  Please open an issue to address this

jcantrill:
@alanconway @vimalk78 can you Shepard the 5.0 and 4.6 back ports for this issue?

PR #959, opened 2021-03-23
LOG-1161: Added events for valid/invalid pipelines

igor-karpukhin:
### Description
CLO will emit events for a valid and invalid CLF pipelines

/cc @jcantrill 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-1161

### DIFF
--- a/pkg/controller/forwarding/forwarding_controller.go
+++ b/pkg/controller/forwarding/forwarding_controller.go
@@ -12,6 +12,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
@@ -28,7 +29,11 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileForwarder{client: mgr.GetClient(), scheme: mgr.GetScheme()}
+	return &ReconcileForwarder{
+		client:   mgr.GetClient(),
+		scheme:   mgr.GetScheme(),
+		recorder: mgr.GetEventRecorderFor("clusterlogforwarder"),
+	}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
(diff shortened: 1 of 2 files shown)


vimalk78:
Please add an example of added Event in the PR description

igor-karpukhin:
@vimalk78 I'll a test with the example event output

igor-karpukhin:
Here is an example event: 
````
Events:
  Type    Reason  Age                From                 Message
  ----    ------  ----               ----                 -------
  Normal  Ready   21s (x2 over 21s)  clusterlogforwarder  All pipelines are valid
```

pkg/controller/forwarding/forwarding_controller.go
```diff
@@ -111,10 +117,13 @@ func (r *ReconcileForwarder) Reconcile(request reconcile.Request) (reconcile.Res
 
 	if instance.Status.IsReady() {
 		instance.Status.Conditions.SetCondition(condReady)
+		r.recorder.Event(instance, "Normal", string(condReady.Type), "All pipelines are valid")
```
alanconway, line 120: We don't want to emit an event for every reconcile, only when a condition changes:

```suggestion
		if instance.Status.Conditions.SetCondition(condReady) {
				r.recorder.Event(instance, "Normal", string(condReady.Type), "All pipelines are valid")
		}
```

pkg/controller/forwarding/forwarding_controller.go
```diff
@@ -111,10 +117,13 @@ func (r *ReconcileForwarder) Reconcile(request reconcile.Request) (reconcile.Res
...
 	}
 
 	if instance.Status.IsDegraded() {
-		instance.Status.Conditions.SetCondition(condDegraded(logging.ReasonInvalid, "Some pipelines are degraded or invalid"))
+		msg := "Some pipelines are degraded or invalid"
+		instance.Status.Conditions.SetCondition(condDegraded(logging.ReasonInvalid, msg))
+		r.recorder.Event(instance, "Error", string(logging.ReasonInvalid), msg)
```
alanconway, line 126: Same again, only event if conition has changed.

alanconway (changes requested):
We only want an event if the condition of the forwarder changes, not on every reconcile (see suggested diff) If we do that, then I think it's ok to have an event for "Ready" since it will only be sent on first start-up, and after recover from some degraded or invalid state.

pkg/controller/forwarding/forwarding_controller.go
igor-karpukhin: Fixed

pkg/controller/forwarding/forwarding_controller.go
igor-karpukhin: Fixed

alanconway (approved):
/lgtm

PR #960, opened 2021-03-23
LOG-1238: Add hack script to create deployment list

jcantrill:
### Description
This PR:
* re-introduces a script from 4.2 to generate a list of kubernetes resources from the CSV to provide an alternate deployment method to hack cluster logging operator

/cc @sichvoge  @alanconway 

### Links
* https://issues.redhat.com/browse/LOG-679

### DIFF
--- a/docs/HACKING.md
+++ b/docs/HACKING.md
@@ -190,3 +190,19 @@ will attempt to clone them to a temporary directory.
 
 Bumping the release and manifest versions typically require updating the `elasticsearch-operator` as well.
 * `dep ensure -update github.com/openshift/elasticsearch-operator`
+
+## Deploying without OLM
+Production relies upon OLM to manage and control the operator deployment, permissions, etc. The manifest defines all the resources needed by OLM.  We can use this same manifest to generate a list of resources to deploy without using OLM.
+```
+make deploy-image
+```
+will produce output that should give you the pullspec on the cluster like:
+```
+image-registry.openshift-image-registry.svc:5000/openshift/origin-cluster-logging-operator:latest
+```
+
+which will allow you to use the script like:
+```                                                                                                                                                                                                                                                                               
+CLO_IMAGE=image-registry.openshift-image-registry.svc:5000/openshift/origin-cluster-logging-operator:latest \
+./hack/gen-olm-artifacts.py manifests/5.1/cluster-logging.v5.1.0.clusterserviceversion.yaml  $CLO_IMAGE | oc create -f -
+```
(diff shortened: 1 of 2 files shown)


docs/HACKING.md
```diff
@@ -190,3 +190,11 @@ will attempt to clone them to a temporary directory.
 
 Bumping the release and manifest versions typically require updating the `elasticsearch-operator` as well.
 * `dep ensure -update github.com/openshift/elasticsearch-operator`
+
+## Deploying without OLM
+Production relies upon OLM to manage and control the operator deployment, permissions, etc. The manifest defines all the resources needed by OLM.  We can use this same manifest to generate a list of resources to deploy without using OLM.
+```
+make deploy-image                                                                                                                                                                                                                                                                               
+
+./hack/gen-olm-artifacts.sh manifests/5.1/cluster-logging.v5.1.0.clusterserviceversion.yaml  $CLO_IMAGE | oc create -f -
```
sichvoge, line 199: Can you describe, for testing purposes, to what we should point the CLO_IMAGE to?

sichvoge (commented):
Small comment.

hack/gen-olm-artifacts.sh
```diff
@@ -0,0 +1,215 @@
+#!/usr/bin/env python
```
alanconway, line 1: Nit, file suffix should be .py not .sh

alanconway (changes requested):
Rename it to .py and it's good to go.

jcantrill:
@sichvoge updated.

docs/HACKING.md
```diff
@@ -190,3 +190,19 @@ will attempt to clone them to a temporary directory.
...
+make deploy-image
+```
+will produce output that should give you the pullspec on the cluster like:
+```
+image-registry.openshift-image-registry.svc:5000/openshift/origin-cluster-logging-operator:latest
+```
+
+which will allow you to use the script like:
+```                                                                                                                                                                                                                                                                               
+CLO_IMAGE=image-registry.openshift-image-registry.svc:5000/openshift/origin-cluster-logging-operator:latest \
```
alanconway, line 206: Would be nice to wrap this up as `make deploy-no-olm`

hack/gen-olm-artifacts.py
```diff
@@ -0,0 +1,215 @@
+#!/usr/bin/env python
```
alanconway, line 1: Just a thought, not an issue here, but should we think about moving towards Go as a single language for stuff like this? It can be run like a scripting language `go run gen-olm-artifacts.go` as well as compiled, and the templates package makes this kind of thing pretty easy.

alanconway (approved):
/lgtm

hack/gen-olm-artifacts.py
jcantrill: my only reservation would be if we want to allow someone to utilize without relying on golang to accomplish deployment :man_shrugging:

PR #961, opened 2021-03-26
LOG-1157: Send Logs to specific ES index

vimalk78:
### Description
 - Update viaq_inde_name field of log record based on indexName/indexKey set in the CLF spec
 - Removed sending flatten_labels to all outputs except elasticsearch
 - Enabled sending labels to all outputs except elasticsearch
 - Modified unit tests
 - Added/Modified functional tests
/cc 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-1157
- Enhancement Proposal: https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/forwarding-json-structured-logs.md

Also fixes https://issues.redhat.com/browse/LOG-1148


pkg/generators/forwarding/fluentd/fluent_conf_test.go
```diff
@@ -1023,6 +1025,10 @@ var _ = Describe("Generating fluentd config", func() {
 			</label>
 			# Ship logs to specific outputs
 			<label @SECUREFORWARD_RECEIVER>
+				<filter **>
```
alanconway, line 1028: Why removing labels from forward receivers?

pkg/generators/forwarding/fluentd/output_conf_syslog_test.go
```diff
@@ -29,6 +29,10 @@ var _ = Describe("Generating external syslog server output store config blocks",
 		  merge_json_log false
 		  replace_json_log true
 		</filter>
+		<filter **>
```
alanconway, line 32: Why removing labels from syslog output?

pkg/generators/forwarding/fluentd/output_fluentd_buffer_test.go
```diff
@@ -605,6 +634,10 @@ var _ = Describe("Generating fluentd config", func() {
 
 		It("should provide a default buffer configuration", func() {
 			kafkaConf := `<label @KAFKA_RECEIVER>
+		<filter **>
```
alanconway, line 637: Why removing labels from kafka receiver?

alanconway (changes requested):
Should only be removing labels for Elasticsearch outputs.

pkg/generators/forwarding/fluentd/fluent_conf_test.go
vimalk78: because earlier the labels were removed earlier in the fluent pipeline, and thus removed from all the outputs

alanconway (commented):
/lgtm

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -598,6 +597,21 @@ const pipelineToOutputCopyTemplate = `{{- define "pipelineToOutputCopyTemplate"
 
 const outputLabelConfTemplate = `{{- define "outputLabelConf" -}}
 <label {{.LabelName}}>
+  {{ if (.NeedChangeElasticsearchIndexName) -}}
+  <filter **>
+    @type record_modifier
+	<record>
+	  indexFromKey     ${record['kubernetes']['labels']['{{.Target.OutputTypeSpec.Elasticsearch.IndexKey}}']}
```
alanconway, line 604: This is incorrect: we want to treat IndexKey as a full key in the record. As perr the PR, it could also be kubernetes.namespace_name or kuberntes.openshift.label.XX

/lgtm cancel

vimalk78: @alanconway PTAL. tried to fix it.

alanconway (commented):
/lgtm

pkg/generators/forwarding/fluentd/fluent_conf.go
```diff
@@ -120,6 +120,25 @@ func (conf *outputLabelConf) StoreID() string {
...
+func (conf *outputLabelConf) NeedChangeElasticsearchIndexName() bool {
+	return conf.Target.Type == logging.OutputTypeElasticsearch &&
+		conf.Target.OutputTypeSpec.Elasticsearch != nil &&
+		(conf.Target.OutputTypeSpec.Elasticsearch.IndexKey != "" || conf.Target.OutputTypeSpec.Elasticsearch.IndexName != "")
+}
+
+func generateRubyDigArgs(path string) string {
+	var args []string
+	for _, s := range strings.Split(path, ".") {
+		args = append(args, fmt.Sprintf("\"%s\"", s))
```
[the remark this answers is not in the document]
vimalk78: I changed as per your suggestion.
About the test case, i need to think how best to add. may be i can add it in another PR?

test/functional/normalization/flatten_labels_test.go
```diff
@@ -17,30 +17,55 @@ var _ = Describe("[Normalization] Fluentd normalization", func() {
...
+
+			//verify the new key exists
+			Expect(logs[0].Kubernetes.FlatLabels).To(Not(BeNil()), fmt.Sprintf("Expected to find the kubernetes.flat_labels key in %#v", logs[0]))
+
+			//verify we removed the old key
+			Expect(logs[0].Kubernetes.Labels).To(BeNil(), fmt.Sprintf("Did not expect to find the kubernetes.labels key in %#v", logs[0]))
+		})
+	})
+	Context("for non-ES output", func() {
+		It("should not remove 'kubernetes.labels' and not add 'kubernetes.flat_labels'", func() {
```
jcantrill, line 54: @alanconway I am on board with this change since I'm confident any users of non-ES outputs would prefer a map of labels and likely don't use "flat_labels", but this does break backwards compatibility.  I am inclined to go forward with the change and risk a bug to fix in the future if needed

test/functional/outputs/forward_to_elasticsearch_index_test.go
```diff
@@ -0,0 +1,190 @@
...
+			var logs []types.ApplicationLog
+			err = types.StrictlyParseLogs(raw, &logs)
+			Expect(err).To(BeNil(), "Expected no errors parsing the logs")
+			// Compare to expected template
+			outputTestLog := logs[0]
+			outputLogTemplate.ViaqIndexName = ""
+			Expect(outputTestLog).To(matchers.FitLogFormatTemplate(outputLogTemplate))
+		})
+	})
+	Context("when sending to ElasticSearch "+elasticSearchTag+" protocol", func() {
```
jcantrill, line 82: It looks like you can remove this context since and pull the "It" into the preceeding one since they are the same

test/functional/outputs/forward_to_elasticsearch_index_test.go
```diff
@@ -0,0 +1,190 @@
...
```
jcantrill, line 108: same here

test/functional/outputs/forward_to_elasticsearch_index_test.go
```diff
@@ -0,0 +1,190 @@
...
```
jcantrill, line 137: same here

test/functional/outputs/forward_to_elasticsearch_index_test.go
```diff
@@ -0,0 +1,190 @@
...
+				err = types.StrictlyParseLogs(raw, &logs)
+				Expect(err).To(BeNil(), "Expected no errors parsing the logs")
+				// Compare to expected template
+				outputTestLog := logs[0]
+				outputLogTemplate.ViaqIndexName = ""
+				Expect(outputTestLog).To(matchers.FitLogFormatTemplate(outputLogTemplate))
+			})
+		})
+	})
+	Context("when sending to ElasticSearch "+elasticSearchTag+" protocol", func() {
```
jcantrill, line 162: same here

test/functional/outputs/forward_to_elasticsearch_test.go
```diff
@@ -1,56 +1,39 @@
...
-
 		// Template expected as output Log
 		outputLogTemplate = functional.NewApplicationLogTemplate()
 	)
 
 	BeforeEach(func() {
 
 		framework = functional.NewFluentdFunctionalFramework()
-		addElasticSearchContainer := newVisitor(framework)
+		//addElasticSearchContainer := newVisitor(framework)
```
jcantrill, line 24: remove me

test/functional/outputs/forward_to_elasticsearch_test.go
```diff
@@ -1,56 +1,39 @@
...
 		functional.NewClusterLogForwarderBuilder(framework.Forwarder).
 			FromInput(logging.InputNameApplication).
 			ToElasticSearchOutput()
-		Expect(framework.DeployWithVisitor(addElasticSearchContainer)).To(BeNil())
+		Expect(framework.Deploy()).To(BeNil())
+		//Expect(framework.DeployWithVisitor(addElasticSearchContainer)).To(BeNil())
```
jcantrill, line 29: remove me

jcantrill (changes requested):
mostly nits with one outstanding question otherwise lgtm

test/functional/normalization/flatten_labels_test.go
k-keiichi-rh: As for the legacy syslog and Fluentd method, I am afraid that the flattening labels code are already removed by this commit 2a19585ad75e34b4a62c99c8f1e93d72a0b7c9fa. It depends on the result of this discussion, but I will fix them to keep the backward compatibility.

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -622,6 +613,28 @@ const pipelineToOutputCopyTemplate = `{{- define "pipelineToOutputCopyTemplate"
 
 const outputLabelConfTemplate = `{{- define "outputLabelConf" -}}
 <label {{.LabelName}}>
+  {{- if (.NeedChangeElasticsearchIndexName)}}
+  <filter **>
+    @type record_modifier
+	<record>
+	  indexFromKey     ${record.dig({{.GetKeyVal .Target.OutputTypeSpec.Elasticsearch.IndexKey}})}
+	  hasIndexName     "{{.Target.OutputTypeSpec.Elasticsearch.IndexName}}"
+	  viaq_index_name  ${if !record['indexFromKey'].nil?; record['indexFromKey']; elsif record['hasIndexName'] != ""; record['hasIndexName']; else record['viaq_index_name']; end;}
```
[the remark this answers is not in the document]
vimalk78: any suggestions for improvement?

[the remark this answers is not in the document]
jcantrill: I think we would need to move this into a plugin that evaluates it there.

test/functional/normalization/flatten_labels_test.go
alanconway: @jcantrill  OK  let's try going with the change and using labels for everything except ES, which gets flat_labels.

pkg/generators/forwarding/fluentd/templates.go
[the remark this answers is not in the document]
vimalk78: `record_modifier` works faster than `record_transformer` 
https://docs.fluentd.org/filter/record_transformer#need-more-performance

PR #979, opened 2021-04-13
Go implementation log file exporter jira ticket log 1188

pmoogi-redhat:
### Description
Summary
Implement an independent process to monitor CRIO container log file rotations that can be used to publish total_bytes_logged metrics in prometheus format. The motivation for moving away from a Ruby implementation is Ruby is not efficient in tracking file rotations. Also collectors can change over time so making metric computation outside of fluentd for logging total_bytes_logged is more preferred.

Exposes metric endpoint that is consumable by Prometheus at port :2112
Exposes metric of total_bytes_logged

/cc @alanconway @jcantrill 
/assign @jcantrill 

### Links

- Related PR(s): origin-aggregated-logging:Metric-for-inbound-log-data-loss-at-the-collector-JIRA-ticket-[LOG-1032](https://issues.redhat.com/browse/LOG-1032)
   Got in_tail and prometheus plugin changes which enable total_bytes_collected metric computed and published in prometheus plugin
- JIRA: https://issues.redhat.com/browse/LOG-1188

### DIFF
--- a/bundle/manifests/cluster-logging.v5.2.0.clusterserviceversion.yaml
+++ b/bundle/manifests/cluster-logging.v5.2.0.clusterserviceversion.yaml
@@ -330,6 +330,8 @@ spec:
                     value: "cluster-logging-operator"
                   - name: FLUENTD_IMAGE
                     value: "quay.io/openshift-logging/fluentd:latest"
+                  - name: LOGFILEMETRICEXPORTER_IMAGE
+                    value: "quay.io/openshift-logging/log-file-metric-exporter:latest"
   customresourcedefinitions:
     owned:
     - name: clusterloggings.logging.openshift.io
--- a/manifests/5.2/cluster-logging.v5.2.0.clusterserviceversion.yaml
+++ b/manifests/5.2/cluster-logging.v5.2.0.clusterserviceversion.yaml
@@ -328,6 +328,8 @@ spec:
                     value: "cluster-logging-operator"
                   - name: FLUENTD_IMAGE
                     value: "quay.io/openshift-logging/fluentd:latest"
+                  - name: LOGFILEMETRICEXPORTER_IMAGE
+                    value: "quay.io/openshift-logging/log-file-metric-exporter:latest"
   customresourcedefinitions:
     owned:
     - name: clusterloggings.logging.openshift.io
--- a/olm_deploy/operatorregistry/registry-deployment.yaml
+++ b/olm_deploy/operatorregistry/registry-deployment.yaml
@@ -41,6 +41,8 @@ spec:
           value: ${IMAGE_ELASTICSEARCH6}
         - name: IMAGE_LOGGING_KIBANA6
           value: ${IMAGE_LOGGING_KIBANA6}
+        - name: IMAGE_LOG_FILE_METRIC_EXPORTER
+          value: ${IMAGE_LOG_FILE_METRIC_EXPORTER}
 
       containers:
       - name: cluster-logging-operator-registry
--- a/olm_deploy/scripts/catalog-deploy.sh
+++ b/olm_deploy/scripts/catalog-deploy.sh
@@ -6,6 +6,7 @@ export IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY=${IMAGE_CLUSTER_LOGGING_OPERATOR_
 export IMAGE_CLUSTER_LOGGING_OPERATOR=${IMAGE_CLUSTER_LOGGING_OPERATOR:-registry.ci.openshift.org/${LOGGING_IS}/${LOGGING_VERSION}:cluster-logging-operator}
 export IMAGE_LOGGING_CURATOR5=${IMAGE_LOGGING_CURATOR5:-registry.ci.openshift.org/${LOGGING_IS}/${LOGGING_VERSION}:logging-curator5}
 export IMAGE_LOGGING_FLUENTD=${IMAGE_LOGGING_FLUENTD:-registry.ci.openshift.org/${LOGGING_IS}/${LOGGING_VERSION}:logging-fluentd}
+export IMAGE_LOG_FILE_METRIC_EXPORTER=${IMAGE_LOG_FILE_METRIC_EXPORTER:-registry.ci.openshift.org/${LOGGING_IS}/${LOGGING_VERSION}:log-file-metric-exporter}
 
 CLUSTER_LOGGING_OPERATOR_NAMESPACE=${CLUSTER_LOGGING_OPERATOR_NAMESPACE:-openshift-logging}
 
@@ -14,6 +15,7 @@ echo "cluster logging operator registry: ${IMAGE_CLUSTER_LOGGING_OPERATOR_REGIST
 echo "cluster logging operator: ${IMAGE_CLUSTER_LOGGING_OPERATOR}"
 echo "curator5: ${IMAGE_LOGGING_CURATOR5}"
 echo "fluentd: ${IMAGE_LOGGING_FLUENTD}"
+echo "log-file-metric-exporter: ${IMAGE_LOG_FILE_METRIC_EXPORTER}"
 
 echo "In namespace: ${CLUSTER_LOGGING_OPERATOR_NAMESPACE}"
 
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -18,16 +18,17 @@ const (
 	FluentdTrustedCAName       = "fluentd-trusted-ca-bundle"
 	KibanaTrustedCAName        = "kibana-trusted-ca-bundle"
 	// internal elasticsearch FQDN to prevent to connect to the global proxy
-	ElasticsearchFQDN   = "elasticsearch.openshift-logging.svc"
-	ElasticsearchName   = "elasticsearch"
-	ElasticsearchPort   = "9200"
-	FluentdName         = "fluentd"
-	KibanaName          = "kibana"
-	KibanaProxyName     = "kibana-proxy"
-	CuratorName         = "curator"
-	LogStoreURL         = "https://" + ElasticsearchFQDN + ":" + ElasticsearchPort
-	MasterCASecretName  = "master-certs"
-	CollectorSecretName = "fluentd"
+	ElasticsearchFQDN          = "elasticsearch.openshift-logging.svc"
+	ElasticsearchName          = "elasticsearch"
+	ElasticsearchPort          = "9200"
+	FluentdName                = "fluentd"
+	KibanaName                 = "kibana"
+	KibanaProxyName            = "kibana-proxy"
+	CuratorName                = "curator"
+	LogfilesmetricexporterName = "logfilesmetricexporter"
+	LogStoreURL                = "https://" + ElasticsearchFQDN + ":" + ElasticsearchPort
+	MasterCASecretName         = "master-certs"
+	CollectorSecretName        = "fluentd"
 	// Disable gosec linter, complains "possible hard-coded secret"
 	CollectorSecretsDir     = "/var/run/ocp-collector/secrets" //nolint:gosec
 	KibanaSessionSecretName = "kibana-session-secret"          //nolint:gosec
@@ -35,8 +36,9 @@ const (
 	LegacySecureforward = "_LEGACY_SECUREFORWARD"
 	LegacySyslog        = "_LEGACY_SYSLOG"
 
-	FluentdImageEnvVar = "FLUENTD_IMAGE"
-	CertEventName      = "cluster-logging-certs-generate"
+	FluentdImageEnvVar        = "FLUENTD_IMAGE"
+	LogfilesmetricImageEnvVar = "LOGFILEMETRICEXPORTER_IMAGE"
+	CertEventName             = "cluster-logging-certs-generate"
 )
 
 var ReconcileForGlobalProxyList = []string{FluentdTrustedCAName}
--- a/pkg/k8shandler/fluentd.go
+++ b/pkg/k8shandler/fluentd.go
@@ -35,6 +35,8 @@ const (
 	fluentdName              = "fluentd"
 	syslogName               = "syslog"
 	fluentdRequiredESVersion = "6"
+	logVolumeMountName       = "varlog"
+	logVolumePath            = "/var/log"
 )
 
 func (clusterRequest *ClusterLoggingRequest) removeFluentd() (err error) {
@@ -150,7 +152,7 @@ func (clusterRequest *ClusterLoggingRequest) reconcileFluentdServiceMonitor() er
 	logMetricExporterEndpoint := monitoringv1.Endpoint{
 		Port:   exporterPortName,
 		Path:   "/metrics",
-		Scheme: "http",
+		Scheme: "https",
 		TLSConfig: &monitoringv1.TLSConfig{
 			CAFile:     prometheusCAFile,
 			ServerName: fmt.Sprintf("%s.%s.svc", fluentdName, cluster.Namespace),
@@ -368,11 +370,23 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
 		}
 	}
 	fluentdContainer := NewContainer("fluentd", "fluentd", v1.PullIfNotPresent, *resources)
+	// deliberately not passing any resources for running the below container process, let it have cpu and memory as the process requires
+	exporterresources := &v1.ResourceRequirements{}
+
+	exporterContainer := NewContainer("logfilesmetricexporter", "logfilesmetricexporter", v1.PullIfNotPresent, *exporterresources)
 
 	fluentdContainer.Ports = []v1.ContainerPort{
 		{
 			Name:          metricsPortName,
 			ContainerPort: metricsPort,
+			Protocol:      v1.ProtocolTCP,
+		},
+	}
+
+	exporterContainer.Ports = []v1.ContainerPort{
+		{
+			Name:          exporterPortName,
+			ContainerPort: exporterPort,
 			Protocol:      v1.ProtocolTCP,
 		},
 	}
@@ -419,6 +433,13 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
 		{Name: "tmp", MountPath: "/tmp"},
 	}
 
+	exporterContainer.VolumeMounts = []v1.VolumeMount{
+		{Name: logVolumeMountName, MountPath: logVolumePath},
+		{Name: metricsVolumeName, MountPath: "/etc/fluent/metrics"},
+	}
+	// Setting up CMD for log-file-metric-exporter
+	exporterContainer.Command = []string{"/usr/local/bin/log-file-metric-exporter", "  -verbosity=2", " -dir=/var/log/containers", " -http=:2112", " -keyFile=/etc/fluent/metrics/tls.key", " -crtFile=/etc/fluent/metrics/tls.crt"}
+
 	// List of _unique_ output secret names, several outputs may use the same secret.
 	unique := sets.NewString()
 	for _, o := range pipelineSpec.Outputs {
@@ -452,6 +473,13 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
 		ReadOnlyRootFilesystem:   utils.GetBool(true),
 		AllowPrivilegeEscalation: utils.GetBool(false),
 	}
+	exporterContainer.SecurityContext = &v1.SecurityContext{
+		SELinuxOptions: &v1.SELinuxOptions{
+			Type: "spc_t",
+		},
+		ReadOnlyRootFilesystem:   utils.GetBool(true),
+		AllowPrivilegeEscalation: utils.GetBool(false),
+	}
 
 	tolerations := utils.AppendTolerations(
 		collectionSpec.Logs.FluentdSpec.Tolerations,
@@ -471,8 +499,9 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
 
 	fluentdPodSpec := NewPodSpec(
 		"logcollector",
-		[]v1.Container{fluentdContainer},
+		[]v1.Container{fluentdContainer, exporterContainer},
 		[]v1.Volume{
+			{Name: logVolumeMountName, VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: logVolumePath}}},
 			{Name: "varlogcontainers", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/var/log/containers"}}},
 			{Name: "varlogpods", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/var/log/pods"}}},
 			{Name: "varlogjournal", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/var/log/journal"}}},
--- a/pkg/utils/utils.go
+++ b/pkg/utils/utils.go
@@ -32,9 +32,10 @@ const (
 
 // COMPONENT_IMAGES are thee keys are based on the "container name" + "-{image,version}"
 var COMPONENT_IMAGES = map[string]string{
-	"curator":             "CURATOR_IMAGE",
-	constants.FluentdName: constants.FluentdImageEnvVar,
-	"kibana":              "KIBANA_IMAGE",
+	"curator":                            "CURATOR_IMAGE",
+	constants.FluentdName:                constants.FluentdImageEnvVar,
+	"kibana":                             "KIBANA_IMAGE",
+	constants.LogfilesmetricexporterName: constants.LogfilesmetricImageEnvVar,
 }
 
 // GetAnnotation returns the value of an annoation for a given key and true if the key was found
--- a/pkg/k8shandler/fluentd_test.go
+++ b/pkg/k8shandler/fluentd_test.go
@@ -45,8 +45,8 @@ func TestNewFluentdPodSpecWhenFieldsAreUndefined(t *testing.T) {
 	cluster := &logging.ClusterLogging{}
 	podSpec := newFluentdPodSpec(cluster, nil, logging.ClusterLogForwarderSpec{})
 
-	if len(podSpec.Containers) != 1 {
-		t.Error("Exp. there to be 1 fluentd container")
+	if len(podSpec.Containers) != 2 {
+		t.Error("Exp. there to be 2 fluentd container")
 	}
 
 	resources := podSpec.Containers[0].Resources
@@ -81,8 +81,8 @@ func TestNewFluentdPodSpecWhenResourcesAreDefined(t *testing.T) {
 	}
 	podSpec := newFluentdPodSpec(cluster, nil, logging.ClusterLogForwarderSpec{})
 
-	if len(podSpec.Containers) != 1 {
-		t.Error("Exp. there to be 1 fluentd container")
+	if len(podSpec.Containers) != 2 {
+		t.Error("Exp. there to be 2 fluentd container")
 	}
 
 	resources := podSpec.Containers[0].Resources
@@ -258,8 +258,8 @@ func TestNewFluentdPodSpecWhenProxyConfigExists(t *testing.T) {
 		},
 	}, logging.ClusterLogForwarderSpec{})
 
-	if len(podSpec.Containers) != 1 {
-		t.Error("Exp. there to be 1 fluentd container")
+	if len(podSpec.Containers) != 2 {
+		t.Error("Exp. there to be 2 fluentd container")
 	}
 
 	checkFluentdProxyEnvVar(t, podSpec, "HTTP_PROXY", httpproxy)
@@ -409,8 +409,8 @@ func TestNewFluentdPodWhenChunkLimitSizeExists(t *testing.T) {
 	}
 	podSpec := newFluentdPodSpec(cluster, nil, logging.ClusterLogForwarderSpec{})
 
-	if len(podSpec.Containers) != 1 {
-		t.Error("Exp. there to be 1 fluentd container")
+	if len(podSpec.Containers) != 2 {
+		t.Error("Exp. there to be 2 fluentd container")
 	}
 
 	checkFluentdForwarderEnvVar(t, podSpec, "BUFFER_SIZE_LIMIT", strconv.FormatInt(chunkLimitSize.Value(), 10))


jcantrill (commented):
/hold

internal/cmd/go-symlink-logfiles-watcher/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,132 @@
+package main
```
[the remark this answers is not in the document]
alanconway: +1
@blockloop can you weigh in on the naming? You have a good familiarity with prometheus conventions.

internal/cmd/go-symlink-logfiles-watcher/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,132 @@
...
+	"github.com/openshift/cluster-logging-operator/internal/pkg/symnotify"
+        "github.com/prometheus/client_golang/prometheus"
+        "github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+
+
+
+
+var  debugOn bool = true
```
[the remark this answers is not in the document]
alanconway: I think golint and/or gofmt should enforce that, are they enabled for the internal code?

internal/cmd/go-symlink-logfiles-watcher/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,132 @@
...
+                return err
+        }
+        stat, err := os.Stat(path)
+        if err != nil {
+                return err
+        }
+        if stat.IsDir() {
+                return nil // Ignore directories
+        }
+        lastSize, size := float64(w.sizes[path]), float64(stat.Size())
```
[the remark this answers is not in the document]
alanconway: Yep, missing array values return as Zero values for the type, in this case 0.0.

internal/cmd/go-symlink-logfiles-watcher/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,132 @@
...
+	debug("logfilespathname= %v",dir)
+	debug("containernames= %v",containernames)
+	debug("debug option= %v",debugOn)
+	debug("listening port address= %v",listeningport)
+
+
+	//Get new watcher
+        w := &FileWatcher {
+                metrics: prometheus.NewCounterVec(prometheus.CounterOpts{
+                        Name: "gofilewatcher_input_status_total_bytes_logged",
```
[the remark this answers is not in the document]
alanconway: Another one for @blockloop to weigh in. There are some specific promtail conventions about the order of bytes, total etc. The fluentd metric name also should be consistent wit hthis.

internal/cmd/go-symlink-logfiles-watcher/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,132 @@
...
+                        Help: "gofilewatcher total bytes logged to disk (log file) ",
+                }, []string{"path"}),
+               sizes: make(map[string]float64),
+               added: make(map[string]bool),
+
+        }
+	prometheus.Register(w.metrics)
+	
+	defer prometheus.Register(w.metrics)
```
alanconway, line 117:
```suggestion
	defer prometheus.Unregister(w.metrics)
```

internal/go-symlink-logfiles-watcher/Docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM openshift/origin-release:golang-1.15 AS build
```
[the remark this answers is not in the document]
alanconway: Do we want a single image with two entrypoints, or two images? I was assuming two images, since we will deploy this as part of the fluentd DaemonSet not with the CLO.

internal/go-symlink-logfiles-watcher/Docker/go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/openshift/cluster-logging-operator/internal
```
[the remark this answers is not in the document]
alanconway: +1 share dependencies to avoid bloating the repo with multiple copies of vendor packages.,

internal/go-symlink-logfiles-watcher/Makefile
```diff
@@ -0,0 +1,9 @@
+all: build
```
alanconway, line 1: Don't need a separate makefile, the Dockerfile (whichever one ends up building this) can just use `go build` and do file copies as needed.

alanconway (changes requested):

internal/cmd/go-symlink-logfiles-watcher/symlink-logfiles-watcher.go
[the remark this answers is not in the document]
alanconway: Based on https://prometheus.io/docs/instrumenting/exporters/ this should be `log_exporter`

alanconway:
Final re-naming!

Executable: `log_exporter`

Metrics:
-    log_logged_bytes_total: total number of bytes written to a single log file path, accounting for rotations.
-    log_collected_bytes_total: total number of bytes collected from a single log file.
-    Both metrics have the label "path" which is the absolute path to the log file.

internal/cmd/go-symlink-logfiles-watcher/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,132 @@
...
+	//directory to be watched out where symlinks to all logs files are present e.g. /var/log/containers/
+	//debug option true or false
+	//listening port where this go-app push prometheus registered metrics for further collected or reading by end prometheus server
+	flag.StringVar(&dir, "logfilespathname", "/var/log/containers/", "Give the dirname where logfiles are going to be located, default /var/log/containers/")
+	flag.StringVar(&containernames,"containernames","log-stress","Given container names e.g. xxx yyy zzz only their log files are followed default is low-stress")
+	flag.BoolVar(&debugOn, "debug", false, "Give debug option false or true, default set to true")
+	flag.StringVar(&listeningport, "listeningport", ":2112", "Give the listening port where metrics can be exposed to and listened by a running prometheus server, default is :2112")
+	flag.Parse()
+
+	debug("logfilespathname= %v",dir)
```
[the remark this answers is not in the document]
pmoogi-redhat: done

internal/logs-files-metric-exporter/Docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM openshift/origin-release:golang-1.15 AS build
```
jcantrill, line 1: remove me.  not required.  Likely you will need to make revisions to the Makefile and root Dockerfile

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
+package main
```
alanconway, line 1: Final rename! internal/cmd/log-exporter/main.go

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+        var dir string
+        var listeningport string
+
+	//directory to be watched out where symlinks to all logs files are present e.g. /var/log/containers/
+	//debug option true or false
+	//listening port where this go-app push prometheus registered metrics for further collected or reading by end prometheus server
+	flag.StringVar(&dir, "logfilespathname", "/var/log/containers/", "Give the dirname where logfiles are going to be located, default /var/log/containers/")
+	flag.StringVar(&containernames,"containernames","log-stress","Given container names e.g. xxx yyy zzz only their log files are followed default is low-stress")
+	flag.BoolVar(&debugOn, "debug", false, "Give debug option false or true, default set to true")
+	flag.StringVar(&listeningport, "listeningport", ":2112", "Give the listening port where metrics can be exposed to and listened by a running prometheus server, default is :2112")
```
alanconway, line 95:
```suggestion
	flag.StringVar(&dir, "dir", "/var/log/containers/", "Diretory containing log files")
	flag.BoolVar(&debugOn, "debug", false, "Enable debug messages")
	flag.StringVar(&addr, "http", ":2112", "HTTP service address where metrics are exposed")
```
Note:
- you don't need to mention the default value in the message, flag.Parse() will add it automatically.
- containernames isn't required users can do a pattern match query on the "path" label to select logs.
- the -http flag is a convention for Go utilities that expose a http service, e.g. see `godoc -h` It is not strictly a port, since it can also include a host name for selective listening.

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
jcantrill, line 1: @alanconway please chime in on the name.  It still seems a little off with multiple pluralization.  Should it be ?:
* log-file-metric-exporter
* log-files-metric-exporter

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+	
+	defer prometheus.Unregister(w.metrics)
+
+	symwatcher, err := symnotify.NewWatcher()
+	w.watcher = symwatcher
+	
+	if err != nil {
+		log.Error(err, "NewFileWatcher error")
+		os.Exit(1)
+	}
```
alanconway, line 124:
```suggestion
	}
	defer w.watcher.Close()
```

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/openshift/cluster-logging-operator/internal/pkg/symnotify"
+        "github.com/prometheus/client_golang/prometheus"
+        "github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+
+
+
+var (
+debugOn bool = true
```
jcantrill, line 17: I presume we can remove this default value since the flags will set it accordingly

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+        var add float64
+        if size > lastSize {
+                // File has grown, add the difference to the counter.
+                add = size - lastSize
+        } else if size < lastSize {
+                // File truncated, starting over. Add the size.
+                add = size
+        }
+	if debugOn { 
+        log.Info("For logfile in...", "path",path,"lastsize",lastSize,"currentsize",size,"addedbytes",add)
```
jcantrill, line 55: Remove debugOn block: the log library allows you to do something like `log.V(2).Info()` to ONLY log at severity 2 (e.g. debug)

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+         if debugOn { 
+          log.Info("Watching out logfiles dir ...", "dir",dir,"containersmatching",containernames,"debug",debugOn,"listeningport",listeningport)
+         } 
+
+
+	//Get new watcher
+        w := &FileWatcher {
+                metrics: prometheus.NewCounterVec(prometheus.CounterOpts{
+                        Name: "logs_files_metric_exporter_input_status_bytes_logged_total",
+                        Help: "logs-files-metric-exporter total bytes logged to disk (log file) ",
```
alanconway, line 108:
```suggestion
                        Name: "log_logged_bytes_total",
                        Help: "Total number of bytes written to a single log file path, accounting for rotations",
```

alanconway (changes requested):
Just a few minor renaming diffs now that we've decided on the metric names finally. Otherwise this is good to go.

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+		//For the cases new log files added, old files moved, old files deleted, you need to add/remove them from watcher as whole dir added to the watcher
+		//For new log files added write event is not getting issued
+		
+                e, err := w.watcher.Event()
+                if err != nil {
+		log.Error(err, "Watcher.Event returning err")
+		os.Exit(1)
+		}
+
+	        if debugOn {
```
jcantrill, line 77: Same here.  switch not needed

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+func main() {
+        var dir string
+        var listeningport string
+
+	//directory to be watched out where symlinks to all logs files are present e.g. /var/log/containers/
+	//debug option true or false
+	//listening port where this go-app push prometheus registered metrics for further collected or reading by end prometheus server
+	flag.StringVar(&dir, "logfilespathname", "/var/log/containers/", "Give the dirname where logfiles are going to be located, default /var/log/containers/")
+	flag.StringVar(&containernames,"containernames","log-stress","Given container names e.g. xxx yyy zzz only their log files are followed default is low-stress")
+	flag.BoolVar(&debugOn, "debug", false, "Give debug option false or true, default set to true")
```
jcantrill, line 94: This should be a numeric severity level we can pass to the logger.  Please see the CLO main

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+
+
+         if debugOn { 
+          log.Info("Watching out logfiles dir ...", "dir",dir,"containersmatching",containernames,"debug",debugOn,"listeningport",listeningport)
+         } 
+
+
+	//Get new watcher
+        w := &FileWatcher {
+                metrics: prometheus.NewCounterVec(prometheus.CounterOpts{
```
jcantrill, line 106: @alanconway speaking with @sichvoge he mentioned we should also expose the: podname, namespace.  Fluentd has a regex where we can grab this info along with container hash from the filename.  Open question here is if the file we are monitoring is the same fail.

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+                        Name: "logs_files_metric_exporter_input_status_bytes_logged_total",
```
jcantrill, line 107: Reminder that @alanconway is still working to lock down this metric nam

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+                }, []string{"path"}),
+               sizes: make(map[string]float64),
+               added: make(map[string]bool),
+
+        }
+	prometheus.Register(w.metrics)
+	
+	defer prometheus.Unregister(w.metrics)
+
+	symwatcher, err := symnotify.NewWatcher()
```
jcantrill, line 118: Is there any reason we can not do this before creating `FileWatcher` and initialize FileWatcher struct with the return value?

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+                add = size
+        }
+	if debugOn { 
+        log.Info("For logfile in...", "path",path,"lastsize",lastSize,"currentsize",size,"addedbytes",add)
+        } 
+        counter.Add(add)
+        return nil
+}
+
+func (w FileWatcher) Watch( ) {
```
jcantrill, line 61: shouldnt this be a pointer `(w FileWatcher)`

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+        "github.com/prometheus/client_golang/prometheus"
+        "github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+
+
+
+var (
+debugOn bool = true
+containernames string = ""
```
jcantrill, line 18: This looks like an unused variable outside of binary flags

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+        stat, err := os.Stat(path)
+        if err != nil {
+                return err
+        }
+        if stat.IsDir() {
+                return nil // Ignore directories
+        }
+        lastSize, size := float64(w.sizes[path]), float64(stat.Size())
+        w.sizes[path] = size
+        var add float64
```
jcantrill, line 46: I think this would be better named as 'change' or 'diff'

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
```diff
@@ -0,0 +1,131 @@
...
+		//All logfiles with containername are added to the watcher
+		//write event for these logfiles are being watched
+		//create event gets issued for all new logfiles appear under logfilepathname /var/log/containers/
+		//For the cases new log files added, old files moved, old files deleted, you need to add/remove them from watcher as whole dir added to the watcher
+		//For new log files added write event is not getting issued
+		
+                e, err := w.watcher.Event()
+                if err != nil {
+		log.Error(err, "Watcher.Event returning err")
+		os.Exit(1)
```
jcantrill, line 74: Is this truly a hard exit condition?  Is it possible there are some errors we can log and simply continue?

internal/pkg/symnotify/symnotify.go
```diff
@@ -0,0 +1,112 @@
...
+	"path/filepath"
+	"time"
+	"github.com/fsnotify/fsnotify"
+)
+
+type Event = fsnotify.Event
+type Op = fsnotify.Op
+
+var (
+debugOn bool = true
```
jcantrill, line 20: should be able to remove

internal/pkg/symnotify/symnotify.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+const (
+	Create Op = fsnotify.Create
```
jcantrill, line 24: Does this Compile with the space in `Create Op`

internal/pkg/symnotify/symnotify.go
```diff
@@ -0,0 +1,112 @@
...
+	case e, ok = <-w.watcher.Events:
+	case err, ok = <-w.watcher.Errors:
+	case <-time.After(timeout):
+		return Event{}, os.ErrDeadlineExceeded
+	}
+	switch {
+	case !ok:
+		return Event{}, io.EOF
+	case e.Op == Create:
+		log.Info("Create Event Detected for file.." ,"e.Name",e.Name)
```
jcantrill, line 60: probably want to set the verbosity to 2 or even 3 here `log.V(2).Info()`

internal/pkg/symnotify/symnotify.go
```diff
@@ -0,0 +1,112 @@
...
+		if info, err := os.Lstat(e.Name); err == nil {
+			if isSymlink(info) {
+				_ = w.watcher.Add(e.Name)
+			}
+		}
+	case e.Op == Remove:
+		log.Info("Create Event Detected for file.." ,"e.Name",e.Name)
```
jcantrill, line 67: same here

internal/pkg/symnotify/symnotify.go
```diff
@@ -0,0 +1,112 @@
...
+
+// Event returns the next event.
+func (w *Watcher) Event() (e Event, err error) {
+	return w.EventTimeout(time.Duration(math.MaxInt64))
+}
+
+// EventTimeout returns the next event or os.ErrDeadlineExceeded if timeout is exceeded.
+func (w *Watcher) EventTimeout(timeout time.Duration) (e Event, err error) {
+	var ok bool
+	select {
```
jcantrill, line 50: I'm confused regarding this entire block.  It seems we should run through all these things in a prescribed fashion and early exist or other as needed if there is a timeout or error

internal/pkg/symnotify/symnotify.go
```diff
@@ -0,0 +1,112 @@
...
+	case err, ok = <-w.watcher.Errors:
+	case <-time.After(timeout):
+		return Event{}, os.ErrDeadlineExceeded
+	}
+	switch {
+	case !ok:
+		return Event{}, io.EOF
+	case e.Op == Create:
+		log.Info("Create Event Detected for file.." ,"e.Name",e.Name)
+		if info, err := os.Lstat(e.Name); err == nil {
```
jcantrill, line 61: What if we have an error?  we should at least log it

internal/pkg/symnotify/symnotify.go
```diff
@@ -0,0 +1,112 @@
...
+			if isSymlink(info) {
+				_ = w.watcher.Add(e.Name)
```
jcantrill, line 63: Is the return value useful or significant here?

internal/pkg/symnotify/symnotify.go
```diff
@@ -0,0 +1,112 @@
...
+			}
+		}
+	case e.Op == Remove:
+		log.Info("Create Event Detected for file.." ,"e.Name",e.Name)
+		w.watcher.Remove(e.Name)
+		w.watcher.Add(e.Name)
```
jcantrill, line 69: Why do we need to re-add ?

internal/pkg/symnotify/symnotify.go
```diff
@@ -0,0 +1,112 @@
...
+func (w *Watcher) Add(name string) error {
+	if err := w.watcher.Add(name); err != nil {
+		return err
+	}
+
+	// Scan directories for existing symlinks, we wont' get a Create for those.
+	if infos, err := ioutil.ReadDir(name); err == nil {
+        for _, info := range infos {
+		if isSymlink(info) {
+		log.Info("Adding file to watcher ...","filename",filepath.Join(name, info.Name()))
```
jcantrill, line 93: verbosity

internal/pkg/symnotify/symnotify_test.go
```diff
@@ -0,0 +1,140 @@
...
+func (f *Fixture) Event() symnotify.Event {
+	f.T.Helper()
+	e, err := f.Watcher.EventTimeout(time.Second)
+	require.NoError(f.T, err)
+	return e
+}
+
+func TestWatchesRealFiles(t *testing.T) {
+	f := NewFixture(t)
+	assert, require := assert.New(t), require.New(t)
```
jcantrill, line 68: Can this be moved into the fixture to reduce code dup

jcantrill (commented):
please make sure to format before pushing changes `make fmt`

internal/pkg/symnotify/symnotify.go
alanconway: We could, but we should exclude "file not found" which is an expected race condition since the files can be deleted by other processes.

internal/pkg/symnotify/symnotify.go
alanconway: Same comment as above.

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
alanconway: +1 as per discussion we should use these labels:
_definitely_:  namspace, pod
_investigate_: container
_verify_: node/hostname - prometheus automatically adds a host label, so may not be needed. I must double check.

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
alanconway: Locked: `log_logged_bytes_total`

alanconway (changes requested):

internal/pkg/symnotify/symnotify.go
```diff
@@ -57,18 +54,18 @@ func (w *Watcher) EventTimeout(timeout time.Duration) (e Event, err error) {
...
 		if info, err := os.Lstat(e.Name); err == nil {
 			if isSymlink(info) {
 				_ = w.watcher.Add(e.Name)
 			}
 		}
 	case e.Op == Remove:
-		log.Info("Create Event Detected for file.." ,"e.Name",e.Name)
+		log.V(2).Info("Remove Event Detected for file.." ,"e.Name",e.Name)
 		w.watcher.Remove(e.Name)
 		w.watcher.Add(e.Name)
```
alanconway:
```suggestion
```

We don't need to handle Remove, the w.watcher already knows it's been removed (it told us) and there's no reason to re-add (it will fail)

pmoogi-redhat: @alanconway 
Without the below 
   case e.Op == Remove:
                log.V(2).Info("Remove Event Detected for file.." ,"e.Name",e.Name)
                w.watcher.Remove(e.Name)
                w.watcher.Add(e.Name)
I couldn't get metric values right. Will retest it by completely removing the above line of code.

pmoogi-redhat: @alanconway working now by removing both the below treatment on the occasion of detected Remove event.
w.watcher.Remove(e.Name)
w.watcher.Add(e.Name)

internal/log-file-metric-exporter/Docker/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM openshift/origin-release:golang-1.15 AS build
```
jcantrill, line 1: This file is not required at all.  We modify the existing Dockerfile to copy the binary output to the desired location in the image.
* change to depend on building the new binary: https://github.com/openshift/cluster-logging-operator/blob/master/Makefile#L68
* copy new binary to /usr/bin like https://github.com/openshift/cluster-logging-operator/blob/master/Dockerfile#L34
* add a newly configured container to: https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/fluentd.go#L451
* probably need to add an endpoint so we can scrape the service https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/fluentd.go#L154

internal/cmd/log-file-metric-exporter/main.go
```diff
@@ -0,0 +1,150 @@
...
+		_,filename := path.Split(e.Name)
+                filename = strings.TrimRight(filename,".log")
+                log.V(2).Info("file name after trimming","filename", filename)
+                filenameslice := strings.Split(filename,"_")
+		if (len(filenameslice) == 3) {
+		  podname = filenameslice[0]
+                  namespace = filenameslice[1]
+                  containername = filenameslice[2]
+	        } else {
+	          namespace="unknown"
```
jcantrill, line 91: I don't see how this will ever be the case since the file format is a known pattern that has existed since the inception of kubernetes.  If we don't match this means:
* there is an error in the logic
* the spec changed and we didn't make adjustments
The only open question I have is container name which based on your analysis is there.  I'm hoping it is always there.

internal/cmd/log-file-metric-exporter/main.go
```diff
@@ -0,0 +1,150 @@
...
+		
+                e, err := w.watcher.Event()
+                if err != nil {
+		log.Error(err, "Watcher.Event returning err")
+		os.Exit(1)
+		}
+
+                log.V(2).Info("Events notified for...", "e.Name",e.Name,"Event",e.Op)
+
+		//Get namespace, podname, containername from e.Name - log file path
```
jcantrill, line 80: We should be able to do this with a compiled regex.   ref: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/blob/master/lib/fluent/plugin/filter_kubernetes_metadata.rb#L56

internal/cmd/log-file-metric-exporter/main.go
```diff
@@ -0,0 +1,150 @@
...
+        "github.com/prometheus/client_golang/prometheus"
+        "github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+
+
+
+var (
+ verbosity int = 0
+ namespace string = "unknown"
```
jcantrill, line 20: I don't believe these are needed

internal/pkg/symnotify/symnotify.go
jcantrill: I would at a minimum log the error at a higher verbosity for trouble shooting

internal/pkg/symnotify/symnotify.go
jcantrill: Same here... log at a higher verbosity as we don't know what useful information it could provide

internal/pkg/symnotify/symnotify.go
```diff
@@ -0,0 +1,107 @@
...
+		return Event{}, io.EOF
+	case e.Op == Create:
+		log.V(2).Info("Create Event Detected for file.." ,"e.Name",e.Name)
+		if info, err := os.Lstat(e.Name); err == nil {
+			if isSymlink(info) {
+				_ = w.watcher.Add(e.Name)
+			}
+		}
+	case e.Op == Remove:
+		log.V(2).Info("Remove Event Detected for file.." ,"e.Name",e.Name)
```
jcantrill, line 64: Should we not actually remove the entry to keep the map from growing?

internal/pkg/symnotify/symnotify.go
```diff
@@ -0,0 +1,107 @@
...
+	if err := w.watcher.Add(name); err != nil {
+		return err
+	}
+
+	// Scan directories for existing symlinks, we wont' get a Create for those.
+	if infos, err := ioutil.ReadDir(name); err == nil {
+        for _, info := range infos {
+		if isSymlink(info) {
+		log.V(2).Info("Adding file to watcher ...","filename",filepath.Join(name, info.Name()))
+				_ = w.watcher.Add(filepath.Join(name, info.Name()))
```
jcantrill, line 89: Log at higher verbosity

jcantrill (changes requested):
Please run `make fmt` before commiting

internal/pkg/symnotify/symnotify.go
pmoogi-redhat: @jcantrill changing it to '3' level of verbosity - is that fine ?

internal/pkg/symnotify/symnotify.go
pmoogi-redhat: @jcantrill   Removing the path on the issue of Remove event didn't work as per expected results.  Metric counter was not showing increasing bytes. If we don't do  anything, then i found it is working as per expected results.

internal/cmd/log-file-metric-exporter/main.go
pmoogi-redhat: @jcantrill we need to initialise these variables somewhere right ?

internal/cmd/log-file-metric-exporter/main.go
pmoogi-redhat: @jcantrill  are you suggesting equivalent regex based parsing to be written in Golang ?

internal/log-file-metric-exporter/Docker/Dockerfile
pmoogi-redhat: @jcantrill  first change in Makefile will allow us to compile binary for log-file-metric-exporter, second change will copy this compiler binary to /usr/bin in CLO container. i am not able to comprehend how this gets executed part of CLO process. The third change - podspec for fluentd - "{Name: "entrypoint", VolumeSource: v1.VolumeSource{ConfigMap: &v1.ConfigMapVolumeSource{LocalObjectReference: v1.LocalObjectReference{Name: "fluentd"}}}}," , do we need to add one more entry point for log-file-metric-exported ?  fourth change also i need to better understand for exposing log_logged_bytes_total from log-file-metric-exporter.

internal/cmd/logs-files-metric-exporter/symlink-logfiles-watcher.go
pmoogi-redhat: @alanconway should os.Exit(1) be removed from error handling ? please guide further.

internal/cmd/log-file-metric-exporter/main.go
pmoogi-redhat: @jcantrill do we have situations where workloads dumping some data to /var/log/containers/  directly using their own filename conventions ? @alanconway kindly guide me here..

internal/pkg/symnotify/symnotify_test.go
pmoogi-redhat: @alanconway seeking your advice here.

internal/cmd/log-file-metric-exporter/main.go
jcantrill: yes. instead of string manipulation

internal/cmd/log-file-metric-exporter/main.go
jcantrill: no. we should outright  do nothing for files that do not match the pattern.  the pattern is the API we have from kubernetes

internal/cmd/log-file-metric-exporter/main.go
jcantrill: i dont think we need them at all because they are values that will always be known from the file name pattern

internal/pkg/symnotify/symnotify.go
jcantrill: but do we end up with a memory link for a map that grows without bounds?

internal/pkg/symnotify/symnotify.go
jcantrill: I was refering to the return code of "Add"... it may not be useful on the surface but logging it at a higher verbosity at least give the capability to inspect it if needed

alanconway:
/approve
I'll let @jcantrill do the LGTM since he's raised a couple more points.

internal/cmd/log-file-metric-exporter/main.go
```diff
@@ -0,0 +1,153 @@
...
+
+                log.V(2).Info("Events notified for...", "e.Name",e.Name,"Event",e.Op)
+
+		//Get namespace, podname, containername from e.Name - log file path
+		//filename := strings.Trim(e.Name,"/var/log/containers/")
+
+		_,filename := path.Split(e.Name)
+                filename = strings.TrimRight(filename,".log")
+                log.V(2).Info("file name after trimming","filename", filename)
+		submatches := kubernetesregexpCompiled.FindAllString(e.Name, -1)
```
jcantrill, line 89: I'm missing what is happening here.  Utilizing regex, you should be able to write a set of capture groups to get all the needed information without  needing to do any string manipulation.  This looks like you are attempting to combine both string manipulation and regex which negates using one or the other

internal/pkg/symnotify/symnotify.go
jcantrill: I dont see we have an answer to this questino

internal/pkg/symnotify/symnotify.go
jcantrill: same here.  what is the verdict.  seems like it would be useful to log that return at some higher verbosity level

jcantrill (commented):
Please run `make fmt` before sumitting

internal/pkg/symnotify/symnotify.go
pmoogi-redhat: @alanconway @jcantrill the below is found not yielding expected results ..  
               case e.Op == Remove:
               log.V(2).Info("Remove Event Detected for file.." ,"e.Name",e.Name)
               w.watcher.Remove(e.Name)
               
I don't know the answers to it... why..

internal/cmd/log-file-metric-exporter/main.go
```diff
@@ -0,0 +1,164 @@
...
+		//podname := filenameslice[0]
+		//namespace := filenameslice[1]
+		//containername := filenameslice[2]
+                //log.V(2).Info("Namespace podname containername...","namespace", namespace, "podname", podname, "containername", containername)
+                for i, n := range r2 {
+                     md[n1[i]] = n
+                }
+		podname := md["pod_name"]
+		namespace := md["namespace"]
+		containername := md["container_name"]
```
alanconway, line 107: Using numbered submatches rather than names:

```suggestion
                md := kubernetesregexpCompiled.FindStringSubmatch(e.Name)
		if len(md != MatchLen {
				log.V(2).Info("filename doesn't conform with k8 logfile path name ...","filename",e.Name)
		} else {
		podname := md[PodNameIndex]
		namespace := md[Namespace]
		containername := md[ContainerNameIndex]
```

alanconway (changes requested):
Made some suggestions about handling the regexp, otherwise looks good.

Dockerfile
```diff
@@ -3,6 +3,7 @@ WORKDIR /go/src/github.com/openshift/cluster-logging-operator
 
 # COPY steps are in the reverse order of frequency of change
 COPY cmd ./cmd
+COPY internal ./internal
```
jcantrill, line 6: will need to move to Dockerfile.in if  https://github.com/openshift/cluster-logging-operator/pull/992  lands first

internal/cmd/log-file-metric-exporter/main.go
```diff
@@ -0,0 +1,164 @@
...
+		
+                e, err := w.watcher.Event()
+                if err != nil {
+		log.Error(err, "Watcher.Event returning err")
+		os.Exit(1)
+		}
+
+                log.V(2).Info("Events notified for...", "e.Name",e.Name,"Event",e.Op)
+
+		//Get namespace, podname, containername from e.Name - log file path
```
jcantrill, line 83: Remove me if I'm commented code

internal/cmd/log-file-metric-exporter/main.go
```diff
@@ -0,0 +1,164 @@
...
+                //filename = strings.TrimRight(filename,".log")
+                //log.V(2).Info("file name after trimming","filename", filename)
+		//submatches := kubernetesregexpCompiled.FindAllString(e.Name, -1)
+
+                r2 := kubernetesregexpCompiled.FindStringSubmatch(e.Name)
+                n1 := kubernetesregexpCompiled.SubexpNames()
+
+		//if submatches == nil {
+		if r2 == nil {
+		log.V(2).Info("filename doesn't conform with k8 logfile path name ...","filename",e.Name)
```
jcantrill, line 95: Should this force skipping any additional evaluation as well, like with a `continue`.  If the file path does not match then we don't know which pod or namespace for association

internal/cmd/log-file-metric-exporter/main.go
```diff
@@ -0,0 +1,164 @@
...
+		} else {
+                //filenameslice := strings.Split(filename,"_")
```
jcantrill, line 97: remove me

pkg/k8shandler/fluentd.go
```diff
@@ -142,6 +149,16 @@ func (clusterRequest *ClusterLoggingRequest) reconcileFluentdServiceMonitor() er
 			// ServerName can be e.g. fluentd.openshift-logging.svc
 		},
 	}
+	logMetricExporterEndpoint := monitoringv1.Endpoint{
+		Port:   exporterPortName,
+		Path:   "/metrics",
+		Scheme: "http",
+		TLSConfig: &monitoringv1.TLSConfig{
```
jcantrill, line 156: Is it valid to configure a CA when scraping with a scheme of http?  Also, the endpoint started in main.go does not load any certs.  There seems like a mismatch here.  For fluentd, we actually mount certs provided by the certgen service.  @alanconway thoughts?

pkg/k8shandler/fluentd.go
```diff
@@ -337,6 +354,26 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateFluentdSecret() error
 	return nil
 }
 
+func newLogMetricExporterContainer() v1.Container {
+	resources := v1.ResourceRequirements{}
```
jcantrill, line 358: @alanconway I left this blank when I provided the PR, but how should we pass in resources here.  I'm not fond of opening up the API, but maybe we need to allow a backdoor for setting mem and cpu, maybe via annotations?

Dockerfile
pmoogi-redhat: I don't find Dockerfile.in in this repo. let me know if its belong to other internal repo..

internal/cmd/log-file-metric-exporter/main.go
pmoogi-redhat: if file doesn't conform with k8 filename conventions.. don't use extra labels for publishing underlying metrics. i shouldn't affect rest of the working.

pmoogi-redhat (commented):
@alanconway incorporated your regex based parsing k8 filename suggestion. please see.

pkg/k8shandler/fluentd.go
alanconway: Any cert files should be passed as command line argumets. The question of where they come from and what is in them is outside the scope of this program. We should follow the same pattern that is used in fluentd for the prometheus plugin.

pkg/k8shandler/fluentd.go
alanconway: We could add a "MetricExporterSpec" to the API, I don't think that would be too bad. The API already has componet-specific requirements (FluentdSpec) so it's not making it any worse. 
Alternatively we could split the Fluentd requirements - e.g. give 90% to fluentd and 10% to the exporter. We'd need to do some benchmarks to make sure that's a reasonable split but we should be doing CPU/Mem benehmarks anyway so it's useful work.
The backdoor approach means our collector is using resources that the user doesn't know about, which is not ideal.

alanconway: I would lean towards just adding to the API. This is something to clean up in the "API v2" effort when we get smarter about scaling and resource use, but it's not worse than whats already there.

pkg/k8shandler/fluentd.go
jcantrill: @alanconway I'm asking specifically these questions:
* is it valid to stand up a port and define certs but also use scheme `http`.  My expectation is it should be `https`
* we should pass certs as a cli arg as you mention if we need them.  I think they can be the same certs fluent uses
* If we need https then the socket needs to be set up for that later in main.go

pkg/k8shandler/fluentd.go
jcantrill: @pmoogi-redhat this means we need to add soming like https://github.com/openshift/cluster-logging-operator/blob/master/pkg/apis/logging/v1/clusterlogging_types.go#L35  with the resource requirements.  Make sure it additionally includes the comments as these are used for the code generation

alanconway:
@pmoogi-redhat This all  looks good. Add unit tests for `FileWatcher`, including the file name parsing and I'll LGTM.
Go has very easy-to-use test coverage tools to check your coverage, see https://blog.golang.org/cover

Dockerfile
jcantrill: yes #992 pulls that file in.  if you rebase on upstream/master you will be able to make the needed changes

jcantrill:
@pmoogi-redhat please rebase on upstream/master and squash all the changes into a single commit for final review

pmoogi-redhat:
> @pmoogi-redhat please rebase on upstream/master and squash all the changes into a single commit for final review

@jcantrill rebased it with upstream/master, made changes in Dockerfile.in to include copying of internal/. Kindly review.

go.mod
```diff
@@ -1,13 +1,14 @@
 module github.com/openshift/cluster-logging-operator
 
-go 1.14
+go 1.15
```
jcantrill, line 3: @bparees any idea if this breaks our CI if the base image is not also1.15?

manifests/5.1/cluster-logging.v5.1.0.clusterserviceversion.yaml
```diff
@@ -327,6 +327,8 @@ spec:
                     value: "quay.io/openshift/origin-logging-fluentd:latest"
                   - name: CURATOR_IMAGE
                     value: "quay.io/openshift/origin-logging-curator5:latest"
+                  - name: IMAGE_LOG_METRIC_EXPORTER
+                    value: quay.io/openshift/origin-cluster-logging-operator:latest
```
jcantrill, line 331: changes are going into 5.2 so this will need to be added to that file instead.. but i guess it will depend if https://github.com/openshift/cluster-logging-operator/pull/1007 lands first.  you might not need to do anything

pkg/constants/constants.go
```diff
@@ -16,17 +16,17 @@ const (
...
-	ElasticsearchPort   = "9200"
-	FluentdName         = "fluentd"
-	KibanaName          = "kibana"
-	KibanaProxyName     = "kibana-proxy"
-	CuratorName         = "curator"
-	LogStoreURL         = "https://" + ElasticsearchFQDN + ":" + ElasticsearchPort
-	MasterCASecretName  = "master-certs"
-	CollectorSecretName = "fluentd"
-
+	ElasticsearchFQDN     = "elasticsearch.openshift-logging.svc.cluster.local"
```
jcantrill, line 19: revert this line.  looks like a bad merge.  it was recently removed

jcantrill (changes requested):
You still have 2 commits so the rebase looks incorrect, the first of which has garbage in it from a merge diff.  Please try rebasing again and squashing commits

pkg/k8shandler/fluentd.go
pmoogi-redhat: @jcantrill @alanconway added changes to support LogFileMetricExportedSpec - please review my changes done in 	 
        modified:   apis/logging/v1/clusterlogging_types.go
	modified:   k8shandler/defaults.go
	modified:   k8shandler/fluentd.go
Hoping above is sufficient for specifying individually resources for LogFileMetricExporter.

pmoogi-redhat: @vimalk78 please review as well

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -192,6 +193,20 @@ type FluentdSpec struct {
...
+        //
+        // +nullable
+        // +optional
+        Resources *v1.ResourceRequirements `json:"resources"`
+
+        // Define which Nodes the Pods are scheduled on.
+        //
+        // +nullable
+        NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+        Tolerations  []v1.Toleration   `json:"tolerations,omitempty"`
```
vimalk78, line 207: The `FluentdSpec` is already contains these fields, as these fields are required at pod level
So no need to duplicate them here, if `LogFileMetricExporterSpec` is only for container resources

pkg/constants/constants.go
```diff
@@ -16,17 +16,17 @@ const (
...
-	CollectorSecretName = "fluentd"
-
+	ElasticsearchFQDN     = "elasticsearch.openshift-logging.svc"
+	ElasticsearchName     = "elasticsearch"
+	ElasticsearchPort     = "9200"
+	FluentdName           = "fluentd"
+	KibanaName            = "kibana"
+	KibanaProxyName       = "kibana-proxy"
+	CuratorName           = "curator"
+	LogMetricExporterName = "exporter"
```
vimalk78, line 26: rename to `LogFileMetricExporterName`

pkg/k8shandler/fluentd.go
```diff
@@ -446,11 +498,13 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
 		},
 	)
 
+	exporterContainer := newLogMetricExporterContainer()
+
 	fluentdPodSpec := NewPodSpec(
 		"logcollector",
-		[]v1.Container{fluentdContainer},
+		[]v1.Container{fluentdContainer, exporterContainer},
```
vimalk78, line 505: is it possible to disable this feature? or is it always enabled from now on

pkg/utils/utils.go
```diff
@@ -32,9 +32,10 @@ const (
 
 // COMPONENT_IMAGES are thee keys are based on the "container name" + "-{image,version}"
 var COMPONENT_IMAGES = map[string]string{
-	"curator":             "CURATOR_IMAGE",
-	constants.FluentdName: constants.FluentdImageEnvVar,
-	"kibana":              "KIBANA_IMAGE",
+	"curator":                       "CURATOR_IMAGE",
+	constants.FluentdName:           constants.FluentdImageEnvVar,
+	"kibana":                        "KIBANA_IMAGE",
+	constants.LogMetricExporterName: "IMAGE_LOG_METRIC_EXPORTER",
```
vimalk78, line 38: inconsistent naming. somewhere it is `LogFileMetric` somewhere it is `LogMetric`

vimalk78, line 38: other image names end in `_IMAGE` , this one starts with `IMAGE_`

internal/cmd/log-file-metric-exporter/main.go
```diff
@@ -0,0 +1,140 @@
...
+
+const (
+	PodNameIndex = iota + 1
+	NamespaceIndex
+	ContainerNameIndex
+	DockerIndex
+	MatchLen
+)
+
+type FileWatcher struct {
```
vimalk78, line 29: The filename is `main.go` , may be we can move this struct in a different file/package? with a package level method `NewFileWatcher()` instead of creating it in main?

pkg/k8shandler/fluentd.go
```diff
@@ -337,6 +354,41 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateFluentdSecret() error
...
+                        Requests: v1.ResourceList{
+                                v1.ResourceMemory: defaultLogFileMetricExporterMemory,
+                                v1.ResourceCPU:    defaultLogFileMetricExporterCpuRequest,
+                        },
+                }
+        }
+
+
+	container := NewContainer(constants.LogMetricExporterName, constants.LogMetricExporterName, v1.PullIfNotPresent, *resources)
+	container.Command = []string{"log-file-metric-exporter"}
```
vimalk78, line 375: this command takes arguments right? why not make it clear that we are only watching application logs by giving an argument `--dir /var/log/containers` 
Since we are mounting complete `/var/log` in the container, it is not clear what are we watching for. giving argument makes it clear, even though that is the default argument

vimalk78: if we want to watch only app logs, then mount the `/var/loc/containers` to be more explicit about it.

manifests/5.2/cluster-logging.clusterserviceversion.yaml
```diff
@@ -325,6 +325,10 @@ spec:
                     value: "cluster-logging-operator"
                   - name: FLUENTD_IMAGE
                     value: "quay.io/openshift/origin-logging-fluentd:latest"
+                  - name: CURATOR_IMAGE
```
jcantrill, line 328: Sorry I broke you here.  @vimalk78 asked me to move the file name so will need a rebase once https://github.com/openshift/cluster-logging-operator/pull/1013 merges

pkg/k8shandler/defaults.go
```diff
@@ -19,4 +19,7 @@ var (
 
 	defaultFluentdMemory     resource.Quantity = resource.MustParse("736Mi")
 	defaultFluentdCpuRequest resource.Quantity = resource.MustParse("100m")
+	
+	defaultLogFileMetricExporterMemory     resource.Quantity = resource.MustParse("736Mi")
```
jcantrill, line 23: this is too much for this go based container and likely does not require any defaults.  I would remove and only apply what a user specs if needed

pkg/k8shandler/fluentd.go
```diff
@@ -337,6 +354,41 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateFluentdSecret() error
 	return nil
 }
 
+func newLogMetricExporterContainer() v1.Container {
+	collectionSpec := logging.CollectionSpec{}
+        if cluster.Spec.Collection != nil {
+                collectionSpec = *cluster.Spec.Collection
+        }
+	resources := collectionSpec.Logs.LogFileMetricExporterSpec.Resources 
+        if resources == nil {
```
jcantrill, line 363: companion to comment above.  no need to apply defaults

pkg/constants/constants.go
pmoogi-redhat: resolved

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -192,6 +193,20 @@ type FluentdSpec struct {
...
+	// The resource requirements for Fluentd
+	//
+	// +nullable
+	// +optional
+	Resources *v1.ResourceRequirements `json:"resources"`
+
+	// Define which Nodes the Pods are scheduled on.
+	//
+	// +nullable
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
jcantrill, line 206: Remove me.  this container will be deployed with fluentd and will be bound to what is defined there

pkg/apis/logging/v1/clusterlogging_types.go
```diff
@@ -192,6 +193,20 @@ type FluentdSpec struct {
...
+	Tolerations  []v1.Toleration   `json:"tolerations,omitempty"`
```
jcantrill, line 207: same as above

pkg/k8shandler/fluentd.go
```diff
@@ -337,6 +354,41 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateFluentdSecret() error
 	return nil
 }
 
+func newLogMetricExporterContainer(cluster *logging.ClusterLogging) v1.Container {
+	collectionSpec := logging.CollectionSpec{}
+	if cluster.Spec.Collection != nil {
+		collectionSpec = *cluster.Spec.Collection
+	}
+	resources := collectionSpec.Logs.LogMetricExporterSpec.Resources
+	if resources == nil {
```
jcantrill, line 363: remove me.  we will not default resources for the container if they are not defined

pkg/k8shandler/fluentd.go
```diff
@@ -337,6 +354,41 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateFluentdSecret() error
 	return nil
 }
 
+func newLogMetricExporterContainer(cluster *logging.ClusterLogging) v1.Container {
+	collectionSpec := logging.CollectionSpec{}
```
jcantrill, line 358: remove me.  this is unused and does not make sense in context of the container for the exporter

pkg/apis/logging/v1/clusterlogging_types.go
pmoogi-redhat: resolved

pkg/k8shandler/fluentd.go
pmoogi-redhat: @jcantrill @alanconway reverted back log-file-metric-exporter from internal\pkg, internal\cmd, taken out resource spec defined specifically for this exporter. Repo is 99% established in here for building , testing log-file-metric-exporter as an independent component - https://github.com/ViaQ/log-file-metric-exporter/tree/master. kindly review all my changes in this PR.

manifests/5.2/cluster-logging.v5.2.0.clusterserviceversion.yaml
```diff
@@ -325,6 +325,10 @@ spec:
                     value: "cluster-logging-operator"
                   - name: FLUENTD_IMAGE
                     value: "quay.io/openshift/origin-logging-fluentd:latest"
+                  - name: CURATOR_IMAGE
```
vimalk78, line 328: please rebase on master, the `CURATOR_IMAGE` has been removed

pkg/k8shandler/fluentd.go
```diff
@@ -353,6 +372,7 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
 		}
 	}
 	fluentdContainer := NewContainer("fluentd", "fluentd", v1.PullIfNotPresent, *resources)
+	exporterContainer := NewContainer("LogFileMetricExporter", "LogFileMetricExporter", v1.PullIfNotPresent, *resources)
```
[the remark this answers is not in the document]
pmoogi-redhat: resolved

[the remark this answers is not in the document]
pmoogi-redhat: shall we also incorporate changes in pkg/apis/logging/v1/clusterlogging_types.go 
to take care of log-file-metric-exporter container resources spec ?

//Collector Spec
type LogCollectionSpec struct {
	// The type of Log Collection to configure
	Type LogCollectionType `json:"type"`

	// Specification of the Fluentd Log Collection component
	FluentdSpec `json:"fluentd,omitempty"`
	// Specification of the log-file-metric-exporter component - for computing and publishing of total bytes logged metric
	//LogFileMetricExporterSpec `json:"logfilemetricexporter,omitempty"`
}

//for log-file-metric-exporter container resources spec

type LogFileMetricExporterSpec struct {
	// The resource requirements for log-file-metric-exporter container

	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
	Tolerations  []v1.Toleration   `json:"tolerations,omitempty"`
}

pkg/k8shandler/fluentd.go
```diff
@@ -428,7 +462,10 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
 
 	fluentdContainer.SecurityContext = &v1.SecurityContext{
 		Privileged: utils.GetBool(true),
-	}
+	} 
+        exporterContainer.SecurityContext = &v1.SecurityContext{
+                Privileged: utils.GetBool(true),
```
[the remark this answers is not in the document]
pmoogi-redhat: ok

bundle/manifests/cluster-logging.v5.2.0.clusterserviceversion.yaml
```diff
@@ -313,7 +313,7 @@ spec:
               containers:
               - name: cluster-logging-operator
                 image: quay.io/openshift-logging/cluster-logging-operator:latest
-                imagePullPolicy: IfNotPresent
+                imagePullPolicy: Always
```
vimalk78, line 316: Please undo this change

vimalk78 (changes requested):
There seem to be unrelated changes that have come in this PR. Need to make changes related to log-file-exporter in a new branch and push again in PR

alanconway (commented):
There are a lot of changes here that are not part of the Go watcher.
Please rebase on master and revert any remaining changes that are not an intentional part of the Go watcher. Note that before you merge any changes from a branch it is important to run `make check` to update generated code and check for format and lint errors, and it is usually wise to rebase on master and clean up any unintentional changes to make sure your branch is current.

alanconway (commented):
/approve

There are a lot of changes here that are not part of the Go watcher.
Please rebase on master and revert any changes that are not an intentional part of the Go watcher. Also make sure to run `make check` to run the formatter, linter etc.

The code that is relevant to the Go watcher looks fine. Bring the branch up to date and get rid of all the unrelated changes and I'll LGTM.

pkg/k8shandler/fluentd.go
```diff
@@ -400,6 +436,12 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
 		{Name: metricsVolumeName, MountPath: "/etc/fluent/metrics"},
 	}
 
+	exporterContainer.VolumeMounts = []v1.VolumeMount{
+		{Name: logVolumeMountName, MountPath: logVolumePath},
+	}
+	// default parameters --> exporterContainer.Command = []string{"/usr/local/bin/log-file-metric-exporter -verbosity=2 -dir=/var/log/containers -http=:2112"}
+	exporterContainer.Command = []string{"/bin/sh", "-c", "/usr/local/bin/log-file-metric-exporter  -verbosity=2  -dir=/var/log/containers -http=:2112"}
```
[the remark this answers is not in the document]
pmoogi-redhat: it is working fine and helped me to debug from container terminal.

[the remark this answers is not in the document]
alanconway: It is an extra process that _might_ cause subtle changes in signal handling behavior now that there are 2 processes instead of one. I can't prove that it _will_, but to quote Murphy's Law: if anything can go wrong, it will. I'm not paranoid, they really are out to get me.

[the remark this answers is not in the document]
alanconway: I say this because we have seen in the past that apparently simple changes that start multiple processes (e.g. letting fluentd fork workers) seem to have caused pods to hang on exit.

olm_deploy/operatorregistry/registry-deployment.yaml
```diff
@@ -41,6 +41,8 @@ spec:
           value: ${IMAGE_ELASTICSEARCH6}
         - name: IMAGE_LOGGING_KIBANA6
           value: ${IMAGE_LOGGING_KIBANA6}
+        - name: IMAGE_LOGGING_LOGFILEMETRICEXPORTER
```
jcantrill, line 44: Please modify to https://github.com/openshift/release/blob/master/ci-operator/config/openshift/cluster-logging-operator/openshift-cluster-logging-operator-master.yaml#L120  `IMAGE_LOG_FILE_METRIC_EXPORTER`

bundle/manifests/cluster-logging.v5.2.0.clusterserviceversion.yaml
```diff
@@ -330,6 +330,8 @@ spec:
                     value: "cluster-logging-operator"
                   - name: FLUENTD_IMAGE
                     value: "quay.io/openshift-logging/fluentd:latest"
+                  - name: LOGFILEMETRICEXPORTER_IMAGE
+                    value: "quay.io/openshift-logging/log-file-metric-exporter:5.2"
```
jcantrill, line 334: please set the tag to "latest" so we do not have to update in future

manifests/5.2/cluster-logging.v5.2.0.clusterserviceversion.yaml
```diff
@@ -330,6 +330,8 @@ spec:
                     value: "cluster-logging-operator"
                   - name: FLUENTD_IMAGE
                     value: "quay.io/openshift-logging/fluentd:latest"
+                  - name: LOGFILEMETRICEXPORTER_IMAGE
+                    value: "quay.io/openshift-logging/log-file-metric-exporter:5.2"
```
jcantrill, line 334: same here please set the tag to "latest" so we do not have to update in future

olm_deploy/scripts/catalog-deploy.sh
```diff
@@ -6,6 +6,7 @@ export IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY=${IMAGE_CLUSTER_LOGGING_OPERATOR_
 export IMAGE_CLUSTER_LOGGING_OPERATOR=${IMAGE_CLUSTER_LOGGING_OPERATOR:-registry.ci.openshift.org/${LOGGING_IS}/${LOGGING_VERSION}:cluster-logging-operator}
 export IMAGE_LOGGING_CURATOR5=${IMAGE_LOGGING_CURATOR5:-registry.ci.openshift.org/${LOGGING_IS}/${LOGGING_VERSION}:logging-curator5}
 export IMAGE_LOGGING_FLUENTD=${IMAGE_LOGGING_FLUENTD:-registry.ci.openshift.org/${LOGGING_IS}/${LOGGING_VERSION}:logging-fluentd}
+export IMAGE_LOGGING_LOGFILEMETRICEXPORTER=${IMAGE_LOGGING_LOGFILEMETRICEXPORTER:-registry.ci.openshift.org/${LOGGING_IS}/${LOGGING_VERSION}:log-file-metric-exporter}
```
jcantrill, line 9: correct the image name to match CI

olm_deploy/scripts/catalog-deploy.sh
```diff
@@ -14,6 +15,7 @@ echo "cluster logging operator registry: ${IMAGE_CLUSTER_LOGGING_OPERATOR_REGIST
 echo "cluster logging operator: ${IMAGE_CLUSTER_LOGGING_OPERATOR}"
 echo "curator5: ${IMAGE_LOGGING_CURATOR5}"
 echo "fluentd: ${IMAGE_LOGGING_FLUENTD}"
+echo "log-file-metric-exporter: ${IMAGE_LOGGING_LOGFILEMETRICEXPORTER}"
```
jcantrill, line 18: same here

pkg/k8shandler/fluentd.go
jcantrill: @pmoogi-redhat  we are defining certs here.  Please update "scheme" to be "https"

pkg/k8shandler/fluentd.go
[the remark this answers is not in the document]
jcantrill: +1 We should be able to run this:
* by defaulting to the image entrypoint
* calling the command directly without needing to wrapper in bash

jcantrill (changes requested):

pkg/k8shandler/fluentd.go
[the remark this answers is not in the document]
pmoogi-redhat: @jcantrill sorry i am not clear on this. Are you saying cmd should execute directly this binary as the below
exporterContainer.Command = []string{"/usr/local/bin/log-file-metric-exporter  -verbosity=2  -dir=/var/log/containers -http=:2112"}

or the entrypoint should be set as []string{"/usr/local/bin/log-file-metric-exporter  -verbosity=2  -dir=/var/log/containers -http=:2112"}

pkg/k8shandler/fluentd.go
```diff
@@ -446,11 +491,16 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
...
 	fluentdPodSpec := NewPodSpec(
 		"logcollector",
-		[]v1.Container{fluentdContainer},
+		[]v1.Container{fluentdContainer, exporterContainer},
 		[]v1.Volume{
-			{Name: "varlog", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/var/log"}}},
+			{Name: logVolumeMountName, VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: logVolumePath}}},
+			{Name: logfilesVolumeMountName, VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: logfilesVolumePath}}},
+			{Name: podlogfilesVolumeMountName, VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: podlogfilesVolumePath}}},
+			{Name: logfilesmetricexporterbinVolumeMountName, VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: logfilesmetricexporterbinVolumePath}}},
```
jcantrill, line 503: Should not need this if we are running as a separate container

pkg/k8shandler/fluentd.go
```diff
@@ -31,10 +31,19 @@ import (
...
+	syslogName                               = "syslog"
+	fluentdRequiredESVersion                 = "6"
+	logVolumeMountName                       = "varlog"
+	logfilesVolumeMountName                  = "varlogcontainers"
+	podlogfilesVolumeMountName               = "varlogpods"
+	logfilesmetricexporterbinVolumeMountName = "usrlocalbin"
+	logVolumePath                            = "/var/log"
+	logfilesVolumePath                       = "/var/log/containers"
+	podlogfilesVolumePath                    = "/var/log/pods"
+	logfilesmetricexporterbinVolumePath      = "/usr/local/bin"
```
jcantrill, line 45: or this constant

pkg/k8shandler/fluentd.go
```diff
@@ -446,11 +494,17 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
...
+	//exporterContainer := newLogFileMetricExporterContainer(cluster)
+
 	fluentdPodSpec := NewPodSpec(
 		"logcollector",
-		[]v1.Container{fluentdContainer},
+		[]v1.Container{fluentdContainer, exporterContainer},
 		[]v1.Volume{
-			{Name: "varlog", VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: "/var/log"}}},
+			{Name: logVolumeMountName, VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: logVolumePath}}},
+			{Name: CrtKeyVolumeMountName, VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: CrtKeyVolumePath}}},
```
eranra, line 504: drop this, we do not need that mount

pkg/k8shandler/fluentd.go
```diff
@@ -31,10 +31,21 @@ import (
...
-	fluentdAlertsFile        = "fluentd/fluentd_prometheus_alerts.yaml"
-	fluentdName              = "fluentd"
-	syslogName               = "syslog"
-	fluentdRequiredESVersion = "6"
+	fluentdAlertsFile                        = "fluentd/fluentd_prometheus_alerts.yaml"
+	fluentdName                              = "fluentd"
+	syslogName                               = "syslog"
+	fluentdRequiredESVersion                 = "6"
+	logVolumeMountName                       = "varlog"
+	CrtKeyVolumeMountName                    = "etcfluentmetrics"
```
eranra, line 39: drop this we do not need this

pkg/k8shandler/fluentd.go
```diff
@@ -400,6 +438,13 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
 		{Name: metricsVolumeName, MountPath: "/etc/fluent/metrics"},
 	}
 
+	exporterContainer.VolumeMounts = []v1.VolumeMount{
+		{Name: logVolumeMountName, MountPath: logVolumePath},
+		{Name: CrtKeyVolumeMountName, MountPath: CrtKeyVolumePath},
```
eranra, line 443: change to "CrtKeyVolumeMountName" to "metricsVolumeName" and "CrtKeyVolumePath" into "/etc/fluent/metrics" (maybe best will be to put that into a variable and change In both)

pkg/k8shandler/fluentd.go
```diff
@@ -142,6 +158,16 @@ func (clusterRequest *ClusterLoggingRequest) reconcileFluentdServiceMonitor() er
 			// ServerName can be e.g. fluentd.openshift-logging.svc
 		},
 	}
+	logMetricExporterEndpoint := monitoringv1.Endpoint{
+		Port:   exporterPortName,
+		Path:   "/metrics",
+		Scheme: "https",
+		TLSConfig: &monitoringv1.TLSConfig{
+			CAFile:     prometheusCAFile,
+			ServerName: fmt.Sprintf("%s.%s.svc", logfilesmetricexporterName, cluster.Namespace),
```
eranra, line 167: change back to `fluentdName`

pmoogi-redhat: resolved

pkg/k8shandler/fluentd.go
pmoogi-redhat: resolved

pkg/k8shandler/fluentd.go
pmoogi-redhat: resolved

pkg/k8shandler/fluentd.go
pmoogi-redhat: resolved

pkg/k8shandler/fluentd.go
```diff
@@ -419,6 +439,13 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
 		{Name: "tmp", MountPath: "/tmp"},
 	}
 
+	exporterContainer.VolumeMounts = []v1.VolumeMount{
+		{Name: logVolumeMountName, MountPath: logVolumePath},
+		{Name: metricsVolumeName, MountPath: "/etc/fluent/metrics"},
+	}
+	// default parameters -->  -verbosity=2 -dir=/var/log/containers -http=:2112"
+	exporterContainer.Command = []string{"/bin/sh", "-c", "/usr/local/bin/log-file-metric-exporter  -verbosity=2  -dir=/var/log/containers -http=:2112 -keyFile=/etc/fluent/metrics/tls.key  -crtFile=/etc/fluent/metrics/tls.crt"}
```
jcantrill, line 447: why do we need to wrapper with "/bin/sh" instead of simply calling /usr/local/bin/log-file-metric-exporter  and passing the args we need

pkg/k8shandler/fluentd.go
```diff
@@ -469,10 +499,17 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
 		},
 	)
 
+	//exporterContainer := newLogFileMetricExporterContainer(cluster)
```
jcantrill, line 502: remove comment

pkg/k8shandler/fluentd.go
```diff
@@ -452,6 +479,9 @@ func newFluentdPodSpec(cluster *logging.ClusterLogging, trustedCABundleCM *v1.Co
 		ReadOnlyRootFilesystem:   utils.GetBool(true),
 		AllowPrivilegeEscalation: utils.GetBool(false),
 	}
+	exporterContainer.SecurityContext = &v1.SecurityContext{
```
jcantrill, line 482: @syedriko what do we need here to comply with your security changes.  i believe the only requirement is we need to be able to read `/var/log/containers/*`

jcantrill (changes requested):

pkg/k8shandler/fluentd.go
syedriko: Just reuse the fluentdContainer.SecurityContext

pmoogi-redhat:
\retest

pkg/k8shandler/fluentd.go
pmoogi-redhat: @syedriko used same as fluentdContainer.SecurityContext for exporterContainer.SecurityContext.  One test case is failing .. need your review again to resolve it in e2e-gcp test.

pkg/k8shandler/fluentd.go
jcantrill: @pmoogi-redhat  nit

pkg/k8shandler/fluentd.go
jcantrill: @pmoogi-redhat ... following up on my other comment.. nit

pmoogi-redhat:
https://github.com/openshift/cluster-logging-operator/pull/979 - PR merged on golang based file watcher named as log-file-metric-exporter. It watches log files registered into its watcher object. It publishes log_logged_bytes_total for these log files.

PR #987, opened 2021-04-21
Bug1891623: Change expression for Fluentd file buffer usage

vparfonov:

### Description

1. Change `FluentdQueueLengthIncreasing` alert expression. Fluentd buffering logs when it can't push to an output. This can be happened at startup because logstore not ready yet or in case log spamming. To prevent firing alerts if this is not necessary and system works well need to tune expression for buffer usage. Going to use [deriv() function](https://prometheus.io/docs/prometheus/latest/querying/functions/#deriv) for buffer increase calculation

2. Remove `FluentdQueueLengthBurst` at all, because this alert was related to our internal store which do not seem to apply for now.

/cc @alanconway @blockloop 
/assign @jcantrill  

/cherry-pick 

### Links

- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1891623

### DIFF
--- a/files/fluentd/fluentd_prometheus_alerts.yaml
+++ b/files/fluentd/fluentd_prometheus_alerts.yaml
@@ -11,27 +11,16 @@
     "labels":
       "service": "fluentd"
       "severity": "critical"
-  - "alert": "FluentdQueueLengthBurst"
-    "annotations":
-      "message": "In the last minute, fluentd {{ $labels.instance }} buffer queue length increased more than 32. Current value is {{ $value }}."
-      "summary": "Fluentd is overwhelmed"
-    "expr": |
-      delta(fluentd_output_status_buffer_queue_length[1m]) > 32
-    "for": "1m"
-    "labels":
-      "service": "fluentd"
-      "severity": "warning"
   - "alert": "FluentdQueueLengthIncreasing"
     "annotations":
-      "message": "In the last 12h, fluentd {{ $labels.instance }} buffer queue length constantly increased more than 1. Current value is {{ $value }}."
-      "summary": "Fluentd file buffer usage issue"
+      "message": "For the last hour, fluentd {{ $labels.instance }} average buffer queue length has increased continuously."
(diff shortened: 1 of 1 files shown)


vparfonov:
Not tested yet on real cluster

files/fluentd/fluentd_prometheus_alerts.yaml
```diff
@@ -14,7 +14,7 @@
   - "alert": "FluentdQueueLengthBurst"
     "annotations":
       "message": "In the last minute, fluentd {{ $labels.instance }} buffer queue length increased more than 32. Current value is {{ $value }}."
-      "summary": "Fluentd is overwhelmed"
+      "summary": "Fluentd unable to keep up with burst in traffic."
     "expr": |
       delta(fluentd_output_status_buffer_queue_length[1m]) > 32
```
alanconway, line 19: Replace with deriv() and delta() as discussed, if that seems to work well.

alanconway (changes requested):

alanconway:
/approved

For LGTM we need tests to make sure this 
- triggers when we want it to (functional test that simulates overload)
- but does not trigger spuriously (not sure how to test this, or if we can. Anyone?)

files/fluentd/fluentd_prometheus_alerts.yaml
```diff
@@ -14,7 +14,7 @@
   - "alert": "FluentdQueueLengthBurst"
```
jcantrill, line 14: @vparfonov  please remove this entire alert.  @alanconway and I had a conversation regarding the value and decided there needs to be more thought around
* What we are trying to convey and monitor
* What is actionable for someone

As-is, this alert originated in issues we had with writing to our internal store which do not seem to apply

vparfonov: done

jcantrill (commented):
/approve
/lgtm

jcantrill:
@vparfonov please create 5.1 and 5.0 JIRA issues to track these if they don't already exist

PR #990, opened 2021-04-23
LOG-1201: Fixed zero minutes/days/etc. retention time

igor-karpukhin:
### Description
This PR fixes CLO behaviour when retention time (`RetentionPolicy.MaxAge`) is set to `0(m | h | d | w | M | Y)`

/cc @jcantrill @vimalk78 @vparfonov 
/assign @jcantrill 

### Links
- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1936998
- JIRA: https://issues.redhat.com/browse/LOG-1201

### DIFF
--- a/pkg/k8shandler/indexmanagement/indexmanagement_test.go
+++ b/pkg/k8shandler/indexmanagement/indexmanagement_test.go
@@ -27,6 +27,104 @@ var _ = Describe("Indexmanagement", func() {
 		}
 	})
 
+	Describe("IndexManagemet Policy with 0 minutes, hours, days, weeks, months", func() {
+		It("should correctly parse 0m", func() {
+			retentionPolicy = &logging.RetentionPoliciesSpec{
+				App: &logging.RetentionPolicySpec{
+					MaxAge: esapi.TimeUnit("0m"),
+				},
+				Infra: &logging.RetentionPolicySpec{
+					MaxAge: esapi.TimeUnit("0m"),
+				},
+				Audit: &logging.RetentionPolicySpec{
+					MaxAge: esapi.TimeUnit("0m"),
+				},
+			}
+			spec := NewSpec(retentionPolicy)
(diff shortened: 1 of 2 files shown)


pkg/k8shandler/indexmanagement/indexmanagement_test.go
```diff
@@ -27,6 +27,104 @@ var _ = Describe("Indexmanagement", func() {
...
+					MaxAge: esapi.TimeUnit("0h"),
+				},
+				Audit: &logging.RetentionPolicySpec{
+					MaxAge: esapi.TimeUnit("0h"),
+				},
+			}
+			spec := NewSpec(retentionPolicy)
+			Expect(len(spec.Policies)).To(Equal(3))
+			Expect(len(spec.Mappings)).To(Equal(3))
+			Expect(spec.Policies[0].Phases.Delete.MinAge).To(Equal(esapi.TimeUnit("0h")))
```
jcantrill, line 65: @ewolinetz the original intent why this bug was exposed was @RiRa12621 was trying to disable collection of a specific log source (I think).  Implication is never delete?  This corrects the parsing error but I think it will mean: delete the indices everytime cron runs because the indices will always be greater then the minage.  Should we restrict you to defining something > 0 or is it sufficient to say "customer told me continually delete so thats what we will do"?

jcantrill (commented):
/approve
/hold

pkg/k8shandler/indexmanagement/indexmanagement_test.go
ewolinetz: agreed, this currently will mean delete every time it will run. if we want it to mean "don't delete" (which, I personally don't think we should offer because it means the cluster will fall over) we should update the logic in the EO cron scripts first so we don't have a gap where we constantly delete logs.

PR #993, opened 2021-04-27
LOG-1012: Set cluster condition state during reconcile  

vparfonov:

### Description

Update the `ClusterLogging` state condition during reconciliation return error in case critical errors and continue in a degraded mode for non blocking errors.
#### Not blocking errors are:

- Unable to create or update metrics dashboards
- Unable to create or update certificates
- Unable to create or update logstore
- Unable to create or update collection

#### Critical errors are:

- Unable to create or update visualization

/cc 
/assign 

/cherry-pick 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-1012


pkg/k8shandler/reconciler.go
```diff
@@ -35,20 +34,27 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
 		clusterLoggingRequest.ForwarderSpec = forwarder.Spec
 	}
 
+	clusterLoggingRequest.Cluster.Status.Conditions.SetCondition(logging.NewCondition("ConditionUnknown",
```
jcantrill, line 37: @alanconway thoughts?  This is a condition that would constantly flip.  Do we want to do this every cycle.  I'm betting it may be difficult for a user to catch

pkg/k8shandler/reconciler.go
```diff
@@ -35,20 +34,27 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
...
+		corev1.ConditionUnknown, "Processing", "Reconciliation in progress"))
 	if clusterLoggingRequest.IncludesManagedStorage() {
 		// Reconcile certs
 		if err = clusterLoggingRequest.CreateOrUpdateCertificates(); err != nil {
+			clusterLoggingRequest.Cluster.Status.Conditions.SetCondition(
+				condInvalid("Unable to create or update certificates"))
```
jcantrill, line 43: @alanconway I don't think this is entirely accurate.  This is a hard failure for the logstore but if we have collection spec'd outside of managed storage... The operator would be degraded providing partial use.

pkg/k8shandler/reconciler.go
```diff
@@ -35,20 +34,27 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
...
 			return fmt.Errorf("Unable to create or update certificates for %q: %v", clusterLoggingRequest.Cluster.Name, err)
```
jcantrill, line 44: This is a hard failure for deploying managed storage but not necessarily log forwarding.  @alanconway thoughts on how best to address?

pkg/k8shandler/reconciler.go
```diff
@@ -35,20 +34,27 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
...
 		}
 
 		// Reconcile Log Store
 		if err = clusterLoggingRequest.CreateOrUpdateLogStore(); err != nil {
+			clusterLoggingRequest.Cluster.Status.Conditions.SetCondition(condInvalid("Unable to create or update logstore"))
```
jcantrill, line 49: same comment here

pkg/k8shandler/reconciler.go
```diff
@@ -35,20 +34,27 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
...
 			return fmt.Errorf("Unable to create or update logstore for %q: %v", clusterLoggingRequest.Cluster.Name, err)
 		}
 
 		// Reconcile Visualization
 		if err = clusterLoggingRequest.CreateOrUpdateVisualization(); err != nil {
-			return fmt.Errorf("Unable to create or update visualization for %q: %v", clusterLoggingRequest.Cluster.Name, err)
+			clusterLoggingRequest.Cluster.Status.Conditions.SetCondition(condDegraded(logging.ReasonMissingResource,
+				"Unable to create or update visualization"))
+			//return fmt.Errorf("Unable to create or update visualization for %q: %v", clusterLoggingRequest.Cluster.Name, err)
```
jcantrill, line 57: remove me if we are not hard erroring

pkg/k8shandler/reconciler.go
```diff
@@ -68,14 +74,20 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
 
 	// Reconcile Collection
 	if err = clusterLoggingRequest.CreateOrUpdateCollection(); err != nil {
+		clusterLoggingRequest.Cluster.Status.Conditions.SetCondition(condInvalid("Unable to create or update collection"))
 		return fmt.Errorf("Unable to create or update collection for %q: %v", clusterLoggingRequest.Cluster.Name, err)
```
jcantrill, line 78: @alanconway this is the flip of logstore, while maybe not getting new logs, having deployed storage may still provide query into already collected logs

files/fluentd/fluentd_prometheus_alerts.yaml
```diff
@@ -14,7 +14,7 @@
   - "alert": "FluentdQueueLengthBurst"
     "annotations":
       "message": "In the last minute, fluentd {{ $labels.instance }} buffer queue length increased more than 32. Current value is {{ $value }}."
-      "summary": "Fluentd is overwhelmed"
+      "summary": "Fluentd unable to keep up with burst in traffic."
```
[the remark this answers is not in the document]
vparfonov: yes, fixed

pkg/k8shandler/reconciler.go
alanconway: +1 drop this condition.
Status is not the right place for transient "in progress" messages, it is for long-lasting conditions that won't change unless there is an external change: either a change to the API or a change in the state of some operand.

pkg/k8shandler/reconciler.go
alanconway: +1. We should also mark all the conditions for all pipelines to "default" as invalid if the are not already.  
If that leaves 0 valid pipelines, then this condition is Invalid, otherwise it is Degraded.

pkg/k8shandler/reconciler.go
alanconway: See above

pkg/k8shandler/reconciler.go
alanconway: +1 degraded

alanconway (changes requested):

pkg/k8shandler/reconciler.go
vparfonov: ok

pkg/k8shandler/reconciler.go
vparfonov: ok

pkg/k8shandler/reconciler.go
vparfonov: ok

pkg/k8shandler/reconciler.go
vparfonov: ok

alanconway:
/approve
This is going the right direction, pleaes @mention me when all the comments are addressed to re-review

vparfonov:
@alanconway Please, take a look

jcantrill (commented):
/hold

pkg/k8shandler/reconciler.go
```diff
@@ -38,17 +38,24 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
...
+			log.Error(err, fmt.Sprintf("Cluster Logging transition to Degraded state: %s", msg))
 		}
 
 		// Reconcile Visualization
 		if err = clusterLoggingRequest.CreateOrUpdateVisualization(); err != nil {
-			return fmt.Errorf("Unable to create or update visualization for %q: %v", clusterLoggingRequest.Cluster.Name, err)
+			msg := fmt.Sprintf("Unable to create or update visualization for %q: %v", clusterLoggingRequest.Cluster.Name, err)
+			clusterLoggingRequest.Cluster.Status.Conditions.SetCondition(condInvalid(msg))
+			log.Error(err, fmt.Sprintf("Cluster Logging transition to Invalid state: %s", msg))
+			return fmt.Errorf(msg)
```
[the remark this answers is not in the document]
vparfonov: Look like I misunderstand your comments [here](https://github.com/openshift/cluster-logging-operator/pull/993#discussion_r621386494), I thought you're saying that this is "hard error" for us.

pkg/k8shandler/reconciler.go
```diff
@@ -68,14 +75,41 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
...
 	if err = clusterLoggingRequest.CreateOrUpdateDashboards(); err != nil {
-		return fmt.Errorf("Unable to create or update metrics dashboards for %q: %w", clusterLoggingRequest.Cluster.Name, err)
+		msg := fmt.Sprintf("Unable to create or update metrics dashboards for %q: %v", clusterLoggingRequest.Cluster.Name, err)
+		clusterLoggingRequest.Cluster.Status.Conditions.SetCondition(condDegraded(logging.ReasonMissingResource, msg))
+		log.Error(err, "Cluster Logging transition to Degraded state: "+msg)
+	}
+
+	spec := &logging.ClusterLogForwarderSpec{}
+	status := &logging.ClusterLogForwarderStatus{}
+	clusterLoggingRequest.verifyInputs(spec, status)
```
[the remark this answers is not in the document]
vparfonov: Trying to follow Alan's [comment](https://github.com/openshift/cluster-logging-operator/pull/993#discussion_r622433383) and yes you are right we aready do it [here](https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/forwarding.go#L123-L134), but if i understand correct `NormalizeForwarder` don't call during collector reconciliation. Please, correct me if i wrong, becpuse logic of recocilation not fully cleare for me at the moment.

pkg/k8shandler/reconciler.go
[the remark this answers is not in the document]
vparfonov: @jcantrill What errors we can call `critical` on your point of view?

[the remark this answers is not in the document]
jcantrill: From my perspective, the critical path which leads to "hard error" with immediate exit is:
* Minimal resource to deploy collector: DS, configmap, secret, priorityclass, SA, permissions, certs for default store if logstore defined

Everything else means we are degraded:
* No elasticsearch store
* No visualization
* Resources associated with collector metrics: service, rulebooks, etc

pkg/k8shandler/reconciler.go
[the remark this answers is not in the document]
jcantrill: The status from CLF are displayed on CLF.  IMO we should refrain from duplicating them on CL  I think its fair to distinguish CL from CLF and it may be possible that CL is degraded while CLF is not and visa versa

vparfonov:
@jcantrill found this [struct](https://github.com/openshift/cluster-logging-operator/blob/master/pkg/apis/logging/v1/clusterlogging_types.go#L50) but looks like it never use, maybe it will be reasonable for this task

alanconway (approved):
/lgtm

jcantrill:
Taking one last look at the logic here, we lose conditional statue as we progress through each check.  The PR uses `SetCondition` which adds or updates based on the conditionType.  This means if multiple pieces fail that only one condition type is updated.  I'll work to expand the expectation here and maybe we need to put this back into the backlog pending further design.  My expectation is:
* we would have some over arching operator condition
* Additional conditions for some specific check that are true/false depending on that smaller conditions

I'll proceed to work on some more design

alanconway:
Regarding @jcantrill comment about losing status, I suggest we replace the direct use of SetCondition with tge wrapper function ClusterLoggingRequest.UpdateCondition but with a few improvements:

- UpdateCondition should create an empty conditions if the conditions is nil, I think I saw a panic caused by a nil Conditions.
- Never replace an "abnormal" condition with a "normal" one. 
  - always add/replace missing or "Unknown" conditions.
  - For most conditions (e.g. Degraded)  replace false with true but not vice versa. 
   - "Ready" is the opposite, can replace true (healthy) with false (broken) but not vice versa.
@vparfonov @jcantrill with that change I think this is in good shape - any other concerns?

PR #999, opened 2021-04-30
Fix built-in application behavior to collect all of logs

k-keiichi-rh:
# Overview
This PR changes the default input behavior to collect  records from all namespaces and pods.
The current log filtering by pod labels and namespaces only collects non matching records("anything left over").

For example, 
```
kind: ClusterLogForwarder
metadata:
  name: instance
  namespace: openshift-logging
spec:
  inputs:
   - name: application-logs1
     application:
       namespaces:
       - my-project
  outputs:
   - name: syslog1
     ...
  pipelines:
   - name: syslog-app-logs1
     inputRefs:
     - application-logs1
     outputRefs:
     - syslog1
   - name: app-default
     inputRefs:
     - application
     outputRefs:
     - default
```
The "app-default" pipeline won't collect logs from "my-project" namespace in the customized input.

The concept of "application" was discussed in https://github.com/openshift/cluster-logging-operator/pull/955/files#r600738769. But I would like to confirm it again.

/cc @alanconway @jcantrill

### Links
- Enhancement proposal: https://github.com/openshift/enhancements/pull/457
- JIRA: https://issues.redhat.com/browse/LOG-660

### DIFF
--- a/pkg/generators/forwarding/fluentd/templates.go
+++ b/pkg/generators/forwarding/fluentd/templates.go
@@ -540,16 +540,20 @@ const inputSelectorToPipelineTemplate = `{{- define "inputSelectorToPipelineTemp
 <label {{sourceTypelabelName .Source}}>
   <match **>
     @type label_router
-{{- if .PipelineNames }}
-    default_route {{sourceTypelabelName .Source}}_DEFAULT
-{{- end }}
     {{- range .InputSelectors }}
     {{ . }}
     {{- end}}
+{{- if .PipelineNames }}
+    <route>
+      @label {{sourceTypelabelName .Source}}_ALL
+      <match>
+      </match>
+    </route>
+{{- end }}
   </match>
 </label>
 {{ if .PipelineNames -}}
-<label {{sourceTypelabelName .Source}}_DEFAULT>
+<label {{sourceTypelabelName .Source}}_ALL>
   <match **>
     @type copy
 {{- range $index, $pipelineLabel := .PipelineNames }}
--- a/pkg/generators/forwarding/fluentd/fluent_conf_test.go
+++ b/pkg/generators/forwarding/fluentd/fluent_conf_test.go
@@ -496,7 +496,6 @@ var _ = Describe("Generating fluentd config", func() {
   <label @_APPLICATION>
     <match **>
       @type label_router
-      default_route @_APPLICATION_DEFAULT
       <route>
         @label @APPS_PIPELINE
         <match>
@@ -509,9 +508,14 @@ var _ = Describe("Generating fluentd config", func() {
           namespaces dev-apple,project2-namespace
         </match>
       </route>
+      <route>
+        @label @_APPLICATION_ALL
+        <match>
+        </match>
+      </route>
     </match>
   </label>
-  <label @_APPLICATION_DEFAULT>
+  <label @_APPLICATION_ALL>
     <match **>
       @type copy
       <store>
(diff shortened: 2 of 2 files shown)


jcantrill (commented):
/approve

pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -540,12 +540,16 @@ const inputSelectorToPipelineTemplate = `{{- define "inputSelectorToPipelineTemp
...
     @type label_router
-{{- if .PipelineNames }}
-    default_route {{sourceTypelabelName .Source}}_DEFAULT
-{{- end }}
     {{- range .InputSelectors }}
     {{ . }}
     {{- end}}
+{{- if .PipelineNames }}
+    <route>
+      @label {{sourceTypelabelName .Source}}_DEFAULT
```
jcantrill, line 548: What is sourcetype_default?  "Default" is whatever the default output type is, currently elasticsearch.  I believe the implication of what you are saying is:  If I have no other routing choices defined but I need to send everything else to the default store do that here, in which case it should be "@DEFAULT"

k-keiichi-rh: > What is sourcetype_default? "Default" is whatever the default output type is, currently elasticsearch.

"_default" is my wrong choice. As you said, "If I have no other routing choices defined but I need to send everything else"
is what I want to say. But it's not limited to "the default store". I think it's same as the built-in input name  `application`.
So is the "_BUILT_IN" the right choice?

Or are there any good words?
To implement pod label filtering, the commit 41bb627a2b332e6297f52920f85c0f94bbc04071 is replaced with the commit 5b7c7f599651b91017412be99532444c51023dae.
I want to define the following "<match **> section in the commit 41bb627a2b332e6297f52920f85c0f94bbc04071.

Quoted from the commit 41bb627a2b332e6297f52920f85c0f94bbc04071:
```
  # Relabel specific sources (e.g. logs.apps) to multiple pipelines
  <label @_APPLICATION>
     <match kubernetes.**_dev-apple_**>
      @type copy
      <store>
            @type relabel
            @label @APPS_PIPELINE2
      </store>
    </match>
    <match kubernetes.**_project1-namespace_**>
      @type copy
      <store>
                @type relabel
                @label @APPS_PIPELINE
      </store>
    </match>
    <match kubernetes.**_project2-namespace_**>
      @type copy
      <store>
                @type relabel
                @label @APPS_PIPELINE
      </store>
      <store>
                @type relabel
                @label @APPS_PIPELINE2
      </store>
    </match>
+   <match **>
|     @type copy
|     <store>
|           @type relabel
|           @label @MY_DEFAULT_PIPELINE
|     </store>
+   </match>                    
   </label>                         
```

alanconway: I think the code is correct but agree the name is confusing.

The bug we are fixing is that the label filter `default_route` only gets _unmatched_ traffic, but we want inputs with no filters to forward _all_ traffic, so @k-keiichi-rh  has replaced `default_route` with a "match all" route. 
How about `{{sourceTypelabelName .Source}}_ALL`, so we have `@_APPLICATION_ALL` meaning the label that gets all the application logs with no filters.

alanconway (changes requested):

pkg/generators/forwarding/fluentd/templates.go
k-keiichi-rh: Thanks. I fixed to `{{sourceTypelabelName .Source}}_ALL`.

alanconway (approved):
/lgtm

alanconway:
@jcantrill are you happy to remove the hold?

PR #1005, opened 2021-05-04
Update the CentOS Dockerfile patch for adjustment

k-keiichi-rh:
Update the CentOS dockerfile patch to align with the latest Dockerfile generated by Dockerfile.in

/cc @alanconway @jcantrill


alanconway (approved):
/lgtm
/approve

jcantrill:
If we really need this change we need to build upon recent changes to generate and check in the results. Similar to what was done for the elasticsearch operator

PR #1012, opened 2021-05-06
Replace OCP_VERSION with LOGGING_VERSION

alanconway:
All versioned artifacts now go under manifests/$LOGGING_VERSION.  Redundant OCP_VERSION was removed.

/assign @jcantrill


jcantrill (commented):
/lgtm

PR #1016, opened 2021-05-11
LOG-1354: No datapoints found on Elastic Query/Fetch Latency dashboard

lukas-vlcek:
### Description

There was a slight change to indices statistics metrics. We also need
to get rid of label filter (it does not make sense now).

Additional change was done to Prometheus metric calculation itself.
We changed "rate()/irate()" to "rate()/rate()".

Context – useful resource about rate and irate metrics: <https://www.robustperception.io/irate-graphs-are-better-graphs>

/cc @blockloop 
/assign @ewolinetz 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-1354

### DIFF
--- a/files/dashboards/openshift-logging-dashboard.json
+++ b/files/dashboards/openshift-logging-dashboard.json
@@ -746,13 +746,13 @@
           "steppedLine": false,
           "targets": [
             {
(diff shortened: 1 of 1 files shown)


blockloop (approved):
/lgtm

PR #1021, opened 2021-05-13
LOG-1355: Support outputDefaults in Log Forwarding

vimalk78:
### Description
This PR attempts to support `outputDefaults` which to allow index settings on the default elasticsearch output, as per the Json structured logs enhancement proposal: https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/forwarding-json-structured-logs.md#default-output-configuration

/cc @jcantrill 
/assign @alanconway 

/cherry-pick release-5.1

### Links
- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-1355
- Enhancement proposal: To allow index settings on the default elasticsearch output,


test/functional/framework.go
```diff
@@ -80,8 +80,8 @@ type FluentdFunctionalFramework struct {
 }
 
 func init() {
-	maxDuration, _ = time.ParseDuration("60*2m")
-	defaultRetryInterval, _ = time.ParseDuration("1ms")
+	maxDuration, _ = time.ParseDuration("10m")
```
alanconway, line 83: FYI: a better way to initialize duration constants is
```
const maxDuration = 60 *time.Minute
```
Not necessary to change for this PR.

pkg/apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -211,6 +216,26 @@ type PipelineSpec struct {
...
+	//
+	// +optional
+	StructuredIndexKey string `json:"structuredIndexKey,omitempty"`
+
+	// StructuredIndexName specifies the name of elasticsearch schema
+	//
+	// +optional
+	StructuredIndexName string `json:"structuredIndexName,omitempty"`
+}
+
```
alanconway, line 238: Very sorry about this, it's my fault. You followed the proposal, I wrote it wrong. Here's what I meant to write:

```suggestion
type OutputDefaults struct {
	// +optional
	Elasticsearch *Elasticsearch `json:"elasticsearch,omitempty"`
}
```
This way we re-use the existing ES settings struct, and it's easy to extend to cover other output types in future if we need to. I'll add a JIRA to cover other output types later - ultimately OutputDefaults == OutputTypeSpec so we automatically allow default settings for any new output type that is added.

vimalk78: updated

alanconway (approved):
/lgtm

PR #1049, opened 2021-06-09
Fix broken Makefile - expose cluster registry route explicitly.

alanconway:
My last Makefile change broke things for AWS users or anywhere that the registry is not exposed by default (it is on CRC, my bad)

Makefile now exposes the registry if needed. Tested on AWS and CRC.


jcantrill (commented):
/lgtm

PR #1062, opened 2021-06-16
LOG-1421: Modify channels to PE recommendations

jcantrill:
### Description
This PR:
* Modifies channel naming to match PE recommendations
* Sets the default channel to use when a dependency is declared

### Links
https://issues.redhat.com/browse/LOG-1421


jcantrill:
cc @alanconway @vimalk78

PR #1064, opened 2021-06-17
Fix CLO registry deployment

vimalk78:
### Description
need to add `LOCAL_IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY` and `IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY` variables to new make targets

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -220,12 +220,15 @@ cluster-logging-cleanup: cluster-logging-operator-uninstall cluster-logging-cata
 # builds an operator-registry image containing the cluster-logging operator
 cluster-logging-catalog-build: .make/cluster-logging-catalog-build
 .make/cluster-logging-catalog-build: $(shell find olm_deploy -type f)
+	LOCAL_IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY=$(REGISTRY_PUBLIC)/openshift/cluster-logging-operator-registry \
 	olm_deploy/scripts/catalog-build.sh
 	touch $@
 
 # deploys the operator registry image and creates a catalogsource referencing it
 cluster-logging-catalog-deploy: .make/cluster-logging-catalog-deploy
 .make/cluster-logging-catalog-deploy: $(shell find olm_deploy -type f)
+	IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY=$(REGISTRY_INTERNAL)/openshift/cluster-logging-operator-registry \
+	IMAGE_CLUSTER_LOGGING_OPERATOR=$(REGISTRY_INTERNAL)/openshift/origin-cluster-logging-operator:latest \
 	olm_deploy/scripts/catalog-deploy.sh
 
 # deletes the catalogsource and catalog namespace


Makefile
```diff
@@ -220,12 +220,15 @@ cluster-logging-cleanup: cluster-logging-operator-uninstall cluster-logging-cata
 # builds an operator-registry image containing the cluster-logging operator
 cluster-logging-catalog-build: .make/cluster-logging-catalog-build
 .make/cluster-logging-catalog-build: $(shell find olm_deploy -type f)
+	LOCAL_IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY=$(REGISTRY_PUBLIC)/openshift/cluster-logging-operator-registry \
```
jcantrill, line 223: shouldnt this be full swap of the image URL

Makefile
```diff
@@ -220,12 +220,15 @@ cluster-logging-cleanup: cluster-logging-operator-uninstall cluster-logging-cata
...
 	olm_deploy/scripts/catalog-build.sh
 	touch $@
 
 # deploys the operator registry image and creates a catalogsource referencing it
 cluster-logging-catalog-deploy: .make/cluster-logging-catalog-deploy
 .make/cluster-logging-catalog-deploy: $(shell find olm_deploy -type f)
+	IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY=$(REGISTRY_INTERNAL)/openshift/cluster-logging-operator-registry \
+	IMAGE_CLUSTER_LOGGING_OPERATOR=$(REGISTRY_INTERNAL)/openshift/origin-cluster-logging-operator:latest \
```
jcantrill, line 231: same here

Makefile
vimalk78: sorry, i didn't get you.

jcantrill: should this be something like `LOCAL_IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY=${SOMEOTHERVAR_TO_A_FULL_URL}`

vimalk78: this line i copied from https://github.com/openshift/cluster-logging-operator/pull/1064/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R151

vimalk78: and just moved it below, i suspected that the environment variables are not getting passed from one make target to another

alanconway:
If we export variables with the most common values at the top of the makefile, then we don't need to set them again in targets unless that target needs a non-default values. All processes forked by the targets will inherit the Makefiles env.

Makefile
alanconway: Make targets don't _pass_ env vars; assignments of the form `export FOO=BAR` modify the env of the make process, and that env is inherited by the shells that execute the commands in a target. The trick is the _order of evaluation_ of assignments and targets. That is black magic that no mere mortal can comprehend.  My latest change  https://github.com/openshift/cluster-logging-operator/pull/1070 should avoid this ordering madness - please try a rebase.

alanconway:
@vimalk78  Is this PR still needed since you reverted the registry changes?

alanconway:
Closing, I think this is no longer needed. Please reopen if I'm wrong.
/close

PR #1065, opened 2021-06-17
Bug 1971003: [release-4.6] Support Adding log source info to log message

vimalk78:
This is a manual cherry-pick of https://github.com/openshift/cluster-logging-operator/pull/790 
 - A fix for bug 1891886.
 - Added a field in LF CRD `AddLogSource`, which when enabled
   adds namespace_name, pod_name, container_name to the `message` field of the record.
 - This can be further used with `PayloadKey` set to `"message"` to achieve the same
   behavior as older syslog plugin.
 - fixed hostname to be cluster's node name, rather than fluentd's podname

### Description

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:

### DIFF
--- a/pkg/generators/forwarding/fluentd/templates.go
+++ b/pkg/generators/forwarding/fluentd/templates.go
@@ -622,6 +622,21 @@ const outputLabelConfJsonParseNoretryTemplate = `{{- define "outputLabelConfJson
 	merge_json_log false
 	replace_json_log true
   </filter>
+{{ if .Target.Syslog.AddLogSource }}
+  <filter **>
+	@type record_modifier
+	<record>
+	  kubernetes_info ${if record.has_key?('kubernetes'); record['kubernetes']; else {}; end}
+	  namespace_info  ${if record['kubernetes_info'] != nil && record['kubernetes_info'] != {}; "namespace_name=" + record['kubernetes_info']['namespace_name']; else nil; end}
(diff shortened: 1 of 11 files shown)


pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -622,6 +622,21 @@ const outputLabelConfJsonParseNoretryTemplate = `{{- define "outputLabelConfJson
 	merge_json_log false
 	replace_json_log true
   </filter>
+{{ if .Target.Syslog.AddLogSource }}
+  <filter **>
+	@type record_modifier
+	<record>
+	  kubernetes_info ${if record.has_key?('kubernetes'); record['kubernetes']; else {}; end}
```
jcantrill, line 629: i realize this is a backport... but should this be empty space and not empty hash?

jcantrill (commented):
/lgtm

pkg/generators/forwarding/fluentd/templates.go
vimalk78: this is an empty hash because this expression tries to fetch the value of `kubernetes` object

syedriko (approved):

mfojtik:
[patch-manager] :rocket: Approved for z-stream by score: 1.00

approved

PR #1071, opened 2021-06-24
Updated Dockerfile.in to not use ART images

vimalk78:
### Description
Since downstream builds do not use ART images anymore, need to update base images in Dockerfile.in to match the images used in CPaaS.
 Also added an extra LABEL step in Dockerfile to update the name LABEL as per downstream requirements

 - updated Dockerfile.in
 - updated origin-meta.yaml
 - updated generated Dockerfile

/cc 
/assign @jcantrill 

/cherry-pick release-5.1


jcantrill (commented):
/lgtm

PR #1072, opened 2021-06-24
LOG-1500: Discard messages after 60m instead of retry forever

jcantrill:
### Description
This PR:
* Modifies buffer retry to drop log messages after 10 minutes
* Adds API to buffer tuning section to allow configuration of retry

/cc @alanconway 

### Links
* Ref: https://issues.redhat.com/browse/LOG-1500

### DIFF
--- a/manifests/5.2/logging.openshift.io_clusterloggings_crd.yaml
+++ b/manifests/5.2/logging.openshift.io_clusterloggings_crd.yaml
@@ -296,6 +296,13 @@ spec:
                               exponential_backoff`.'
                             pattern: ^([0-9]+)([smhd]{0,1})$
                             type: string
+                          retryTimeout:
+                            description: RetryTimeout represents the maxixum time
+                              interval to attempt retries before giving up and the
+                              record is disguarded.  If unspecified, the default will
+                              be used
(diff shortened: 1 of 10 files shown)


alanconway (commented):
This seems like a generic setting not a fluentd one. What do you think of putting it on the CLF output? We have a global outputDefaults field on the CLF now so you could still set a global default but also specific over-rides e.g. because an off-site output is expected to have more network problems than a local-store output.

jcantrill:
> This seems like a generic setting not a fluentd one. What do you think of putting it on the CLF output? We have a global outputDefaults field on the CLF now so you could still set a global default but also specific over-rides e.g. because an off-site output is expected to have more network problems than a local-store output.

I agree this is a generic setting but I think the correct way to expose this is to roll it into a policy of some kind related to the buffer settings.  The "policy" would  spec the tradeoff regarding delivery guarantee and however we abstract file vs memory buffers.  I have mentioned this several times before and believe that would be a better approach.  The change here is more of a temporary fix.

Also, @sichvoge and I have talked about needing a more generic way to provide overrides which I need to discuss with you

PR #1039, opened 2021-05-25
LOG-1400: elasticsearch output should add app- prefix to index names

vimalk78:
### Description
This is an attempt to fix implementation of json log forwarding based on updated enhancement proposal.

/cc @jcantrill @alanconway 
/assign @jcantrill 

/cherry-pick 

### Links
- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal: PR https://github.com/openshift/enhancements/pull/780

PR #1078, opened 2021-06-29
[release-5.1] LOG-1400: elasticsearch output should add app- prefix to index names.

vimalk78:
### Description
A manual cherry-pick of https://github.com/openshift/cluster-logging-operator/pull/1039

/cc 
/assign

### DIFF
--- a/bundle/manifests/logging.openshift.io_clusterlogforwarders_crd.yaml
+++ b/bundle/manifests/logging.openshift.io_clusterlogforwarders_crd.yaml
@@ -137,13 +137,13 @@ spec:
                       specified here will be used as default values for Elasticsearch
                       Output spec"
                     properties:
-                      structuredIndexKey:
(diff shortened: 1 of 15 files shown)


alanconway:
@vimalk78  Is the cherry-pick bot giving us any real convenience compared to treating cherry-picks as normal merges? I haven't done any cherry-picks so this is a neutral question...

vimalk78:
> @vimalk78 Is the cherry-pick bot giving us any real convenience compared to treating cherry-picks as normal merges? I haven't done any cherry-picks so this is a neutral question...

I have done many cherry-picks in CLO, but recently i have seen 3 cases in CLO where a cheery-pick missed files in manifest folder. From next time i will manually check to make sure all changes are there.

It could be due to a sub-folder in `manifest` which changes name with version. So bot may fail to cherry-pick from `master`, which has sub-folder 5.2,  to release-5.1 which has subfolder 5.1

PR #1079, opened 2021-06-30
Fix race condition in registry.sh

alanconway:
### Description

Sporadically on AWS `oc registry info --public` returns 0 exit status,
but the next call fails to return a registry string and login fails.

Modify the script to retry until everything works together:
registry info 0 exit status, registry name non-empty, and login succeeds.

/cc @vimalk78
/assign @alanconway

### DIFF
--- a/hack/registry.sh
+++ b/hack/registry.sh
@@ -3,18 +3,20 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


jcantrill (commented):
/lgtm

PR #1090, opened 2021-07-08
LOG-1499: Force encoding to utf-8 and replace invalid chars

jcantrill:
### Description
This PR:
* Explicitly forces messages to be converted to utf-8
* Replaces any invalid characters with a char that can be consumed by Elasticsearch

Tested by:
* using hexedit to change a message line to have 0x92 in it
* setup fluent.conf to write to elasticsearch
* include filter in conf to explicitly convert before writing to elasticsearch
* demonstrated ES rejection without conversion and acceptance with conversion

### Links
* https://issues.redhat.com/browse/LOG-1499


alanconway (approved):
/lgtm

PR #1092, opened 2021-07-09
LOG-1102: Finalize Log forwarding to Cloudwatch for GA

syedriko:
### Description
A series of cherrypicks off the tech-preview branch with support for log forwarding to Cloudwatch

/cc @igor-karpukhin 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-1102

Commits will be squashed before merge.


igor-karpukhin (approved):
Looks good. But E2E test failed.

internal/pkg/generator/forwarder/generator.go
```diff
@@ -44,15 +45,21 @@ func Generate(clfYaml string, includeDefaultLogStore, includeLegacyForward bool)
...
 	}
 	if includeDefaultLogStore {
 		clRequest.Cluster.Spec.LogStore = &logging.LogStoreSpec{
 			Type: logging.LogStoreTypeElasticsearch,
 		}
 	}
+
 	spec, status := clRequest.NormalizeForwarder()
 	log.V(2).Info("Normalization", "spec", spec)
+	_, _ = utils.ToJson(status)
```
jcantrill, line 60: @igor-karpukhin is there a purpose to this?  Looks like we should remove as there is overhead.  @syedriko maybe we can add this and any other changes into a "clean up" commit

jcantrill (commented):
/approve

internal/pkg/generator/forwarder/generator.go
syedriko: I was wondering the same thing, do this check for well-formed JSON?

jcantrill: Looks like we need a rebase but please remove.  I don't see any value here unless @igor-karpukhin can articulate otherwise

syedriko: Rebased

PR #1096, opened 2021-07-12
LOG-1267: Loki output implementation.

alanconway:
Issue: https://issues.redhat.com/browse/LOG-1267
Enhancement: https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/forward-to-loki.md

Added Loki receiver for functional and e2e tests, using Loki single-process mode.

NOTES:

Added "log_type" field to the log record: "application", "infrastructure", "audit".
The log type is related but not the same as the elasticsearch index,
which has ES-specific prefix/suffix and can be changed by the ES output.

Uses standard go testing.T tests. More compact than Ginkgo tests.
Demonstrates similar before/after, variable scope, and naming as Ginkgo tests.

/assign @jcantrill
/cc @periklis


pkg/apis/logging/v1/output_types.go
```diff
@@ -136,3 +138,23 @@ type Elasticsearch struct {
...
+	//
+	// If LabelKeys is not set, the default labels are:
+	// [log_type, kubernetes_namespace_name, kubernetes_pod_name, kubernetes_host]
+	//
+	// Note: the label kubernetes_host is always included, even if not requested.
+	// Loki requires log streams to be correctly ordered by timestamp.
+	// Including kubernetes_host in the label set ensures each stream originates from a single host,
+	// and timestamps cannot become disordered due to clock differences on different hosts.
+	//
+	// +optional
```
periklis, line 158: nit. Awesome introduction on validation of label keys. If you want to give API docs readers a full insight, I recommend to leave 1 or 2 sentences that the user selection is mandated on loki side by the rate limit config. Precisely these two flags mandate if logs are accepted due to label keys/values length:
```
# Maximum length of a label name.
# CLI flag: -validation.max-length-label-name
[max_label_name_length: <int> | default = 1024]

# Maximum length of a label value.
# CLI flag: -validation.max-length-label-value
[max_label_value_length: <int> | default = 2048]
```
Source: https://grafana.com/docs/loki/latest/configuration/#limits_config

pkg/apis/logging/v1/output_types.go
```diff
@@ -136,3 +138,23 @@ type Elasticsearch struct {
 	// +optional
 	StructuredTypeName string `json:"structuredTypeName,omitempty"`
 }
+
+// Loki provides optional extra properties for `type: loki`
+type Loki struct {
```
periklis, line 143: General question, are we omitting the ability for user to set the Tenant ID for the simple forwarding case in custom managed Loki installation. If yes, why?

jcantrill:
@alanconway  please link the needed JIRA to this task for tracking

pkg/apis/logging/v1/output_types.go
alanconway: I think I misunderstood a conversation we had. We talked about it being sufficient to use the URL path for our own managed Loki. I assumed that extended to all Loki installations. Should I re-instate that ability for user-managed Lokis?

periklis: Yes user-managed Loki's should support the TenantID field.

alanconway: /hold
Restoring tenantID field.

pkg/generators/forwarding/fluentd/loki_conf.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"fmt"
+	"strings"
+)
+
+var defaultLokiLabelKeys = []string{
+	"log_type",
+	"kubernetes.namespace_name",
+	"kubernetes.pod_name",
```
4n4nd, line 11: I don't think we should keep `kubernetes.pod_name` as a default label. Since openshift/k8s clusters can have constantly changing pod names this might lead to a high cardinality and reduce querying performance ([Docs](https://grafana.com/blog/2020/04/21/how-labels-in-loki-can-make-log-queries-faster-and-easier/#cardinality)).

pkg/apis/logging/v1/output_types.go
```diff
@@ -136,3 +138,34 @@ type Elasticsearch struct {
...
+	// TenantKey is a meta-data key field to use as the TenantID,
+	// For example: 'TenantKey: kubernetes.namespace_name` will use the kubernetes
+	// namespace as the tenant ID.
+	//
+	// +optional
+	TenantKey string `json:"tenantKey,omitempty"`
```
periklis, line 149: As mentioned on the enhancement doc, TenantKey should not be a key on one of the labels but rather an arbitrary string that forwarder users can pick based on their needs.

alanconway:
/unhold
tenantID restored, added tag for uniqueness of non-container logs.
ready for internal trials, may be some changes before release.

periklis (approved):
LGTM

periklis (changes requested):
@alanconway  This needs a `make bundle` before merging.

alanconway:
@periklis need final review and lgtm.
I had to make some changes to the label set to avoid out-of-order problems.
Explained in the  updated proposal: https://github.com/openshift/enhancements/pull/842 (also could use an LGTM)
Thanks!

alanconway:
/hold
update log_type to log_source based on  https://issues.redhat.com/browse/LOG-1592

PR #1100, opened 2021-07-14
LOG-1576: Revert utf-8 encoding and revert LOG-1569

jcantrill:
### Description
This PR:
* Reverts UTF-8 conversion introduced by [LOG-1569](https://issues.redhat.com/browse/LOG-1569)
* Drops any utf-8 conversion under the assumptions:
  * The existing filter uses #force_encoding that fundamentally has no impact
  * The point at which error occurs there is no place where a fix can be made by the collector
  * The record will be dropped for ES based on the latest changes to the fluent es plugin
  * Trying to force the encoding assuming ascii-8bit results in [LOG-1569](https://issues.redhat.com/browse/LOG-1569)

### Links
* https://issues.redhat.com/browse/LOG-1576
* Revert utf8 change of https://issues.redhat.com/browse/LOG-1569

### DIFF
--- a/pkg/generators/forwarding/fluentd/templates.go
+++ b/pkg/generators/forwarding/fluentd/templates.go
@@ -140,11 +140,6 @@ const fluentConfTemplate = `{{- define "fluentConf" -}}
 <label @INGRESS>
 
   ## filters
-  <filter **>
-    @type record_modifier
-    char_encoding utf-8
-  </filter>
-
   <filter journal>
     @type grep
     <exclude>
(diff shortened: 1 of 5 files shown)


pkg/generators/forwarding/fluentd/templates.go
```diff
@@ -138,6 +138,7 @@ const fluentConfTemplate = `{{- define "fluentConf" -}}
 #syslog input config here
 
 <label @INGRESS>
+
```
alanconway, line 141: Looks like a stray + character here. Otherwise LGTM

alanconway (approved):
/lgtm

PR #1109, opened 2021-07-20
Remove login as system:admin from test scripts.

alanconway:
The tests should do everything as kubeadmin.
Multiple users is confusing and there is no system:admin user on CRC.

/cc @jcantrill


alanconway:
/close
Not clear if this is correct or needed, AWS and CRC behave differently.
If necessary this will be re-introduced as part of LOG-1727

PR #1118, opened 2021-07-26
LOG-1444 New collector conf generation

vimalk78:
### Description
This PR attempts to improve the fluentd configuration generation.
#### Main Features:
 - Based on a tiny framework which lets define small configuration Elements
 - The Element wraps a golang template along with the data type to hold the data template needs
 - An Element addresses a small part of configuration. e.g. Container Log Source, Elasticsearch Output, UsernamePassword, CertKey etc, defined based on basic fluentd config abstractions like Match, ReLabel, Copy, Store etc
 - Using the data type + template avoids adding logic in golang templates
 - The Elements can be composed into other Elements to give rise to make a conf Section
 - A Section is a collection of Elements at a high level division of configuration. e.g. Inputs, Outputs, Ingress etc

The same basic idea can be used to generate conf for other Collector types (vector/fluent-bit)

/cc 

### Links

- JIRA: https://issues.redhat.com/secure/Dashboard.jspa


pkg/k8shandler/forwarding.go
```diff
@@ -52,27 +53,13 @@ func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig() (config s
...
 	}
 
-	generator, err := forwarding.NewConfigGenerator(
-		clusterRequest.Cluster.Spec.Collection.Logs.Type,
-		clusterRequest.includeLegacyForwardConfig(),
-		clusterRequest.includeLegacySyslogConfig(),
-		clusterRequest.useOldRemoteSyslogPlugin(),
-	)
-
-	if err != nil {
```
jcantrill, line 62: Was this moved someplace?  We lost validation that returns status that there is no collector

pkg/generator/vector/sources_test.go
```diff
@@ -0,0 +1,186 @@
...
+# Logs from containers (including openshift containers)
+[sources.container_logs]
+  auto_partial_merge = true
+  exclude_paths_glob_patterns = ["/var/log/containers/vector-*_openshift-logging_*.log", "/var/log/containers/elasticsearch-*_openshift-logging_*.log", "/var/log/containers/kibana-*_openshift-logging_*.log"]
+
+[sources.journal_logs]
+  type = "journald"
+
+# Logs from host audit
+[sources.host_audit_logs]
```
jcantrill, line 116: I don't know if vector supports wild cards like fluent for routing but we should consider flipping the naming convention to something like "audit_logs_host".  This would allow us to filter from generic to specific

pkg/generator/vector/sources_test.go
```diff
@@ -0,0 +1,186 @@
...
+						OutputRefs: []string{logging.OutputNameDefault},
+						Name:       "pipeline",
+					},
+				},
+			},
+			ExpectedConf: `
+# Logs from containers (including openshift containers)
+[sources.container_logs]
+  auto_partial_merge = true
+  exclude_paths_glob_patterns = ["/var/log/containers/vector-*_openshift-logging_*.log", "/var/log/containers/elasticsearch-*_openshift-logging_*.log", "/var/log/containers/kibana-*_openshift-logging_*.log"]
```
jcantrill, line 110: We'll need to eval how to make this useful since vector follows `/var/log/pods`

pkg/generator/fluentd/output/buffer_conf_test.go
```diff
@@ -0,0 +1,116 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/ginkgo/extensions/table"
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	. "github.com/openshift/cluster-logging-operator/pkg/generator"
+	"github.com/openshift/cluster-logging-operator/pkg/generator/fluentd/output"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var buffer_test = Describe("Generate fluentd conf", func() {
```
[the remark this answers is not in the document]
vimalk78: the variable name is there just as a documentation for a test case.

pkg/generator/fluentd/output/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,223 @@
...
+				OutputConf(bufspec, secret, o, op),
+			),
+		},
+	}
+}
+
+func OutputConf(bufspec *logging.FluentdBufferSpec, secret *corev1.Secret, o logging.OutputSpec, op *Options) []Element {
+	es := ESOutput(bufspec, secret, o, op)
+	need_retry := true
+	if need_retry {
```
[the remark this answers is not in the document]
vimalk78: i wanted to know more about this retry, do we do this based on some condition? or should this always be enabled? then i can remove the variable and `if` check

pkg/generator/generator.go
```diff
@@ -0,0 +1,140 @@
...
+type Element interface {
+	Name() string
+	Template() string
+	Create(*template.Template) *template.Template
+	Data() interface{}
+}
+
+type Elements []Element
+
+type Section struct {
```
alanconway, line 19: Hierarchy and indenting: we may need more than 2 levels of hierarchy in future.
Suggestion: Make Section implement the Element interface. Then we can have sections within sections with correct indenting. Here's a sketch:
```
package main

import (
	"bufio"
	"fmt"
	"strings"
)

type Element interface {
	Template() string
}

func indentTemplates(indent int, es []Element) string {
	pad := strings.Repeat(" ", indent)
	b := &strings.Builder{}
	for _, e := range es {
		scanner := bufio.NewScanner(strings.NewReader(e.Template()))
		for scanner.Scan() {
			fmt.Fprintf(b, "%v%v\n", pad, scanner.Text())
		}
	}
	return b.String()
}

type Section []Element

func (s Section) Template() string { return indentTemplates(0, []Element(s)) }

type Indented []Element

func (s Indented) Template() string { return indentTemplates(2, []Element(s)) }

// Dummy element, just a string
type Text struct{ Text string }

func (t Text) Template() string { return t.Text }

func main() {
	conf := Section{
		Text{"hello"},
		Indented{
			Text{"foo"},
			Indented{
				Text{"inside"},
			},
			Text{"bar"},
		},
		Text{"bye"},
	}
	fmt.Println(conf.Template())
}

```

alanconway (commented):
Looks good. I made some suggestions to simplify and make the code smaller, but the approach is good.

pkg/generator/fluentd/conf.go
```diff
@@ -0,0 +1,56 @@
+package fluentd
+
+import (
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	. "github.com/openshift/cluster-logging-operator/pkg/generator"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var Header = `
```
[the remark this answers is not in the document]
vimalk78: made into a constant

pkg/generator/fluentd/output/buffer_conf_test.go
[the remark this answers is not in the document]
vimalk78: updated

pkg/generator/fluentd/output/elasticsearch/elasticsearch.go
```diff
@@ -0,0 +1,223 @@
...
+func ESOutput(bufspec *logging.FluentdBufferSpec, secret *corev1.Secret, o logging.OutputSpec, op *Options) Elasticsearch {
+	// URL is parasable, checked at input sanitization
+	u, _ := urlhelper.Parse(o.URL)
+	port := u.Port()
+	if port == "" {
+		port = defaultElasticsearchPort
+	}
+	return Elasticsearch{
+		Desc:           "Elasticsearch store",
+		StoreID:        strings.ToLower(fmt.Sprintf("%v", helpers.Replacer.Replace(o.Name))),
```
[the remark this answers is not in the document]
vimalk78: updated

pkg/generator/vector/sources_test.go
vimalk78: i didnt get you, container log path remains same for both vector and fluent

pkg/generator/generator.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+type Element interface {
+	Name() string
+	Template() string
+	Create(*template.Template) *template.Template
+	Data() interface{}
+}
+
+type Elements []Element
```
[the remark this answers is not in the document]
vimalk78: removed this unused type

pkg/generator/generator.go
```diff
@@ -0,0 +1,140 @@
...
+
+func (g *Generator) GenerateConf(es ...Element) (string, error) {
+	conf, err := g.generate(es)
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(conf), nil
+}
+
+func (g *Generator) GenerateRec(t *template.Template, e Element, b *bytes.Buffer) error {
```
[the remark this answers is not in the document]
vimalk78: removed method

pkg/generator/generator.go
```diff
@@ -0,0 +1,140 @@
...
+	t = e.Create(t)
+	err := t.ExecuteTemplate(b, e.Name(), e.Data())
+	if err != nil {
+		fmt.Printf("error occured %v\n", err)
```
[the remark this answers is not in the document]
vimalk78: removed `fmt.Printf`

pkg/generator/generator.go
```diff
@@ -0,0 +1,140 @@
...
+		return err
+	}
+	return nil
+}
+
+func (g *Generator) compose(es []Element) (string, error) {
```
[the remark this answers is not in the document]
vimalk78: removed method, used closure

pkg/generator/generator.go
```diff
@@ -0,0 +1,140 @@
...
+	return g.generate(es)
+}
+
+func (g *Generator) compose_one(e Element) (string, error) {
```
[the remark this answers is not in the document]
vimalk78: removed method, used closure

pkg/generator/generator.go
```diff
@@ -0,0 +1,140 @@
...
+	return g.generate([]Element{
+		e,
+	})
+}
+
+func (g *Generator) generate(es []Element) (string, error) {
```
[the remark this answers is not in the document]
vimalk78: these are to be called from template

pkg/generator/generator.go
```diff
@@ -0,0 +1,140 @@
...
+var NoOptions = map[string]string{}
+
+type Generator struct {
+}
+
+func MakeGenerator() *Generator {
+	return &Generator{}
+}
+
+func (g *Generator) GenerateConfWithHeader(header string, es ...Element) (string, error) {
```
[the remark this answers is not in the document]
vimalk78: Header is added only when full conf is generated. so i added a new Element for Header. and this function is not needed.
removed `GenerateConfWithHeader` method

pkg/generator/generator.go
```diff
@@ -0,0 +1,140 @@
...
+func (g *Generator) generate(es []Element) (string, error) {
+	if len(es) == 0 {
+		return "", nil
+	}
+	t := template.New("generate")
+	t.Funcs(template.FuncMap{
+		"compose":         g.compose,
+		"compose_one":     g.compose_one,
+		"indent":          indent,
+		"comma_separated": comma_separated,
```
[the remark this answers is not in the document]
vimalk78: Modified as suggested

pkg/generator/generator.go
```diff
@@ -0,0 +1,140 @@
...
+		// 	outlines[i] = ""
+		// } else {
+		// 	outlines[i] = pad + inline
+		// }
+		outlines[i] = pad + inline
+	}
+	return strings.Join(outlines, "\n")
+}
+
+func comma_separated(arr []string) string {
```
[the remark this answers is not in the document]
vimalk78: Updated

pkg/generator/literal.go
```diff
@@ -0,0 +1,29 @@
...
+	return b.TemplateStr
+}
+
+func (b ConfLiteral) Create(t *template.Template) *template.Template {
+	return template.Must(t.Parse(b.Template()))
+}
+
+func (b ConfLiteral) Data() interface{} {
+	return b
+}
```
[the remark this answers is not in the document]
vimalk78: removed this method from the interface

pkg/generator/generator.go
```diff
@@ -0,0 +1,140 @@
...
+	"fmt"
+	"strings"
+	"text/template"
+)
+
+type Element interface {
+	Name() string
+	Template() string
+	Create(*template.Template) *template.Template
+	Data() interface{}
```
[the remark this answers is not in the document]
vimalk78: removed this method from the interfave

pkg/generator/nil.go
```diff
@@ -0,0 +1,24 @@
...
+func (n NilElement) Create(t *template.Template) *template.Template {
+	return template.Must(t.Parse(n.Template()))
+}
+
+func (n NilElement) Data() interface{} {
+	return n
+}
+
+var Nil NilElement
+var Nils []Element = []Element{Nil}
```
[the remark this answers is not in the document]
vimalk78: done

pkg/generator/generator.go
```diff
@@ -0,0 +1,140 @@
...
+
+import (
+	"bytes"
+	"fmt"
+	"strings"
+	"text/template"
+)
+
+type Element interface {
+	Name() string
```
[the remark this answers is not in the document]
vimalk78: `Name` seems to be needed , need name of the template to instantiate it.

pkg/generator/generator.go
vimalk78: `Section` is a convenience type, all the heavy lifting is done by `Element` and `[]Element` 
`Section` is used only in the first level, to show a breakdown of generated configuration + add a comment along with the declared section to document the meaning of the section.
e.g.
```
{
	PrometheusMetrics(clfspec, op),
	"Section to add measurement, and dispatch to Concat or Ingress pipelines",
},
{
	Concat(clfspec, op),
	`Concat pipeline 
	section`,
},
```
Need your and @jcantrill help to fill up the comments for each `Section`

vimalk78: for
```
type Text struct{ Text string }
func (t Text) Template() string { return t.Text }
```

we already have a type `ConfLiteral` which is used to combine a template defined as a string
https://github.com/openshift/cluster-logging-operator/pull/1118/files#diff-f589689c1969e276096b90ed4bf632a068e506f1509ffafe8d9dc3273239a775
usage: https://github.com/openshift/cluster-logging-operator/pull/1118/files#diff-5d532129029d7d81e3ad40ecb9e6b36b09c30771e0727ba74a04009197141f0cR32

vimalk78: About multiple levels, we already support that with Element

pkg/generator/vector/sources.go
```diff
@@ -0,0 +1,114 @@
...
+		[]string{
+			fmt.Sprintf("%q", fmt.Sprintf(CollectorLogsPath(), VectorCollectorPodNamePrefix())),
+			fmt.Sprintf("%q", fmt.Sprintf(LogStoreLogsPath(), ESLogStorePodNamePrefix())),
+			fmt.Sprintf("%q", VisualizationLogsPath()),
+		},
+		", ",
+	))
+}
+
+func LoggingNamespace() string {
```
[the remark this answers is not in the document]
vimalk78: updated

pkg/generator/url/url.go
```diff
@@ -0,0 +1,49 @@
+// package url provides operations on URLs and URL schemes for log forwarding destinations.
```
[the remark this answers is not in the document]
vimalk78: working on this change

pkg/generator/clo.go
```diff
@@ -0,0 +1,36 @@
+package generator
+
+import (
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+)
+
+/**
+ These unexported methods are copied frm CLO
```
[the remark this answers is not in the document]
vimalk78: working on this change

pkg/generator/fluentd/output/elasticsearch/elasticsearch.go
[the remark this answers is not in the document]
jcantrill: Elasticsearch AFAIK is the only output which has a batch submission that will reject the batch if one record fails.  The plugin has logic to remove failed records from the batch and allow retrying them.  This is only accomplished by re-routing the failed messages which in this case is to reroute them where configured.  We should always have retry for Elasticsearch but I don't know it necessarily applies to other output types.

pkg/generator/vector/sources_test.go
jcantrill: The container log path is different AFAIK between the two collectors.  We  configure fluent to use `/var/log/containers` but vector uses `/var/log/pods`.  This means the exclude pattern would change because the file name changes the order of podname and namespace.

pkg/generator/fluentd/inputs_to_pipeline.go
```diff
@@ -0,0 +1,233 @@
...
+	// routed by namespace, or labels
+	routedPipelines := ApplicationsToPipelines{}
+	unRoutedPipelines := []string{}
+	for _, pipeline := range spec.Pipelines {
+		for _, inRef := range pipeline.InputRefs {
+			if input, ok := userDefined[inRef]; ok {
+				// user defined input
+				if input.Application != nil {
+					app := input.Application
+					var a *ApplicationToPipeline = nil
```
jcantrill, line 124: I would consider initializing the type straight away since that seems to me to be the default expectation and reset it nil if needed

pkg/k8shandler/forwarding.go
```diff
@@ -59,28 +60,32 @@ func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig() (config s
...
 
-	generator, err := forwarding.NewConfigGenerator(
-		clusterRequest.Cluster.Spec.Collection.Logs.Type,
-		clusterRequest.includeLegacyForwardConfig(),
-		clusterRequest.includeLegacySyslogConfig(),
-		clusterRequest.useOldRemoteSyslogPlugin(),
-	)
+	op := generator.Options{}
+	if clusterRequest.includeLegacyForwardConfig() {
+		op[generator.IncludeLegacyForwardConfig] = ""
```
jcantrill, line 65: Does the value matter here?

pkg/k8shandler/forwarding.go
```diff
@@ -59,28 +60,32 @@ func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig() (config s
...
 				corev1.ConditionTrue,
 			)
 	}
 
-	clfSpec := &clusterRequest.ForwarderSpec
-	fwSpec := clusterRequest.Cluster.Spec.Forwarder
-
-	generatedConfig, err := generator.Generate(clfSpec, clusterRequest.OutputSecrets, fwSpec)
+	g := generator.MakeGenerator()
+	conf_sections := fluentd.Conf(&clusterRequest.Cluster.Spec, clusterRequest.OutputSecrets, &clusterRequest.ForwarderSpec, op)
```
jcantrill, line 87: Rename to camelCase which (I think) is general convention for golang

jcantrill, line 87: Seems as though this should be hidden in "MakeGenerator()" as at some point we will need to support multiple collectors based on a factory hint

jcantrill (commented):
In general, I like the molecularity offered by the changes.  It does seem, however, we repeat many things that could be pushed into helper methods that would be easier to unit test.  Think of MVC pattern where the "view" in our case is the golang template and the field to be rendered is hidden in a method on the "model" (which is our struct).  This is the general pattern by which the first version was written and I believe it still holds true here.  I don't know that it is necessary to always have a concrete struct for any given template; there may be places where we can use anonymous or common ones to achieve the same result and reuse helper logic

pkg/k8shandler/forwarding.go
vimalk78: this cant be hidden in `MakeGenerator` because the `fluentd` is the package name, not an object which can be returned from a function.
when we support multiple collectors, based on the hint we need to invoke from different package

```
        var collector = clusterRequest.Cluster.Spec.Collection.Logs.Type
	switch collector {
	case logging.LogCollectionTypeFluentd:
		confSections = fluentd.Conf(&clusterRequest.Cluster.Spec, clusterRequest.OutputSecrets, &clusterRequest.ForwarderSpec, op)
	case logging.LogCollectionTypeVector:
		confSections = vector.Conf(&clusterRequest.Cluster.Spec, clusterRequest.OutputSecrets, &clusterRequest.ForwarderSpec, op)
	case "fluent-bit":
	}

```

vimalk78: updated

pkg/k8shandler/forwarding.go
vimalk78: no, only the key is used.

internal/cmd/forwarder-generator/main.go
```diff
@@ -32,6 +32,7 @@ func main() {
 	yamlFile := flag.String("file", "", "ClusterLogForwarder yaml file. - for stdin")
 	includeDefaultLogStore := flag.Bool("include-default-store", true, "Include the default storage when generating the config")
 	includeLegacyForward := flag.Bool("include-legacy-forward", false, "Include the legacy forward when generating the config")
+	debugOutput := flag.Bool("debug-output", false, "Generate config normally, but replace output plugins with @stdout plugin,, so that records can be printed in collector logs.")
```
syedriko, line 35: Two commas

internal/pkg/generator/forwarder/generator.go
```diff
@@ -18,15 +20,22 @@ const (
 	useOldRemoteSyslogPlugin = false
 )
 
-func Generate(clfYaml string, includeDefaultLogStore, includeLegacyForward bool) (string, error) {
+func Generate(clfYaml string, includeDefaultLogStore, includeLegacyForward, debug_output bool) (string, error) {
```
syedriko, line 23: please rename debug_output to debugOutput

pkg/generator/fluentd/conf.go
```diff
@@ -0,0 +1,106 @@
...
+
+	logging "github.com/openshift/cluster-logging-operator/pkg/apis/logging/v1"
+	. "github.com/openshift/cluster-logging-operator/pkg/generator"
+	corev1 "k8s.io/api/core/v1"
+)
+
+var (
+	ErrNoValidInputs    = errors.New("No valid inputs found in ClusterLogForwarder")
+	ErrNoOutputs        = errors.New("No outputs defined in ClusterLogForwarder")
+	ErrInvalidOutputURL = errors.New("Invalid URL in one of the outputs in ClusterLogForwarder")
```
syedriko, line 16: If this is intended to be user-visible, it would be nice to say which output has the invalid URL

pkg/generator/generator.go
```diff
@@ -0,0 +1,128 @@
...
+type OutLabel = string
+
+// ComponentID is used to define a component in vector configuration, a source, sink, etc
+type ComponentID = string
+
+// Options is a map of Options used to customize the config generation. E.g. Debugging, legacy config generation
+type Options map[string]interface{}
+
+// NoOptions is used to pass empty options
+var NoOptions = map[string]interface{}{}
```
syedriko, line 35: var NoOptions = Options{}

pkg/generator/fluentd/output/security/security.go
```diff
@@ -0,0 +1,111 @@
...
+import (
+	"fmt"
+	"path/filepath"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+type TLS bool
+
+type TLSKeyCert struct {
```
syedriko, line 12: Cert then Key would be more natural

pkg/generator/fluentd/output/kafka/kafka.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+}
+
+//Brokers returns the list of broker endpoints of a kafka cluster.
+//The list represents only the initial set used by the collector's kafka client for the
+//first connention only. The collector's kafka client fetches constantly an updated list
+//from kafka. These updates are not reconciled back to the collector configuration.
+//The list of brokers are populated from the Kafka OutputSpec `Brokers` field, a list of
+//valid URLs. If none provided the target URL from the OutputSpec is used as fallback.
+//Finally, if neither approach works the current collector process will be terminated.
```
syedriko, line 89: This seems a little heavy-handed. This Kafka output can be one of many. Signaling this error in CLF status seems like a better option.

pkg/generator/fluentd/output/kafka/sasl.go
```diff
@@ -0,0 +1,22 @@
+package kafka
+
+type SaslOverSSL bool
+
+func (s SaslOverSSL) Name() string {
+	return "kafkaSaslOverSSLTemplate"
+}
+
+func (s SaslOverSSL) Template() string {
+	enabled := `{{define "kafkaSaslOverSSLTemplate" -}}
```
syedriko, line 10: Or to put it a bit shorter

return fmt.Sprintln(){{`define "kafkaSaslOverSSLTemplate" -}}
sasl_over_ssl %t
{{- end}}`, s)

pkg/generator/fluentd/output/kafka/kafka.go
vimalk78: this comment was lifted from old code as it is

pkg/generator/fluentd/sources.go
```diff
@@ -0,0 +1,96 @@
...
+	. "github.com/openshift/cluster-logging-operator/pkg/generator"
+	"github.com/openshift/cluster-logging-operator/pkg/generator/fluentd/source"
+)
+
+func Sources(spec *logging.ClusterLogForwarderSpec, o Options) []Element {
+	return MergeElements(
+		MetricSources(spec, o),
+		LogSources(spec, o),
+	)
+}
```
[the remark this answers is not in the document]
vimalk78: @pmoogi-redhat can you add some formatting to this line?

pkg/generator/fluentd/output/syslog/syslog.go
```diff
@@ -0,0 +1,395 @@
...
+	if matchtags == source.JournalTags {
+		keys = append(keys, "$.systemd.u.SYSLOG_IDENTIFIER")
+	}
+	if IsKeyExpr(s.Tag) {
+		keys = append(keys, s.Tag)
+	}
+	if IsTagExpr(s.Tag) && !tagAdded {
+		keys = append(keys, "tag")
+		tagAdded = true
+	}
```
ajaygupta978, line 269: Can we change above to following for better readabilty
       

```
if IsKeyExpr(s.Tag) {
		keys = append(keys, s.Tag)
                if  !tagAdded {
		  keys = append(keys, "tag")
		  tagAdded = true
              }
	}
	
```

pkg/generator/fluentd/output/syslog/syslog.go
```diff
@@ -0,0 +1,395 @@
...
+	return keys
+}
+
+func ParseJson(o logging.OutputSpec, op Options) Element {
+	return Filter{
+		MatchTags: "**",
+		Element: ConfLiteral{
+			TemplateName: "syslogParseJson",
+			TemplateStr: `{{define "syslogParseJson" -}}
+@type parse_json_field
```
ajaygupta978, line 305: looks like indentation issue

pkg/generator/fluentd/output/syslog/syslog.go
```diff
@@ -0,0 +1,395 @@
...
+		}
+		return conf
+	} else {
+		return nil
+	}
+}
+
+// The Syslog output fields can be set to an expression of the form $.abc.xyz
+// If an expression is used, its value will be taken from corresponding key in the record
+var keyre = regexp.MustCompile(`^\$(\.[[:word:]]*)+$`)
```
ajaygupta978, line 385: can we add one example of key and tag here?

vimalk78: Added

pkg/generator/fluentd/output/syslog/syslog.go
vimalk78: it is not an indentation issue. but added a new line in the template first line

pkg/generator/fluentd/output/syslog/syslog.go
vimalk78: no, this line is same as old code

pkg/generator/fluentd/sources_to_inputs.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/openshift/cluster-logging-operator/pkg/generator/fluentd/source"
+)
+
+func SourcesToInputs(spec *logging.ClusterLogForwarderSpec, o Options) []Element {
+	var el []Element = make([]Element, 0)
+	types := GatherSources(spec, o)
+	types = AddLegacySources(types, o)
+	if types.Has(logging.InputNameInfrastructure) {
+		el = append(el, Match{
+			Desc:      "Dont discard Infrastructure logs",
```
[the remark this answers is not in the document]
vimalk78: changed. thanks !!

pkg/generator/fluentd/output/kafka/sasl.go
vimalk78: done.

pkg/generator/fluentd/output/security/security.go
vimalk78: changed.

internal/pkg/generator/forwarder/generator.go
vimalk78: changed

internal/cmd/forwarder-generator/main.go
vimalk78: changed

pkg/generator/fluentd/inputs_to_pipeline.go
```diff
@@ -0,0 +1,233 @@
...
+					if len(app.Namespaces) != 0 {
+						if a == nil {
```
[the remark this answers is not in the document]
vimalk78: added

pkg/generator/fluentd/sources.go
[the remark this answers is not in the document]
vimalk78: you can add more paths in `@tail` plugin by adding them in the `Paths` fields of `ContainerLogs` type
```
    if types.Has(logging.InputNameApplication) || types.Has(logging.InputNameInfrastructure) {      
        el = append(el,      
            source.ContainerLogs{      
                Desc:         "Logs from containers (including openshift containers)",      
                Paths:        ContainerLogPaths(),    <<< can add more paths here  
                ExcludePaths: ExcludeContainerPaths(),      
                PosFile:      "/var/lib/fluentd/pos/es-containers.log.pos",      
                OutLabel:     "MEASURE",      
            })      
    }      
```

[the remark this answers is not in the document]
vimalk78: we don't yet support `@type apache2` as a log source.

pkg/generator/fluentd/sources.go
```diff
@@ -0,0 +1,96 @@
...
+			})
+	}
+	if types.Has(logging.InputNameApplication) || types.Has(logging.InputNameInfrastructure) {
+		el = append(el,
+			source.ContainerLogs{
+				Desc:         "Logs from containers (including openshift containers)",
+				Paths:        ContainerLogPaths(),
+				ExcludePaths: ExcludeContainerPaths(),
+				PosFile:      "/var/lib/fluentd/pos/es-containers.log.pos",
+				OutLabel:     "MEASURE",
```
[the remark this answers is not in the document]
vimalk78: a new parse plugin can be added as an `Element` anywhere in the conf

pkg/generator/fluentd/conf.go
vimalk78: updated to add the output name in the error

PR #1120, opened 2021-07-27
LOG-1411: Calculate recommended queued_chunks_limit_size with fluentd tuning parameters

k-keiichi-rh:
### Description
This PR changes:

- Encapsulate ChunkLimitSize and TotalLimitSize to remove if/else statements from template for clean-up purpose
- Echo correct chunkLimitSize which is set in ClusterLogging CR
- Add a 'TOTAL_LIMIT_SIZE' environment variable on fluentd pod
- Use the specified 'TOTAL_LIMIT_SIZE' to calculate 'queued_chunks_limit_size' when specifying 'totalLimitSize' in ClusterLogging CR

/cc @alanconway @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-1411

### DIFF
--- a/pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
+++ b/pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
@@ -27,19 +27,26 @@ const (
 )
 
 func (olc *outputLabelConf) ChunkLimitSize() string {
-	if hasBufferConfig(olc.forwarder) {
+	if hasBufferConfig(olc.forwarder) && olc.forwarder.Fluentd.Buffer.ChunkLimitSize != "" {
 		return string(olc.forwarder.Fluentd.Buffer.ChunkLimitSize)
+	} else {
+		size := ""
+		switch olc.Target.Type {
+		case loggingv1.OutputTypeFluentdForward:
+			size = "1m"
+		default:
+			size = "8m"
+		}
+		return fmt.Sprintf("\"#{ENV['BUFFER_SIZE_LIMIT'] || '%s'}\"", size)
 	}
-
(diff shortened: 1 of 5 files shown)


pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
```diff
@@ -27,19 +27,30 @@ const (
...
+	} else {
+		size := ""
+		switch olc.Target.Type {
+		case loggingv1.OutputTypeFluentdForward:
+			size = "1m"
+		case loggingv1.OutputTypeElasticsearch, loggingv1.OutputTypeKafka, loggingv1.OutputTypeSyslog:
+			size = "8m"
+		default:
+			// This should not normally happen.
+			panic(fmt.Sprintf("unknown or mismatched type: %s", olc.Target.Type))
```
jcantrill, line 41: Why the panic? Why can't the default be every thing but forward?

jcantrill (commented):
/approve

pkg/generators/forwarding/fluentd/output_fluentd_buffer.go
k-keiichi-rh: I was not sure whether or not '8m' as default chunk_limit_size is suitable or not.
But it looks like forward is only an exceptional case and the others including Loki take '8m'. 
So I will change this part to "everything but forward is '8m' as default value" as you suggested.

jcantrill:
/hold
Please squash into single commit

k-keiichi-rh:
> Please squash into single commit

I did it.

PR #1122, opened 2021-07-29
LOG-1464: Disable audit log source when only legacy syslog forwarding enabled

k-keiichi-rh:
### Description
The current legacy syslog forwarding keeps sending empty audit messages to an external rsyslog server.
This issue is caused by the behavior of the legacy method which doesn't skip to send an empty string.
As for k8s audit records, the value of "message" field is always empty string and the legacy syslog forwarding 
tries to use the "message" field from records to set payload by default. 
As a result, the external rsyslog server receives the empty audit log. This is the reason why this issue happens.

This PR disables audit log source when only legacy syslog forwarding is enabled.
There are several options to fix this issue:
  1. Fix the behavior of out_syslog and out_syslog_buffered plugins to skip to send empty string
  2. Transform audit fluentd records to inject audit information into the "message" field
  3. Disable audit source not to send audit messages to an external rsyslog server

The approach I took is the "3" option because the legacy syslog forwarding doesn't have TLS support to send logs securely.

/cc @jcantrill @alanconway 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-1464

### DIFF
--- a/pkg/generators/forwarding/fluentd/output_conf_legacy_test.go.rej
+++ b/pkg/generators/forwarding/fluentd/output_conf_legacy_test.go.rej
@@ -1,0 +1,18 @@
+--- pkg/generators/forwarding/fluentd/output_conf_legacy_test.go
++++ pkg/generators/forwarding/fluentd/output_conf_legacy_test.go
+@@ -939,15 +878,6 @@ var _ = Describe("Generating fluentd leg
+             </store>
+           </match>
+         </label>
+-        <label @_AUDIT>
+-          <match **>
+-            @type copy
+-            <store>
+-              @type relabel
+-              @label @_LEGACY_SYSLOG
+-            </store>
+-          </match>
+-        </label>
+         <label @_INFRASTRUCTURE>
+           <match **>
+             @type copy
--- a/pkg/generators/forwarding/fluentd/generators.go
+++ b/pkg/generators/forwarding/fluentd/generators.go
@@ -62,10 +62,12 @@ func (engine *ConfigGenerator) Generate(clfSpec *logging.ClusterLogForwarderSpec
 		inputs.Insert(
 			logging.InputNameInfrastructure,
 			logging.InputNameApplication,
-			logging.InputNameAudit,
 		)
+		if engine.includeLegacyForwardConfig {
+			inputs.Insert(logging.InputNameAudit)
+		}
 		for _, logType := range inputs.List() {
-			if engine.includeLegacySyslogConfig {
+			if engine.includeLegacySyslogConfig && logType != logging.InputNameAudit {
 				routeMap.Insert(logType, constants.LegacySyslog)
 			}
 			if engine.includeLegacyForwardConfig {
--- a/pkg/generators/forwarding/fluentd/output_conf_legacy_test.go
+++ b/pkg/generators/forwarding/fluentd/output_conf_legacy_test.go
@@ -645,66 +645,6 @@ var _ = Describe("Generating fluentd legacy output store config blocks", func() 
             </pattern>
           </parse>
         </source>
-        # linux audit logs
-        <source>
-          @type tail
-          @id audit-input
-          @label @MEASURE
-          path "/var/log/audit/audit.log"
-          pos_file "/var/lib/fluentd/pos/audit.log.pos"
-          tag linux-audit.log
-          <parse>
-            @type viaq_host_audit
-          </parse>
-        </source>
-        # k8s audit logs
-        <source>
-          @type tail
-          @id k8s-audit-input
-          @label @MEASURE
-          path "/var/log/kube-apiserver/audit.log"
-          pos_file "/var/lib/fluentd/pos/kube-apiserver.audit.log.pos"
-          tag k8s-audit.log
-          <parse>
-            @type json
-            time_key requestReceivedTimestamp
-            # In case folks want to parse based on the requestReceivedTimestamp key
-            keep_time_key true
-            time_format %Y-%m-%dT%H:%M:%S.%N%z
-          </parse>
-        </source>
-
-        # Openshift audit logs
-        <source>
-          @type tail
-          @id openshift-audit-input
-          @label @MEASURE
-          path /var/log/oauth-apiserver/audit.log,/var/log/openshift-apiserver/audit.log
-          pos_file /var/lib/fluentd/pos/oauth-apiserver.audit.log
-          tag openshift-audit.log
-          <parse>
-            @type json
-            time_key requestReceivedTimestamp
-            # In case folks want to parse based on the requestReceivedTimestamp key
-            keep_time_key true
-            time_format %Y-%m-%dT%H:%M:%S.%N%z
-          </parse>
-        </source>
-        # Openshift Virtual Network (OVN) audit logs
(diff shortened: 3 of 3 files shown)


jcantrill:
> ### Description
> The current legacy syslog forwarding keeps sending empty audit messages to an external rsyslog server.
> This issue is caused by the behavior of the legacy method which doesn't skip to send an empty string.
> As for k8s audit records, the value of "message" field is always empty string and the legacy syslog forwarding

Why is the field empty?  Are we improperly transforming the k8s record?  Do we not have the same output contract?

> 1. Fix the behavior of out_syslog and out_syslog_buffered plugins to skip to send empty string
> 2. Transform audit fluentd records to inject audit information into the "message" field
> 3. Disable audit source not to send audit messages to an external rsyslog server
> 
> The approach I took is the "3" option because the legacy syslog forwarding doesn't have TLS support to send logs securely.

Is this considered a requirement of shipping audit logs to have TLS support?  That seems reasonable  but I am not certain I have seen that expressed anywhere.

@alanconway @vimalk78 do you have thoughts about the "audit" envelope? This solution drops messages which seems like instead we should figure out why the messages are empty to begin with

k-keiichi-rh:
> Why is the field empty? Are we improperly transforming the k8s record? Do we not have the same output contract?

The original k8s audit records themselves don't have the "message" field and the fluentd pipeline in OpenShift logging basically doesn't change the k8s audit records except for [changing the value of "level" key and adding "{k8s|openshift}_audit_level" key](https://github.com/openshift/cluster-logging-operator/blob/4c3f8099af3fec5ed6b3cc4ad2752bae16053a58/pkg/generators/forwarding/fluentd/templates.go#L278).

Now there are the following patterns to send k8s audit records via syslog:
1. Send it by the legacy syslog way
    Empty audit messages are sent because it sends the "message" key field only by default and empty string is not skipped to send
2. Send it by the new syslog way
    Empty audit messages are not sent because it sends whole audit records by default.
3. Send it by the new syslog way with "payloadKey: message" in CLF
    Empty audit messages are not sent because it sends the "message" key field only and the value of "message" key is empty, but the new syslog way tries to skip empty string.

FYI, the k8s audit records originates from /var/log/kube-apiserver/audit.log, /var/log/oauth-apiserver/audit.log and /var/log/openshift-apiserver/audit.log. And they have ["Event" json objects](https://kubernetes.io/docs/reference/config-api/apiserver-audit.v1/#audit-k8s-io-v1-Event).

The following example is one of k8s audit records from /var/log/kube-apiserver/audit.log.
```
{
  "kind": "Event",
  "apiVersion": "audit.k8s.io/v1",
  "level": "Metadata",
  "auditID": "28db3fcf-f86b-4d43-bb34-e6cb08ca5829",
  "stage": "ResponseComplete",
  "requestURI": "/api/v1/namespaces/kube-system/configmaps?allowWatchBookmarks=true&fieldSelector=metadata.name%3Dextension-apiserver-authentication&resource
Version=1651669&timeout=7m11s&timeoutSeconds=431&watch=true",
  "verb": "watch",
  "user": {
    "username": "system:serviceaccount:openshift-console-operator:console-operator",
    "uid": "dcd9f592-0bc5-4455-940e-3a9802d16b41",
    "groups": [
      "system:serviceaccounts",
      "system:serviceaccounts:openshift-console-operator",
      "system:authenticated"
    ]
  },
  "sourceIPs": [
    "10.217.0.19"
  ],
  "userAgent": "console/v0.0.0 (linux/amd64) kubernetes/$Format",
  "objectRef": {
    "resource": "configmaps",
    "namespace": "kube-system",
    "name": "extension-apiserver-authentication",
    "apiVersion": "v1"
  },
  "responseStatus": {
    "metadata": {},
    "status": "Success",
    "message": "Connection closed early",
    "code": 200
  },
  "requestReceivedTimestamp": "2021-08-02T20:05:35.718263Z",
  "stageTimestamp": "2021-08-02T20:05:45.718737Z",
  "annotations": {
    "authentication.k8s.io/legacy-token": "system:serviceaccount:openshift-console-operator:console-operator",
    "authorization.k8s.io/decision": "allow",
    "authorization.k8s.io/reason": "RBAC: allowed by RoleBinding \"console-operator/kube-system\" of Role \"extension-apiserver-authentication-reader\" to ServiceAccount \"console-operator/openshift-console-operator\""
  }
}

```

> Is this considered a requirement of shipping audit logs to have TLS support? That seems reasonable but I am not certain I have seen that expressed anywhere.

No, I have not seen it. I jumped to conclusions without having a discussion. I thought this way was reasonable from [the following sentense](https://docs.openshift.com/container-platform/4.7/logging/config/cluster-logging-log-store.html#cluster-logging-elasticsearch-audit_cluster-logging-store), but we need to have a discussion whether no TLS support is a reasonable reason same as not providing secure storage or not.  
```
Because the internal OpenShift Container Platform Elasticsearch log store does not provide secure storage for audit logs, by default audit logs are not stored in the internal Elasticsearch instance. 
```
from

alanconway:
What is the format of our forwarded log record? Is it the unmodified audit log, or is it our format with an empty `message` and the audit log in the `event` field? A short term fix might be to just add the content to the `message` field as well. Need to think about the proper long-term fix...

k-keiichi-rh:
> What is the format of our forwarded log record? Is it the unmodified audit log, or is it our format with an empty `message` and the audit log in the `event` field? A short term fix might be to just add the content to the `message` field as well. Need to think about the proper long-term fix...

Quoted from other channel we discussed:
As for the new syslog forwarding, we forward the raw audit log with minor transforming things. So the proper long-term fix is required. As a workaround, we can set up a pipeline for audit logs only and remove "payload:message" in CLF.
As for the legacy forwarding,  just removing audit source is fine. Because we can't split audit log source with the others in the legacy way.

alanconway:
@k-keiichi-rh I'm going on vacation next 2 weeks, will let @jcantrill follow up on this. 
It is not a long-term solution, but since it only affects legacy syslog it probably is an acceptable workaround for the short term. 
I've created https://issues.redhat.com/browse/LOG-1654 to track the long term solution. Please comment there if you have any insights from your perspective.

jcantrill:
In interest of not overwhelming QE for closing our 5.2 release, I'm going to refrain from release the hold until after we branch.  We should cherry pick this back to 5.2 in the first z stream (and earlier where needed).

pkg/generators/forwarding/fluentd/output_conf_legacy_test.go.rej
```diff
@@ -0,0 +1,18 @@
...
+-            @type copy
+-            <store>
+-              @type relabel
+-              @label @_LEGACY_SYSLOG
+-            </store>
+-          </match>
+-        </label>
+         <label @_INFRASTRUCTURE>
+           <match **>
+             @type copy
```
vimalk78, line 18: @k-keiichi-rh please remove this file.
Cc: @jcantrill

k-keiichi-rh: I am sorry for my fault. I submitted https://github.com/openshift/cluster-logging-operator/pull/1140.

k-keiichi-rh:
I will fix the patch for release-5.2.

PR #1125, opened 2021-07-30
Add Dockerfile to build and promote source to gating jobs

periklis:
### Description
This PR addresses preparation issues to resolve openshift/release#20604. In detail it provides a Dockerfile that will enable our upstream CI to build and promote the source code to other jobs, e.g. to EO gating jobs.

/cc @jcantrill 
/assign @jcantrill


periklis:
Not the right approach to promote the bundle, we should let the CI promote the full bundle.

PR #1130, opened 2021-08-05
LOG-1587: functional test for kafka - PR attempt

pmoogi-redhat:
### Description
Moving kafka output testing to the functional framework,
so that it can speed up e2e testing and isolate kafka output testing from uncontrollable cluster variables.

Scope of this functional test - validating if we can produce a fluentd configuration that allows fluent to write messages to kafka. This only requires that fluent be able to: 1. connect to kafka 2. push a message

/cc @jcantrill @vimalk78
/assign @jcantrill 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-1587
- Enhancement proposal:

### DIFF
--- a/test/functional/framework.go
+++ b/test/functional/framework.go
@@ -53,14 +53,20 @@ var outputLogFile = map[string]map[string]string{
 	logging.OutputTypeFluentdForward: {
 		applicationLog: ApplicationLogFile,
 		auditLog:       "/tmp/audit-logs",
-		ovnAuditLog:    "/tmp/audit-logs",
-		k8sAuditLog:    "/tmp/audit-logs",
+		ovnAuditLog:    "/tmp/ovnaudit-logs",
+		k8sAuditLog:    "/tmp/k8audit-logs",
 	},
 	logging.OutputTypeSyslog: {
-		applicationLog: "/var/log/infra.log",
-		auditLog:       "/var/log/infra.log",
-		k8sAuditLog:    "/var/log/infra.log",
-		ovnAuditLog:    "/var/log/infra.log",
+		applicationLog: "/var/log/app.log",
+		auditLog:       "/var/log/audit.log",
+		k8sAuditLog:    "/var/log/k8audit.log",
+		ovnAuditLog:    "/var/log/ovnaudit.log",
+	},
+	logging.OutputTypeKafka: {
+		applicationLog: "/var/log/app.log",
+		auditLog:       "/var/log/audit.log",
+		k8sAuditLog:    "/var/log/k8audit.log",
+		ovnAuditLog:    "/var/log/ovnaudit.log",
 	},
 }
 
@@ -251,7 +257,7 @@ done
 	role := runtime.NewRole(f.Test.NS.Name, f.Name,
 		v1.PolicyRule{
 			Verbs:     []string{"list", "get"},
-			Resources: []string{"pods", "namespaces"},
+			Resources: []string{"nodes", "pods", "namespaces"},
 			APIGroups: []string{""},
 		},
 	)
@@ -355,6 +361,10 @@ func (f *FluentdFunctionalFramework) addOutputContainers(b *runtime.PodBuilder, 
 			}
 		case logging.OutputTypeSyslog:
 			if err := f.addSyslogOutput(b, output); err != nil {
+				return err
+			}
+		case logging.OutputTypeKafka:
+			if err := f.addKafkaOutput(b, output); err != nil {
 				return err
 			}
 		case logging.OutputTypeElasticsearch:
@@ -469,7 +479,11 @@ func (f *FluentdFunctionalFramework) WriteMessagesToLog(msg string, numOfLogs in
 func (f *FluentdFunctionalFramework) ReadApplicationLogsFrom(outputName string) ([]string, error) {
 	return f.ReadLogsFrom(outputName, applicationLog)
 }
-
+func (f *FluentdFunctionalFramework) ReadApplicationLogsFromKafka(topic string, brokerlistener string, consumercontainername string) (string, error) {
+	cmd := fmt.Sprintf(`./bin/kafka-verifiable-consumer.sh --topic %s --group-id 0  --broker-list %s`, topic, brokerlistener)
+	results, err := f.RunCommand(consumercontainername, "bash", "-c", cmd)
+	return results, err
+}
 func (f *FluentdFunctionalFramework) ReadAuditLogsFrom(outputName string) ([]string, error) {
 	return f.ReadLogsFrom(outputName, auditLog)
 }
--- a/test/functional/framework.go
+++ b/test/functional/framework.go
@@ -53,20 +53,20 @@ var outputLogFile = map[string]map[string]string{
 	logging.OutputTypeFluentdForward: {
 		applicationLog: ApplicationLogFile,
 		auditLog:       "/tmp/audit-logs",
-		ovnAuditLog:    "/tmp/ovnaudit-logs",
-		k8sAuditLog:    "/tmp/k8audit-logs",
+		ovnAuditLog:    "/tmp/audit-logs",
+		k8sAuditLog:    "/tmp/audit-logs",
 	},
 	logging.OutputTypeSyslog: {
-		applicationLog: "/var/log/app.log",
-		auditLog:       "/var/log/audit.log",
-		k8sAuditLog:    "/var/log/k8audit.log",
-		ovnAuditLog:    "/var/log/ovnaudit.log",
+		applicationLog: "/var/log/infra.log",
+		auditLog:       "/var/log/infra.log",
+		k8sAuditLog:    "/var/log/infra.log",
+		ovnAuditLog:    "/var/log/infra.log",
 	},
 	logging.OutputTypeKafka: {
 		applicationLog: "/var/log/app.log",
-		auditLog:       "/var/log/audit.log",
-		k8sAuditLog:    "/var/log/k8audit.log",
-		ovnAuditLog:    "/var/log/ovnaudit.log",
+		auditLog:       "/var/log/infra.log",
+		k8sAuditLog:    "/var/log/audit.log",
+		ovnAuditLog:    "/var/log/ovnaudit-log.log",
 	},
 }
 
--- a/test/functional/output_kafka.go
+++ b/test/functional/output_kafka.go
@@ -1,0 +1,148 @@
+package functional
+
+import (
+	"fmt"
+	"github.com/ViaQ/logerr/log"
+	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/test/helpers/kafka"
+	"github.com/openshift/cluster-logging-operator/test/runtime"
+	"strconv"
+	"strings"
+)
+
+const (
+	// Kafka deployment definitions
+	ImageRemoteKafka            = kafka.KafkaImage
+	ImageRemoteKafkaInit        = kafka.KafkaInitUtilsImage
+	kafkaBrokerContainerName    = "broker"
+	kafkaZookeeperContainerName = "zookeeper"
+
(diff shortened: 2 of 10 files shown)


test/functional/framework.go
```diff
@@ -60,6 +60,11 @@ var outputLogFile = map[string]map[string]string{
 		k8sAuditLog:    "/var/log/infra.log",
 		ovnAuditLog:    "/var/log/infra.log",
 	},
+	logging.OutputTypeKafka: {
+		applicationLog: "/var/log/infra.log",
```
jcantrill, line 64: I assume you want to send these to different log files to facilitate verification?

test/functional/output_kafka.go
```diff
@@ -0,0 +1,85 @@
...
+	//steps to deploy kafka single node cluster a. Deploy Zookeeper b. Deploy Broker c. Deploy kafka Consumer
+
+	//step a
+	// Deploy Zookeeper steps : create configmap, create statefulset?, create zookeeper service
+	zookeepercm := kafka.NewZookeeperConfigMap(OpenshiftLoggingNS)
+	log.V(2).Info("Creating zookeeper configmap", "namespace", zookeepercm.Namespace, "name", zookeepercm.Name)
+	if err := f.Test.Client.Create(zookeepercm); err != nil {
+		return err
+	}
+	zookeeperapp := kafka.NewZookeeperStatefuleSet(OpenshiftLoggingNS)
```
jcantrill, line 30: We likely only need the container definition to mount into the functional test pod.  There should be no need for a SS since we can use ephemeral storage.  This should be a small, controlled test that does not need storage

test/functional/output_kafka.go
```diff
@@ -0,0 +1,85 @@
...
+	log.V(2).Info("Creating zookeeper StatefulSet", "namespace", zookeeperapp.Namespace, "name", zookeeperapp.Name)
+	if err := f.Test.Client.Create(zookeeperapp); err != nil {
+		return err
+	}
+	zookeepersvc := kafka.NewZookeeperService(OpenshiftLoggingNS)
```
jcantrill, line 35: Similar to the statefulset comment, you may not require a service since the container will live in the same pod as the functional test.  It may require config changes that would be reflected in the configmap

test/functional/output_kafka.go
```diff
@@ -0,0 +1,85 @@
...
+	log.V(2).Info("Creating zookeeper ZookeeperService", "namespace", zookeepersvc.Namespace, "name", zookeepersvc.Name)
+	if err := f.Test.Client.Create(zookeepersvc); err != nil {
+		return err
+	}
+
+	//step b
+	// Deploy Broker steps : create RBAC, create configmap, create secret, create broker service, create statefulset
+	brokercr, _ := kafka.NewBrokerRBAC(OpenshiftLoggingNS)
```
jcantrill, line 43: You should not need RBAC if the broker container is added to the functional pod

test/functional/output_kafka.go
```diff
@@ -0,0 +1,85 @@
...
+	if err := f.Test.Client.Create(brokercm); err != nil {
+		return err
+	}
+	brokersecret := kafka.NewBrokerSecret(OpenshiftLoggingNS)
+	log.V(2).Info("Creating Broker Secret", "namespace", brokersecret.Namespace, "name", brokersecret.Name)
+	if err := f.Test.Client.Create(brokersecret); err != nil {
+		return err
+	}
+
+	brokersvc := kafka.NewBrokerService(OpenshiftLoggingNS)
```
jcantrill, line 60: Same comment regrading zookeeper

test/functional/output_kafka.go
```diff
@@ -0,0 +1,85 @@
...
+	log.V(2).Info("Creating Broker broker service", "namespace", brokersvc.Namespace, "name", brokersvc.Name)
+	if err := f.Test.Client.Create(brokersvc); err != nil {
+		return err
+	}
+
+	brokerapp := kafka.NewBrokerStatefuleSet(OpenshiftLoggingNS)
```
jcantrill, line 66: Should only require the container definition to be added to the functional pod

test/functional/output_kafka.go
```diff
@@ -0,0 +1,85 @@
...
+	if err := f.Test.Client.Create(brokerapp); err != nil {
+		return err
+	}
+
+	//step c
+	topics := []string{kafka.DefaultTopic}
+
+	for _, topic := range topics {
+		// Deploy consumer
+		consumerapp := kafka.NewKafkaConsumerDeployment(OpenshiftLoggingNS, topic)
```
jcantrill, line 77: should only need the container definition for the pod

test/functional/outputs/forward_to_kafka_test.log
```diff
@@ -0,0 +1,175 @@
+package outputs
```
jcantrill, line 1: should this be named "*.go"

test/functional/outputs/forward_to_kafka_test.log
```diff
@@ -0,0 +1,175 @@
...
+	setKafkaSpecValues := func(outspec *logging.OutputSpec) {
+		outspec.Kafka = &logging.Kafka{
+			Topic:  kafka.AppLogsTopic,
+		}
+	}
+
+
+	timestamp := "2013-03-28T14:36:03.243000+00:00"
+
+	Context("Application Logs", func() {
```
jcantrill, line 39: I do not believe we need to ship more then one logtype unless you find logic in the kafka config that is specific to a given log type.  Maybe there is a need for multiple types to ensure we configure multiple topics  correctly?

test/functional/outputs/forward_to_kafka_test.log
```diff
@@ -0,0 +1,175 @@
...
+		It("should send large message over Kafa", func() {
```
jcantrill, line 40: This may be a valid alternate case if we need to account for "large message"

test/functional/outputs/forward_to_kafka_test.log
```diff
@@ -0,0 +1,175 @@
...
+			Expect(framework.Deploy()).To(BeNil())
+
+			// Log message data
+			for _, log := range K8sAuditLogs {
+				log = test.Escapelines(log)
+				//log = functional.NewFullCRIOLogMessage(timestamp, log)
+				Expect(framework.WriteMessagesTok8sAuditLog(log, 1)).To(BeNil())
+			}
+
+			time.Sleep(time.Minute * 5)
```
jcantrill, line 75: We should have no need for sleeps in functional tests

test/functional/outputs/forward_to_kafka_test.log
```diff
@@ -0,0 +1,175 @@
...
+			// Read line from Kafka output
+			outputlogs, err := framework.ReadAuditLogsFrom(logging.OutputTypeKafka)
+			Expect(err).To(BeNil(), "Expected no errors reading the logs")
+			Expect(outputlogs).ToNot(BeEmpty())
+			for _, o := range outputlogs {
+				fmt.Printf("log received %s\n", o)
+			}
+			fields := strings.Split(outputlogs[0], " ")
+			Expect(getAppName(fields)).To(Equal("myapp"))
```
jcantrill, line 84: Are these syslog expectations?

test/functional/outputs/forward_to_kafka_test.log
```diff
@@ -0,0 +1,175 @@
...
+			Expect(getProcID(fields)).To(Equal("myproc"))
+			Expect(getMsgID(fields)).To(Equal("mymsg"))
+		})
+	})
+})
+
+var (
+	JSONApplicationLogs = []string{
```
jcantrill, line 114: I assume we dont need these var samples?

test/helpers/kafka_conf.go
```diff
@@ -0,0 +1,298 @@
+package helpers
```
jcantrill, line 1: If we need to make these specific to the functional framework they may need to be added into test/functional  It depends if there is network info that makes specific assumptions about the containers and how the servers are found

test/functional/framework.go
pmoogi-redhat: @jcantrill i just followed - https://github.com/openshift/cluster-logging-operator/blob/ed0ac043934a256886613975f242163b16a9b697/test/functional/framework.go ... you are right ..these should different log files ...

test/functional/outputs/forward_to_kafka_test.log
eranra: Looks like this file is still in progress... multiple errors etc .

eranra: Also, without this file it is impossible to "execute" any tests so can't really test the functionality of this PR

test/functional/output_kafka.go
```diff
@@ -0,0 +1,102 @@
...
+	kafkaNodeReaderBinding      = "kafka-node-reader-binding"
+	kafkaInsidePort             = int32(9093)
+	kafkaOutsidePort            = int32(9094)
+	kafkaJMXPort                = int32(5555)
+	zookeeperClientPort         = int32(2181)
+	zookeeperPeerPort           = int32(2888)
+	zookeeperLeaderElectionPort = int32(3888)
+)
+
+func (f *FluentdFunctionalFramework) addKafkaOutput(b *runtime.PodBuilder, output logging.OutputSpec) error {
```
eranra, line 30: @pmoogi-redhat you are using two different "creation" approaches in this function (a) and (b) directly using the builder and (c) using the "old" helper ... I suggest moving to one of the options ... either rely on the "old" way of creating objects like (c) or more to new ... haveing a mix will be support hell in the future.

test/functional/output_kafka.go
```diff
@@ -0,0 +1,102 @@
...
+	// Deploy Broker steps : create configmap, create container, create broker service
+
+	brokercm := kafka.NewBrokerConfigMap(OpenshiftLoggingNS)
+	log.V(2).Info("Creating Broker ConfigMap", "namespace", brokercm.Namespace, "name", brokercm.Name)
+	if err := f.Test.Client.Create(brokercm); err != nil {
+		return err
+	}
+
+	//standup pod with container running broker
+	log.V(2).Info("Adding container", "name", name)
```
eranra, line 71: it is the same name as in https://github.com/openshift/cluster-logging-operator/pull/1130/files#diff-356bda7067f17250b18d697ae80688cdaedf875ce56114ec1215413676ff97c8R46 you want to update name before creating this second container

test/functional/outputs/forward_to_kafka_test.log
jcantrill: @pmoogi-redhat  please also move to `test/functional/outputs/kafka/....`

test/functional/output_kafka.go
```diff
@@ -0,0 +1,102 @@
...
+		AddVolumeMount("brokerlogs", "/opt/kafka/logs", "", false).
+		AddEnvVar("extensions", "/opt/kafka/libs/extensions").
+		AddEnvVar("data", "/var/lib/kafka/data").
+		AddEnvVar("KAFKA_LOG4J_OPTS", "-Dlog4j.configuration=file:/etc/kafka/log4j.properties").
+		AddEnvVar("JMX_PORT", strconv.Itoa(int(kafkaJMXPort))).
+		AddContainerPort("inside", kafkaInsidePort).
+		AddContainerPort("outside", kafkaOutsidePort).
+		AddContainerPort("inside", kafkaJMXPort).
+		WithCmdArgs([]string{"./bin/kafka-server-start.sh", "/etc/kafka/server.properties"}).
+		WithPrivilege().
```
syedriko, line 84: Out of curiosity, why does the Kafka container need to be privileged?

pmoogi-redhat:
closing this as kafka-functional-test-LOG-1587 got the implementation of kafka functional test rebased on latest upstream/master
https://github.com/openshift/cluster-logging-operator/pull/1183 so this PR #1130 no longer needed.

PR #1131, opened 2021-08-05
LOG-1390: Allow optional fields in message test framework

jcantrill:
### Description
This PR:
* adds optional fields to the message test framework in order to support consuming plugins (e.g. metadata) that provide new fields
* Without this change we will be required to manual merge the changes because of circular deps

### Links
https://issues.redhat.com/browse/LOG-1390

cc @vimalk78 @eranra


jcantrill:
/hold

for 5.2 FF

PR #1141, opened 2021-08-17
Added command to fetch indices size and creation time in file indices…

rhn-support-achakrat:
…_size

### Description
Added command to fetch indices size and creation time in file indices

Modified file "gather_logstore_resources" to capture the indices information which includes indices size in human-readable format, indices date of creation.


jcantrill (commented):
/approve
/lgtm

PR #1142, opened 2021-08-18
Collecting buffer size inside fluentd pods for every output pipeline

devenkulkarni:
### Description

Added a command to capture all fluentd pods buffer sizes for every output pipeline as after configuring CLF, multiple output pipelines are created with respect to buffer directory. This helps in troubleshooting.

Modified file: gather_collection_resources
Added commands: 

~~~
    echo "$pod" >> $collector_folder/output-buffer-size.txt
    echo "---------------" >> $collector_folder/output-buffer-size.txt
    oc -n openshift-logging exec $pod -- bash -c ' du -sh /var/lib/fluentd/*' >> $collector_folder/output-buffer-size.txt
~~~

### DIFF
--- a/must-gather/collection-scripts/gather_collection_resources
+++ b/must-gather/collection-scripts/gather_collection_resources
@@ -68,4 +68,7 @@ do
     check_collector_persistence $pod
     oc -n openshift-logging exec -- ls -l /var/lib/fluentd/clo_default_output_es > $collector_folder/$pod.es-buffers.txt
     oc -n openshift-logging exec -- ls -l /var/lib/fluentd/retry_clo_default_output_es > $outdir/$pod.buffers.es-retry.txt
+    echo "$pod" >> $collector_folder/output-buffer-size.txt
+    echo "---------------" >> $collector_folder/output-buffer-size.txt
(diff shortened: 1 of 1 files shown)


must-gather/collection-scripts/gather_collection_resources
```diff
@@ -68,4 +68,7 @@ do
     check_collector_persistence $pod
     oc -n openshift-logging exec -- ls -l /var/lib/fluentd/clo_default_output_es > $collector_folder/$pod.es-buffers.txt
     oc -n openshift-logging exec -- ls -l /var/lib/fluentd/retry_clo_default_output_es > $outdir/$pod.buffers.es-retry.txt
+    echo "$pod" >> $collector_folder/all-pod-buffer-size.txt
```
jcantrill, line 71: this would be more appropriately named something related to output buffers.  These are buffers associated with fluentd pods but the buffers are for each output configured in the CLF.

devenkulkarni: Yes, hence I have used a "*" to capture buffer sizes since fluentd mostly errors out with "buffer" related messages and this is the first thing that can be checked. Using wildcard would capture all the files under `/var/lib/fluentd` accordingly.

devenkulkarni: I have renamed the output file, please review and let me know if you have any other suggestions.

PR #1143, opened 2021-08-23
Bug 1976692: fluentd total_limit_size fix  

ajaygupta978:
### Description
This PR fixes buffer memory size issue.

/cc @jcantrill 
/cc @vimalk78 
/assign @jcantrill

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1976692
- Github issue:
- JIRA: https://issues.redhat.com/browse/OCPBUGSM-31512
- Enhancement proposal:

### DIFF
--- a/files/fluentd/run.sh
+++ b/files/fluentd/run.sh
@@ -44,10 +44,9 @@ fi
 # Calculate the max allowed for each output buffer given the number of
 # buffer file paths
 ###
-
 NUM_OUTPUTS=$(grep "path.*'$FILE_BUFFER_PATH" $FLUENT_CONF | wc -l)
 if [ $NUM_OUTPUTS -eq 0 ]; then
-    # Reset to default single output if log forwarding outputs all invalid
+    # Reset to default single output if log forwarding outputs all invalid.
     NUM_OUTPUTS=1
 fi
 
@@ -68,17 +67,32 @@ fi
 # Determine allowed total given the number of outputs we have.
 ALLOWED_DF_LIMIT=$(expr $DF_LIMIT \* $ALLOWED_PERCENT_OF_DISK / 100) || :
 
-# TOTAL_LIMIT_SIZE per buffer
-TOTAL_LIMIT_SIZE=$(expr $ALLOWED_DF_LIMIT / $NUM_OUTPUTS) || :
-echo "Setting each total_size_limit for $NUM_OUTPUTS buffers to $TOTAL_LIMIT_SIZE bytes"
-export TOTAL_LIMIT_SIZE
+# total limit size allowed  per buffer
+TOTAL_LIMIT_SIZE_ALLOWED_PER_BUFFER=$(expr $ALLOWED_DF_LIMIT / $NUM_OUTPUTS) || :
+TOTAL_LIMIT_SIZE_ALLOWED_PER_BUFFER=${TOTAL_LIMIT_SIZE_ALLOWED_PER_BUFFER:-0}
+
+TOTAL_LIMIT_SIZE_PER_BUFFER=${TOTAL_LIMIT_SIZE_PER_BUFFER:-0}
+TOTAL_LIMIT_SIZE_PER_BUFFER=$(echo $TOTAL_LIMIT_SIZE_PER_BUFFER |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc) || :
+
+if [[ $TOTAL_LIMIT_SIZE_PER_BUFFER -lt $TOTAL_LIMIT_SIZE_ALLOWED_PER_BUFFER ]];
+then
+   if [[ $TOTAL_LIMIT_SIZE_PER_BUFFER -eq 0 ]]; then
+       TOTAL_LIMIT_SIZE_PER_BUFFER=$TOTAL_LIMIT_SIZE_ALLOWED_PER_BUFFER
+   fi
+else
+    echo "Requested buffer size per output $TOTAL_LIMIT_SIZE_PER_BUFFER for $NUM_OUTPUTS buffers exceeds maximum available size  $TOTAL_LIMIT_SIZE_ALLOWED_PER_BUFFER bytes per output"
+    TOTAL_LIMIT_SIZE_PER_BUFFER=$TOTAL_LIMIT_SIZE_ALLOWED_PER_BUFFER
+fi
+
+echo "Setting each total_size_limit for $NUM_OUTPUTS buffers to $TOTAL_LIMIT_SIZE_PER_BUFFER bytes"
+export TOTAL_LIMIT_SIZE_PER_BUFFER
 
 ##
 # Calculate the max number of queued chunks given the size of each chunk
 # and the max allowed space per buffer
 ##
 BUFFER_SIZE_LIMIT=$(echo ${BUFFER_SIZE_LIMIT:-8m} |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc)
-BUFFER_QUEUE_LIMIT=$(expr $TOTAL_LIMIT_SIZE / $BUFFER_SIZE_LIMIT)
+BUFFER_QUEUE_LIMIT=$(expr $TOTAL_LIMIT_SIZE_PER_BUFFER / $BUFFER_SIZE_LIMIT)
 echo "Setting queued_chunks_limit_size for each buffer to $BUFFER_QUEUE_LIMIT"
 export BUFFER_QUEUE_LIMIT
 echo "Setting chunk_limit_size for each buffer to $BUFFER_SIZE_LIMIT"
--- a/pkg/generators/forwarding/fluentd/fluent_conf_test.go
+++ b/pkg/generators/forwarding/fluentd/fluent_conf_test.go
@@ -621,7 +621,7 @@ var _ = Describe("Generating fluentd config", func() {
           retry_forever true
           queued_chunks_limit_size "#{ENV['BUFFER_QUEUE_LIMIT'] || '32' }"
           
-          total_limit_size "#{ENV['TOTAL_LIMIT_SIZE'] ||  8589934592 }" #8G
+          total_limit_size "#{ENV['TOTAL_LIMIT_SIZE_PER_BUFFER'] ||  8589934592 }" #8G
           
           
           chunk_limit_size "#{ENV['BUFFER_SIZE_LIMIT'] || '8m'}"
@@ -671,7 +671,7 @@ var _ = Describe("Generating fluentd config", func() {
           retry_forever true
           queued_chunks_limit_size "#{ENV['BUFFER_QUEUE_LIMIT'] || '32' }"
           
-          total_limit_size "#{ENV['TOTAL_LIMIT_SIZE'] ||  8589934592 }" #8G
+          total_limit_size "#{ENV['TOTAL_LIMIT_SIZE_PER_BUFFER'] ||  8589934592 }" #8G
           
           
           chunk_limit_size "#{ENV['BUFFER_SIZE_LIMIT'] || '8m'}"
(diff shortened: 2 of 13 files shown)


ajaygupta978:
@jcantrill: I did tested manually. Please suggest what kind of tests needs to be added.

vimalk78 (changes requested):
PR seems to be adding a new file `run.sh` at repo root
`run.sh` should be generated by operator at run time.

files/fluentd/run.sh
```diff
@@ -68,8 +67,31 @@ fi
 # Determine allowed total given the number of outputs we have.
 ALLOWED_DF_LIMIT=$(expr $DF_LIMIT \* $ALLOWED_PERCENT_OF_DISK / 100) || :
 
+TOTAL_LIMIT_SIZE_CONF=$(grep "total_limit_size" $FLUENT_CONF |  head -1 | awk '{print $2}' )
```
jcantrill, line 70: This is unrealiable if there are multiple outputs configured.  You need to pass the value from the spec as an env var from here: https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/fluentd.go#L432

jcantrill, line 70: Maybe we should name this either `TOTAL_LIMIT_SIZE_REQUEST_PER_BUFFER` or `TOTAL_LIMIT_SIZE_SPEC_PER_BUFFER`

files/fluentd/run.sh
```diff
@@ -68,8 +67,31 @@ fi
...
+   TOTAL_LIMIT_SIZE_CONF=''
+else
+   TOTAL_LIMIT_SIZE_CONF=$(echo $TOTAL_LIMIT_SIZE_CONF |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc) || :
+fi
+
+TOTAL_LIMIT_SIZE_CONF=${TOTAL_LIMIT_SIZE_CONF:-0}
+
 # TOTAL_LIMIT_SIZE per buffer
-TOTAL_LIMIT_SIZE=$(expr $ALLOWED_DF_LIMIT / $NUM_OUTPUTS) || :
+TOTAL_LIMIT_SIZE_ALLOWED=$(expr $ALLOWED_DF_LIMIT / $NUM_OUTPUTS) || :
```
jcantrill, line 81: lets also name this `_PER_BUFFER`

files/fluentd/run.sh
```diff
@@ -68,8 +67,31 @@ fi
...
+
+if [[ $TOTAL_LIMIT_SIZE_CONF -lt $TOTAL_LIMIT_SIZE_ALLOWED ]];
+then
+   if [[ $TOTAL_LIMIT_SIZE_CONF -gt 0 ]]; then
+       TOTAL_LIMIT_SIZE=$TOTAL_LIMIT_SIZE_CONF
+   else
+       TOTAL_LIMIT_SIZE=$TOTAL_LIMIT_SIZE_ALLOWED
+   fi
+else
+    TOTAL_LIMIT_SIZE=$TOTAL_LIMIT_SIZE_ALLOWED
```
jcantrill, line 92: lets echo a message why we are disregarding... "Requested exceeds the maximum available...." or something like that

files/fluentd/run.sh
```diff
@@ -68,8 +67,31 @@ fi
 # Determine allowed total given the number of outputs we have.
 ALLOWED_DF_LIMIT=$(expr $DF_LIMIT \* $ALLOWED_PERCENT_OF_DISK / 100) || :
 
+TOTAL_LIMIT_SIZE_CONF=$(grep "total_limit_size" $FLUENT_CONF |  head -1 | awk '{print $2}' )
+
+if [[ $TOTAL_LIMIT_SIZE_CONF == *"TOTAL_LIMIT_SIZE"* ]]; then
```
jcantrill, line 72: I'll defer to bash experts here.  not sure what this block is doing other then converting to bytes in the else block

ajaygupta978:
> PR seems to be adding a new file `run.sh` at repo root
> `run.sh` should be generated by operator at run time.

@vimalk78  https://github.com/openshift/cluster-logging-operator/blob/master/pkg/k8shandler/fluentd.go#L341 Code uses run.sh script from files/fluentd/run.sh

files/fluentd/run.sh
ajaygupta978: How to access these variable in run.sh ?

files/fluentd/run.sh
ajaygupta978: I observed fluent.conf has ` total_limit_size "#{ENV['TOTAL_LIMIT_SIZE'] ||  8589934592 }" #8G`  when CLO starts so if block just ignores it if it finds `TOTAL_LIMIT_SIZE` in the line. Once user changes this value using ClusterLogging instance `total_limit_size` set to actual value like 8G, 8m which needs to be converted into bytes.

ajaygupta978:
Here is output:
1) Default output: When ClusterLogging instance is not changed.

`[root@ibm-p8-kvm-03-guest-02 cluster-logging-operator]# oc logs fluentd-5j5m7
Setting each total_size_limit for 2 buffers to 9621286195 bytes
Setting queued_chunks_limit_size for each buffer to 1146
Setting chunk_limit_size for each buffer to 8388608`

2) When totalLimitSize is less than Allowed Limit in ClusterLogging instance. In this case totalLimitSize per output `2147483648` is less than `9621286195` 

`Setting each total_size_limit for 2 buffers to 2147483648 bytes
Setting queued_chunks_limit_size for each buffer to 256
Setting chunk_limit_size for each buffer to 8388608`

3)  When totalLimitSize is more than Allowed Limit in ClusterLogging instance. In this case totalLimitSize per output `12884901888` is less than `9621286195` 

`Requested buffer size per output 12884901888 for 2 buffers exceeds maximum available size  9621286195 bytes per output
Setting each total_size_limit for 2 buffers to 9621286195 bytes
Setting queued_chunks_limit_size for each buffer to 1146
Setting chunk_limit_size for each buffer to 8388608`

ajaygupta978:
@jcantrill : Could you please review it?

pkg/utils/quantity.go
```diff
@@ -0,0 +1,90 @@
...
+		"m":  "Mi",
+		"Gi": "Gi",
+		"G":  "Gi",
+		"g":  "Gi",
+	}
+)
+
+// parseQuantityString is a fast scanner for quantity values.
+// it extracts the numerical value and suffix from the quantity string
+func parseQuantityString(str string) (value, suffix string, err error) {
```
jcantrill, line 27: Why would duplicate any of this code which looks to be copied from the kubernetes instead of simply using "MustParse" https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L125

pkg/utils/quantity_test.go
```diff
@@ -0,0 +1,39 @@
+package utils
```
jcantrill, line 1: Reuse of https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L125 looks like it would negate this test

files/fluentd/run.sh
```diff
@@ -67,18 +66,36 @@ if [ $ALLOWED_PERCENT_OF_DISK -gt 100 ] || [ $ALLOWED_PERCENT_OF_DISK -le 0 ] ;
 fi
 # Determine allowed total given the number of outputs we have.
 ALLOWED_DF_LIMIT=$(expr $DF_LIMIT \* $ALLOWED_PERCENT_OF_DISK / 100) || :
+TOTAL_LIMIT_SIZE_PER_BUFFER=${TOTAL_LIMIT_SIZE_PER_BUFFER:-0}
+
+TOTAL_LIMIT_SIZE_PER_BUFFER=$(echo $TOTAL_LIMIT_SIZE_PER_BUFFER |  sed -e "s/[Kk]/*1024/g;s/[Mm]/*1024*1024/g;s/[Gg]/*1024*1024*1024/g;s/i//g" | bc) || :
+
+# total limit size requested per buffer
+TOTAL_LIMIT_SIZE_PER_BUFFER=$(expr $TOTAL_LIMIT_SIZE_PER_BUFFER / $NUM_OUTPUTS) || :
```
jcantrill, line 74: Each output can have a maximum of 1 buffer which means there is no need to divide here since  we are specifiying the max of the buffer via an env var.

files/fluentd/run.sh
```diff
@@ -67,18 +66,36 @@ if [ $ALLOWED_PERCENT_OF_DISK -gt 100 ] || [ $ALLOWED_PERCENT_OF_DISK -le 0 ] ;
...
+
+# total limit size allowed  per buffer
```
jcantrill, line 76: Please move https://github.com/openshift/cluster-logging-operator/pull/1143/files#diff-7883a5579a471846a9b8e5aafeaceccfb4a98af2a279837c4812bf87d5dc33baL69 here closer to where it is first used ... `ALLOWED_DF_LIMIT`

jcantrill (changes requested):

pkg/utils/quantity.go
ajaygupta978: @jcantrill : Code in quantity.go doing `pre-processing` to convert suffix like `G->Gi` `M->Mi` and then using code from https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L264 to convert to bytes. So **1G** is converted to **1,073,741,824**.  If we directly invoke MustParse then 1G will be converted to 1,000,000,000.

pkg/utils/quantity_test.go
ajaygupta978: I didn't follow  what you mean ? I am testing **ParseQuantity** method of quantity.go from utils package. Also I didn't write quantity.go and quantity_test.go code. I took  code from latest master branch

files/fluentd/run.sh
jcantrill: @ajaygupta978  these was not addressed

files/fluentd/run.sh
jcantrill: @ajaygupta978 this was not addressed

pkg/utils/quantity.go
jcantrill: The best I can tell, the code introduced here almost exactly duplicates https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L125 which means we are supporting a copy that can vary and be subject to defects.  The code from APIMachinery has significantly larger visibility to a broader range of projects and is more likely to be correct or fixed if errors occur.   Please refactor the code to reuse the existing library

pkg/utils/quantity_test.go
jcantrill: Reusing the existing library means this test is not required. I'm not certain why this is in the master branch but IMO it should be removed from there as well

files/fluentd/run.sh
ajaygupta978: @jcantrill I was waiting for your comments for go code. I already tested run.sh with your suggested changes. I will push go and run.sh changes together.

files/fluentd/run.sh
ajaygupta978: @jcantrill I was waiting for your comments for go code. I already tested run.sh with your suggested changes. I will push go and run.sh changes together.

pkg/utils/quantity.go
ajaygupta978: This file is removed now.

pkg/utils/quantity_test.go
ajaygupta978: This file is removed now.

files/fluentd/run.sh
ajaygupta978: Done.

files/fluentd/run.sh
ajaygupta978: Done.

jcantrill:
@ajaygupta978  Please prepare  PRs and associated JIRAs for all 5.x streams

ajaygupta978:
@jcantrill : There is minor change in `pkg/utils/utils.go` in new commit. Please review

jcantrill (commented):
/lgtm

hardys:
[patch-manager] Approving for cherry-pick with a score of 0.10

PR #1147, opened 2021-08-30
LOG-1702: Loki getting out-of-order audit logs.

alanconway:
Cause: We were using k8s event requestReceivedTimestamp for fluent timestamp.
K8s does not log audit events until a response is sent, so it's
possible that the logs are not ordered by requestReceivedTimestamp.

Fix: Use the local system time for audit logs.  Users can still aget
the requestReceivedTimestamp from the event record.

/assign @jcantrill


jcantrill:
/lgtm

cc @vimalk78

jcantrill:
@vimalk78 I assume a cherryypick will fail on this for `master` because of the changes to the generator?

PR #1159, opened 2021-09-03
LOG-1733: Improve must-gather for ES store

lukas-vlcek:
### Description

Adding new resources:

- Adding [_nodes](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/cluster-nodes-info.html)
- Adding [hot_threads](https://www.elastic.co/guide/en/elasticsearch/reference/6.8/cluster-nodes-hot-threads.html)

Extending curl timout to 20 sec:

We are experiencing empty resource logs in must-gather archives. Instead of expected content they only contain message referencing to "timeout error code (28)".

Completing the documentation.

Adding suffix to each type of resource.
- .json is a json content (can be directly processed by jq)
- .cat is output of _cat endpoint (can be processed by awk or sed)
- .txt is non-structural text format

/cc @btaani
/assign @periklis 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-1733

### DIFF
--- a/must-gather/README.md
+++ b/must-gather/README.md
@@ -44,13 +44,20 @@ Example must-gather for cluster-logging output:
 │  │  └── elasticsearch-operator-7dc7d97b9d-jb4r4
 │  ├── es
 │  │  ├── cluster-elasticsearch
-│  │  │  ├── aliases
-│  │  │  ├── health
-│  │  │  ├── indices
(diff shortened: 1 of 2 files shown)


jcantrill (commented):
/approve

jcantrill (commented):
/lgtm

PR #1162, opened 2021-09-07
LOG-1479: Add functional test for the CloudWatch output

syedriko:
### Description
Added a functional test for the CloudWatch logforwarding output. The test bounces against an AWS mock server https://github.com/spulec/moto.
/cc @jcantrill 
/assign @igor-karpukhin 

### Links
- Depending on PR(s): https://github.com/openshift/origin-aggregated-logging/pull/2176
- JIRA: https://issues.redhat.com/browse/LOG-1479


syedriko:
The new CloudWatch test in ci/prow/functional will fail until the fluentd patch for ssl_verify_peer merges.

test/functional/outputs/forward_to_cloudwatch_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+		maxDuration, _          = time.ParseDuration("10m")
+		defaultRetryInterval, _ = time.ParseDuration("10s")
+	)
+
+	BeforeEach(func() {
+		framework = functional.NewFluentdFunctionalFramework()
+
+		log.V(2).Info("Creating service moto")
+		service := runtime.NewService(framework.Namespace, "moto")
```
jcantrill, line 67: This looks like a pattern we are repeating from the Loki tests and were started as part of the kafka functional tests.  Functional framework intent is to create a single POD that includes all needed containers for the test.  This means they should share the same network address space.  

* Why do we need a service and route to create a test that writes from the forwarder to cloudwatch mock.   
* What are we missing in the framework that keeps us from using it as designed.  
* What do we gain be introducing the additional complexities of OpenShift

test/runtime/pod.go
```diff
@@ -107,6 +107,31 @@ func (builder *PodBuilder) AddConfigMapVolume(name, configMapName string) *PodBu
...
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: secretName,
+			},
+		},
+	})
+	return builder
+}
+
+func (builder *PodBuilder) GetContainer(name string) *ContainerBuilder {
```
jcantrill, line 122: I'm asking but don't know the answer.  Is there any other way to get a container? Does it makes sense to name it "GetContainerByName" ?

test/functional/outputs/forward_to_cloudwatch_test.go
```diff
@@ -0,0 +1,191 @@
...
+
+	AfterEach(func() {
+		framework.Cleanup()
+	})
+
+	Context("When sending a sequence of log messages to CloudWatch", func() {
+		FIt("should get there", func() {
+
+			var logGroupsOutput *cwl.DescribeLogGroupsOutput
+			err := wait.PollImmediate(defaultRetryInterval, maxDuration, func() (done bool, err error) {
```
jcantrill, line 139: Any reason not to encapsulate this block in a similar framework method like https://github.com/openshift/cluster-logging-operator/blob/master/test/functional/outputs/forward_to_syslog_test.go#L94

jcantrill (changes requested):

test/functional/outputs/forward_to_cloudwatch_test.go
syedriko: It's not the writing to the mock, it's the reading from the mock. The test, running outside of cluster, talks to moto to check that the logs made it there. That's what the `Context("When sending a sequence of log messages to CloudWatch"` clause does.
To fit this into the framework we could
- fork moto and trying to merge it upstream to get it to log what we need into its own log which we can `oc rsh cat`
- put the aws cli client or write our own simple aws cli client that dumps into its log what it reads from moto, put it into a container, deploy into the test pod. Dealing with that container image looked like a pain to me.

test/runtime/pod.go
syedriko: Not really. The name is the only attribute of a container that's required to be unique in a Pod, so we will need to return a collection of containers if we look up by any other attribute.  https://github.com/kubernetes/api/blob/5c8efac7a372214f8b0673cb92073d73f3881818/core/v1/types.go#L2229

test/functional/outputs/forward_to_cloudwatch_test.go
syedriko: This code is specific to CloudWatch and this is the only place it's used, so I didn't think about moving it to the framework. We can do that when we need to call it from somewhere else.

jcantrill (commented):
/lgtm

PR #1164, opened 2021-09-07
Add internal to test-unit after sdk bump

jcantrill:
### Description
This PR:
* adds `internal` package into unit tests that was missed after sdk bump
* Skips vector generator tests

cc @vimalk78 @vparfonov


alanconway (commented):
/lgtm

PR #1182, opened 2021-09-20
LOG-1725 Summarize Recognized Auth Keys.

alanconway:
Common keys are documented in the Secret section of the CLF Go spec.
Output-specific keys are documented in configuration field for that output.

/hold

Open questions:
- is it too late to rename "passphrase" as "tls.key.passphrase"?
  - just "passphrase" doensn't tell you what it is a passphrase for...
- is it too late to drop "sasl_over_ssl" in favour of "sasl.enable"?
  - if not we can maybe keep sasl_over_ssl as a hidden alias of sasl.enable.
  - SASL can be used with or without SSL, some sasl mechs do their own encryption.
- should output specific keys be listed at the end of the main Secret doc?
  - I put them with the output-config field, but I dithered over which to do.

/assign @jcantrill
/cc @periklis
/cc @vimalkum

### DIFF
--- a/apis/logging/v1/cluster_log_forwarder_types.go
+++ b/apis/logging/v1/cluster_log_forwarder_types.go
@@ -147,11 +147,39 @@ type OutputSpec struct {
 	// Secret for authentication.
 	// Name of a secret in the same namespace as the cluster logging operator.
 	//
-	// For client authentication, set secret keys `tls.crt` and `tls.key` to the client certificate and private key.
+	// All sensitive authentication information is provided via a kubernetes Secret object.
+	// A Secret is a key:value map, common keys are described here.
+	// Some output types support additional specialized keys, documented with the output-specific configuration field.
+	// All secret keys are optional, enable the security features you want by setting the relevant keys.
 	//
-	// To use your own certificate authority, set secret key `ca-bundle.crt`.
+	// Transport Layer Security (TLS)
 	//
-	// Depending on the `type` there may be other secret keys that have meaning.
+	// Using a TLS URL ('https://...' or 'ssl://...') without any secret enables basic TLS:
+	// client authenticates server using system default certificate authority.
+	//
+	// Additional TLS features are enabled by including a Secret and setting the following optional fields:
+	//
+	//   `tls.crt`: (string) File name containing a client certificate.
+	//     Enables mutual authentication. Requires `tls.key`.
+	//   `tls.key`: (string) File name containing the private key to unlock the client certificate.
+	//     Requires `tls.crt`
+	//   `passphrase`: (string) Passphrase to decode an encoded TLS private key.
+	//     Requires tls.key.
+	//   `ca-bundle.crt`: (string) File name of a custom CA for server authentication.
+	//
+	// Username and Password
+	//
+	//   `username`: (string) Authentication user name. Requires `password`.
+	//   `password`: (string) Authentication password. Requires `username`.
+	//
+	// Simple Authentication Security Layer (SASL)
+	//
+	//   `sasl.enable`: (boolean) Explicitly enable or disable SASL.
+	//     If missing, SASL is automatically enabled when any of the other `sasl.` keys are set.
+	//   `sasl.mechanisms`: (array) List of allowed SASL mechanism names.
+	//     If missing or empty, the system defaults are used.
+	//   `sasl.allow-insecure`: (boolean) Allow mechanisms that send clear-text passwords.
+	//     Default false.
 	//
 	// +optional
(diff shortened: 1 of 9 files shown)


apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -147,11 +147,33 @@ type OutputSpec struct {
...
+	// Additional TLS features are enabled by selectively setting the following optional fields.
+	//
+	//   `tls.crt`: (string) File name of a client certificate to authenticate the client. Enables mutual authentication. Requires `tls.key`.
+	//   `tls.key`: (string) File name containing the private key  to unlock the client certificate. Requires `tls.crt`
+	//   `passphrase`: (string) Passphrase to decode an encoded private key. Requires tls.key.
+	//   `ca-bundle.crt`: (string) File name of a custom CA for server authentication.
+	//
+	// Username and Password
+	//
+	//   `user`: (string) Authentication user name. `password` must also be set.
```
jcantrill, line 169: We already are using `username` for the ES password feature that landed.  This means we either: change it here or honor both with one taking priority

apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -147,11 +147,33 @@ type OutputSpec struct {
...
+	//   `password`: (string) Authentication password. `user` must also be set.
+	//
+	// Simple Authentication Security Layer (SASL)
+	//
+	//   `sasl.enable`: (boolean) Enable or disable SASL.
```
jcantrill, line 174: Do we need to explicitly "enable" or is it sufficient the presence of something else  means it is enabled?

jcantrill:
> 
> * is it too late to rename "passphrase" as "tls.key.passphrase"?
>   

We can deprecate the original one, honor both with preference of one over the other

>   * just "passphrase" doensn't tell you what it is a passphrase for...
> * is it too late to drop "sasl_over_ssl" in favour of "sasl.enable"?
>   

Same here though I don't believe we currently support any SASL features.

>   * if not we can maybe keep sasl_over_ssl as a hidden alias of sasl.enable.
>   * SASL can be used with or without SSL, some sasl mechs do their own encryption.
> * should output specific keys be listed at the end of the main Secret doc?
>

apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -147,11 +147,33 @@ type OutputSpec struct {
 	// Secret for authentication.
 	// Name of a secret in the same namespace as the cluster logging operator.
 	//
-	// For client authentication, set secret keys `tls.crt` and `tls.key` to the client certificate and private key.
+	// All sensitive authentication information is provided via a kubernetes Secret object.
+	// A Secret is a key:value map, common keys are described here.
+	// Some outputs support additional pecialized keys, documented with the output-specific configuration field.
```
periklis, line 152: `pecialized` -> `specialized`?

periklis (commented):
Besides one small nit, looks good to me

apis/logging/v1/cluster_log_forwarder_types.go
alanconway: I don't think that's a problem - in the context of an "elasticsearch" output it means what it means now. In the context of other outputs it may mean something else (basic auth, or user/pass over TLS) or not be applicable at all. The idea is to find a minimal  set of generic keys that can be interpreted appropriately in the context of what a given output supports - and add specific ones (e.g. aws_blah) where an output does something thta doesn't fit the generic pattern. 
Am I missing something?

alanconway:
Updated the doc text, please re-review @jcantrill . I'm writing the backwards-compat code now....

alanconway:
@jcantrill @vimalk78  hopefully final - please look it over.
- I didn't change "passphrase", it's already public and we also have "ca-bundle" so we wouldn't have consistent "tls." prefixes anyway. I don't think it's a problem.
- I wrote the code to support sasl.enable and sasl_over_ssl for Kafka. Future SASL outputs can stick to the sasl. variables.
- I shortened the security. go code, I had to write some wrappers to make deprecation simple, so I simplified some of the other funcs there.

alanconway:
/hold
Wait for merge of https://github.com/openshift/cluster-logging-operator/pull/1189
Separate generated file changes.

alanconway:
/hold
Needs rebase after #1189 has merged to remove unrelated generated code changes.

alanconway:
/unhold
@jcantrill can I get a final LGTM? All the generated weirdness is gone.

jcantrill (commented):
/lgtm

PR #1186, opened 2021-09-23
LOG-1168: disabled hostname verification in TLS mode

vimalk78:
### Description
Disabled hostname verification when TLS is used. Hostname verification creates problems when logs are sent to external syslog server.

/cc 
/assign 

/cherry-pick 
should be cherry-picked to earlier versions

### Links

- JIRA: https://issues.redhat.com/browse/LOG-1168


syedriko (commented):
I thought we were going to surface this flag in the LF output config?

vimalk78:
> I thought we were going to surface this flag in the LF output config?

you mean set it from the API?

syedriko:
> > I thought we were going to surface this flag in the LF output config?
> 
> you mean set it from the API?

Right, so it's not hardcoded. May be default it to false if that's more appropriate, but still expose it.

vimalk78:
> > > I thought we were going to surface this flag in the LF output config?
> > 
> > 
> > you mean set it from the API?
> 
> Right, so it's not hardcoded. May be default it to false if that's more appropriate, but still expose it.

It will be difficult to abstract it out to an api level for all outputs. 
@alanconway  Do we need consistent means to control hostname verification in SSL config?

syedriko:
> > > > I thought we were going to surface this flag in the LF output config?
> > > 
> > > 
> > > you mean set it from the API?
> > 
> > 
> > Right, so it's not hardcoded. May be default it to false if that's more appropriate, but still expose it.
> 
> It will be difficult to abstract it out to an api level for all outputs.
> @alanconway Do we need consistent means to control hostname verification in SSL config?

I didn't mean all outputs, just syslog.

alanconway (changes requested):
This has security implications. Sugggest we this an additional configurable key the Secret: 
```
  // tls.verfiy (bool) if true verify the server's hostname is valid and matchse the certificate.
  //    If false, skip the hostname check. Default true.
```
I'd rather leave the default as true in general. What's the problem with syslog? Maybe we could switch the default just for syslog.

internal/generator/fluentd/output/syslog/tls.go
```diff
@@ -11,13 +13,26 @@ func (t TLS) Name() string {
...
+{{- end}}`, t)
+}
+
+type HostnameVerify security.HostnameVerify
+
+func (h HostnameVerify) Name() string {
+	return "syslogHostNameVerify"
+}
+
+func (h HostnameVerify) Template() string {
```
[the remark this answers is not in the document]
vimalk78: the function which is there as per your earlier comment
https://github.com/openshift/cluster-logging-operator/pull/1118#discussion_r691723275

internal/generator/fluentd/output/syslog/output_conf_syslog_test.go
```diff
@@ -357,6 +355,47 @@ var _ = Describe("Generating external syslog server output store config blocks",
...
+					results, err := g.GenerateConf(c...)
+					Expect(err).To(BeNil())
+					Expect(results).To(EqualTrimLines(`
+						tls true
+						verify_mode 1 #VERIFY_NONE:0, VERIFY_PEER:1
+						ca_file '/var/run/ocp-collector/secrets/some-secret/ca-bundle.crt'
+					`))
+				})
+				It("should produce security config with host name verification disabled", func() {
+					secret.Data["syslog_hostname_verify"] = []byte("false")
```
syedriko, line 388: Can we refer to `const SyslogHostnameVerify` here?

internal/generator/fluentd/output/security/security.go
```diff
@@ -10,6 +10,8 @@ import (
 
 type TLS bool
 
+type HostnameVerify bool
```
syedriko, line 13: I'm not sure I see the need for a HostnameVerify type. Just like the TLS type above.

vimalk78: Just like other types in this package, this is a common type which captures the requirements. then each output package adds the template needed for generating conf.

internal/generator/fluentd/output/syslog/output_conf_syslog_test.go
vimalk78: done

PR #1191, opened 2021-09-24
LOG-1422: Remove legacy CLF Support

jcantrill:
### Description
This PR:
* Removes support for legacy CLF

@alanconway @rolfedh this means users of legacy CLF will LF stop on upgrade to 5.3 as we have no way of converting their BYO config to something we understand.  I'm not sure how to best document other then we already stated it was deprecated

### Links
* https://issues.redhat.com/browse/LOG-1422


vimalk78:
do you want to delete this folder also? https://github.com/openshift/cluster-logging-operator/tree/master/internal/generator/fluentd/output/legacy

alanconway:
What's missing if we convert legacy to the nearest equivalent supported forwarder output?

jcantrill:
> What's missing if we convert legacy to the nearest equivalent supported forwarder output?

Currently I believe we only allow configuration of a "store" stanza which means conceptually a user could make an equivalent CLF output declaration.  I'm not certain it's worth the effort for us to figure out what to upgrade into given we have listed this feature as deprecated for several releases so users should already be on notice

vimalk78 (approved):
/lgtm

PR #1196, opened 2021-09-30
LOG-1730: Refactor ClusterLogging API, Add Vector

vimalk78:
### Description
This PR updates ClusterLogging API to add vector collector.
  - Added Vector to ClusterLogging API
  - Modified ClusterLogging API to rename ForwarderSpec to FluentdTuningSpec

/cc 
/assign 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: 
- Enhancement proposal:


apis/logging/v1/clusterlogging_types.go
```diff
@@ -178,34 +173,67 @@ type ElasticsearchSpec struct {
...
+	//
+	// +nullable
+	// +optional
+	FluentdSpec *FluentdSpec `json:"fluentd,omitempty"`
+
+	// Specification of the Vector Log Collection component
+	//
+	// +nullable
+	// +optional
+	VectorSpec *VectorSpec `json:"vector,omitempty"`
```
alanconway, line 202: This is an additive, backwards compatible change, OK to add this for vector support.

apis/logging/v1/clusterlogging_types.go
```diff
@@ -54,11 +54,6 @@ type ClusterLoggingSpec struct {
 	//
 	// +nullable
 	Curation *CurationSpec `json:"curation,omitempty"`
-
-	// Specification for Forwarder component for the cluster
-	//
-	// +nullable
-	Forwarder *ForwarderSpec `json:"forwarder,omitempty"`
```
jcantrill, line 61: We can not remove this but will need to mark it as deprecated for upgrade and backwards compatibility reasons

apis/logging/v1/clusterlogging_types.go
```diff
@@ -178,34 +173,67 @@ type ElasticsearchSpec struct {
 
 // This is the struct that will contain information pertinent to Log and event collection
 type CollectionSpec struct {
-	// Specification of Log Collection for the cluster
-	Logs LogCollectionSpec `json:"logs,omitempty"`
```
jcantrill, line 182: Same here.  This will need to be marked deprecated for for backwards compatibility reasons

apis/logging/v1/clusterlogging_types.go
```diff
@@ -178,34 +173,67 @@ type ElasticsearchSpec struct {
...
+type VectorSpec struct {
+	//DataDir is the directory used for persisting Vector state
 	// +optional
-	Resources *v1.ResourceRequirements `json:"resources"`
+	DataDir string `json:"data_dir"`
 
-	// Define which Nodes the Pods are scheduled on.
-	//
-	// +nullable
-	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
alanconway, line 207: Why do we not need node selectors & tolerations for Vector? It's great if we don't - just curious why.

apis/logging/v1/clusterlogging_types.go
```diff
@@ -178,34 +173,67 @@ type ElasticsearchSpec struct {
...
+	// +optional
 	// +nullable
+	Tuning *FluentdTuningSpec `json:"tuning,omitempty"`
+}
+
+type VectorSpec struct {
+	//DataDir is the directory used for persisting Vector state
 	// +optional
-	Resources *v1.ResourceRequirements `json:"resources"`
+	DataDir string `json:"data_dir"`
```
jcantrill, line 220: Why does this need to be configurable for users?

apis/logging/v1/clusterlogging_types.go
```diff
@@ -178,34 +173,67 @@ type ElasticsearchSpec struct {
...
 
-	// Define which Nodes the Pods are scheduled on.
-	//
-	// +nullable
-	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
-	Tolerations  []v1.Toleration   `json:"tolerations,omitempty"`
+	//HostKey sets the event key to use for the event host field.
```
jcantrill, line 222: Why do any of these fields need to be modified via API?

jcantrill (changes requested):
We will need to be able to upgrade existing deployments into the new, desired format, likely with a webhook.  This means we can not get rid of the old structure but be capable of transforming the old into the new.

alanconway (changes requested):
I think we need to clearly separate the introduction of vector from introducing a backwards-incompatible v2 AP. Vector first - it's fine to add a vector section to the v1 API, that is still backward compatible. The other changes  here are good improvements for the future; but breaking the API  at the same time as introducing vector will be a nightmare.

alanconway (changes requested):
Don't make incompatible changes to Forwarder & Fluentd sections.  We
don't want to break existing fluentd config at the same time we're
trying to convince people to migrate to Vector.  We can go ahead with
the new Collector structure for the Vector.

Do we absolutely need the Collector/Vector settings? Unless vector can't function
without them I'd rather not commit to new API until we have some experience.
Also I'd be unhappy if we can't make Vector work out of the box without
user configuring settings they probably won't understand.

Otherwise LGTM.

alanconway:
@vimalk78  can this be closed?

jcantrill:
@vimalk78 can we close this?

PR #1201, opened 2021-10-04
LOG-1730: Refactoring CLO for new collector

vimalk78:
### Description
Refactor CLO to decouple collector deployment from fluentd specific details.
ranamed functions to be independent of collector type, moved common function from fluentd.go to collector.go. fluentd.go now only has functions specific to fluentd.

/cc 
/assign 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-1730


internal/k8shandler/collection.go
```diff
@@ -126,6 +142,297 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection() (err err
...
+		if err = clusterRequest.RemoveConfigMap(caName); err != nil {
+			return
+		}
+
+		if err = clusterRequest.RemoveDaemonset(name); err != nil {
+			return
+		}
+
+		// Wait longer than the terminationGracePeriodSeconds
+		time.Sleep(12 * time.Second)
```
jcantrill, line 174: Why do we need to sleep and simply not let it continue?

jcantrill (commented):
/approve

PR #1202, opened 2021-10-04
LOG-1771: Refactor test/runtime package into internal/runtime

jcantrill:
### Description
This PR:
* Refactors the `test/runtime` code into `internal/runtime` for reuse in our handlers
* Removes the `factory` package in favor of using builders and runtime

@alanconway please re-review this.  The previous PR I am unable to get e2e results to figure out if there is actually something wrong

### Ref
* [LOG-1771](https://issues.redhat.com/browse/LOG-1771): Refactor test/runtime package into internal/runtime


test/functional/outputs/forward_to_cloudwatch_test.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/json"
+	runtime2 "github.com/openshift/cluster-logging-operator/internal/runtime"
```
alanconway, line 7: Why runtime2 and not runtime?

alanconway (changes requested):
There are some "runtime2" imports that look like they were part of an intermediate stage, I don't see any reason why they can't be "runtime". Otherwise all good

test/functional/outputs/forward_to_cloudwatch_test.go
jcantrill: Updated to:
* remove ref to `runtime2`
* use `testruntime` where there was conflict

PR #1204, opened 2021-10-05
LOG-1575: Add deep_copy to fan out to outputs

jcantrill:
### Description
This PR:
* deep copies message to outputs to resolve unintended parsing of messages to be delivered to Elasticsearch

/assign @vimalk78 

### Links
* https://issues.redhat.com/browse/LOG-1575

### DIFF
--- a/internal/generator/fluentd/elements/copy.go
+++ b/internal/generator/fluentd/elements/copy.go
@@ -5,7 +5,8 @@ import (
 )
 
 type Copy struct {
-	Stores []generator.Element
+	DeepCopy bool
+	Stores   []generator.Element
 }
 
 func (c Copy) Name() string {
@@ -15,6 +16,9 @@ func (c Copy) Name() string {
 func (c Copy) Template() string {
 	return `{{define "` + c.Name() + `"  -}}
 @type copy
+{{if .DeepCopy -}}
+deep_copy true
+{{end -}}
 {{compose .Stores}}
 {{end}}`
 }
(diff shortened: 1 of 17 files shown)


internal/generator/fluentd/pipeline_to_output.go
```diff
@@ -81,7 +81,8 @@ func PipelineToOutputs(spec *logging.ClusterLogForwarderSpec, op Options) []Elem
 				Match{
 					MatchTags: "**",
 					MatchElement: Copy{
-						Stores: CopyToLabels(helpers.LabelNames(p.OutputRefs)),
+						DeepCopy: true,
```
vimalk78, line 84: instead of always setting deep copy to true, we can add it selectively by detecting if we need deep copying. e.g. using syslog and ES together

internal/generator/fluentd/elements/copy.go
```diff
@@ -15,6 +16,9 @@ func (c Copy) Name() string {
 func (c Copy) Template() string {
 	return `{{define "` + c.Name() + `"  -}}
 @type copy
+{{if .DeepCopy -}}
+deep_copy true
```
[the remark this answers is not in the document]
jcantrill: We'll have to change it if/when we bump the version of fluent.  The only alternative we have now is this flag unless we do something custom/specific for syslog.  Thoughts?

internal/generator/fluentd/pipeline_to_output.go
vimalk78: agree

PR #1241, opened 2021-11-04
LOG-1766: Makefile changes to prepare for preview releases.

alanconway:
Does not enable preview builds yet, changes to pave the way:

- Set default LOGGING_VERSION explicitly in Makefile
  With previews there may be more than one manifests directory.
  Can be over-ridden from command line or environment variables.

- Improve dependencies in bundle build, deploy and install to automate
  builds and avoid un-necessary work.

/cc @jcantrill
/cc @periklis


Makefile
```diff
@@ -17,7 +17,7 @@ export GODEBUG=x509ignoreCN=0
 export APP_NAME=cluster-logging-operator
 export IMAGE_TAG?=127.0.0.1:5000/openshift/origin-$(APP_NAME):latest
 
-export LOGGING_VERSION?=$(shell basename $(shell ls -d manifests/[0-9]*))
+export LOGGING_VERSION?=5.3
```
jcantrill, line 20: We purposely did away with an explicit version so this file did not need update on change in release.  I'm not certain how @vparfonov changes play into this but my understanding is in future we should be able to rely on the `bundle` dir to populate the manifest dir instead of the otherway around.  I believe that is a change currently by EO

periklis: In general we should retire the old manifests dir entirely as we have done in EO since 5.2. Imho the version be static here per y-stream release and inform the bundle creation regarding channel naming, e.g. `5.3` -> `stable,stable-5.3`

Makefile
```diff
@@ -190,12 +192,15 @@ test-cluster:
...
-.PHONY: bundle
+bundle: $(OPM) $(shell find manifests -type f)
+	@rm -rf bundle
+	$(MAKE) regenerate
+	MANIFEST_VERSION=${LOGGING_VERSION} \
+	OPENSHIFT_VERSIONS=${OPENSHIFT_VERSIONS} \
+	CHANNELS=${CHANNELS} \
+	DEFAULT_CHANNEL=${DEFAULT_CHANNEL} \
+	hack/generate-bundle.sh
+	@touch $@
```
periklis, line 203: @jcantrill @vparfonov: **Note**: Since CLO is using a modern version of the `operator-sdk` you should replace `opm` with the `operator-sdk` bundle commands.

Makefile
alanconway: @jcantrill please confirm if you are OK with the change @periklis suggests. Anything that simplifies the release artifacts is good IMO. I will make the changes in separate PRs: 
- remove old manifests dir, make sure everything works
- proceed with upstream release & changes to enable previews
@periklis please point me to the PR that made this change in EO so I can learn from that.

alanconway: @periklis assuming your change, should previews be treated similar to Z-stream? I.e. 
- single directory, static per Y stream
- PRE release: active release is always X.NextY.0-preview.N, channel is 'preview'
- POST release: active release is X.Y.Z, channel is 'stable, stable-X.Y'

Makefile
alanconway: @vparfonov what's  holding up your merge? I'd much rather commit the preview version work **after** your changes.
Is it still  just minor lint issues? If your sure they're harmless then maybe get your changes in immediately and we sort out lint later...

alanconway:
Moved to https://github.com/openshift/cluster-logging-operator/pull/1241#issuecomment-970403625

PR #1243, opened 2021-11-07
Added retry logic around test points in the e2e container security to fix a flake

syedriko:
### Description

- Added retry logic around test points in the container security test to account for possible collector DaemonSet rollout concurrent with the test. 
- Updated github.com/onsi/gomega to v1.17.0

/cc @jcantrill 
/assign @vimalk78


jcantrill (approved):
/approve 
/lgtm

syedriko:
Proper fix for this container security test flake. The root cause is a change to the collector DaemonSet _after_ it's been created, which triggers a rolling update of the Pods in the DaemonSet. Thus is is possible for the container security test to connect to a terminating Pod, causing a sporadic test failure.
The fix adds retry logic around all the test points that reconnects to another Pod in the DaemonSet instead of the terminating one.
Gomega has been updated to v1.17.0 to pick up an elegant way of specifying the retry test logic with Eventually() async assertions.

PR #1263, opened 2021-11-23
Log 2012: add vector config to Secret rather than ConfigMap

ajaygupta978:
### Description
This PR generates configuration for vector log collector.

/cc @vimalk78 
/assign @jcantrill 

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-2012
- Enhancement proposal:


internal/generator/vector/output/kafka/kafka_test.go
```diff
@@ -59,8 +59,8 @@ timestamp_format = "rfc3339"
 
 # Sasl Config
 [sinks.kafka_receiver.sasl]
-username = "" # inject code to read from filename
-password = "" # inject code to read from filename
+user = ""
```
vimalk78, line 62: please add tests which generate proper values of username/password

vimalk78: same for elasticsearch

vimalk78 (commented):
left some comments

internal/k8shandler/collection.go
```diff
@@ -396,27 +397,36 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateCollectorPrometheusRu
 	})
 }
 
-func (clusterRequest *ClusterLoggingRequest) createOrUpdateCollectorConfigMap(collectorType logging.LogCollectionType, collectorConfig string) error {
-	var collectorConfigMap *corev1.ConfigMap = nil
+func (clusterRequest *ClusterLoggingRequest) createOrUpdateCollectorConfigMapSecret(collectorType logging.LogCollectionType, collectorConfig string) error {
```
jcantrill, line 400: This name should be more generic to not ref "configmap" or "secret".  What is it doing?  its "ReconcilingCollectorConfig".  Actually this is exactly what I referred to where we should consider refactoring this into a collector specific function that does one thing or the other depending upon the collector impl instead of using an if block

internal/constants/constants.go
```diff
@@ -60,10 +60,11 @@ const (
 	CollectorSecretsDir     = "/var/run/ocp-collector/secrets" //nolint:gosec
 	KibanaSessionSecretName = "kibana-session-secret"          //nolint:gosec
 
-	CollectorName             = "collector"
-	CollectorMetricSecretName = "collector-metrics"
-	CollectorMonitorJobLabel  = "monitor-collector"
-	CollectorTrustedCAName    = "collector-trusted-ca-bundle"
+	CollectorName                = "collector"
+	CollectorConfigMapSecretName = "collector-configmap"
```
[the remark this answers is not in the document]
ajaygupta978: Ok done

internal/k8shandler/collection.go
```diff
@@ -396,27 +397,36 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateCollectorPrometheusRu
...
 	} else if collectorType == logging.LogCollectionTypeVector {
-		collectorConfigMap = NewConfigMap(
-			constants.CollectorName,
-			clusterRequest.Cluster.Namespace,
-			map[string]string{
-				"vector.toml": collectorConfig,
-			},
-		)
+
+		secrets := map[string][]byte{
```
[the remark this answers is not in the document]
ajaygupta978: ok

internal/k8shandler/collection.go
```diff
@@ -438,10 +448,21 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateCollectorConfigMap(co
...
 }
 
+func (clusterRequest *ClusterLoggingRequest) createSecret(secretName, namespace string, secrets map[string][]byte) error {
+	collectorSecret := NewSecret(secretName, namespace, secrets)
+	utils.AddOwnerRefToObject(collectorSecret, utils.AsOwner(clusterRequest.Cluster))
+	err := clusterRequest.CreateOrUpdateSecret(collectorSecret)
+	return err
+}
+
+func (clusterRequest *ClusterLoggingRequest) createOrUpdateCollectorConfigMap(collectorType logging.LogCollectionType, collectorConfig string) error {
```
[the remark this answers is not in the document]
ajaygupta978: Yes this method internally calls `createOrUpdateCollectorConfigMapSecret` which creates configmap for fluentd and secret for vector.

internal/generator/vector/output/kafka/kafka_test.go
ajaygupta978: Conf generator reads user/password from `/var/run/ocp-collector/secrets` but in case of unit test case this file doesn't exist so user/password is empty in unit test case. Even incase of fluentd value is `username "#{File.exists?('/var/run/ocp-collector/secrets/kafka-receiver-1/username') ? open('/var/run/ocp-collector/secrets/kafka-receiver-1/username','r') do |f|f.read end : ''}"`

internal/generator/vector/output/security/security.go
```diff
@@ -95,6 +96,15 @@ func SecretPath(name string, file string) string {
 	return fmt.Sprintf("%q", filepath.Join("/var/run/ocp-collector/secrets", name, file))
 }
 
+func GetValueFromFile(name string, file string) string {
+	filename := filepath.Join("/var/run/ocp-collector/secrets", name, file)
```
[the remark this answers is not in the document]
ajaygupta978: done. Thanks for suggestion.

internal/k8shandler/collection.go
ajaygupta978: @jcantrill  This method creates configmap(fluentd) or secret (vector).  I think in Java we can do using `interface` but in golang how do we achieve this without using if block. Can you please share an example ?

internal/generator/vector/output/security/security.go
```diff
@@ -95,6 +96,16 @@ func SecretPath(name string, file string) string {
 	return fmt.Sprintf("%q", filepath.Join("/var/run/ocp-collector/secrets", name, file))
 }
 
+func GetValueFromFile(name string, file string) string {
+	filename := SecretPath(name, file)
+	b, err := ioutil.ReadFile(filename)
+
+	if err != nil {
```
alanconway, line 103: We really need to stop writing functions that hide errors. I know we have a bunch of code that doesn't have good error reporting paths, but we need to fix that rather than perpetuate it. These errors should bubble up to where we can put them in a status message and log them so users can debug the problem.

internal/k8shandler/collection.go
alanconway: Golang interface can be used similar to Java to provide alternate implementations of the same methods.

alanconway (changes requested):

internal/k8shandler/collection.go
ajaygupta978: thanks @alanconway. Yes Golang interface can work like Java interface. 
@jcantrill  You are right we would need to refactor code to make code extensible and have separate implementations  for  each collector type. This will require some effort to understand in terms of flow to generate collector instance using factory and pass it to these methods. Can we avoid it in this PR as we are about to finish sprint ?  We can do it in separate PR.

internal/generator/vector/output/security/security.go
ajaygupta978: thanks @alanconway for suggestion. I agree that its not good practice. I was probably following other examples which also needs to be fixed.

internal/constants/constants.go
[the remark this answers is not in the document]
vimalk78: this is still
```
CollectorConfigSecretName = "collector-configmap"
```

internal/k8shandler/collection.go
vimalk78: @ajaygupta978 even if re factoring is not possible, please consider renaming the method

internal/constants/constants.go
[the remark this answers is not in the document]
ajaygupta978: forgot to change value. Now done.

internal/k8shandler/collection.go
ajaygupta978: Ok done.

internal/k8shandler/collection.go
```diff
@@ -396,27 +397,38 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateCollectorPrometheusRu
...
+			secrets["vector.toml"] = []byte(collectorConfig)
+			return nil
+		})
+		err = clusterRequest.createSecret(constants.CollectorConfigSecretName, clusterRequest.Cluster.Namespace, secrets)
 	}
+	if err != nil {
+		return err
+	}
+
+	return nil
```
vimalk78, line 428: why not just return `err` ?

vimalk78:
/approve
/lgtm
hold cancel

PR #1265, opened 2021-11-26
LOG-1923: Refactor CLO to request certificates from EO instead of creating them

vparfonov:

### Description
Refactor CLO to request certificates from EO instead of creating them.
1 .Remove files:
-  internal/k8shandler/certificates.go
-  internal/k8shandler/certificates_test.go

2. Avoid creation secrets with certificates on CLO side for elasicsearch, kibana and collector(Fluentd).

3. Add annotations: 
```
"logging.openshift.io/elasticsearch-cert-management": "true",
"logging.openshift.io/elasticsearch-cert.collector":  "system.logging.fluentd",

```
4. Keep script `scripts/cert_generation.sh` because it used for functional and e2e tests.

/cc 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s): https://github.com/openshift/elasticsearch-operator/pull/832
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-1923
- Enhancement proposal:


jcantrill:
Looks fine but we need to consider and test the upgrade scenerio:
* Deploy older operator and test upgrade to new CLO
* How do we remove our old secrets when they are owned by CLO but leave them alone when they are owned by EO (ownerref?)

vparfonov:
> Looks fine but we need to consider and test the upgrade scenerio:
> 
> * Deploy older operator and test upgrade to new CLO
> * How do we remove our old secrets when they are owned by CLO but leave them alone when they are owned by EO (ownerref?)

Hm, i think it will be recreated, name of secret the same. I will test

internal/k8shandler/logstore.go
```diff
@@ -30,7 +32,8 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateLogStore() (err error
 
 		cluster := clusterRequest.Cluster
 
-		if err = clusterRequest.createOrUpdateElasticsearchSecret(); err != nil {
+		if err = clusterRequest.removeElasticsearchIfSecretOwnedByCLO(); err != nil {
```
pavolloffay, line 35: @vparfonov could you please explain upgrade behavior? What happens with already provisioned ES instances for which certs are managed by CLO?  

If am reading this correctly the CLO removes ES if the certs  were managed by CLO. And then a new ES CR is created with cert-management from EO?

vparfonov: @pavolloffay I going to check ownership of secret if not ES so i need to remove it and re-create deployments

pavolloffay: does this imply ~no~ zero time downtime during the upgrade?

vparfonov: yep, in this case

pavolloffay: In this case I am doing something similar for Jaeger https://github.com/jaegertracing/jaeger-operator/pull/1708

pavolloffay: @vparfonov how does the log-forwarder/cliens (e.g. fluentd?) behave in this case? They will fail to export the data and fail/drop if they cannot export?

PR #1267, opened 2021-11-29
Makefile: tag images with branch name

vimalk78:
### Description
For local development and debugging, many times it is needed to deploy logging from different branches to compare a feature across branches, and need to switch back again. Since the images are tagged `latest`, the same image pull spec cannot be used across branches even though the image has been pushed to cluster.

This PR attempts to tag the images in the cluster with branch name so that images in cluster do not get overwritten.

```
# do testing/debugging on 5.3
> git checkout release-5.3
> make deploy


# delete the deployments
> oc delete ns openshift-logging openshift-operators-redhat

# do testing/debugging on master
> git checkout master
> make deploy

# now cluster has two CLO images tagged 
$ oc -n openshift get is/origin-cluster-logging-operator -o jsonpath='{.status.tags[*].tag}' 
master release-5.3

# delete the deployments
> oc delete ns openshift-logging openshift-operators-redhat

# easily switch back to release-5.3, without building and pushing image to cluster again
> git checkout release-5.3
> make deploy-catalog
> make install

# now cluster has logging deployed from release-5.3

```

/cc 
/assign 

/cherry-pick release-5.3

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:

### DIFF
--- a/olm_deploy/operatorregistry/registry-deployment.yaml
+++ b/olm_deploy/operatorregistry/registry-deployment.yaml
@@ -18,7 +18,7 @@ spec:
       initContainers:
       - name: mutate-csv-and-generate-sqlite-db
         image: ${IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY}
-        imagePullPolicy: IfNotPresent
+        imagePullPolicy: Always
         command:
         - sh
         args:
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,8 @@ export GO111MODULE=on
 export GODEBUG=x509ignoreCN=0
 
 export APP_NAME=cluster-logging-operator
-export IMAGE_TAG?=127.0.0.1:5000/openshift/origin-$(APP_NAME):latest
+export CURRENT_BRANCH=$(shell git rev-parse --abbrev-ref HEAD)
+export IMAGE_TAG?=127.0.0.1:5000/openshift/origin-$(APP_NAME):$(CURRENT_BRANCH)
 
 export LOGGING_VERSION?=$(shell basename $(shell ls -d manifests/[0-9]*))
 export NAMESPACE?=openshift-logging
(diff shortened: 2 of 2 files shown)


jcantrill (commented):
/lgtm

olm_deploy/operatorregistry/registry-deployment.yaml
```diff
@@ -18,7 +18,7 @@ spec:
       initContainers:
       - name: mutate-csv-and-generate-sqlite-db
         image: ${IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY}
-        imagePullPolicy: IfNotPresent
+        imagePullPolicy: Always
```
alanconway, line 21: I don't think we want Always here unconditionally - I can see why it's helpful in development but it's expensive/fragile in a production system. I suggest using the short git commit hash as part of the image name - that way the image name will always be different if it was built from different code and there's no need for Always.

alanconway (changes requested):

olm_deploy/operatorregistry/registry-deployment.yaml
alanconway: /hold

vimalk78: using git hash in development may not that useful, a readable branch name is more effective.
moreover this imagePullPolicy was already set to `Always` , i had changed it to `IfNotPresent` in earlier PR

vimalk78: in development we use env variables to set up the images to be used in logging deployment `IMAGE_CLUSTER_LOGGING_OPERATOR` , setting an env variable with git hash in the image name will not be developer friendly

vimalk78: @alanconway does this address your concerns

alanconway: @vimalk78 Yep, it makes sense for development. 
Not sure why we have Always but it sounds like that is a separate question from this PR.

PR #1269, opened 2021-11-29
LOG-1925: Fix cpu graph on dashboard

jcantrill:
### Description
This PR fixes the CPU dashboard graph:
* fixing the metrics
* correcting the container name for collectors

### Links
* https://issues.redhat.com/browse/LOG-1925

### DIFF
--- a/files/dashboards/openshift-logging-dashboard.json
+++ b/files/dashboards/openshift-logging-dashboard.json
@@ -947,7 +947,7 @@
       "steppedLine": false,
       "targets": [
         {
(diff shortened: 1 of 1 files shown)


ajaygupta978 (commented):
@jcantrill : Any specific reason to use irate instead of rate.  Based on the documentation it says 
"irate should only be used when graphing volatile, fast-moving counters. Use rate for alerts and slow-moving counters".

jcantrill:
> @jcantrill : Any specific reason to use irate instead of rate. Based on the documentation it says "irate should only be used when graphing volatile, fast-moving counters. Use rate for alerts and slow-moving counters".

The only reason I could find is that "rate" doesn't exist; the graph is blank. The "rate" version wasn't even suggested as part of the auto-complete.  When changed to "rate" it provided a graph

PR #1273, opened 2021-11-30
LOG-2028: Enable vector in functional framework

jcantrill:
### Description
This PR:

* Initially enables the functional framwork to support vector
* Consolidates the collector sections into a common factory

/assign @vimalk78 

One test is enabled (which fails) and can be run with 'make test-functional-vector'

### Links
* https://issues.redhat.com/browse/LOG-2028


alanconway (approved):
/lgtm

PR #1282, opened 2021-12-09
LOG-2046: fix wrong certs when OutputDefaults are defined

jcantrill:
### Description
This PR:
* Fixes how outputdefaults are applied to only set them if not defined on the output
* Fixes modifying the output secret if OutputDefaults are applied
* Fixes the forward-generator tool 

/assign @alanconway 

### Links
* https://issues.redhat.com/browse/LOG-2046


alanconway (approved):
/lgtm

PR #1289, opened 2021-12-16
LOG-1910: Mount host filesystem to leverage CRIO rollover

jcantrill:
### Description
This PR:
* Refactors the functional-benchmarker to mount the host filesystem to take advantage of CRIO rollover instead of mimicking it

This will make our tests more accurate and additionally allow us to add logic to check for log loss due to load and file rotation.

/cc @alanconway @syedriko 

### Links
* https://issues.redhat.com/browse/LOG-1910


alanconway (approved):
/lgtm

PR #1292, opened 2022-01-06
LOG-2117 Remove deprecated Ginkgo custom reporters

alanconway:
/cc @jcantril


jcantrill:
Why not migrate  to v2 and use the built in junit reports so we can see failures in CI https://onsi.github.io/ginkgo/MIGRATING_TO_V2

jcantrill:
@alanconway one of the ideas for the reporters is the results appear in the CI properly instead of having to review the log.  I don't think we ever had it working properly. We should consider upgrading to v2 and adding the reporters back in.

PR #1293, opened 2022-01-07
LOG-2066: Modify client to also watch cluster config NS

jcantrill:
### Description
This PR:
* Fixes configmap dashboard reconciliation by configuring API client to watch openshift-config-managed NS

### Links
ref: https://issues.redhat.com/browse/LOG-2066

### DIFF
--- a/controllers/clusterlogging/clusterlogging_controller.go
+++ b/controllers/clusterlogging/clusterlogging_controller.go
@@ -58,7 +58,12 @@ var _ reconcile.Reconciler = &ReconcileClusterLogging{}
 type ReconcileClusterLogging struct {
 	// This Client, initialized using mgr.Client() above, is a split Client
 	// that reads objects from the cache and writes to the apiserver
-	Client   client.Client
+	Client client.Client
+
+	// Reader is an initialized client.Reader that reads objects directly from the apiserver
+	// instead of the cache. Useful for cases where need to read/write to a namespace other than
+	// the deployed namespace (e.g. openshift-config-managed)
+	Reader   client.Reader
 	Scheme   *runtime.Scheme
 	Recorder record.EventRecorder
 }
@@ -93,7 +98,7 @@ func (r *ReconcileClusterLogging) Reconcile(ctx context.Context, request reconci
 		return reconcile.Result{}, nil
 	}
 
-	if err = k8shandler.Reconcile(instance, r.Client, r.Recorder); err != nil {
+	if err = k8shandler.Reconcile(instance, r.Client, r.Reader, r.Recorder); err != nil {
 		log.Error(err, "Error reconciling clusterlogging instance")
 	}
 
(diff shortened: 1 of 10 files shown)


controllers/clusterlogging/clusterlogging_controller.go
```diff
@@ -58,7 +58,12 @@ var _ reconcile.Reconciler = &ReconcileClusterLogging{}
 type ReconcileClusterLogging struct {
 	// This Client, initialized using mgr.Client() above, is a split Client
 	// that reads objects from the cache and writes to the apiserver
-	Client   client.Client
+	Client client.Client
+
+	//Reader is an initialized client.Reader that reads objects directly from the apiserver
+	//instead of the cache. Useful for cases where need to read/write to a namespace other than
+	//the deployed namespace (e.g. openshift-config-managed)
```
xperimental, line 65:
```suggestion
	// Reader is an initialized client.Reader that reads objects directly from the apiserver
	// instead of the cache. Useful for cases where need to read/write to a namespace other than
	// the deployed namespace (e.g. openshift-config-managed)
```

internal/utils/utils.go
```diff
@@ -191,14 +191,18 @@ const defaultShareDir = "/usr/share/logging"
 func GetShareDir() string {
 	// shareDir is <repo-root>/files - try to find from working dir.
 	const sep = string(os.PathSeparator)
-	const repoRoot = sep + "cluster-logging-operator" + sep
+	const repoRoot = sep + "cluster-logging-operator"
 	wd, err := os.Getwd()
 	if err != nil {
+		log.V(3).Error(err, "Error determining share directory.  Returning default", "default", defaultShareDir)
```
xperimental, line 197:
```suggestion
		log.V(3).Error(err, "Error determining share directory. Returning default", "default", defaultShareDir)
```

internal/utils/comparators/configmaps/comparator.go
```diff
@@ -15,6 +16,9 @@ const (
 //AreSame compares configmaps for equality and return true equal otherwise false.  This comparison
 //only compares the data of the configmap by default unless otherwise configured
 func AreSame(actual *corev1.ConfigMap, desired *corev1.ConfigMap, options ...ComparisonOption) bool {
+	log.V(3).Info("Compare configmaps", "actual", actual)
+	log.V(3).Info("Compare configmaps", "desired", desired)
+	log.V(3).Info("Compare configmaps", "options", options)
```
xperimental, line 21: This seems very verbose. Does it make sense to remove it or increase the verbosity needed?

internal/utils/utils.go
```diff
@@ -191,14 +191,18 @@ const defaultShareDir = "/usr/share/logging"
...
 		return defaultShareDir
 	}
+	log.V(3).Info("GetShareDir", "workingdir", wd)
```
xperimental, line 200: This seems very verbose. Does it make sense to remove it or increase the verbosity needed?

xperimental (commented):
Thanks for the quick chat about the questions. Unfortunately this means that only the nit-picky comments remain.

PR #1334, opened 2022-02-15
LOG-2099: Add optionalint in prep for fix

jcantrill:
### Description
This PR:
* Adds an OptionalInt type for the message tests in preparation to add an openshift.sequence number

### Links
* https://issues.redhat.com/browse/LOG-2099

### DIFF
--- a/test/helpers/types/optional_int.go
+++ b/test/helpers/types/optional_int.go
@@ -1,0 +1,75 @@
+package types
+
+import (
+	"fmt"
+	"github.com/ViaQ/logerr/log"
+	"regexp"
+	"strconv"
+	"strings"
+)
+
+//OptionalInt allows passing an arbitrary int as well matching conditional values
+//for message comparison (e.g. >=6)
+type OptionalInt string
+
+func (oi *OptionalInt) MarshalJSON() ([]byte, error) {
+	return []byte(*oi), nil
+}
+
+func (oi *OptionalInt) UnmarshalJSON(data []byte) error {
+	*oi = OptionalInt(string(data))
+	return nil
+}
+
+func (oi OptionalInt) getParts() (string, int) {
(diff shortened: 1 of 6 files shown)


test/helpers/types/optional_int.go
```diff
@@ -0,0 +1,75 @@
...
+
+func (oi *OptionalInt) UnmarshalJSON(data []byte) error {
+	*oi = OptionalInt(string(data))
+	return nil
+}
+
+func (oi OptionalInt) getParts() (string, int) {
+	comparison := "="
+	value := 0
+	optionalIntRE := regexp.MustCompile(`(?P<comparison>[><=]{1,2})?(?P<value>\d*)`)
```
syedriko, line 27: should this be in the package scope?

jcantrill: I purposely made it private as i didn't see an immediate need for it to be utilized publicly

syedriko: I meant lifetime, not export status:
```var optionalIntRE = regexp.MustCompile(`(?P<comparison>[><=]{1,2})?(?P<value>\d*)`)

func (oi OptionalInt) getParts() (string, int) {

...
```

jcantrill: I left it as-is to ensure the group names stayed close to where they were used.  Those could also be factored into private constants.  I'm indifferent here because it's a test type not being used in a production scenario; efficiency isn't of main concern.  I defer to you how firm your belief is that it should be elsewhere for the change, otherwise I'm find as-is

jcantrill: @syedriko I think we will need follow-up PRs anyway to verify we are getting the sequence number.  I'll try to clean up in subsequent PRs

PR #1342, opened 2022-02-17
LOG-2255: Vector: Support Namespace filtering

vimalk78:
### Description
Add support for namespace filtering in Vector config.

/cc @jcantrill @alanconway 
/assign 

/cherry-pick 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-2255

### DIFF
--- a/internal/generator/vector/sources_to_inputs.go
+++ b/internal/generator/vector/sources_to_inputs.go
@@ -2,6 +2,7 @@ package vector
 
 import (
 	"fmt"
+	"strings"
 
 	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
 	"github.com/openshift/cluster-logging-operator/internal/generator"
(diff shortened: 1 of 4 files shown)


internal/generator/vector/sources_to_inputs.go
```diff
@@ -74,7 +81,31 @@ func SourcesToInputs(spec *logging.ClusterLogForwarderSpec, o generator.Options)
...
+				if input.Application != nil {
+					app := input.Application
+					if len(app.Namespaces) != 0 {
+						matchNS := []string{}
+						for _, ns := range app.Namespaces {
+							matchNS = append(matchNS, fmt.Sprintf(IsNamespaceLog, ns))
+						}
+						el = append(el, Route{
+							ComponentID: RouteApplicationLogs,
+							Inputs:      helpers.MakeInputs("application"),
```
jcantrill, line 99: nit.  dont we have a logtype constant for "application"

vimalk78: actually, this is not a log type, this is name of the config block which generates application logs.
May be i can add another constant

alanconway: Seems like you could reasonably use the same constant, since the config block is related to the log type.

PR #1346, opened 2022-02-18
LOG-2255: Vector: Support Pod Label based routing

vimalk78:
### Description
Support Pod Label based routing of Application logs

/cc @jcantrill @alanconway 
/assign 

/cherry-pick 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-2255

### DIFF
--- a/internal/generator/vector/vrl.go
+++ b/internal/generator/vector/vrl.go
@@ -1,0 +1,49 @@
+package vector
+
+import (
+	"fmt"
+	"strings"
+)
+
+var (
+	SkipEmpty = func(in []string) []string {
+		vals := []string{}
+		for _, v := range in {
+			if strings.TrimSpace(v) != "" {
+				vals = append(vals, v)
+			}
+		}
+		return vals
+	}
+	Paren = func(in string) string {
+		return fmt.Sprintf("(%s)", in)
+	}
+	ParenAll = func(in []string) []string {
+		if len(in) == 1 {
+			return in
+		}
+		vals := []string{}
+		for _, v := range in {
+			vals = append(vals, Paren(v))
+		}
+		return vals
+	}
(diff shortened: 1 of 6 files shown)


internal/generator/vector/vrl.go
```diff
@@ -0,0 +1,49 @@
...
+			}
+		}
+		return vals
+	}
+	Paren = func(in string) string {
+		return fmt.Sprintf("(%s)", in)
+	}
+	ParenAll = func(in []string) []string {
+		if len(in) == 1 {
+			return in
```
ajaygupta978, line 23: shouldn't this be Paren(in) ?

internal/generator/vector/vrl.go
```diff
@@ -0,0 +1,49 @@
...
+		return fmt.Sprintf("starts_with!(%s,%q)", x, y)
+	}
+	Eq = func(x, y string) string {
+		return fmt.Sprintf("%s == %q", x, y)
+	}
+	Quote = func(expr string) string {
+		return fmt.Sprintf("'%s'", expr)
+	}
+	OR = func(nsExpr ...string) string {
+		return strings.Join(ParenAll(SkipEmpty(nsExpr)), " || ")
```
ajaygupta978, line 41: what if nsExpr is empty, this will result in " || " ? Similarly for AND

internal/generator/vector/vrl.go
vimalk78: if there is one element only, no need to parenthesize

internal/generator/vector/vrl.go
vimalk78: where ever the `OR` and `AND` are being used, the non-zero length is being checked.

PR #1350, opened 2022-02-23
LOG-2259: Vector: Add Route transform for application routing only once

vimalk78:
### Description
The Route transform was added for each input with user defined inputs. Just one Route transform is sufficient with a map for each user defined input.

/cc @jcantrill 
/assign 

/cherry-pick 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-2259

### DIFF
--- a/internal/generator/vector/sources_to_inputs.go
+++ b/internal/generator/vector/sources_to_inputs.go
@@ -96,6 +96,7 @@ func SourcesToInputs(spec *logging.ClusterLogForwarderSpec, o generator.Options)
 		el = append(el, r)
 	}
 	userDefined := spec.InputMap()
+	routeMap := map[string]string{}
 	for _, pipeline := range spec.Pipelines {
 		for _, inRef := range pipeline.InputRefs {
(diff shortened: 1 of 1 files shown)


internal/generator/vector/sources_to_inputs.go
```diff
@@ -96,6 +96,7 @@ func SourcesToInputs(spec *logging.ClusterLogForwarderSpec, o generator.Options)
 		el = append(el, r)
 	}
 	userDefined := spec.InputMap()
+	RouteMap := map[string]string{}
```
jcantrill, line 99: nit.   should be `routeMap`

jcantrill:
@vimalk78 I have one nit.  I defer to you if you want to fix it in a subsequent PR

internal/generator/vector/sources_to_inputs.go
vimalk78: updated

jcantrill (approved):
/approve

PR #1356, opened 2022-02-28
Vector: Add vector to forwarder-generator

vimalk78:
### Description
Added an option "collector" to forwarder-generator, and supported two values `fluentd` and `vector`

```
$ ./bin/forwarder-generator -h
Usage of ./bin/forwarder-generator:
      --collector string        collector type: fluentd or vector (default "fluentd")
      --debug-output            Generate config normally, but replace output plugins with @stdout plugin, so that records can be printed in collector logs.
      --file string             ClusterLogForwarder yaml file. - for stdin
      --help                    This message
      --include-default-store   Include the default storage when generating the config (default true)

```

/cc @jcantrill 
/assign 

/cherry-pick

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -183,7 +183,8 @@ test-functional-vector:
 .PHONY: test-functional-vector
 
 test-forwarder-generator: bin/forwarder-generator
-	@bin/forwarder-generator --file hack/logforwarder.yaml > /dev/null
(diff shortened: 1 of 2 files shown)


jcantrill (commented):
/lgtm

PR #1359, opened 2022-03-01
WIP LOG-2014:Support user defined application inputs

ajaygupta978:
Adding E2E test case which verifies logs are filtered based on name space.

- JIRA: https://issues.redhat.com/browse/LOG-2014


vimalk78:
i will suggest to change in [namespace_filtering_test.go](https://github.com/openshift/cluster-logging-operator/pull/1359/files#diff-c178e5371d8680cc29f21bcb089f3d27a60cd7bfa340cd7e204b6d66f8a7339c) and enable the same test in it for vector.

The test has nothing specific about fluentd, not sure why it got added to fluentd folder. you can move the file to proper place if you think.

test/e2e/collection/namespace_filtering_test.go
```diff
@@ -0,0 +1,188 @@
...
+			docs := jsonMap["hits"].(map[string]interface{})["hits"].([]interface{})
+			for _, document := range docs {
+				namespace_value := document.(map[string]interface{})["_source"].(map[string]interface{})["kubernetes"].(map[string]interface{})["namespace_name"].(string)
+				if namespace_value != namespace {
+					return namespace_value, nil
+				}
+			}
+		}
+		return namespace, nil
+	}
```
vimalk78, line 138: The function is named `GetNamespace` while taking a namespace name as argument.
why not change the function to return all namespaces from app logs, and then check which namespaces should be present.

test/e2e/collection/namespace_filtering_test.go
```diff
@@ -0,0 +1,188 @@
...
+			cr := helpers.NewClusterLogging(helpers.ComponentTypeCollectorVector)
+			if err := e2e.CreateClusterLogging(cr); err != nil {
+				Fail(fmt.Sprintf("Unable to create an instance of cluster logging: %v", err))
+			}
+			if err := e2e.WaitFor(helpers.ComponentTypeCollectorVector); err != nil {
+				Fail(fmt.Sprintf("Failed waiting for component %s to be ready: %v", helpers.ComponentTypeCollectorVector, err))
+			}
+		})
+		AssertBehaviours()
+	})
```
vimalk78, line 174: is it possible to write this as a for loop? better if we can, and then update other tests also.

alanconway:
DescribeTable/Entry is the Ginkgo construct for writing repetative "table tests", there are examples in the tests.

test/e2e/collection/collector_suite_test.go
```diff
@@ -0,0 +1,13 @@
...
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestCollector(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "ClusterLogging E2E Suite - Collectors")
```
jcantrill, line 12: Lets change this description to "[e2e][Collection] Suite"

test/e2e/collection/namespace_filtering_test.go
jcantrill: Logic to retrieve app logs via E2E test is useful outside this test and should be implemented here: https://github.com/openshift/cluster-logging-operator/blob/master/test/framework/e2e/elasticsearch.go#L95  It may be possible to do something clever by refactoring this logic https://github.com/openshift/cluster-logging-operator/blob/master/test/framework/functional/elasticsearch.go#L26  into a place that can be commonly used

test/e2e/collection/namespace_filtering_test.go
jcantrill: Agree  here and with @alanconway .  we should convert this to a DescribeTable

ajaygupta978:
@jcantrill  regarding https://github.com/openshift/cluster-logging-operator/pull/1359#discussion_r828040624  in our test cases we needed only namespace list which is specific to test case. So I wrote method in the test case itself rather than common file. Another way is to implement https://github.com/openshift/cluster-logging-operator/blob/master/test/framework/e2e/elasticsearch.go#L95. which returns logs and then in test case we can parse to extract namespace list. 
Please advise.

alanconway (approved):
/lgtm

test/e2e/collection/namespace_filtering_test.go
```diff
@@ -0,0 +1,182 @@
+package collection
```
jcantrill, line 1: Why did we need to completely rewrite this test instead of doing a `git mv` of the old to the new namespace and modify it accordingly

test/e2e/collection/namespace_filtering_test.go
```diff
@@ -0,0 +1,182 @@
...
+			return nil, errors.New("No pods found for elasticsearch")
+		}
+		log.V(3).Info("Pod ", "PodName", pods.Items[0].Name)
+		query := "--query=" + app_index + "/_search?size=10000"
+		stdout, err := e2e.PodExec(constants.OpenshiftNS, pods.Items[0].Name, "elasticsearch", []string{"es_util", query})
+		if err != nil {
+			return nil, errors.New("Pods exec error")
+		}
+		// Get total hits
+		var jsonMap map[string]interface{}
```
jcantrill, line 121: All of this code is a duplicate to that which lives in a the functional framework.  Please refactor into a common place if needed (e.g. test/helpers/elasticsearch)

test/e2e/collection/namespace_filtering_test.go
```diff
@@ -0,0 +1,182 @@
...
+	DescribeTable("Running collector tests",
+		func(collectorType helpers.LogComponentType) {
+			cr := helpers.NewClusterLogging(collectorType)
+			if err := e2e.CreateClusterLogging(cr); err != nil {
+				Fail(fmt.Sprintf("Unable to create an instance of cluster logging: %v", err))
+			}
+			if err := e2e.WaitFor(collectorType); err != nil {
+				Fail(fmt.Sprintf("Failed waiting for component %s to be ready: %v", collectorType, err))
+			}
+			AssertBehaviours()
```
jcantrill, line 164: Consider replacing this with the function content;  I don't see the advantage to move it into a separate function

jcantrill (commented):
/hold

test/e2e/collection/namespace_filtering_test.go
jcantrill: @ajaygupta978 :point_up:

test/framework/e2e/elasticsearch.go
```diff
@@ -93,7 +93,63 @@ type ElasticLogStore struct {
...
+		result, err := es.Framework.PodExec(constants.OpenshiftNS, pods.Items[0].Name, "elasticsearch", []string{"es_util", cmd})
+		if err != nil {
+			return false, err
+		}
+		if result != "" && err == nil {
+			var elasticResult map[string]interface{}
+			err = json.Unmarshal([]byte(result), &elasticResult)
+			if err == nil {
+				if elasticResult["timed_out"] == false {
+					resultHits := elasticResult["hits"].(map[string]interface{})
```
jcantrill, line 122: This same/similar code lives in test/functional/elasticsearch.go.  Please find commonality that can be refactored into test/helpers/elasticsearch/elasticsearch.go

test/helpers/elastichelper/elastichelper.go
```diff
@@ -0,0 +1,51 @@
+package elastichelper
```
jcantrill, line 1: please name this package 'elasticsearch'

test/framework/functional/elasticsearch.go
```diff
@@ -20,54 +19,26 @@ func (f *CollectorFunctionalFramework) GetLogsFromElasticSearch(outputName strin
 	if !ok {
 		return "", fmt.Errorf(fmt.Sprintf("can't find log of type %s", outputLogType))
 	}
-	return f.GetLogsFromElasticSearchIndex(outputName, index)
+	return f.GetLogsFromElasticSearchIndex(f.Test.NS.Name, f.Pod.Name, outputName, index)
 }
 
-func (f *CollectorFunctionalFramework) GetLogsFromElasticSearchIndex(outputName string, index string) (result string, err error) {
+func (f *CollectorFunctionalFramework) GetLogsFromElasticSearchIndex(namespaceName string, podName string, outputName string, index string) (result string, err error) {
```
jcantrill, line 25: The method arg does not need to be changed.  These can be obtained from 'f' as part of the framework

jcantrill (changes requested):

jcantrill:
> @jcantrill [#1359 (comment)](https://github.com/openshift/cluster-logging-operator/pull/1359#discussion_r837548165) Do you mean why we wrote new test case ? In the old package we had only fluentd test case which was using fluent forward as end point. Since vector doesn't support it so we have to elastic search as common end point for fluentd and vector. This required new code to be written in elasticsearch.go.

this is mostly the same testcase which means you could "git mv" it to a new package and edit it accordingly.  This preserves history which can be important during investigation

PR #1371, opened 2022-03-08
LOG-2333: Fix ES DeDot lua script dropping Journal logs

vimalk78:
### Description
Before the events are sent to Elasticsearch sink, a Lua script is used to substitute `.` with `_` in the pod_labels.
The lua script logic does a check for presence of `kubernetes` field in the record, and proceeds to do `DeDot` only if this field is present. The event was not emitted if this field was not present.

Fixed lua script to emit the event if `kubernetes` field is not found.

/cc @jcantrill 
/assign 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-2333

### DIFF
--- a/internal/generator/vector/normalize.go
+++ b/internal/generator/vector/normalize.go
@@ -12,13 +12,13 @@ import (
 const (
 	FixLogLevel = `
 level = "unknown"
-if match(.message,r'(Warning|WARN|W[0-9]+|level=warn|Value:warn|"level":"warn")'){
+if match!(.message,r'(Warning|WARN|W[0-9]+|level=warn|Value:warn|"level":"warn")'){
   level = "warn"
-} else if match(.message, r'Info|INFO|I[0-9]+|level=info|Value:info|"level":"info"'){
+} else if match!(.message, r'Info|INFO|I[0-9]+|level=info|Value:info|"level":"info"'){
   level = "info"
-} else if match(.message, r'Error|ERROR|E[0-9]+|level=error|Value:error|"level":"error"'){
+} else if match!(.message, r'Error|ERROR|E[0-9]+|level=error|Value:error|"level":"error"'){
   level = "error"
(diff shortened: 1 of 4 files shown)


internal/generator/vector/normalize.go
```diff
@@ -12,13 +12,13 @@ import (
 const (
 	FixLogLevel = `
 level = "unknown"
-if match(.message,r'(Warning|WARN|W[0-9]+|level=warn|Value:warn|"level":"warn")'){
+if match!(.message,r'(Warning|WARN|W[0-9]+|level=warn|Value:warn|"level":"warn")'){
```
jcantrill, line 15: How does this change relate to that described in the bug?

internal/generator/vector/sources_to_inputs.go
```diff
@@ -35,6 +35,7 @@ var (
 	AddLogTypeApp   = fmt.Sprintf(".log_type = %q", logging.InputNameApplication)
 	AddLogTypeInfra = fmt.Sprintf(".log_type = %q", logging.InputNameInfrastructure)
 	AddLogTypeAudit = fmt.Sprintf(".log_type = %q", logging.InputNameAudit)
+	AddInfraType    = fmt.Sprintf(".infra_type=%q", "journal")
```
jcantrill, line 38: Is this field new in the data model?  I don't recall this existing before

internal/generator/vector/normalize.go
vimalk78: not related directly. 
Since i was suspecting that journal source is not working in vector version i was using (the metrics also showed 0),  I wanted to check this bug on latest version of vector. 
The latest vector gave an error for these lines.

internal/generator/vector/sources_to_inputs.go
vimalk78: I added this field, since journal logs can easily get lost in the infrastructure logs

vimalk78: removed it

jcantrill (commented):
/lgtm

PR #1379, opened 2022-03-11
LOG-1659: Update dev images to use public bases

jcantrill:
### Description
This PR:
* Replaces ocp specific base images with those that are publicly available with RH or other subscriptions

### Links
* https://issues.redhat.com/browse/LOG-1659

cc @alanconway

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 ### This is a generated file from Dockerfile.in ###
 #@follow_tag(registry-proxy.engineering.redhat.com/rh-osbs/openshift-golang-builder:rhel_8_golang_1.16)
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.8 AS builder
+FROM registry.redhat.io/ubi8/go-toolset:1.16.12 AS builder
 
 ENV BUILD_VERSION=${CI_CONTAINER_VERSION}
 ENV OS_GIT_MAJOR=${CI_X_VERSION}
@@ -29,14 +29,15 @@ COPY ${REMOTE_SOURCE}/apis ./apis
 COPY ${REMOTE_SOURCE}/controllers ./controllers
 COPY ${REMOTE_SOURCE}/internal ./internal
 
+USER 0
 RUN make build
 
 
 #@follow_tag(registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-cli:v4.8)
-FROM registry.ci.openshift.org/ocp/4.8:cli AS origincli
+FROM quay.io/openshift/origin-cli:4.8 AS origincli
 
 #@follow_tag(registry.redhat.io/ubi8:latest)
-FROM registry.ci.openshift.org/ocp/4.8:base
+FROM registry.redhat.io/ubi8:8.5
 RUN INSTALL_PKGS=" \
       openssl \
       rsync \
--- a/Dockerfile.patch
+++ b/Dockerfile.patch
@@ -1,3 +1,0 @@
-22a23,24
-> RUN rm /etc/yum.repos.d/*
-> ADD art.repo /etc/yum.repos.d
--- a/origin-meta.yaml
+++ b/origin-meta.yaml
@@ -1,7 +1,7 @@
 from:
   - source: registry-proxy.engineering.redhat.com/rh-osbs/openshift-golang-builder\:v(?:[\.0-9\-]*).*
-    target: registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.8 AS builder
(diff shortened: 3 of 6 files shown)


Dockerfile
```diff
@@ -1,6 +1,6 @@
 ### This is a generated file from Dockerfile.in ###
 #@follow_tag(registry-proxy.engineering.redhat.com/rh-osbs/openshift-golang-builder:rhel_8_golang_1.16)
-FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.8 AS builder
+FROM registry.access.redhat.com/ubi8/go-toolset AS builder
```
vimalk78, line 3: suggest to use
```
registry.redhat.io/ubi8/go-toolset:1.16.12 
```

Dockerfile
```diff
@@ -29,14 +29,15 @@ COPY ${REMOTE_SOURCE}/apis ./apis
 COPY ${REMOTE_SOURCE}/controllers ./controllers
 COPY ${REMOTE_SOURCE}/internal ./internal
 
+USER 0
```
vimalk78, line 32: why we need privileged access in CLO?

Dockerfile
```diff
@@ -29,14 +29,15 @@ COPY ${REMOTE_SOURCE}/apis ./apis
...
 RUN make build
 
 
 #@follow_tag(registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-cli:v4.8)
-FROM registry.ci.openshift.org/ocp/4.8:cli AS origincli
+FROM quay.io/openshift/origin-cli:4.8 AS origincli
```
vimalk78, line 37: suggest to use 
```
registry.redhat.io/openshift4/ose-cli:v4.10 
```

Dockerfile
```diff
@@ -29,14 +29,15 @@ COPY ${REMOTE_SOURCE}/apis ./apis
...
 
 #@follow_tag(registry.redhat.io/ubi8:latest)
-FROM registry.ci.openshift.org/ocp/4.8:base
+FROM registry.access.redhat.com/ubi8
```
vimalk78, line 40: suggest to use
```
registry.redhat.io/ubi8:8.5 
```

origin-meta.yaml
```diff
@@ -1,7 +1,7 @@
 from:
   - source: registry-proxy.engineering.redhat.com/rh-osbs/openshift-golang-builder\:v(?:[\.0-9\-]*).*
-    target: registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.16-openshift-4.8 AS builder
+    target: registry.access.redhat.com/ubi8/go-toolset AS builder
   - source: registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-cli:v4.8.0-([0-9]*).*
-    target: registry.ci.openshift.org/ocp/4.8:cli AS origincli
+    target: quay.io/openshift/origin-cli:4.8 AS origincli
   - source: registry.redhat.io/ubi8:8.(\d)-([\.0-9])*
-    target: registry.ci.openshift.org/ocp/4.8:base
+    target: registry.access.redhat.com/ubi8
```
vimalk78, line 7: same as above comments

Dockerfile
jcantrill: Build failed making a directory upon change.  I'm open to suggestions.  If this is only in the "builder" image then maybe it's a non-issue as I would expect the "run" base does not use this user.  Do you know?

Dockerfile
jcantrill: We know for certain these images are public?

jcantrill: this one requires agreement to subscription: ```Before downloading or using this Container, you must agree to the Red Hat subscription agreement located at [redhat.com/licenses](https://www.redhat.com/en/about/agreements). If you do not agree with these terms, do not download or```

Dockerfile
jcantrill: this one is fine `Terms & conditions
The Red Hat Universal Base Image is free to deploy on Red Hat or non-Red Hat platforms and freely redistributable. S`

Dockerfile
vimalk78: how do you get this error?

Dockerfile.patch
```diff
@@ -1,3 +0,0 @@
-22a23,24
-> RUN rm /etc/yum.repos.d/*
-> ADD art.repo /etc/yum.repos.d
```
vimalk78, line 3: we can remove `art.repo` also if it is not needed

Dockerfile
jcantrill: This is not an error but the verbiage displayed on the catalog UI

PR #1383, opened 2022-03-14
LOG-1713: fix prometheus roles to be namespaced scoped

jcantrill:
### Description
This PR:
* Fixes the prometheus roles to be namespace scoped

### Links
* https://issues.redhat.com/browse/LOG-1713

### DIFF
--- a/bundle/manifests/cluster-logging-operator-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
+++ b/bundle/manifests/cluster-logging-operator-metrics-monitor_monitoring.coreos.com_v1_servicemonitor.yaml
@@ -6,7 +6,7 @@ metadata:
   name: cluster-logging-operator-metrics-monitor
 spec:
   endpoints:
-    - port: http-metrics
+  - port: http-metrics
   namespaceSelector: {}
   selector:
     matchLabels:
--- a/bundle/manifests/log-collector-privileged-binding_rbac.authorization.k8s.io_v1_rolebinding.yaml
+++ b/bundle/manifests/log-collector-privileged-binding_rbac.authorization.k8s.io_v1_rolebinding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  creationTimestamp: null
+  name: log-collector-privileged-binding
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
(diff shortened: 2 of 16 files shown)


jcantrill:
> LGTM, but did you forget to generate the bundle maybe?

Ran make target but it appears it updated files outside of the changes I was targeting

periklis (commented):
One more thing, you need to ensure that you manual delete the old clusterrole and clusterrolebinding from the bundle and manifests dir.

jcantrill:
> One more thing, you need to ensure that you manual delete the old clusterrole and clusterrolebinding from the bundle and manifests dir.

@periklis I believe this change does accomplish removing those roles/bindings.  Existing deployments I do not believe will be updated to remove them.  I think this change also moves us one step closer to being able to remove the manifest directory

periklis:
> > One more thing, you need to ensure that you manual delete the old clusterrole and clusterrolebinding from the bundle and manifests dir.
> 
> @periklis I believe this change does accomplish removing those roles/bindings. Existing deployments I do not believe will be updated to remove them. I think this change also moves us one step closer to being able to remove the manifest directory

Nope the `make bundle` command does not remove the ClusterRole and ClusterRoleBinding files at all. At the current SHA1 for this PR, the file still exists: https://github.com/openshift/cluster-logging-operator/blob/ba4354322a6be05457b25543af1752462bad2a24/bundle/manifests/cluster-logging-metadata-reader_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml

jcantrill:
> > > One more thing, you need to ensure that you manual delete the old clusterrole and clusterrolebinding from the bundle and manifests dir.
> > 
> > 
> > @periklis I believe this change does accomplish removing those roles/bindings. Existing deployments I do not believe will be updated to remove them. I think this change also moves us one step closer to being able to remove the manifest directory
> 
> Nope the `make bundle` command does not remove the ClusterRole and ClusterRoleBinding files at all. At the current SHA1 for this PR, the file still exists: https://github.com/openshift/cluster-logging-operator/blob/ba4354322a6be05457b25543af1752462bad2a24/bundle/manifests/cluster-logging-metadata-reader_rbac.authorization.k8s.io_v1_clusterrolebinding.yaml

This is the role the collector uses to talk to the API server, not for prometheus to scrap the collector.  Interestingly though I wonder if we can reduce it to a namespaces role or if you require clusterwide in order to get pod meta from any pod across the cluster

PR #1386, opened 2022-03-15
adding a check on invalid json file /var/lib/fluentd/pos/journal_pos.…

pmoogi-redhat:
### Description
On certain platform like POWER, when a node crashes and if there is a left over /var/lib/fluentd/pos/journal_pos.json file which got invalid json format then fluentd pod goes into crashloopbackoff state. This fix introduces a small ruby script using the yajl parser to catch error during parsing. If it finds invalid json file then it just delete it before it starts the fluentd process.

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-2159
- Enhancement proposal:


internal/components/fluentd/run_script.go
```diff
@@ -132,6 +132,46 @@ do
...
 `
+
+const CleanInValidJson = `
+#!/usr/bin/env ruby
+
+require 'yajl'
+require 'json'
+
+
+FILE = "/var/lib/fluentd/pos/journal_pos.json"
```
vimalk78, line 161: better to generate the file name from code. since this path is not a standard path. This must be same as path generated in fluent.conf.

internal/components/fluentd/run_script.go
```diff
@@ -132,6 +132,46 @@ do
...
+
+    echo "$FILE exists."
+
+if test -f "$FILE"; then
+    echo "$FILE exists."
+    ruby -v /etc/fluent/configs.d/user/cleanInValidJson.rb  2> /tmp/error-json-parser.log
+
+    if [ -s /tmp/error-json-parser.log ]; then
+      echo "its invalid json"
+      sudo rm $FILE
```
vimalk78, line 145: should we use `rm -f $FILE`
add a message here, saying which file is removed

internal/components/fluentd/run_script.go
```diff
@@ -132,6 +132,46 @@ do
     fi
 done
 
+FILE="/var/lib/fluentd/pos/journal_pos.json"
+
+    echo "$FILE exists."
+
+if test -f "$FILE"; then
+    echo "$FILE exists."
+    ruby -v /etc/fluent/configs.d/user/cleanInValidJson.rb  2> /tmp/error-json-parser.log
```
syedriko, line 141: it'd be much more UNIX-like if the check simply returned 0 for valid and 1 for invalid JSON

syedriko: and the JSON file to check to be a parameter

syedriko: and cleanInValidJson.rb can be a here doc that can be fed to ruby on stdin

internal/components/fluentd/run_script.go
pmoogi-redhat: now passing FILE name as an argument to run.sh bash script. i can possibly grep pos files from fluent.conf and run over them a loop to check if they have a valid json file content or not.

internal/components/fluentd/run_script.go
IshwarKanse: There is no sudo in the image, we can use the rm -f $FILE without sudo.

PR #1388, opened 2022-03-15
LOG-1415: Fix in-file tuning that causes chunk byte limit exceeded

jcantrill:
### Description
This PR:
* fixes the issue where there is a mismatch between buffer tuning and in tail readline limit
* Introduces InTail tuning similar to Buffer tuning for fluentd

/assign @alanconway 

### Links
* https://issues.redhat.com/browse/LOG-1415


alanconway (approved):
/lgtm

PR #1402, opened 2022-03-21
LOG-2393: Fix loki output url parse and add bearer_token support

sasagarw:
### Description

This PR:
- fixes loki output url parse method.
- add support for bearer_token in secrets.
- rename `elasticsearch` -> `loki` in all loki templates.

/cc @periklis 
/assign @periklis 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-2393, https://issues.redhat.com/browse/LOG-2392

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -12,6 +12,7 @@ const (
 	ClientPrivateKey   = "tls.key"
 	TrustedCABundleKey = "ca-bundle.crt"
 	Passphrase         = "passphrase"
+	BearerTokenFileKey = "token"
 
 	// Username/Password keys, used by any output with username/password authentication.
 
--- a/internal/generator/fluentd/output/loki/bearer-token-file.go
+++ b/internal/generator/fluentd/output/loki/bearer-token-file.go
@@ -1,0 +1,18 @@
+package loki
+
+import (
+	"github.com/openshift/cluster-logging-operator/internal/generator/fluentd/output/security"
+)
+
+type BearerTokenFile security.BearerTokenFile
+
+func (bt BearerTokenFile) Name() string {
+	return "lokiBearerTokenFileTemplate"
+}
+
+func (bt BearerTokenFile) Template() string {
+	return `{{define "` + bt.Name() + `" -}}
+bearer_token_file "{{ .BearerTokenFilePath }}"
+{{- end}}
+`
+}
--- a/internal/generator/fluentd/output/loki/ca-file.go
+++ b/internal/generator/fluentd/output/loki/ca-file.go
@@ -7,7 +7,7 @@ import (
 type CAFile security.CAFile
 
 func (ca CAFile) Name() string {
-	return "elasticsearchCAFileTemplate"
+	return "lokiCAFileTemplate"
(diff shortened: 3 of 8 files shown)


periklis (commented):
/lgtm

internal/constants/constants.go
```diff
@@ -12,6 +12,7 @@ const (
 	ClientPrivateKey   = "tls.key"
 	TrustedCABundleKey = "ca-bundle.crt"
 	Passphrase         = "passphrase"
+	BearerTokenFileKey = "bearer_token_file"
```
jcantrill, line 15: Do we have precedence where we use this undscore ruby type syntax for keys instead of CamelCase?  Placing on hold pending comment from @alanconway

jcantrill (commented):
/hold

internal/constants/constants.go
alanconway: I would call it just "token". 
- it is not a file, its the token data that's in the secret.
- "token" isn't likely  ambiguous in any case involving a bearer token.
- short, single words are less likely to be mis-spelled so fewer config errors.
Otherwise LGTM

alanconway (changes requested):

periklis (approved):
/lgtm

internal/constants/constants.go
```diff
@@ -12,6 +12,7 @@ const (
 	ClientPrivateKey   = "tls.key"
 	TrustedCABundleKey = "ca-bundle.crt"
 	Passphrase         = "passphrase"
+	BearerTokenFileKey = "token"
```
alanconway, line 15: Please change to BearerTokenKey - the secret contains the actual token data, not a file name, unlike the config flie.
Not a big deal but it causes confusion sometimes.

internal/generator/fluentd/output/loki/bearer-token-file.go
```diff
@@ -0,0 +1,18 @@
+package loki
+
+import (
+	"github.com/openshift/cluster-logging-operator/internal/generator/fluentd/output/security"
+)
+
+type BearerTokenFile security.BearerTokenFile
```
alanconway, line 7: Same here: global search/replace BearerTokenFile with BearerToken.

alanconway (commented):
One more name change, then I'm LGTM.

internal/generator/fluentd/output/loki/bearer-token-file.go
sasagarw: @alanconway we just point to the file path. It doesn't contain the actual token.
The secret will look like:
`token=/var/run/secrets/kubernetes.io/serviceaccount/token`

internal/constants/constants.go
sasagarw: We just point to the file path. It doesn't contain the actual token.
The secret will look like:
`token=/var/run/secrets/kubernetes.io/serviceaccount/token`
Fluentd loki plugin accepts the file path rather than actual token

sasagarw:
/retitle LOG-2412: Fix loki output url parse and add bearer_token support

PR #1407, opened 2022-03-22
LOG-1099: Fluentd firing far too many dns requests

syedriko:
### Description
Make Elastic's DNS name just "elasticsearch" as opposed to "elasticsearch.openshift-logging.svc" in order to avoid extra DNS queries from fluentd.

/cc @alanconway 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-1909

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -44,8 +44,7 @@ const (
 	TrustedCABundleHashName    = "logging.openshift.io/hash"
 	SecretHashPrefix           = "logging.openshift.io/"
 	KibanaTrustedCAName        = "kibana-trusted-ca-bundle"
-	// internal elasticsearch FQDN to prevent to connect to the global proxy
-	ElasticsearchFQDN          = "elasticsearch.openshift-logging.svc"
+	ElasticsearchFQDN          = "elasticsearch"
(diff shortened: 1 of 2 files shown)


internal/constants/constants.go
```diff
@@ -44,8 +44,7 @@ const (
 	TrustedCABundleHashName    = "logging.openshift.io/hash"
 	SecretHashPrefix           = "logging.openshift.io/"
 	KibanaTrustedCAName        = "kibana-trusted-ca-bundle"
-	// internal elasticsearch FQDN to prevent to connect to the global proxy
```
jcantrill, line 47: Is this because when the global proxy is enabled there is no way to filter traffic via NO_PROXY? meaning we will always be writing through the global proxy

syedriko: Would "NO_PROXY=cluster.local" work?

alanconway (approved):
/lgtm

internal/constants/constants.go
jcantrill: I don't but I will defer to @alanconway here since he gave it a :lgtm

PR #1419, opened 2022-03-31
Log 2434 revert

vimalk78:
### Description
5.4 should be 1.7.4

/cc @syedriko @jcantrill 
/assign


vimalk78:
need this change in 5.4, so no need to revert the commit.
Closing the PR.

PR #1421, opened 2022-04-06
LOG-2213: Enable sending container logs to separate indices

jcantrill:
### Description
This PR:
* Enables forwarding container logs from the same pod to separate indices
* Modifies the CLF API to make behavior "opt-in"
* Consumes the changes to the viaq plugin to move all "structured" index logic from config to the plugin
* Moves the index evaluation from "INGRESS" processing to add it only for Elasticsearch outputs

cc @alanconway @vimalk78 

### Links
* https://issues.redhat.com/browse/LOG-2213


alanconway (approved):
/lgtm

PR #1425, opened 2022-04-10
fix for healthStatus and managedStatus update

pmoogi-redhat:
### Description

fix for updating healthStatus and managedStatus of CL
1. CL healthStatus and managedStatus now treated with 0 value representing 'true' status for healthy and managed status of CL and with 1 value representing 'false" status for healthy and managed status for CL
2. version is set to operator_version =  version.Version value as per the telemetry specification for CLO
3. Any early exit from reconciler with error is treated as un healthy status for CL.  Fixed this in the internal/k8handler/reconciler.go

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-2439 
- Enhancement proposal:


pmoogi-redhat (commented):
@alanconway please review

internal/k8shandler/reconciler.go
```diff
@@ -30,8 +30,15 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
 	}
 
 	if !clusterLoggingRequest.isManaged() {
+		telemetry.Data.CLInfo.Set("managedStatus", "1")
+		erru := telemetry.UpdateCLMetrics()
```
alanconway, line 34: There's no need to create uniquely named errx variables. It is more idiomatic to re-use the same 'err' variable multiple times.

internal/k8shandler/reconciler.go
```diff
@@ -42,11 +49,21 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
...
 			return fmt.Errorf("unable to create or update logstore for %q: %v", clusterLoggingRequest.Cluster.Name, err)
 		}
 
 		// Reconcile Visualization
 		if err = clusterLoggingRequest.CreateOrUpdateVisualization(); err != nil {
+			telemetry.Data.CLInfo.Set("healthStatus", "1")
+			erru := telemetry.UpdateCLMetrics()
+			if erru != nil {
+				log.V(1).Error(erru, "Error in updating clo metrics for telemetry")
+			}
```
alanconway, line 66: This pattern recurs many times. If failure to update telemetry is not a real error then perhaps encapsulate this in 
```
func telementry.UpdateCLMetricsNoErr() {
erru := telemetry.UpdateCLMetrics()
			if erru != nil {
				log.V(1).Error(erru, "Error in updating clo metrics for telemetry")
			}
}
```

PR #1426, opened 2022-04-12
LOG-2294: Fix vector internal metrics not exposed via HTTPS

Pranjal-Gupta2:
### Description

This PR:
-  Adds TLS support for vector internal metrics
- Fix `logfilesmetricexporter` not taking command line arguments
- Enable `logfilesmetricexporter` to use normalized paths (/etc/collector...)
- Adds a test to check if logfilesmetricexporter metrics are being published

/cc @jcantrill @vimalk78 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-2294


internal/generator/vector/conf_test.go
```diff
@@ -222,6 +222,11 @@ type = "prometheus_exporter"
 inputs = ["internal_metrics"]
 address = "0.0.0.0:24231"
 default_namespace = "collector"
+
+[sinks.prometheus_output.tls]
+enabled = true
+key_file = "${METRICS_KEY:-/etc/collector/metrics/tls.key}"
```
jcantrill, line 228: We can fix this and not need env vars

test/e2e/logfilesmetricexporter/metrics_test.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/bash
```
jcantrill, line 1: Can we do this as a functional test to make testing easier?

jcantrill (changes requested):
/hold
/approve 

Please make sure to squash your commits before final submit

test/e2e/logfilesmetricexporter/metrics_test.sh
Pranjal-Gupta2: There is no support for `logfilesmetricexporter` in the functional framework. We will need to determine the viability of adding/creating it for our test.

jcantrill: Given the work required, I will defer to leaving this as an e2e test

jcantrill (commented):
/hold cancel
/lgtm

PR #1431, opened 2022-04-20
LOG-2455: fix infra logs showing in app log group

cahartma:
### Description
fixes issue where infra container logs were being captured by the application log group

/assign @vimalk78 @jcantrill 

### Links
- https://issues.redhat.com/browse/LOG-2455
- related to: https://issues.redhat.com/browse/LOG-2160
- related to: https://issues.redhat.com/browse/LOG-2275

### DIFF
--- a/test/framework/functional/write.go
+++ b/test/framework/functional/write.go
@@ -9,7 +9,7 @@ import (
 	"time"
 )
 
-func (f *CollectorFunctionalFramework) WriteMessagesToApplicationLogInNamespace(msg, namespace string, numOfLogs int) error {
+func (f *CollectorFunctionalFramework) WriteMessagesToNamespace(msg, namespace string, numOfLogs int) error {
 	filename := fmt.Sprintf("%s/%s_%s_%s/%s/0.log", fluentdLogPath[applicationLog], namespace, f.Pod.Name, f.Pod.UID, constants.CollectorName)
 	return f.WriteMessagesToLog(msg, numOfLogs, filename)
 }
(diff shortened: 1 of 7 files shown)


cahartma:
I wanted to comment, that I wasn't sure of the impact of utilizing a long regex like this.   Let me know.

jcantrill (commented):
/hold
/approve

Seems like we could write a functional test to expose this issue and then verify it does not get re-introduced

test/framework/functional/write.go
```diff
@@ -9,7 +9,7 @@ import (
 	"time"
 )
 
-func (f *CollectorFunctionalFramework) WriteMessagesToApplicationLogInNamespace(msg, namespace string, numOfLogs int) error {
+func (f *CollectorFunctionalFramework) WriteMessagesToContainerLogInNamespace(msg, namespace string, numOfLogs int) error {
```
jcantrill, line 12: Nack this change. In the context of logging "application" has meaning that container does not

jcantrill (changes requested):

test/framework/functional/write.go
cahartma: I need a method to pass in a specific namespace from the infra index ("openshift-authentication-operator"), and realized later that this method did that already.  Since namespace not bound, do you have a suggestion for a common name in order to avoid duplicate methods?

cahartma: update to `WriteMessagesToNamespace`

PR #1440, opened 2022-04-27
LOG-2525: enable metadata watch to fix spliting logs

jcantrill:
### Description
This PR:
* fixes sending logs to separate indices from the same container when the pod is annotated after the collector boots
* enables watch for the metadata plugin to refresh the info
* Adds env var to restrict the watch to the pods on the node

### Links
* https://issues.redhat.com/browse/LOG-2525

### DIFF
--- a/internal/generator/fluentd/conf_test.go
+++ b/internal/generator/fluentd/conf_test.go
@@ -409,8 +409,8 @@ var _ = Describe("Testing Complete Config Generation", func() {
     @type kubernetes_metadata
     kubernetes_url 'https://kubernetes.default.svc'
     annotation_match ["^containerType\.logging\.openshift\.io\/.*$"]
+    allow_orphans false
     cache_size '1000'
-    watch 'false'
     use_journal 'nil'
     ssl_partial_chain 'true'
   </filter>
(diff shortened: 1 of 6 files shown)


internal/generator/fluentd/conf_test.go
```diff
@@ -410,7 +410,7 @@ var _ = Describe("Testing Complete Config Generation", func() {
     kubernetes_url 'https://kubernetes.default.svc'
     annotation_match ["^containerType\.logging\.openshift\.io\/.*$"]
     cache_size '1000'
-    watch 'false'
+    watch true
```
vimalk78, line 413: default value of watch is `true`
https://github.com/ViaQ/logging-fluentd/blob/main/fluentd/vendored_gem_src/fluent-plugin-kubernetes_metadata_filter/lib/fluent/plugin/filter_kubernetes_metadata.rb#L48

internal/generator/fluentd/ingress.go
```diff
@@ -197,8 +197,8 @@ const KubernetesMetadataPlugin string = `
   @type kubernetes_metadata
   kubernetes_url 'https://kubernetes.default.svc'
   annotation_match ["^containerType\.logging\.openshift\.io\/.*$"]
+  allow_orphans false
```
jcantrill, line 200: @alanconway please confirm you on board with this change.  Essentially this makes it so regress of fetch meta you will minimum get: namespace, podname, container name.  There is no way to retrieve pod meta by the more unique id that is available from orphaning.  This will lead us to resorting to cluster policy to control the "deleted" namespace issue.  I will only be able to fix it otherwise by making some change in the actual plugin, otherwise we will need to hack the tests some how

PR #1445, opened 2022-04-29
LOG-2439: fix for healthStatus and managedStatus update

pmoogi-redhat:
### Description
[LOG-2439](https://issues.redhat.com//browse/LOG-2439): fix for healthStatus and managedStatus update #1445
 issues notified by QE

CL healthStatus and managedStatus now treated as below
	UnHealthyStatus = "0"
	HealthyStatus   = "1"
	UnManagedStatus = "0"
	ManagedStatus   = "1"

version is set to operator_version =  version.Version value as per the telemetry specification for CLO
Any early exit from reconciler with error is treated as un healthy status for CL.  Fixed this in the internal/k8handler/reconciler.go

/cc @jcantrill 
/assign  @alanconway 

Links
Depending on PR(s):
Bugzilla:
Github issue:
JIRA: https://issues.redhat.com/browse/LOG-2439
Enhancement proposal:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -130,11 +130,13 @@ MANIFESTS=manifests/$(LOGGING_VERSION)
 
 # Do all code/CRD generation at once, with timestamp file to check out-of-date.
 GEN_TIMESTAMP=.target/codegen
+DEFAULT_VERSION=5.5
 generate: $(GEN_TIMESTAMP)
 $(GEN_TIMESTAMP): $(shell find apis -name '*.go')  $(OPERATOR_SDK) $(CONTROLLER_GEN) $(KUSTOMIZE) .target
 	@$(CONTROLLER_GEN) object paths="./apis/..."
 	@$(CONTROLLER_GEN) crd:crdVersions=v1 rbac:roleName=clusterlogging-operator paths="./..." output:crd:artifacts:config=config/crd/bases
 	@bash ./hack/generate-crd.sh
+	echo 'package version; var Version = "$(or $(CI_CONTAINER_VERSION),$(LOGGING_VERSION), DEFAULT_VERSION)"' > version/version.go
 	@$(MAKE) fmt
 	@touch $@
 
--- a/internal/k8shandler/reconciler.go
+++ b/internal/k8shandler/reconciler.go
@@ -30,7 +30,16 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
 	}
 
 	if !clusterLoggingRequest.isManaged() {
+		// if cluster is set to unmanaged then set managedStatus as 0
+		telemetry.Data.CLInfo.Set("managedStatus", constants.UnManagedStatus)
+		telemetry.UpdateCLMetricsNoErr()
 		return nil
+	}
+	// CL is managed by default set it as 1
+	telemetry.Data.CLInfo.Set("managedStatus", constants.ManagedStatus)
+	updateCLInfo := UpdateInfofromCL(&clusterLoggingRequest)
+	if updateCLInfo != nil {
+		log.V(1).Info("Error in updating CL Info for CL specific metrics", "updateCLInfo", updateCLInfo)
 	}
 
 	forwarder := clusterLoggingRequest.getLogForwarder()
@@ -42,11 +51,15 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
 	if clusterLoggingRequest.IncludesManagedStorage() {
 		// Reconcile Log Store
 		if err = clusterLoggingRequest.CreateOrUpdateLogStore(); err != nil {
+			telemetry.Data.CLInfo.Set("healthStatus", constants.UnHealthyStatus)
+			telemetry.UpdateCLMetricsNoErr()
 			return fmt.Errorf("unable to create or update logstore for %q: %v", clusterLoggingRequest.Cluster.Name, err)
 		}
 
 		// Reconcile Visualization
 		if err = clusterLoggingRequest.CreateOrUpdateVisualization(); err != nil {
+			telemetry.Data.CLInfo.Set("healthStatus", constants.UnHealthyStatus)
+			telemetry.UpdateCLMetricsNoErr()
 			return fmt.Errorf("unable to create or update visualization for %q: %v", clusterLoggingRequest.Cluster.Name, err)
 		}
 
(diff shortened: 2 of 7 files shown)


version/version.go
```diff
@@ -1,5 +1,5 @@
 package version
 
 var (
-	Version = "0.0.1"
+	Version = "5.5"
```
jcantrill, line 4: How will this be populated other then a manual bump?

jcantrill (commented):
Generally fine but:
* we should see if there is a way to get version elsewhere.  Maybe it exists in an env, check a running pod
* when commiting a bug fix  ALWAYS use : `LOG-xxxx:  description`  please update the commit

version/version.go
pmoogi-redhat: Makefile got a target version-update where i am doing the below  :

version-update:
	sed -i 's/"[0-9]*\.[0-9]*\.[0-9]*"/\"${LOGGING_VERSION}\"/g' version/version.go
bin/cluster-logging-operator: version-update force
	@oc -n openshift-cluster-version scale deployment/cluster-version-operator --replicas=$(REPLICAS)

jcantrill: This is not a sustainable process.  Please review a running container and see if there is maybe an environment variable from which we can retrieve this info.  There are many variables that are injected into the container at runtime

jcantrill: Example of where to find the information.  We will have to fix our CI to bake this into the `version.go`
```
 containers:
    - resources: {}
      terminationMessagePath: /dev/termination-log
      command:
        - cluster-logging-operator
      env:
        - name: OPERATOR_CONDITION_NAME
          value: cluster-logging.5.4.0-138
```

alanconway:
I suggest something like this:
```
modified   Makefile
@@ -123,8 +123,7 @@ lint-dockerfile:
 .PHONY: lint-dockerfile
 
 fmt:
-	@echo gofmt		# Show progress, real gofmt line is too long
-	find test internal controllers apis -name '*.go' | xargs gofmt -s -l -w
+	find -name '*.go' | xargs gofmt -s -l -w
 
@@ -135,6 +134,7 @@ $(GEN_TIMESTAMP): $(shell find apis -name '*.go')  $(OPERATOR_SDK) $(CONTROLLER_
 	@$(CONTROLLER_GEN) object paths="./apis/..."
 	@$(CONTROLLER_GEN) crd:crdVersions=v1 rbac:roleName=clusterlogging-operator paths="./..." output:crd:artifacts:config=config/crd/bases
 	@bash ./hack/generate-crd.sh
+	echo 'package version; var Version = "$(or $(CI_CONTAINER_VERSION),$(LOGGING_VERSION).0)"' > version/version.go
 	@$(MAKE) fmt
 	@touch $@
 
modified   version/version.go
@@ -1,5 +1,3 @@
 package version
 
-var (
-	Version = "0.0.1"
-)
+var Version = "5.5.0"
```

Makefile
```diff
@@ -135,6 +135,7 @@ $(GEN_TIMESTAMP): $(shell find apis -name '*.go')  $(OPERATOR_SDK) $(CONTROLLER_
 	@$(CONTROLLER_GEN) object paths="./apis/..."
 	@$(CONTROLLER_GEN) crd:crdVersions=v1 rbac:roleName=clusterlogging-operator paths="./..." output:crd:artifacts:config=config/crd/bases
 	@bash ./hack/generate-crd.sh
+	echo 'package version; var Version = "$(or $(CI_CONTAINER_VERSION),$(LOGGING_VERSION).0)"' > version/version.go
```
jcantrill, line 138: what happens when these variables are not available?

pmoogi-redhat: @alanconway could you please guide here. Jeff is asking about situations when CI_CONTAINER_VERSION and LOGGING_VERSION is not made available. Is that could be a situation in general in CI pipeline ? We also have env variable through which we are setting up version information here. I have made both provision in this PR.

jcantrill: @pmoogi-redhat This target AFAIK does not get run during CI. Regardless it is always good practice to provide a default otherwise this step out right fails when doing local development

pmoogi-redhat:
@alanconway @jcantrill please review all changes as i have incorporated all the previous review comments. default fallback option is also added in the Makefile.

internal/k8shandler/reconciler.go
```diff
@@ -67,37 +80,31 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
...
 		telemetry.Data.CollectorErrorCount.Inc("CollectorErrorCount")
+		telemetry.UpdateCLMetricsNoErr()
 		return fmt.Errorf("unable to create or update collection for %q: %v", clusterLoggingRequest.Cluster.Name, err)
 	}
 
 	// Reconcile metrics Dashboards
 	if err = metrics.ReconcileDashboards(clusterLoggingRequest.Client, reader); err != nil {
+		telemetry.Data.CLInfo.Set("healthStatus", constants.UnHealthyStatus)
+		telemetry.UpdateCLMetricsNoErr()
 		log.Error(err, "Unable to create or update metrics dashboards", "clusterName", clusterLoggingRequest.Cluster.Name)
```
alanconway, line 93: This doesn't return an error, will fall thru and overwrite with HealthyStatus. Suggest setting HealthyStatus at the top of this function, and it will get over-written with UnhealthyStatus if an error is encountered.

alanconway: @jcantrill Is it correct that no error is returned here? If so is this really an unhealthy condition?

internal/k8shandler/reconciler.go
```diff
@@ -67,37 +80,31 @@ func Reconcile(requestCluster *logging.ClusterLogging, requestClient client.Clie
...
-		}
-		erru := telemetry.UpdateCLMetrics()
-		if erru != nil {
-			log.V(1).Error(erru, "Error in updating clo metrics for telemetry")
-		}
-	}
-	///////
-	//if it reaches till this point without any errors than mark CL in healthy state
-	telemetry.Data.CLInfo.Set("healthStatus", "0")
+	//if there is no early exit from reconciler then new CL spec is applied successfully hence healthStatus is set to true or 1
```
alanconway, line 96: Suggest setting at the start, see previous comment.

alanconway (changes requested):
Minor change

PR #1451, opened 2022-05-03
LOG-2063: fix deploying a collector with no CLF

jcantrill:
### Description
This PR:
* fixes collector deployment when only a collector is defined and there is no CLF
* Stops deployment of either fluent or vector 
* Refactors one test to be ginkgo table driven

### Links
* https://issues.redhat.com/browse/LOG-2063


jcantrill:
cc @alanconway @vimalk78

alanconway (approved):
/lgtm

PR #1454, opened 2022-05-08
LOG-2587: Fix ES functional tests for vector

vimalk78:
### Description
This PR
- fixes functional framework to make kubernetes_logs start scraping application logs created by functional framework
- changes generator to add a flag for metrics, so that functional tests do not generate metrics collection. 
- fixes app log data model to conform to functional test app log template
- refactors functional test to separate the message templates for fluentd and vector
- enables `outputs/elasticsearch` tests to run for both fluentd and vector

/cc @jcantrill @cahartma 
/assign 

/cherry-pick 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-2587

### DIFF
--- a/test/framework/functional/vector/message_templates.go
+++ b/test/framework/functional/vector/message_templates.go
@@ -1,4 +1,4 @@
-package functional
+package vector
 
 import (
 	"time"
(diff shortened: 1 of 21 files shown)


test/helpers/types/types.go
```diff
@@ -31,7 +31,8 @@ type ContainerLog struct {
 	Hostname         string                 `json:"hostname"`
 	PipelineMetadata PipelineMetadata       `json:"pipeline_metadata"`
 	LogType          string                 `json:"log_type,omitempty"`
-	ViaqIndexName    string                 `json:"viaq_index_name"`
+	ViaqIndexName    string                 `json:"viaq_index_name,omitempty"`
+	WriteIndex       string                 `json:"write_index,omitempty"`
```
jcantrill, line 35: Why do we need "write" index?

test/helpers/types/types.go
```diff
@@ -44,12 +45,14 @@ type Docker struct {
 type Kubernetes struct {
 	Annotations       map[string]string `json:"annotations,omitempty"`
 	ContainerName     string            `json:"container_name,omitempty"`
+	ContainerID       string            `json:"container_id,omitempty"`
 	NamespaceName     string            `json:"namespace_name,omitempty"`
 	PodName           string            `json:"pod_name,omitempty"`
-	ContainerImage    string            `json:"container_image,omitempty"`
-	ContainerImageID  string            `json:"container_image_id,omitempty"`
+	PodNodeName       string            `json:"pod_node_name,omitempty"`
```
jcantrill, line 51: Why do we need pod node name?   how is this different from "host"?

test/framework/functional/vector/message_templates.go
```diff
@@ -0,0 +1,85 @@
+package vector
```
jcantrill, line 1: Why does vector require its own set of message templates if our goal is to adhere to the same Viaq data model?

test/helpers/types/types.go
vimalk78: this is the field we used for vector, as the name `viaq-index-name` didnt make much sense for vector.
Also vector ES sink does not delete the key which contains the index name, as fluentd ES plugin can do. So if we keep `viaq-index-name` , it will be added to the record also.

test/helpers/types/types.go
vimalk78: `pod_node_name` contains value from `spec.nodeName`

test/framework/functional/vector/message_templates.go
vimalk78: there are some differences in the data model, as captured in the google docs sheet.

test/helpers/types/types.go
vimalk78: @alanconway any comments?

alanconway (approved):
LGTM but I'll let @jcantrill  finish his deliberations before merging.

vimalk78:
Closing this in favor of #1466

PR #1461, opened 2022-05-10
LOG-2609: fluentd and vector are unable to ship logs to elasticsearch when cluster-wide proxy is in effect

syedriko:
### Description
fluentd and vector are unable to ship logs to elasticsearch when cluster-wide proxy is in effect. This is due to https://github.com/openshift/cluster-logging-operator/pull/1407. The fix is to add `elasticsearch,` to the `NO_PROXY` and `no_proxy` env variables in case those are set and make sure we do not try to talk to elasticsearch through the proxy.

/cc @alanconway 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-2609

### DIFF
--- a/internal/collector/collector_test.go
+++ b/internal/collector/collector_test.go
@@ -171,7 +171,7 @@ var _ = Describe("Factory#NewPodSpec", func() {
 
 				verifyEnvVar(collector, "HTTP_PROXY", httpproxy)
 				verifyEnvVar(collector, "HTTPS_PROXY", httpproxy)
-				verifyEnvVar(collector, "NO_PROXY", noproxy)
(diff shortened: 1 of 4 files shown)


cahartma:
@syedriko the only possible issue with adding "elasticsearch" to the noProxy, is if external ES instances have this name in their host.   A customer with an external ES would most likely need to further edit the proxy.  Any reason you're no longer going to change the ES endpoint back to include ".openshift-logging.svc", as mentioned last we spoke?

syedriko:
elasticsearch.openshift-logging.svc causes more DNS lookups than necessary. "elasticsearch" in NO_PROXY is not going to match like `*elasticsearch*`.

PR #1466, opened 2022-05-13
LOG-2587: Fix ES functional tests for vector, using go tags

vimalk78:
### Description
This PR

  - fixes functional framework to make `kubernetes_logs` start scraping application logs created by functional framework, enables debug logs for vector
  - adds go test tags based selection of collector for functional test framework.
    - use `go test --tag=fluentd` to set a global variable `"FrameworkType"` to "fluentd"
    - use `go test --tag=vector` to set a global variable `"FrameworkType"` to "vector"
    which can be used to create framework in the tests.
  - fixes app log data model to conform to functional test app log template
  - relaxes message template to accommodate both fluentd and vector based record creation
  - enables outputs/elasticsearch tests to run for both fluentd and vector

/cc @jcantrill @cahartma 

/assign @vimalk78 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-2587

### DIFF
--- a/internal/generator/vector/conf_test.go
+++ b/internal/generator/vector/conf_test.go
@@ -98,6 +98,9 @@ auto_partial_merge = true
 exclude_paths_glob_patterns = ["/var/log/pods/openshift-logging_collector-*/*/*.log", "/var/log/pods/openshift-logging_elasticsearch-*/*/*.log", "/var/log/pods/openshift-logging_kibana-*/*/*.log"]
 pod_annotation_fields.pod_labels = "kubernetes.labels"
 pod_annotation_fields.pod_namespace = "kubernetes.namespace_name"
+pod_annotation_fields.pod_annotations = ""
+pod_annotation_fields.pod_uid = "kubernetes.pod_id"
+pod_annotation_fields.pod_node_name = "hostname"
 
 [sources.raw_journal_logs]
 type = "journald"
@@ -153,6 +156,8 @@ source = """
   del(.stream)
   
   del(.kubernetes.pod_ips)
+  
+  ."@timestamp" = del(.timestamp)
 """
 
 [transforms.journal_logs]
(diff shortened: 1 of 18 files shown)


jcantrill (changes requested):
/hold

I think this is a good direction that would actually be better solved with gingkgo 2.0 and labels, but that being said I think it's still missing some meta to properly use tags.  The way I read the use of tags is that we would wish to annotate the test files with the desired tag so that it is compiled and runs the way we require.  I only see tags in the marker files for which collector type; that does not look to be enough. 

Another point is why do we need "if" statements to skip if we have the tags;  we should be able to "skip" those tests by marking them ready for fluentd and vector

internal/runtime/pod.go
```diff
@@ -100,7 +100,7 @@ func (builder *PodBuilder) AddContainer(name, image string) *ContainerBuilder {
 			Name:            strings.ToLower(name),
 			Image:           image,
 			Env:             []corev1.EnvVar{},
-			ImagePullPolicy: corev1.PullAlways,
+			ImagePullPolicy: corev1.PullIfNotPresent,
```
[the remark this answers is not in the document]
vimalk78: i observed test speedup once i changed this to `IfNotPresent` 
since tests run in parallel, multiple pods start pulling the same image, and it seemed to slow down the tests.

i also used a "warm up" pod which has containers for each image tests needed. 
```
  containers:    
    - name: c0    
      image: elasticsearch:7.10.1    
    - name: c1    
      image: quay.io/openshift/origin-logging-rsyslog:latest    
    - name: c2    
      image: quay.io/openshift-logging/fluentd:1.14.5    
etc
```
then tests didnt have to wait for pods to pull the images.

test/functional/fluentd.go
```diff
@@ -0,0 +1,12 @@
...
+// +build fluentd
+
+package functional
+
+import (
+	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
+)
+
+func init() {
+	FrameworkType = logging.LogCollectionTypeFluentd
```
alanconway, line 11:
```suggestion
	LogCollectionType = logging.LogCollectionTypeFluentd
```

alanconway (changes requested):
I like it, just change "FrameworkType" to "LogCollectionType" - framework is confusing in this context.

test/functional/fluentd.go
vimalk78: @alanconway Updated

internal/generator/vector/conf_test.go
```diff
@@ -153,6 +156,10 @@ source = """
   del(.stream)
   
   del(.kubernetes.pod_ips)
+  
+  ts = .timestamp
+  del(.timestamp)
+  ."@timestamp" = ts
```
cahartma, line 162: you should be able to shorten this to:  `."@timestamp = del(.timestamp) `, since del() returns the value of the deleted field.

alanconway (approved):
/lgtm

internal/generator/vector/conf_test.go
vimalk78: @cahartma updated PR

vimalk78: will need re-lgtm though

jcantrill (commented):
/approve
/hold cancel

PR #1475, opened 2022-05-17
LOG-2553: Always send infra logs to infra indices

jcantrill:
### Description
This PR:
* fixes the case where structured logging is enabled and infra source logs are sent to an application structured index
* enables watch of the api server
* disables allowing_orphan

ref: https://issues.redhat.com/browse/LOG-2553

### DIFF
--- a/internal/generator/vector/conf_test.go
+++ b/internal/generator/vector/conf_test.go
@@ -3,6 +3,7 @@ package vector
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/openshift/cluster-logging-operator/test/framework/unit"
 	"strings"
 
 	"github.com/openshift/cluster-logging-operator/internal/generator"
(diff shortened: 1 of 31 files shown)


jcantrill:
likely blocked by https://github.com/openshift/cluster-logging-operator/pull/1476

internal/generator/vector/conf_test.go
```diff
@@ -3,6 +3,7 @@ package vector
 import (
 	"encoding/json"
 	"fmt"
+	helpers2 "github.com/openshift/cluster-logging-operator/test/framework/unit"
```
cahartma, line 6: Should be used as `unit` below.

jcantrill: i don't remember doing this but I'll fix it

PR #1478, opened 2022-05-24
sync Dockerfile.in from downstream

jcantrill:
### Description
This PR:
* syncs the Dockerfile.in from downstream for consistency

cc @Red-GV @alanconway

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,34 +8,28 @@ ENV OS_GIT_MINOR=${CI_Y_VERSION}
 ENV OS_GIT_PATCH=${CI_Z_VERSION}
 ENV SOURCE_GIT_COMMIT=${CI_CLUSTER_LOGGING_OPERATOR_UPSTREAM_COMMIT}
 ENV SOURCE_GIT_URL=${CI_CLUSTER_LOGGING_OPERATOR_UPSTREAM_URL}
-ENV REMOTE_SOURCE=${REMOTE_SOURCE:-.}
+ENV REMOTE_SOURCES=${REMOTE_SOURCES:-.}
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -9,33 +9,22 @@ ENV OS_GIT_PATCH=${CI_Z_VERSION}
...
 RUN make build
 
-
-#@follow_tag(registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-cli:v4.8)
+#@follow_tag(registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-cli:v4.9)
 FROM quay.io/openshift/origin-cli:4.8 AS origincli
 
 #@follow_tag(registry.redhat.io/ubi8:latest)
 FROM registry.redhat.io/ubi8:8.5
+ENV REMOTE_SOURCE_DIR=${REMOTE_SOURCE_DIR:-/remote_source}
```
Red-GV, line 27: We actually use `REMOTE_SOURCES` instead of `REMOTE_SOURCE_DIR`.

jcantrill: This is the same as we do for MOLO and as advised in CPaaS users group. Ref:
* https://source.redhat.com/groups/public/container-build-system/container_build_system_wiki/containers_from_source_multistage_builds_in_osbs#jive_content_id_Cachito_Integration_for_gomod
* https://chat.google.com/room/AAAATiT7PR8/CqYbPOM1Ip8

This seems wrong and I'm surprised it works https://gitlab.cee.redhat.com/openshift-logging/cpaas-config/-/blob/openshift-logging-5.4-rhel-8/distgit/containers/cluster-logging-operator/Dockerfile.in#L11

jcantrill: I'll update per our conversation with team

PR #1480, opened 2022-05-25
LOG-2596: Fix the condition in [transforms.route_container_logs] in the configuration file for Vector 

vparfonov:

### Description
Fix the condition in [transforms.route_container_logs] in the configuration file for Vector. 
All container logs from namespaces which name start with  `kube-`, `openshift-` or equals `default`, `openshift` should be treated as infra logs,  otherwise are application logs.

/cc @jcantrill 
/assign @vimalk78 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-2596
- Enhancement proposal:

### DIFF
--- a/internal/generator/fluentd/conf_test.go
+++ b/internal/generator/fluentd/conf_test.go
@@ -537,7 +537,7 @@ var _ = Describe("Testing Complete Config Generation", func() {
   </filter>
   
   # Include Infrastructure logs
-  <match kubernetes.var.log.pods.openshift-*_** kubernetes.var.log.pods.default_** kubernetes.var.log.pods.kube-*_** journal.** system.var.log**>
+  <match kubernetes.var.log.pods.openshift_** kubernetes.var.log.pods.openshift-*_** kubernetes.var.log.pods.default_** kubernetes.var.log.pods.kube-*_** journal.** system.var.log**>
     @type relabel
     @label @_INFRASTRUCTURE
   </match>
@@ -628,7 +628,7 @@ var _ = Describe("Testing Complete Config Generation", func() {
 	elasticsearch_index_prefix_field 'viaq_index_name'
 	<elasticsearch_index_name>
 	  enabled 'true'
-	  tag "kubernetes.var.log.pods.openshift-*_** kubernetes.var.log.pods.default_** kubernetes.var.log.pods.kube-*_** journal.system** system.var.log**"
+	  tag "kubernetes.var.log.pods.openshift_** kubernetes.var.log.pods.openshift-*_** kubernetes.var.log.pods.default_** kubernetes.var.log.pods.kube-*_** journal.system** system.var.log**"
 	  name_type static
 	  static_index_name infra-write
 	</elasticsearch_index_name>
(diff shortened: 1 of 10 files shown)


internal/generator/vector/conf_test.go
```diff
@@ -170,8 +170,8 @@ source = """
 [transforms.route_container_logs]
 type = "route"
 inputs = ["container_logs"]
-route.app = '!((starts_with!(.kubernetes.namespace_name,"kube")) || (starts_with!(.kubernetes.namespace_name,"openshift")) || (.kubernetes.namespace_name == "default"))'
-route.infra = '(starts_with!(.kubernetes.namespace_name,"kube")) || (starts_with!(.kubernetes.namespace_name,"openshift")) || (.kubernetes.namespace_name == "default")'
+route.app = '!((starts_with!(.kubernetes.namespace_name,"kube-")) || (starts_with!(.kubernetes.namespace_name,"openshift-")) || (.kubernetes.namespace_name == "default") || (.kubernetes.namespace_name == "openshift"))'
```
jcantrill, line 173: I think we miss "kube" namespace here

jcantrill (changes requested):

internal/generator/vector/conf_test.go
vimalk78: there is no `kube` namespace described in https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/

vimalk78: perhaps we have missed `openshift` namespace in fluentd also?
```
InfraContainerTags            = "kubernetes.var.log.pods.openshift-*_** kubernetes.var.log.pods.default_** kubernetes.var.log.pods.kube-*_**"
```

https://github.com/openshift/cluster-logging-operator/blob/master/internal/generator/fluentd/source/tags.go#L7

internal/generator/fluentd/conf_test.go
```diff
@@ -537,7 +537,7 @@ var _ = Describe("Testing Complete Config Generation", func() {
   </filter>
   
   # Include Infrastructure logs
-  <match kubernetes.var.log.pods.openshift-*_** kubernetes.var.log.pods.default_** kubernetes.var.log.pods.kube-*_** journal.** system.var.log**>
+  <match kubernetes.var.log.pods.openshift kubernetes.var.log.pods.openshift-*_** kubernetes.var.log.pods.default_** kubernetes.var.log.pods.kube-*_** journal.** system.var.log**>
```
jcantrill, line 540: will this match something in openshift namespace?  it may still need `_**` in order to match

vparfonov: I don't think so, according to this [wildcards-expansions-and-other-tips](https://docs.fluentd.org/configuration/config-file#wildcards-expansions-and-other-tips), will match `openshift` workspace

vparfonov: But yes, looks like you are right, need to add `_**`.

jcantrill:
> All appears good to me, but we should be creating a separate ticket for the fluentd changes.

probably but I don't believe its worth the effort

PR #1482, opened 2022-05-27
NO-JIRA: Upgrade golangci-lint, fix issues raised.

alanconway:
Upgrade to golangci-lint 1.45.0
- required to work with Go 1.18, the default on latest Fedora. 
- no immediate need to upgrade Go version but we are ready. 
- disabled gosec complaints:
  - use of math/rand,  does not seem dangerous
  - no minimum TLS version, we should revisit this

Other changes
- let golangci-lint automatically fix any issues it can, including format
- drop separate gofmt run, faster to let golangci-lint to it
- rename golangci-lint.yaml to the default .golangci-lint.yaml
- ran make generate-bundle to get up-to-date


hack/lib/test/junit.sh
```diff
@@ -183,19 +183,20 @@ function os::test::junit::generate_report() {
 #  export JUNIT_REPORT_NUM_FAILED
 function os::test::junit::internal::generate_report() {
     local report_type="$1"
-    os::util::ensure::system_binary_exists 'junitreport'
+    source .bingo/variables.env	# For JUNITREPORT
+    os::util::ensure::system_binary_exists $(JUNITREPORT)
```
xperimental, line 187: Bash syntax is _slightly_ different from Make:

```suggestion
    os::util::ensure::system_binary_exists ${JUNITREPORT}
```

This also applies to the two changes below.

PR #1498, opened 2022-06-09
LOG-2169: Fix Kafka sasl-plaintext/sasl-ssl

vimalk78:
### Description
Fixes Kafka Sasl for vector.

Tested scenarios
 - sasl
 - sasl + ssl
 - ssl
 - mutual ssl
 - without tls

/cc @alanconway @jcantrill 
/assign @alanconway 

/cherry-pick 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-2169, https://issues.redhat.com/browse/LOG-2167

### DIFF
--- a/internal/generator/vector/output/kafka/kafka.go
+++ b/internal/generator/vector/output/kafka/kafka.go
@@ -130,22 +130,18 @@ timestamp_format = "rfc3339"
 
 func TLSConf(o logging.OutputSpec, secret *corev1.Secret) []Element {
 	conf := []Element{}
-	// FIXME ignored URL parse error.
-	u, _ := url.Parse(o.URL)
-	if !urlhelper.IsTLSScheme(u.Scheme) { // Not a TLS URL
-		// FIXME error if not TLS but has secret, probable misconfiguration.
+	if o.Secret == nil {
 		return conf
 	}
-	conf = append(conf, security.TLSConf{
-		Desc:        "TLS Config",
-		ComponentID: strings.ToLower(helpers.Replacer.Replace(o.Name)),
-	})
-	if o.Secret != nil {
+	hasTLS := false
+	u, _ := url.Parse(o.URL)
+	if urlhelper.IsTLSScheme(u.Scheme) {
 		if security.HasPassphrase(secret) {
 			pp := Passphrase{
 				PassphrasePath: security.SecretPath(o.Secret.Name, constants.Passphrase),
 			}
 			conf = append(conf, pp)
+			hasTLS = true
 		}
 		if security.HasTLSCertAndKey(secret) {
 			kc := TLSKeyCert{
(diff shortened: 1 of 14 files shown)


internal/generator/vector/output/kafka/kafka.go
```diff
@@ -128,68 +128,69 @@ timestamp_format = "rfc3339"
...
-	if o.Secret != nil {
+	hasTls := false
+	u, _ := url.Parse(o.URL)
+	if urlhelper.IsTLSScheme(u.Scheme) {
 		if security.HasPassphrase(secret) {
 			pp := Passphrase{
 				PassphrasePath: security.SecretPath(o.Secret.Name, constants.Passphrase),
 			}
 			conf = append(conf, pp)
+			hasTls = true
```
alanconway, line 144:
```suggestion
			hasTLS = true
```

internal/generator/vector/output/kafka/sasl.go
```diff
@@ -1,34 +1,29 @@
...
-func (t SaslConf) Template() string {
-	return `
-{{define "kafkaSasl" -}}
+func (t Sasl) Template() string {
+	return `{{define "vectorKafkaSasl"}}
 # {{.Desc}}
 [sinks.{{.ComponentID}}.sasl]
-{{- end}}`
+enabled = true
+username = "{{.Username}}"
```
alanconway, line 25: What happens if username/password are missing/empty? Is `username = ""` the same as no username?

alanconway (changes requested):
One concern - username, password and mechanism are all optional, it looks like this may generate bad config when they are missing - e.g. in vector configuration I don't think `username=""` is equivalent to having no username. Otherwise looks good.

internal/generator/vector/output/kafka/sasl.go
vimalk78: this configuration is mainly for `SASL/PLAIN` , which needs username/password
We could add a verification step to check username/password is always configured in secret, if `SASL/PLAIN` is configured

internal/generator/vector/output/kafka/kafka.go
```diff
@@ -58,8 +58,8 @@ func Conf(o logging.OutputSpec, inputs []string, secret *corev1.Secret, op Optio
 			Output(o, inputs, secret, op),
 			Encoding(o, op),
 		},
-		TLSConf(o, secret),
-		SASLConf(o, secret),
+		TlsConf(o, secret),
+		SaslConf(o, secret),
```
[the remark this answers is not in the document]
vimalk78: updated

internal/generator/vector/output/kafka/kafka.go
vimalk78: updated

PR #1472, opened 2022-05-17
Bump operator-sdk to v1.20 & logerr to v2

btaani:
### Description
This PR:

1. Upgrades operator-sdk to v.1.20 (according to [this guide](https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.20.0/))
2. Upgrades the `logerr` library to v2
3. Modifies the operator's source code to incorporate the new changes of `logerr`

 
### Links
- JIRA: [LOG-2520](https://issues.redhat.com/browse/LOG-2520) & [LOG-2521](https://issues.redhat.com/browse/LOG-2521)

PR #1503, opened 2022-06-10
Revert back to usage of static logger instance

xperimental:
### Description

The update to logerr changed the logging concept from a static logger instance to passing around a, possibly scoped, logger instance manually. This change was not introduced well in this codebase and caused some confusion regarding the intended usage. logerr v2 now has a static-logger implementation for compatibility, which can be used to gradually transition to the new concept instead.

This PR reverts the usage of the non-static logger back to using the static logger instead. Aside from some changes to imports this should make the logging code look identical to the one before the logerr v2 update.

Reverts parts of #1472

/cc @alanconway 
/assign @jcantrill 

### Links

- JIRA: [LOG-2520](https://issues.redhat.com/browse/LOG-2520) & [LOG-2521](https://issues.redhat.com/browse/LOG-2521)


alanconway (approved):
/lgtm

PR #1509, opened 2022-06-15
LOG-2014: Support user defined app inputs for vector

cahartma:
### Description
Verify users can forward application logs by specifying a namespace.  
Verify users can forward application logs by specifying pod labels.  
Verify comparable functionality with fluentd.   

Note:
All vector forwarding has been tested using cloudwatch.  We do not currently have a functional test for forwarding by specifying namespace since vector lacks the meta `namespace.uid`.  An upcoming release of vector will contain the additional data needed to write the test.

/assign @vimalk78 @jcantrill @syedriko 

### Links
- https://issues.redhat.com/browse/LOG-2014

### DIFF
--- a/test/functional/outputs/cloudwatch/forward_to_cloudwatch_test.go
+++ b/test/functional/outputs/cloudwatch/forward_to_cloudwatch_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


test/functional/outputs/cloudwatch/forward_to_cloudwatch_test.go
```diff
@@ -144,6 +145,43 @@ var _ = Describe("[Functional][Outputs][CloudWatch] Forward Output to CloudWatch
...
+				labelValue = "test1"
+			)
+			// Add our new label
+			framework.Labels[labelKey] = labelValue
+
+			// Generate a pod spec with a selector for labels
+			functional.NewClusterLogForwarderBuilder(framework.Forwarder).
+				FromInputWithVisitor("app-logs",
+					func(spec *logging.InputSpec) {
+						spec.Application = &logging.Application{
```
vimalk78, line 160: suggest to add a Namespace name also here. this is how Application Input is usually defined, for a specific namespace. otherwise lgtm

PR #1522, opened 2022-06-21
LOG-2751: Using related images naming scheme

Red-GV:
This PR adds the images under the `relatedImages` section. 

/cc @syedriko 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-2751

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,9 +93,9 @@ RUN_CMD?=go run
 run:
 	@ls ./bundle/manifests/logging.openshift.io_*.yaml | xargs -n1 oc apply -f
 	@mkdir -p $(CURDIR)/tmp
-	FLUENTD_IMAGE=$(IMAGE_LOGGING_FLUENTD) \
-	VECTOR_IMAGE=$(IMAGE_LOGGING_VECTOR) \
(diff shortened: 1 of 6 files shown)


vimalk78:
need change the variable name in Makefile as well?
https://github.com/openshift/cluster-logging-operator/blob/576cf90d497744bc53bdab2f2efe7aeebb068583/Makefile#L195

Red-GV:
I do not believe so, @vimalk78. Only the name of the image had changed, not the image's source.

Red-GV:
Actually, I think you're right, @vimalk78. I didn't realize exactly how the test framework worked.

bundle/manifests/logging.openshift.io_clusterloggings.yaml
```diff
@@ -502,7 +502,9 @@ spec:
                         nullable: true
                         properties:
                           diskThresholdPercent:
-                            description: The threshold percentage of ES disk usage that when reached, old indices should be deleted (e.g. 75)
+                            description: The threshold percentage of ES disk usage
```
vparfonov, line 505: @Red-GV  Looks like needless changes, probably related to your OS. (I have same with my MacOS)

PR #1526, opened 2022-06-22
NO-JIRA: Remove matchExpressions from labelSelector

alanconway:
We don't yet support matchExpressions so we should not include it in the API.
When we do support it (LOG-1126) matchExpressions can be restored.


jcantrill (commented):
/lgtm

PR #1529, opened 2022-06-24
Do not treat token as path in fluentd Loki config generator

xperimental:
### Description

Currently the fluentd and vector config generators for the Loki output treat the contents of the `token` value of the provided secret differently. The vector generator treats the value as the literal token, while the fluentd generator treats it as a path to a file containing the token.

Looking at how secrets are usually used, the approach the vector configuration generator uses seems more adequate, making it possible to give the collector a secret without having additional files in the container. It also makes the secret configuration compatible with secrets backed by a ServiceAccount which, to me, sounds like a often-used use-case.

As this changes existing behaviour it might be problematic to be backported to 5.4, but I'd like to open discussion on that topic anyway, because I would see this as an improvement of usability over the current behaviour.

### Links

- JIRA: Related to [LOG-1803](https://issues.redhat.com/browse/LOG-1803)

### DIFF
--- a/internal/generator/fluentd/output/loki/bearer-token-file.go
+++ b/internal/generator/fluentd/output/loki/bearer-token-file.go
@@ -12,7 +12,7 @@ func (bt BearerTokenFile) Name() string {
 
 func (bt BearerTokenFile) Template() string {
 	return `{{define "` + bt.Name() + `" -}}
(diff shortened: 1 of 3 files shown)


periklis (commented):
/lgtm

PR #1530, opened 2022-06-24
Fix handling of custom bearer token for vector Loki output

xperimental:
### Description

Currently the vector configuration generator for the Loki output does not provide an option for setting a custom bearer token. The only way of setting a bearer-token to authenticate to Loki is to leave the secret blank and use the token issues to the `logcollector` ServiceAccount.

This PR enables the use of bearer-tokens with custom secrets.

### Links

- JIRA: Related to [LOG-2323](https://issues.redhat.com/browse/LOG-2323)


periklis (commented):
/lgtm

PR #1541, opened 2022-06-30
LOG-1803: Integrate LokiStack as supported log store

xperimental:
### Description

This PR introduces a new `LogStoreType` for the `LokiStack` operated using [loki-operator](https://github.com/grafana/loki/tree/main/operator). Compared to the ElasticSearch log store, the integration of LokiStack is much more light-weight as the cluster-logging-operator will not manage the LokiStack itself but trusts the user to provide the configuration for the log store.

The implementation expects the LokiStack to be in the same namespace ("openshift-logging") and will only forward application and infrastructure logs to the instance. The generated configuration will work with both fluentd and vector.

### Links

- JIRA: [LOG-1803](https://issues.redhat.com/browse/LOG-1803) [LOG-2323](https://issues.redhat.com/browse/LOG-2323)


hack/cr-lokistack.yaml
```diff
@@ -0,0 +1,14 @@
...
+  name: instance
+  namespace: openshift-logging
+spec:
+  logStore:
+    type: lokistack
+    lokistack:
+      name: lokistack-dev
+  collection:
+    logs:
+      type: fluentd
```
xperimental, line 13: I wonder if we should have `vector` in our example for LokiStack. :slightly_smiling_face:

apis/logging/v1/clusterlogging_types.go
```diff
@@ -123,11 +123,16 @@ type ProxySpec struct {
 // This is the struct that will contain information pertinent to Log storage (Elasticsearch)
 type LogStoreSpec struct {
 	// The type of Log Storage to configure
+	//
+	// +kubebuilder:validation:Enum=elasticsearch;lokistack
 	Type LogStoreType `json:"type"`
```
jcantrill, line 128: I think we need some comment here about how the CLO will not request a lokistack for you like it does for ES.  The assumption here is someone else has already deployed the store.  We need to ensure we are clear that it varies from the previous option

apis/logging/v1/clusterlogging_types.go
```diff
@@ -123,11 +123,16 @@ type ProxySpec struct {
 // This is the struct that will contain information pertinent to Log storage (Elasticsearch)
```
jcantrill, line 123: Maybe drop "Elasticsearch" from this comment

jcantrill (commented):
/approve

please squash the commits

config/rbac/kustomization.yaml
```diff
@@ -4,6 +4,8 @@ resources:
 #- role_binding.yaml
 - log-collector-privileged_role.yaml
 - log_collector_privileged_binding.yaml
+- log-collector-loki-clusterrole.yaml
+- log-collector-loki-clusterrolebinding.yaml
```
periklis, line 8: IMHO these should be reconciled by one of CLO's controllers since we don't need them installed on a cluster w/o using lokistack.

periklis (changes requested):

config/rbac/kustomization.yaml
xperimental: Updated the implementation of the controller to reconcile the two resources when a "LokiStack" log store is in use. There's a caveat to this though:

There can be no owner-references from a namespaced object ("ClusterLogging") to a non-namespaced one. This means that the CLO will only remove the ClusterRole/ClusterRoleBinding when the ClusterLogging still exists but the log-store is either removed or updated to not include a LokiStack log store.

I wonder if a Role/RoleBinding would be more fitting for this authorization use-case, after all the LokiStack is a namespaced object as well, so I think it would make sense that the authorization is namespaced as well. I have not checked yet if the gateway also recognises a RoleBinding, but I will do that soon.

apis/logging/v1/clusterlogging_types.go
```diff
@@ -199,6 +213,15 @@ type ElasticsearchSpec struct {
...
 }
 
+// LokiStackStoreSpec is used to set up cluster-logging to use a LokiStack as logging storage.
+// It points to an existing LokiStack in the same cluster.
+type LokiStackStoreSpec struct {
+	// Name of the LokiStack resource.
+	//
+	// +required
+	Name string `json:"name"`
+}
```
periklis, line 223: Why not simply a k8s native LocalObjectReference instead of a custom spec? I come along LORs very often lately in other operators eg the Prometheus-Operator and it comes in handy

internal/k8shandler/logstore_lokistack.go
```diff
@@ -0,0 +1,206 @@
...
+			{
+				Kind:      "ServiceAccount",
+				Name:      "logcollector",
+				Namespace: "openshift-logging",
+			},
+		},
+	}
+}
+
+func compareLokiStackClusterRole(got, want *rbacv1.ClusterRole) bool {
```
periklis, line 158: Why not simply semantic.DeepEqual?

internal/k8shandler/logstore_lokistack.go
```diff
@@ -0,0 +1,206 @@
...
+
+		if !equal {
+			return false
+		}
+	}
+
+	return true
+}
+
+func compareLokiStackClusterRoleBinding(got, want *rbacv1.ClusterRoleBinding) bool {
```
periklis, line 184: Same here why not simply semantic.DeepEqual?

internal/k8shandler/logstore_lokistack_test.go
```diff
@@ -0,0 +1,162 @@
+package k8shandler
```
periklis, line 1: I believe using semantic.DeepEqual makes writing the following testcases obsolete

periklis (changes requested):

internal/k8shandler/logstore_lokistack_test.go
xperimental: Seems I forgot about `equality.Semantic` yesterday, but I think I would keep the tests anyway to make it clear what we are actually comparing.

xperimental:
/test e2e

The last run timed out after 4h but shows no other error message.

PR #1536, opened 2022-06-28
Default to collector service-account for fluentd Loki output

xperimental:
### Description

There's currently a difference in behavior between the configuration generators for fluentd and vector when generating configuration for a Loki output: When generating the configuration for a Loki output on vector, the cluster-logging-operator will automatically fall back to the credentials in the `logcollector` ServiceAccount when no user-provided secret is available while the fluentd generator will not use credentials at all.

This PR changes the behavior of the fluentd configuration generator to match the one of the vector config generator, so that for both collectors it is possible to use a Loki output and authenticate using the `logcollector` ServiceAccount.

### Links

- JIRA: Related to [LOG-1803](https://issues.redhat.com/browse/LOG-1803)

PR #1542, opened 2022-06-30
Temporarily skip insecure fluentd forwarder test

xperimental:
### Description

A [previous PR](https://github.com/openshift/cluster-logging-operator/pull/1536) broke this e2e test by automatically injecting a default secret. While it is possible to inject the default secret only for specific output types to keep the behavior of `fluentdForwarder` as it currently is, during the discussion about the workaround we came to the conclusion that it would be worth investing a bit of time making the "disable verify" flag an explicit configuration option.

This PR only disables the offending e2e test without actually introducing the new option just to unblock the CI for the other PRs as quickly as possible.

/assign @jcantrill


periklis (commented):
/approve
/lgtm

PR #1544, opened 2022-06-30
LOG-2667: Install logging UI automatically with logging operators

alanconway:
Reconciler to create or update the logging view plugin and related objects.

Note: WIP, not yet integrated into the CLO's reconcile loop. PR up for review.


internal/consoleplugin/config.go
```diff
@@ -0,0 +1,223 @@
...
+		replicas:    1,
+	}
+}
+
+func (cf *Config) secret() string              { return cf.Name + "-secret" }
+func (cf *Config) configMapName() string       { return cf.Name + "-conf" }
+func (cf *Config) selector() map[string]string { return map[string]string{"app": cf.Name} }
+
+const (
+	labelApp             = "app"
```
jcantrill, line 49: These are generally labels we should add to all components. Please move to internal/constants.

internal/consoleplugin/config.go
```diff
@@ -0,0 +1,223 @@
...
+	if l == nil {
+		l = map[string]string{}
+	}
+	l[labelApp] = cf.Name
+	l[labelAppK8sName] = cf.Name
+	l[labelAppK8sCreatedBy] = cf.CreatedBy
+	o.SetLabels(l)
+}
+
+func (cf *Config) configMap() *corev1.ConfigMap {
```
jcantrill, line 66: Please use the utility functions in our runtime package in lieu of writing a new one

internal/consoleplugin/config.go
```diff
@@ -0,0 +1,223 @@
...
+			Type:            corev1.ServiceType("ClusterIP"),
+			SessionAffinity: corev1.ServiceAffinity("None"),
+		},
+	}
+	cf.setup(o)
+	return o
+}
+
+func (cf *Config) deployment() *appv1.Deployment {
+	o := &appv1.Deployment{
```
jcantrill, line 120: Same here. We have utilities and builders to do this

jcantrill, line 120: Console plug-ins really require a server instead of providing config that gets loaded by the console?

internal/consoleplugin/console_test.go.bad
```diff
@@ -0,0 +1,42 @@
+package console_test
```
jcantrill, line 1: Please fix me or delete me

internal/consoleplugin/reconciler.go
```diff
@@ -0,0 +1,170 @@
...
+}
+
+func (r *Reconciler) reconcileConsolePlugin(ctx context.Context) error {
+	want := r.consolePlugin
+	got := &consolev1alpha1.ConsolePlugin{}
+	return r.reconcileObject(ctx, want, got, func() (bool, error) {
+		if got.Labels == nil || got.Labels[labelAppK8sCreatedBy] != r.createdBy {
+			// ConsolePlugin exists but was not created by us. Leave it alone.
+			gvk, _ := apiutil.GVKForObject(want, r.c.Scheme())
+			rm, _ := r.c.RESTMapper().RESTMapping(gvk.GroupKind(), gvk.Version)
```
jcantrill, line 93: These seems like the wrong client to do these. It was my understanding there are higher level abstractions then RESTMapper. See anywhere else we create objects

internal/consoleplugin/reconciler.go
```diff
@@ -0,0 +1,170 @@
...
+			return false, errors.NewAlreadyExists(rm.Resource.GroupResource(), want.Name)
+		}
+		if reflect.DeepEqual(want.Spec, got.Spec) {
```
jcantrill, line 96: I don't think it is enough to only compare spec. What about the labels which we want to control

internal/consoleplugin/reconciler.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+		return true, nil
+	})
+}
+
+func (r *Reconciler) reconcileConfigMap(ctx context.Context) error {
+	want := r.configMap
+	got := &corev1.ConfigMap{}
+	return r.reconcileObject(ctx, want, got, func() (bool, error) {
+		return !reflect.DeepEqual(want.Data, got.Data), nil
```
jcantrill, line 109: Same question here about the labels

internal/consoleplugin/reconciler.go
```diff
@@ -0,0 +1,170 @@
...
+	return r.reconcileObject(ctx, want, got, func() (bool, error) {
+		return !reflect.DeepEqual(want.Spec, got.Spec), nil
+	})
+}
+
+func (r *Reconciler) reconcileService(ctx context.Context) error {
+	want := r.service
+	got := &corev1.Service{}
+	return r.reconcileObject(ctx, want, got, func() (bool, error) {
+		if reflect.DeepEqual(want.Spec.Ports, got.Spec.Ports) &&
```
jcantrill, line 125: We likely already have a service reconciliation that does this

jcantrill (changes requested):
There looks to be a lot of work to fully realize this integration

internal/consoleplugin/config.go
jgbernalp: Hi @jcantrill, yes the dynamic console plugins require a server and some configuration on the console to enable them. https://github.com/openshift/enhancements/blob/master/enhancements/console/dynamic-plugins.md#delivering-plugins

internal/cmd/logging-view/main.go
```diff
@@ -0,0 +1,38 @@
+package main
```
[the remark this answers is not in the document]
alanconway: It does, the binary is just for manual testing. I can drop it since I have automated tests, but I thought it might be handy.

internal/consoleplugin/config.go
alanconway: This is not a generic utility function, it populates the configmap with data required for the console. In this case it seems simpler to just create a corev1.ConfigMap{...} with the necessary data than to use the more indirect builders.

internal/consoleplugin/console_test.go.bad
alanconway: Oops.

internal/consoleplugin/reconciler.go
alanconway: I've dropped this, it was just to construct an "AlreadyExists" error, which is surprisingly hard. I'm not returning that error anymore, since having an already existing console plugin isn't an error, it just means some other controller set it up and there's nothing for us to do.

internal/consoleplugin/reconciler.go
alanconway: where would I look for that?

alanconway:
@jcantrill thanks for comments, all the ones I haven't resolved I will fix in the next pass.

internal/consoleplugin/config.go
alanconway: Fixed in next push.

internal/consoleplugin/config.go
```diff
@@ -0,0 +1,223 @@
...
+}
+
+// NewConfig returns a config with default settings.
+// Modify exported fields to change settings.
+func NewConfig(namespace, createdBy string) *Config {
+	return &Config{
+		CreatedBy: createdBy,
+		Name:      "logging-view-plugin",
+		Namespace: namespace,
+		Image:     "quay.io/gbernal/logging-view-plugin:latest", // FIXME
```
jcantrill, line 34: There is a utility for fetching this.  search `RELATED_` as this also requires manifest changes

internal/consoleplugin/config.go
jcantrill: This is the function of which I speak https://github.com/openshift/cluster-logging-operator/blob/master/internal/k8shandler/configmap.go#L11

internal/consoleplugin/reconciler.go
alanconway: Fixed next push

internal/consoleplugin/reconciler.go
alanconway: Fixed next push

internal/consoleplugin/config.go
```diff
@@ -0,0 +1,223 @@
...
+						root                /usr/share/nginx/html;
+					}
+				}
+				`, cf.nginxPort)},
+	}
+	cf.setup(o)
+	return o
+}
+
+func (cf *Config) service() *corev1.Service {
```
jcantrill, line 94: https://github.com/openshift/cluster-logging-operator/blob/master/internal/factory/service.go

internal/consoleplugin/config.go
jcantrill: https://github.com/openshift/cluster-logging-operator/blob/master/internal/k8shandler/deployment.go#L15

internal/consoleplugin/reconciler.go
jcantrill: here but doesnt look generic.  https://github.com/openshift/cluster-logging-operator/blob/master/internal/k8shandler/collection.go#L197  I know EO has made many of these reconcilers generic

alanconway: Can't find it, e.g. the collector service is reconciled directly:  https://github.com/alanconway/cluster-logging-operator/blob/logging-view-plugin/internal/k8shandler/collection.go#L197

test/client/client.go
```diff
@@ -253,6 +254,9 @@ func (c *Client) Host() string { return c.cfg.Host }
 // ControllerRuntimeClient returns the underlying controller runtime Client
 func (c *Client) ControllerRuntimeClient() crclient.Client { return c.c }
 
+// HTTPClient returns a plain http.Client with same config as this client.
+func (c *Client) HTTPClient() (*http.Client, error) { return rest.HTTPClientFor(c.cfg) }
```
jcantrill, line 258: I assume we no longer need this?

internal/consoleplugin/config.go
jcantrill: Can we correct it "WHEN" and "IF" it moves and reduce the code duplication in the code base?

test/client/client.go
alanconway: You're right - I can construct that client in the e2e test without messing with test.Client(). Fixed next push.

alanconway:
@jcantrill I think this covers all your issues except for using builders/runtime helpers for creating objects. The e2e test is failing but I'll sort that tomorrow.
I'm leaving the object creation as is because: 
- we will probably want to move this code to the logging console repo when the dust settles so all operatos can use it. it'll be easier to do that if we minimise deps on CLO helpers.
- object construction in this case is very simple (just a straight struct literal) so there's limited gain from using the helpers.
What do you think?

internal/consoleplugin/config.go
alanconway: See latest comments..

jcantrill:
> @jcantrill I think this covers all your issues except for using builders/runtime helpers for creating objects. The e2e test is failing but I'll sort that tomorrow. I'm leaving the object creation as is because:
> 
> * we will probably want to move this code to the logging console repo when the dust settles so all operatos can use it. it'll be easier to do that if we minimise deps on CLO helpers.

Who are "all the operators" in this use case?  Is the intent to provide a console plugin for a generic deployment of Loki?  Isn't the plugin highly coupled to the "logging" streams (e.g. application, infra).  I say this because console integration seem to be only valid for the "logging" usecase of loki store and no other.  

> * object construction in this case is very simple (just a straight struct literal) so there's limited gain from using the helpers.
>   What do you think?

I'm thinking more of the maintenance usecase where we need to mass change a resource.  My preference is to minimize code duplication regardless of the simplicity but its not pressing.  I fear though this code will not actually move and we will be left with more tech debt.

alanconway:
@jcantrill done - I accept your arguments, used the helpers. It's not a big deal to change back if/when we need to. Also did consistent env.vars & manifest changes, a better e2e test, and a fakeclient unit test. LGTM?

internal/consoleplugin/config.go
```diff
@@ -0,0 +1,205 @@
...
+
+// NewConfig returns a config with default settings.
+// Modify exported fields to change settings.
+func NewConfig(namespace, createdBy string) *Config {
+	return &Config{
+		CreatedBy: createdBy,
+		Name:      "logging-view-plugin",
+		Namespace: namespace,
+		Image:     utils.GetComponentImage(constants.ConsolePluginName),
+		LokiName:  "lokistack-dev-query-frontend-http",
```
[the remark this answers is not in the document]
alanconway: Fixed - taking the name from the default output URL.

apis/logging/v1/groupversion_info.go
```diff
@@ -35,7 +35,7 @@ var (
 	AddToScheme = SchemeBuilder.AddToScheme
 )
 
-// +kubebuilder:rbac:groups=console.openshift.io,resources=consolelinks;consoleexternalloglinks,verbs=get;create;update;delete
+// +kubebuilder:rbac:groups=console.openshift.io,resources=consolelinks;consoleexternalloglinks;consoleplugins,verbs=get;create;update;delete
```
xperimental, line 38: For me there's an error when using the ConsolePlugin as an OwnerRef:

```
cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on
```

Adding this to the role has fixed this issue:

```suggestion
// +kubebuilder:rbac:groups=console.openshift.io,resources=consolelinks;consoleexternalloglinks;consoleplugins,verbs=get;create;update;delete
// +kubebuilder:rbac:groups=console.openshift.io,resources=consoleplugins/finalizers,verbs=update
```

internal/consoleplugin/config.go
```diff
@@ -0,0 +1,205 @@
...
+					include            /etc/nginx/mime.types;
+					default_type       application/octet-stream;
+					keepalive_timeout  65;
+					server {
+						listen              %v ssl;
+						ssl_certificate     /var/serving-cert/tls.crt;
+						ssl_certificate_key /var/serving-cert/tls.key;
+						root                /usr/share/nginx/html;
+					}
+				}
```
xperimental, line 81: Does it make sense to have this as a "file embed" instead of a string constant?

internal/k8shandler/logstore_lokistack.go
```diff
@@ -28,6 +37,35 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateLokiStackLogStore() e
...
+			}
+			u, err := url.Parse(o.URL)
+			if err != nil {
+				return "", err
+			}
+			return u.Host, nil
+		}
+	}
+	return "", errors.New("no default output")
+}
```
xperimental, line 54: I think this could be simplified by generating the hostname based on the LokiStack name and ClusterLogging namespace, similar to the code in [`forwarding.go`](https://github.com/openshift/cluster-logging-operator/blob/b09b736e639b24cf1e28b7391111726e7745e46c/internal/k8shandler/forwarding.go#L384):

```suggestion
func findDefaultLokiHost(clusterLogging *logging.ClusterLogging) (string, error) {
	logStore := clusterLogging.Spec.LogStore
	if logStore == nil || logStore.Type != logging.LogStoreTypeLokiStack {
		return "", fmt.Errorf("logstore does not use LokiStack: %v", logStore)
	}

	return fmt.Sprintf("%s-query-frontend-http.%s.svc", logStore.LokiStack.Name, clusterLogging.Namespace), nil
}
```

xperimental (commented):
Not done with the review yet, had no success with the plugin showing up in my console so far.

Makefile
```diff
@@ -24,6 +24,8 @@ export NAMESPACE?=openshift-logging
 IMAGE_LOGGING_FLUENTD?=quay.io/openshift-logging/fluentd:1.14.6
 IMAGE_LOGGING_VECTOR?=quay.io/openshift-logging/vector:0.21-rh
 IMAGE_LOGFILEMETRICEXPORTER?=quay.io/openshift-logging/log-file-metric-exporter:1.1
+IMAGE_LOGGING_CONSOLE_PLUGIN?=quay.io/gbernal/logging-view-plugin:latest
```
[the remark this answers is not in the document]
alanconway: done

apis/logging/v1/groupversion_info.go
alanconway: Fixed

internal/consoleplugin/config.go
alanconway: ??

internal/k8shandler/logstore_lokistack.go
alanconway: Have a look at the latest.

bundle/manifests/clusterlogging.clusterserviceversion.yaml
```diff
@@ -414,6 +415,8 @@ spec:
                   value: quay.io/openshift-logging/fluentd:1.14.6
                 - name: RELATED_IMAGE_LOG_FILE_METRIC_EXPORTER
                   value: quay.io/openshift-logging/log-file-metric-exporter:latest
+                - name: RELATED_IMAGE_LOGGING_CONSOLE_PLUGIN
+                  value: quay.io/openshift-logging/logging-view-plugin:latest
```
jcantrill, line 419: Do we have CI building this image that should be configured in CI for this repo?

internal/k8shandler/forwarding.go
```diff
@@ -313,6 +314,23 @@ func (clusterRequest *ClusterLoggingRequest) verifyInputs(spec *logging.ClusterL
...
+	}
+	return constants.LokiStackName
+}
+
+// DefaultLokiStackURL returns the URL of the default lokistack log store.
+// The tenant and request path will be added.
+func (clusterRequest *ClusterLoggingRequest) DefaultLokiStackURL(tenantPath string) string {
+	name := clusterRequest.LogStoreLokiStackName()
+	path := path.Join("/api/logs/v1", tenantPath)
+	return fmt.Sprintf("https://%s-gateway-http.%s.svc:8080%s", name, constants.OpenshiftNS, path)
```
jcantrill, line 331: Are we certain the default stack has to be in openshift-logging ?  @xperimental @periklis

internal/consoleplugin/config.go
alanconway: Maybe, but it's pretty short and this is simple. A separate file is one more thing to keep track of.

internal/k8shandler/forwarding.go
alanconway: If not then we need to add a Namespace fileld to LogStore.LokiStack. @xperimental ?

config/rbac/role.yaml
```diff
@@ -58,6 +58,7 @@ rules:
   resources:
   - consoleexternalloglinks
   - consolelinks
+  - consoleplugins
```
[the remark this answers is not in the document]
alanconway: I'm a bit confused about why we have those two role files now - added the missing resources to cluster-logging-operator_clusterrole.yaml just in case.

bundle/manifests/clusterlogging.clusterserviceversion.yaml
alanconway: I believe so @periklis ? The upstream quay.io image is one I pushed manually for now, @jgbernalp is going to automate it.

internal/k8shandler/forwarding.go
periklis: There is no mandatory check except docs, but since our Lokistack tenant mode is called out as `openshift-logging` I want to keep our stack for in cluster operations on this namespace only, else folks won't get any logs

jcantrill: @periklis does this mean this implementation is acceptable?

bundle/manifests/clusterlogging.clusterserviceversion.yaml
jcantrill: https://github.com/openshift/release/pull/30506

internal/k8shandler/forwarding.go
xperimental: I think this is the last thread still open on the "old PR": The namespace field has been removed from the LokiStack log-store configuration on purpose. It operates on the assumption that the LokiStack used for logging is in the same namespace as the `ClusterLogging` resource (which will be `openshift-logging` in all cases that I know of).

In general, it is not a big change to add a `Namespace` field to the resource, I just don't know if this is necessary.

alanconway:
/close
replaced by #1578

PR #1546, opened 2022-07-04
LOG-2784: checking is symbols valid 'utf-8', if not replace to the valid one

vparfonov:

### Description
Checks and replace not valid UTF-8 symbols 
```
<filter pattern>
   @type record_modifier
   # will check is symbol's valid 'utf-8', if not replace to the valid one
   # e.g. japanese 'こんにちは' and ukrainian 'привіт' are valid 'utf-8' string, so will be store without modification
   # but BINARY will be replaced with '?'
   char_encoding utf-8:utf-8
</filter>
```

/cc @jcantrill 
/assign 

/cherry-pick release-5.4 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-2784
- Enhancement proposal:

### DIFF
--- a/internal/generator/fluentd/fluent_conf_test.go
+++ b/internal/generator/fluentd/fluent_conf_test.go
@@ -707,7 +707,7 @@ var _ = Describe("Generating fluentd config", func() {
   #remove structured field if present
   <filter **>
 	@type record_modifier
-    char_encoding ascii-8bit:utf-8
+    char_encoding utf-8:utf-8
 	remove_keys structured
   </filter>
   
@@ -836,7 +836,7 @@ var _ = Describe("Generating fluentd config", func() {
   #remove structured field if present
   <filter **>
 	@type record_modifier
-    char_encoding ascii-8bit:utf-8
+    char_encoding utf-8:utf-8
 	remove_keys structured
   </filter>
   
@@ -1510,7 +1510,7 @@ var _ = Describe("Generating fluentd config", func() {
   #remove structured field if present
   <filter **>
 	@type record_modifier
-    char_encoding ascii-8bit:utf-8
+    char_encoding utf-8:utf-8
 	remove_keys structured
   </filter>
   
@@ -1639,7 +1639,7 @@ var _ = Describe("Generating fluentd config", func() {
   #remove structured field if present
   <filter **>
 	@type record_modifier
-    char_encoding ascii-8bit:utf-8
+    char_encoding utf-8:utf-8
 	remove_keys structured
   </filter>
   
@@ -2317,7 +2317,7 @@ var _ = Describe("Generating fluentd config", func() {
   #remove structured field if present
   <filter **>
 	@type record_modifier
-    char_encoding ascii-8bit:utf-8
+    char_encoding utf-8:utf-8
 	remove_keys structured
   </filter>
   
@@ -2446,7 +2446,7 @@ var _ = Describe("Generating fluentd config", func() {
   #remove structured field if present
   <filter **>
 	@type record_modifier
-    char_encoding ascii-8bit:utf-8
+    char_encoding utf-8:utf-8
 	remove_keys structured
   </filter>
   
@@ -3622,7 +3622,7 @@ var _ = Describe("Generating fluentd config", func() {
   #remove structured field if present
   <filter **>
 	@type record_modifier
-    char_encoding ascii-8bit:utf-8
+    char_encoding utf-8:utf-8
 	remove_keys structured
   </filter>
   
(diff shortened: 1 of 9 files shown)


internal/generator/fluentd/fluent_conf_test.go
```diff
@@ -707,7 +707,7 @@ var _ = Describe("Generating fluentd config", func() {
   #remove structured field if present
   <filter **>
 	@type record_modifier
-    char_encoding ascii-8bit:utf-8
+    char_encoding utf-8:utf-8
```
jcantrill, line 710: ascii-8bit is fluents default encoding which from my understanding is just "binary'.  The reason we introduced this encoding change was to address bad characters being shipped to Elasticsearch where the chars were invalid.  I don't believe this conversion is a no-op.  You could test by removing this char encoding all-to-gether

jcantrill (commented):
/hold

internal/generator/fluentd/fluent_conf_test.go
vparfonov: I tested, and it works well, at least for my test case. 
You can take a look at this method https://github.com/repeatedly/fluent-plugin-record-modifier/blob/master/lib/fluent/plugin/filter_record_modifier.rb#L145. 
If i understand correct it will check is it valid utf8, if not replace.

alanconway (changes requested):
Needs unit tests to verify:
- valid utf8 is unchanged (an example that was broken before)
- utf8 with invalid chars gets correct replacements

alanconway:
/hold
needs test.

alanconway:
I think this approach is correct, but I'd still like to have a test checked in for posterity. Here's my experiment to confirm:
```
# The problem: ASCII-8BIT to UTF8
irb(main):159:0> "こんにちは".encode(Encoding::UTF_8, Encoding::ASCII_8BIT, :invalid => :replace, :undef => :replace) 
=> "���������������"

# The fix: utf8->utf8, leaves valid UTF8 alone:
irb(main):162:0> "こんにちは".encode(Encoding::UTF_8, Encoding::UTF_8, :invalid => :replace, :undef => :replace)
=> "こんにちは"

# The fix: utf8->utf8, replaces invalid characters correctly:
irb(main):163:0> "こんに\xffちは".encode(Encoding::UTF_8, Encoding::UTF_8, :invalid => :replace, :undef => :replace)
=> "こんに�ちは"
```
```

File: test/framework/functional/write.go [hunk 2]
```diff
@@ -121,3 +121,17 @@ func (f *CollectorFunctionalFramework) WriteMessagesToLog(msg string, numOfLogs
 	log.V(3).Info("WriteMessagesToLog", "namespace", f.Pod.Namespace, "result", result, "err", err)
 	return err
 }
+
+// WriteMessagesWithNotUTF8SymbolsToLog write 12 symbols in ISO-8859-1 encoding
+func (f *CollectorFunctionalFramework) WriteMessagesWithNotUTF8SymbolsToLog() error {
+	filename := fmt.Sprintf("%s/%s_%s_%s/%s/0.log", fluentdLogPath[applicationLog], f.Pod.Namespace, f.Pod.Name,
+		f.Pod.UID, constants.CollectorName)
+	logPath := filepath.Dir(filename)
+	cmd := fmt.Sprintf("mkdir -p %s; echo \"$(ruby -e 'bad_chars = Array.new; bad_chars = [0xC0, 0xC1, 0xF5, "+
```
jcantrill, line 130: There should be no need to initialize the array in this way: 
``` 
$ irb
2.7.2 :002 > bad_chars = [0xC0, 0xC1]
 => [192, 193] 

```

jcantrill (approved):
one nit

test/framework/functional/write.go
vparfonov: Thanks, will fix. Don't understand why `loki/audit_logs_test.go` failed, something strange

internal/generator/fluentd/fluent_conf_test.go
alanconway: +1 the goal of the utf-8:utf-8 re-encoding is not to actually change the encoding, but to force the collector to decode the input as utf-8, so it can replace invalid sequences. Then its re-encoded as utf-8 again, but now with invalid sequences replaced.

alanconway (approved):
/lgtm

test/framework/functional/factory.go
```diff
@@ -44,7 +44,7 @@ func NewCRIOLogMessage(timestamp, message string, partial bool) string {
 }
 
 // CRIOTime returns the CRIO string format of time t.
-func CRIOTime(t time.Time) string { return time.Now().UTC().Format(time.RFC3339Nano) }
+func CRIOTime(t time.Time) string { return t.UTC().Format(time.RFC3339Nano) }
```
alanconway, line 47: Good catch.

alanconway (approved):
/lgtm

vparfonov:
cherry pick manually https://github.com/openshift/cluster-logging-operator/pull/1571

PR #1548, opened 2022-07-04
LOG-2285: Add insecureSkipVerify option to output spec

xperimental:
### Description

This PR implements a new global setting on OutputSpec that can be used to control verification of certificates for TLS client connections. This is for enabling easier testing/debugging scenarios.

/assign @alanconway 

### Links

- JIRA: [LOG-2285](https://issues.redhat.com/browse/LOG-2285)


internal/generator/fluentd/output/fluentdforward/fluentdforward.go
```diff
@@ -94,7 +94,7 @@ func SecurityConfig(o logging.OutputSpec, secret *corev1.Secret) []generator.Ele
 		conf = []generator.Element{
 			TLS(true),
 		}
-		if secret == nil {
+		if o.TLS != nil && o.TLS.InsecureSkipVerify {
 			conf = append(conf, TLS(false))
 		}
 	}
```
alanconway: SkipVerify does not mean turn off TLS, it means add the required configuration to enable TLS but skip cert verify. Also the old code looks wrong anyway - TLS on/off depends on the schema (http vs. https) not on the presence or absence of a secret. I think it should be:
```
	if url.IsTLSScheme(u.Schema) {
		conf = append(conf, TLS(true))
		conf = append(conf, TLSSkipVerify(o.TLS != nil && o.TLS.InsecureSkipVerify))
	} else {
		conf = append(conf, TLS(false))
	}
```

test/e2e/logforwarding/fluent/fluent_secure_test.go
```diff
@@ -103,13 +98,18 @@ var _ = Describe("[ClusterLogForwarder]", func() {
 		clf.Spec.Pipelines = []loggingv1.PipelineSpec{{InputRefs: []string{loggingv1.InputNameApplication}}}
 		for _, s := range f.Receiver.Sources {
 			secret := &loggingv1.OutputSecretSpec{Name: s.Name}
+			var tls *loggingv1.OutputTLSSpec
 			if s.Name == "no-auth" {
 				secret = nil
+				tls = &loggingv1.OutputTLSSpec{
+					InsecureSkipVerify: true,
+				}
```
alanconway, line 106: The no-auth case doesn't need skipverify, there's nothing wrong with the cert. I think the the problem was that the absence of a Secret was disabling TLS, which was incorrect. 
You'll need an additional mutual-auth case where the server has a bad cert to validate the skipverify flag works.

alanconway (changes requested):
Comments in line.

internal/generator/fluentd/output/fluentdforward/fluentdforward.go
xperimental: The version you commented on is outdated by now. `TLS(false)` was a bit unclear in what it was actually doing. I assumed that it turns off TLS at first as well (and it does for other output implementations). In this case the template for `TLS(false)` just generated the `insecure_mode true` flag, that's why both `TLS(true)` and `TLS(false)` were added to the configuration.

I have refactored the `TLS` type to be less ambiguous in the current state of the PR.

test/e2e/logforwarding/fluent/fluent_secure_test.go
xperimental: The code never disabled TLS, the `TLS(false)` option was just a bit misleading.

I think I had pushed the last change just when you were doing the review. It seems one comment is on the old version and one on the new state of the PR. Maybe this caused some confusion?

apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -183,10 +186,6 @@ type OutputSpec struct {
 	//   `passphrase`: Passphrase to decode an encoded TLS private key. Requires tls.key.
 	//   `ca-bundle.crt`: Custom CA to validate certificates.
 	//
-	//   `tls.insecure`: If this key is present (with any value) accept server certificates that are self-signed or have an incorrect hostname.
```
jcantrill, line 186: @alanconway do we need to continue to honor this arrangement for backwards compatibility?

alanconway:
@xperimental  I think we should remove the backwards compat. Its use will not be common yes, only occurs in dev/test environments, and will be easily spotted by "invalid cert" messages in collector log, so we can quickly route users to the new method.

apis/logging/v1/cluster_log_forwarder_types.go
xperimental: @jcantrill Alan has answered this question in a separate comment. Okay to un-hold?

PR #1551, opened 2022-07-07
[release-5.4] LOG-2664: suppress type_name deprecation warnings

vparfonov:

### Description
Suppress high number of type deprecation logs, like:
`[types removal] Specifying types in bulk requests is deprecated.`

More details here: https://github.com/uken/fluent-plugin-elasticsearch/issues/766

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:

### DIFF
--- a/internal/generator/fluentd/conf_test.go
+++ b/internal/generator/fluentd/conf_test.go
@@ -685,6 +685,8 @@ var _ = Describe("Testing Complete Config Generation", func() {
     type_name _doc
     http_backend typhoeus
     write_operation create
+    # https://github.com/uken/fluent-plugin-elasticsearch#suppress_type_name
+    suppress_type_name 'true'
     reload_connections 'true'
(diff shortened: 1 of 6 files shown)


Moscagus:
Because **verify_es_version_at_startup is false** and by default **default_elasticsearch_version is 5**, not enough with just setting **suppress_type_name 'true'**

https://github.com/uken/fluent-plugin-elasticsearch/blob/v5.2.1/lib/fluent/plugin/out_elasticsearch.rb#L336

the best solution is:
**verify_es_version_at_startup false** 
**default_elasticsearch_version [6,7,8]** --> **adding this in ClusterLogForwarder kind**
**suppress_type_name true**

At the moment I achieve this by putting in **unmanaged mode the ClusterLogging kind** and **changing Collector ConfigMap**.

PR #1557, opened 2022-07-12
LOG-2767: sts cloudwatch secret should accept format from credrequest

cahartma:
### Description
When using sts to authenticate with cloudwatch, we should additionally accept the secret format created from a credential request.   We will  accept `credentials` as a valid secret key.    The original story and feature implementation only accepted the `role_arn` as a valid key.  (must also contain a correctly formatted valid role arn)

/assign @jcantrill @vimalk78 
/cc @syedriko 

### Links
- https://issues.redhat.com/browse/LOG-2767


jcantrill (commented):
/approve

syedriko (approved):

PR #1558, opened 2022-07-12
LOG-2662: Losing audit logs may happen when using Vector

Clee2691:
#### Summary:
- Removed `ignore_older_secs` from audit log [source] configuration for vector. 

#### Details

This PR addresses an issue with audit logs not being forwarded when using Vector. The problem stemmed from a configuration parameter for Vector called `ignore_older_secs` which "Ignores files with a data modification date older than the specified number of seconds." 

Originally this was set to 600 seconds. Removing the parameter from the audit log configurations should resolve this issue.

/cc @cahartma @jcantrill 

### Links

- [LOG-2662](https://issues.redhat.com/browse/LOG-2662)

### DIFF
--- a/internal/generator/vector/conf_test.go
+++ b/internal/generator/vector/conf_test.go
@@ -106,28 +106,24 @@ journal_directory = "/var/log/journal"
 # Logs from host audit
 [sources.raw_host_audit_logs]
 type = "file"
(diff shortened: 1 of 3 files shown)


cahartma (commented):
/lgtm

PR #1561, opened 2022-07-14
LOG-2778: Create secret for logcollector serviceaccount manually

xperimental:
### Description

Starting with Kubernetes 1.24 (OCP 4.11), ServiceAccounts will no longer automatically have an associated secret containing a token. The cluster-logging-operator currently looks for a secret containing a token associated with the `logcollector` ServiceAccount it creates for the collector DaemonSet. This secret is used for "default credentials" for the log forwarders. Failing to find the secret results in some forwarder configurations no longer having valid credentials.

This PR introduces a manually created secret to the cluster-logging-operator which then can be used like the automatically-created one before.

### Links

- JIRA: https://issues.redhat.com/browse/LOG-2778


jcantrill (commented):
/approve

PR #1562, opened 2022-07-14
LOG-2648: glob_minimum_cooldown_ms of vector is too big (60000ms)

Clee2691:
### Description

Summary:
- Changed `glob_minimum_cooldown_ms` from the default of 60s to 15s for audit and kubernetes logs when forwarding with Vector.

Details:

This PR addresses the issue where Vector sometimes reports "Failed to annotate event with pod metadata" errors. The proposed fix is to change `glob_minimum_cooldown_ms` to 15s instead of the default of 60s.

/cc @jcantrill @cahartma 

### Links

- https://issues.redhat.com/browse/LOG-2648


cahartma (commented):
/lgtm

PR #1563, opened 2022-07-14
LOG-2770: optimize fluent conf for throughput

jcantrill:
### Description
This PR:
* Optimizes the fluent conf to address performance issues
* Depends on https://github.com/ViaQ/logging-fluentd/pull/37
* Depends on a fluent image with https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/pull/347

### Links
* https://issues.redhat.com/browse/LOG-2770

based on rudamentary testing of viaq optimizations which enable only features that are being used:
```
filtering                rate m/s
---------                ---------
no filtering             100,000
as-is                     28,000
remove es_index           35,000   *
remove debug state        30,000
remove prune_labels       28,000
remove flat_labels        31,000   *
remove rename_time        15,000   ?
remove delete emtpy       45,000    
remove openshift.sequenc  30,000
remove handle_undefined   28,000   
remove pipeline_meta      45,000   *
remove formatter          35,000
```
These changes remove much of the fluent viaq design that predates the use of CRIO where there was no standard incoming record format.
Additionally, the improvemets to the kubernetes_metadata plugin are mostly realized by no longer serializing and loading the meta attached to each record

### DIFF
--- a/internal/generator/fluentd/ingress.go
+++ b/internal/generator/fluentd/ingress.go
@@ -52,7 +52,7 @@ func Ingress(spec *logging.ClusterLogForwarderSpec, o Options) []Element {
 					TemplateStr:  RetagJournalLogs,
 				},
 				ConfLiteral{
-					Desc:         "Invoke kubernetes apiserver to get kunbernetes metadata",
+					Desc:         "Invoke kubernetes apiserver to get kubernetes metadata",
 					TemplateName: "kubernetesMetadata",
 					TemplateStr:  KubernetesMetadataPlugin,
 				},
@@ -60,11 +60,6 @@ func Ingress(spec *logging.ClusterLogForwarderSpec, o Options) []Element {
 					Desc:         "Parse Json fields for container, journal and eventrouter logs",
 					TemplateName: "parseJsonFields",
 					TemplateStr:  ParseJsonFields,
-				},
-				ConfLiteral{
-					Desc:         "Clean kibana log fields",
-					TemplateName: "cleanKibanaLogs",
-					TemplateStr:  CleanKibanaLogs,
 				},
 				ConfLiteral{
 					Desc:         "Fix level field in audit logs",
@@ -198,7 +193,6 @@ const KubernetesMetadataPlugin string = `
   kubernetes_url 'https://kubernetes.default.svc'
   allow_orphans false
   cache_size '1000'
-  use_journal 'nil'
   ssl_partial_chain 'true'
 </filter>
 {{end}}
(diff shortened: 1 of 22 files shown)


internal/generator/fluentd/ingress.go
```diff
@@ -207,43 +201,14 @@ const KubernetesMetadataPlugin string = `
...
-{{define "cleanKibanaLogs" -}}
-# {{.Desc}}
-<filter **kibana**>
-  @type record_transformer
-  enable_ruby
-  <record>
-    log ${record['err'] || record['msg'] || record['MESSAGE'] || record['log']}
-  </record>
-  remove_keys req,res,msg,name,level,v,pid,err
+  json_fields 'message'
```
vparfonov, line 208: previously was 'MESSAGE' in upper case

internal/generator/fluentd/ingress.go
```diff
@@ -367,7 +295,7 @@ const ConcatLines string = `
 {{define "concatLines" -}}
 <filter kubernetes.**>
   @type concat
-  key log
+  key message
```
vparfonov, line 298: just for my understanding, this changing related to the optimize data?

test/framework/functional/message_templates.go
```diff
@@ -49,13 +49,12 @@ var (
...
-		Message:       "*",
-		LogType:       "application",
-		ViaqIndexName: "app-write",
-		Level:         "*",
-		Hostname:      "*",
-		ViaqMsgID:     "*",
+		Timestamp: time.Time{},
+		Message:   "*",
+		LogType:   "application",
+		Level:     "*",
```
vparfonov, line 55: Why we remove `ViaqIndexName`?

internal/generator/fluentd/ingress.go
jcantrill: this comes from 3.11 where we were pulling container logs using the journal driver.  It is no longer relevant since we only pull container logs and I have changed the group capture parameters: https://github.com/openshift/cluster-logging-operator/pull/1563/files#diff-cdd5a7217e8d3eeb415a8f7f720e09303a82c2f9225ee7484d8bdabc11901f1eR148

internal/generator/fluentd/ingress.go
jcantrill: This does not relate to optimization other then I made modifications to the incoming parsing expression and have removed and modified a number of filter configurations.

test/framework/functional/message_templates.go
jcantrill: it is only needed for Elasticsearch so I removed it from ingress and moved it to the ES blocks of configuration.

internal/generator/fluentd/ingress.go
vparfonov: thanks

PR #1568, opened 2022-07-19
NO-JIRA: Get rid of system:admin login from e2e cleanup

alanconway:
No longer needed and a nuisance, it logs you out of kube:admin on each test run.


alanconway:
CI hangs during ` INFO[2022-08-17T02:37:43Z] Running step e2e-ipi-deprovision-deprovision. `, killed after 4hours.
GIving up the chase.

PR #1587, opened 2022-08-03
LOG-2864: Compatibility with ES default output

xperimental:
### Description

Currently the LokiStack integration always creates two outputs, one for each of the log streams that is forwarded by default (`application` and `infrastructure`). While these outputs are called `default` and `default-infra`, they do not behave like the `default` output that is available when using the ElasticSearch log storage.

This PR changes the behaviour of the `default` output reference when using LokiStack, so that it is internally split into different outputs depending on the log stream forwarded into it. This should ensure that an input is always forwarded into the correct tenant, thus enforcing the correct authorization. Additionally this enables the use of custom `ClusterLogForwarder` configurations that were usable with the ElasticSearch log storage, for example to forward `audit` logs into the LokiStack log store.

Known limitations:

- The `ClusterRole` assigned to the `logcollector` currently always has write access to _all tenants_ not just the ones it needs for the current configuration.

/cc @cahartma 
/assign @jcantrill 

### Links

- JIRA: [LOG-2864](https://issues.redhat.com//browse/LOG-2864) [LOG-2867](https://issues.redhat.com//browse/LOG-2867)

### DIFF
--- a/internal/k8shandler/forwarding.go
+++ b/internal/k8shandler/forwarding.go
@@ -123,37 +123,28 @@ func (clusterRequest *ClusterLoggingRequest) NormalizeForwarder() (*logging.Clus
 	clusterRequest.OutputSecrets = make(map[string]*corev1.Secret, len(clusterRequest.ForwarderSpec.Outputs))
 
 	// Check for default configuration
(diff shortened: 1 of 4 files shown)


cahartma (commented):
/lgtm

PR #1590, opened 2022-08-05
LOG-2866: Add k8s recommended labels

vimalk78:
### Description
Added kubernetes recommended labels to resources created by Cluster Logging Operator.

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:

/approve cancel
/hold

### DIFF
--- a/internal/factory/daemonset.go
+++ b/internal/factory/daemonset.go
@@ -1,6 +1,7 @@
 package factory
 
 import (
+	"github.com/openshift/cluster-logging-operator/internal/constants"
 	apps "k8s.io/api/apps/v1"
 	core "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -9,10 +10,19 @@ import (
 
 //NewDaemonSet stubs an instance of a daemonset
 func NewDaemonSet(daemonsetName, namespace, loggingComponent, component string, podSpec core.PodSpec) *apps.DaemonSet {
-	labels := map[string]string{
+	labelSelectors := map[string]string{
 		"provider":      "openshift",
 		"component":     component,
 		"logging-infra": loggingComponent,
+	}
+	labels := map[string]string{
+		"app.kubernetes.io/name":       daemonsetName,
+		"app.kubernetes.io/component":  constants.CollectorName,
+		"app.kubernetes.io/created-by": constants.ClusterLoggingOperator,
+		"app.kubernetes.io/managed-by": constants.ClusterLoggingOperator,
+	}
+	for k, v := range labelSelectors {
+		labels[k] = v
 	}
 	return &apps.DaemonSet{
 		TypeMeta: metav1.TypeMeta{
@@ -26,7 +36,7 @@ func NewDaemonSet(daemonsetName, namespace, loggingComponent, component string, 
 		},
 		Spec: apps.DaemonSetSpec{
 			Selector: &metav1.LabelSelector{
-				MatchLabels: labels,
+				MatchLabels: labelSelectors,
 			},
 			Template: core.PodTemplateSpec{
 				ObjectMeta: metav1.ObjectMeta{
--- a/internal/factory/daemonset_test.go
+++ b/internal/factory/daemonset_test.go
@@ -22,11 +22,21 @@ func TestNewDaemonsetSetsAllLabelsToBeTheSame(t *testing.T) {
 	daemonSet := NewDaemonSet("thenname", "thenamespace", "thecomponent", "thecomponent", podspec)
 
 	expLabels := daemonSet.ObjectMeta.Labels
-	if !reflect.DeepEqual(expLabels, daemonSet.Spec.Selector.MatchLabels) {
-		t.Errorf("Exp. the ObjectMeta.Labels %q to be the same as spec.selector.matchlabels: %q", expLabels, daemonSet.Spec.Selector.MatchLabels)
+	for k, v := range daemonSet.Spec.Selector.MatchLabels {
+		if _, ok := expLabels[k]; !ok {
+			t.Errorf("spec.selector.MatchLabel key: %q does not exist in ObjectMeta.Labels", k)
+		}
+		if !reflect.DeepEqual(expLabels[k], v) {
+			t.Errorf("spec.selector.MatchLabel[%q] value %v is not same as ObjectMeta.Labels[%q]: %v", k, expLabels[k], k, v)
+		}
 	}
-	if !reflect.DeepEqual(expLabels, daemonSet.Spec.Template.ObjectMeta.Labels) {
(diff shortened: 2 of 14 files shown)


cahartma (commented):
/lgtm

internal/factory/daemonset.go
```diff
@@ -10,9 +11,13 @@ import (
 //NewDaemonSet stubs an instance of a daemonset
 func NewDaemonSet(daemonsetName, namespace, loggingComponent, component string, podSpec core.PodSpec) *apps.DaemonSet {
 	labels := map[string]string{
-		"provider":      "openshift",
-		"component":     component,
-		"logging-infra": loggingComponent,
+		"provider":                     "openshift",
```
xperimental, line 14: The `labels` variable is used for both the labels of the DaemonSet and Pod resources as well as in the `LabelSelector` in the DaemonSet.

Selectors are write-once attributes, so this will create an error when the CLO tries to update an existing DaemonSet, which would make updates from a previous version impossible.

I think you need to split up the labels and the LabelSelector, so that the selector can stay with the current value to avoid this error. Alternatively you need to make sure that the old resource is deleted before the update (I would not go that route, though).

internal/k8shandler/deployment.go
```diff
@@ -22,27 +23,39 @@ func NewDeployment(deploymentName string, namespace string, loggingComponent str
...
 			},
 		},
 		Spec: apps.DeploymentSpec{
 			Replicas: utils.GetInt32(1),
 			Selector: &metav1.LabelSelector{
 				MatchLabels: map[string]string{
-					"provider":      "openshift",
-					"component":     component,
-					"logging-infra": loggingComponent,
+					"provider":                     "openshift",
```
xperimental, line 39: The comment about `LabelSelector` in the `DaemonSet` also applies here.

vimalk78:
@xperimental i separated `labelSelectors` and `labels` , but this test case fails
`TestNewDaemonsetSetsAllLabelsToBeTheSame` https://github.com/openshift/cluster-logging-operator/blob/a9ec0df3e3deb05881946b6539baf92aa682233f/internal/factory/daemonset_test.go#L25

xperimental:
@vimalk78 The test was probably written with "selector labelset and labels will be the same" in mind and this needs updating now.

vimalk78:
@xperimental updated the PR. please check

internal/factory/daemonset_test.go
```diff
@@ -22,11 +22,21 @@ func TestNewDaemonsetSetsAllLabelsToBeTheSame(t *testing.T) {
...
 	}
-	if !reflect.DeepEqual(expLabels, daemonSet.Spec.Template.ObjectMeta.Labels) {
-		t.Errorf("Exp. the ObjectMeta.Labels %q to be the same as spec.template.objectmeta.labels: %q", expLabels, daemonSet.Spec.Selector.MatchLabels)
+	for k, v := range daemonSet.Spec.Template.ObjectMeta.Labels {
+		if _, ok := expLabels[k]; !ok {
+			t.Errorf("spec.template.ObjectMeta.Labels key: %q does not exist in ObjectMeta.Labels", k)
+		}
+		if !reflect.DeepEqual(expLabels[k], v) {
+			t.Errorf("spec.template.ObjectMeta.Labels[%q] value %v is not same as ObjectMeta.Labels[%q]: %v", k, expLabels[k], k, v)
+		}
```
xperimental, line 39: The labels of the DaemonSet and Pod are still the same using the new code, so you can keep the simple `reflect.DeepEqual` from before:

```suggestion
	if !reflect.DeepEqual(expLabels, daemonSet.Spec.Template.ObjectMeta.Labels) {
		t.Errorf("Exp. the ObjectMeta.Labels %q to be the same as spec.template.objectmeta.labels: %q", expLabels, daemonSet.Spec.Selector.MatchLabels)
```

internal/factory/daemonset_test.go
```diff
@@ -22,11 +22,21 @@ func TestNewDaemonsetSetsAllLabelsToBeTheSame(t *testing.T) {
 	daemonSet := NewDaemonSet("thenname", "thenamespace", "thecomponent", "thecomponent", podspec)
 
 	expLabels := daemonSet.ObjectMeta.Labels
```
xperimental, line 24: Not on this line: The name of the test `TestNewDaemonsetSetsAllLabelsToBeTheSame` does not match the content anymore.

internal/k8shandler/deployment.go
```diff
@@ -22,9 +23,13 @@ func NewDeployment(deploymentName string, namespace string, loggingComponent str
...
 			Namespace: namespace,
 			Labels: map[string]string{
-				"provider":      "openshift",
-				"component":     component,
-				"logging-infra": loggingComponent,
+				"provider":                     "openshift",
+				"component":                    component,
+				"logging-infra":                loggingComponent,
+				"app.kubernetes.io/name":       loggingComponent,
+				"app.kubernetes.io/component":  constants.CollectorName,
```
xperimental, line 30: I think the value should be `component`.

internal/k8shandler/deployment.go
```diff
@@ -40,9 +45,13 @@ func NewDeployment(deploymentName string, namespace string, loggingComponent str
...
-						"provider":      "openshift",
-						"component":     component,
-						"logging-infra": loggingComponent,
+						"provider":                     "openshift",
+						"component":                    component,
+						"logging-infra":                loggingComponent,
+						"app.kubernetes.io/name":       loggingComponent,
+						"app.kubernetes.io/component":  constants.CollectorName,
+						"app.kubernetes.io/created-by": constants.ClusterLoggingOperator,
+						"app.kubernetes.io/managed-by": constants.ClusterLoggingOperator,
```
xperimental, line 54: These are the same labels as on the `Deployment`. Can be replaced by a variable.

xperimental (changes requested):
I think the applied labels do not match the [documented](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) content and so do not achieve being able to query for different resources belonging to a common application:

- `app.kubernetes.io/name` should identify the "application" the resources belong to. As the purpose of this operator is to provide a "cluster logging solution", I'd say the application could be `cluster-logging`. This would be the same for all generated resources. Currently the code sets this label mostly to the name of the resource which is not what it is meant for.
- `app.kubernetes.io/component` should identify parts of the application. This is now mostly set to `constants.CollectorName` which is okay, since most of the objects are part of the collector except for the Elasticsearch log store (which correctly uses a different component value).

As the labels are very similar between the different resources it might make sense to write a function for applying them to a resource in a central location.

vimalk78:
closing PR. Purpose of PR was adding labels (preferably uniform) which can be configured in vector to filter out log collection of logging components, so that it doesnt go into a positive feedback, not to comply with kubernetes recommendations.
Will open a separate PR for adding these labels, or work with existing labels.

PR #1591, opened 2022-08-10
LOG-2285: Fix disabling certificate validation for Kafka output

xperimental:
### Description

During #1548 I failed to realize that the Kafka output for vector does not use the common `tls.*` options for disabling validation of the certificate.

This PR fixes two issues with the Kafka output:

- TLS was not enabled if the URL contained a `tls://` scheme but the secret did not contain any other TLS options
- Use Kafka's custom option to disable verification of the certificate

/cc @cahartma 
/assign @alanconway 

### Links

- JIRA: [LOG-2285](https://issues.redhat.com//browse/LOG-2285)

PR #1605, opened 2022-08-22
LOG-2988: Implement the output API to support sending logs to the Splunk

vparfonov:

### Description
Add support for sending logs to Spunk (HTTP Event Collector) (https://docs.splunk.com/Documentation/Splunk/latest/Data/FormateventsforHTTPEventCollector)

- generator configuration for Vector Splunk sink 
- new output type `splunk` in the Cluster Logging Forwarder API
- e2e test

Depends on PR https://github.com/ViaQ/vector/pull/109 which will enable Splunk sink feature in Vector

/cc @vimalk78 
/assign @jcantrill  

/cherry-pick 

### Links

- Depending on PR(s): https://github.com/ViaQ/vector/pull/109
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-2988
- Enhancement proposal:


apis/logging/v1/output_types.go
```diff
@@ -225,3 +228,24 @@ type Loki struct {
...
+	// The name of the index where to send the events to. If not specified, the default index is used.
+	// Note: This parameter supports Vector's template syntax, which enables you to use dynamic per-event values.
+	//
+	// +optional
+	Index string `json:"index,omitempty"`
+
+	// Fields to be added to Splunk index. https://docs.splunk.com/Documentation/Splunk/8.0.0/Data/IFXandHEC
+	//
+	// +optional
+	IndexedFields []string `json:"indexed_fields"`
```
jcantrill, line 244: Should this be "fields" since that is what the spllunk doc refs?

apis/logging/v1/output_types.go
```diff
@@ -225,3 +228,24 @@ type Loki struct {
...
 }
+
+// Splunk Deliver log data to Splunk’s HTTP Event Collector
+// Provides optional extra properties for `type: splunk_hec` ('splunk_hec_logs' after Vector 0.23
+type Splunk struct {
+	// The name of the index where to send the events to. If not specified, the default index is used.
+	// Note: This parameter supports Vector's template syntax, which enables you to use dynamic per-event values.
+	//
+	// +optional
+	Index string `json:"index,omitempty"`
```
jcantrill, line 239: @alanconway do we wish to expose this to admins?  When we say "default" in our notes is this whatever their default is or do we intend to use dynamic expansion to set the value to something like what we do for elasticsearch?  (e.g. app/infra/audit)

apis/logging/v1/output_types.go
```diff
@@ -225,3 +228,24 @@ type Loki struct {
...
+	// Fields to be added to Splunk index. https://docs.splunk.com/Documentation/Splunk/8.0.0/Data/IFXandHEC
+	//
+	// +optional
+	IndexedFields []string `json:"indexed_fields"`
+
+	// The sourcetype of events sent to this sink. If unset, Splunk will default to httpevent.
+	// Note: This parameter supports Vector's template syntax, which enables you to use dynamic per-event values.
+	//
+	// +optional
+	SourceType string `json:"sourcetype,omitempty"`
```
jcantrill, line 250: @alanconway I believe we want to remove this based upon https://docs.splunk.com/Documentation/Splunk/8.0.0/Data/Listofpretrainedsourcetypes which defines a known format. Ours will always be JSON so my expectation is we may have some integration issues if we allow it to be set

hack/cr-vector-unmanaged.yaml
```diff
@@ -0,0 +1,8 @@
+apiVersion: "logging.openshift.io/v1"
+kind: "ClusterLogging"
+metadata:
+  name: "instance"
+spec:
+  managementState: "Unmanaged"
```
jcantrill, line 6: If we have an initial file that is unmanaged the CLO will do nothing with it

internal/generator/vector/output/splunk/splunk.go
```diff
@@ -0,0 +1,129 @@
...
+func TLSConf(o logging.OutputSpec, secret *corev1.Secret) []Element {
+	conf := []Element{}
+	if o.Secret != nil {
+		hasTLS := false
+		conf = append(conf, security.TLSConf{
+			ComponentID:        strings.ToLower(vectorhelpers.Replacer.Replace(o.Name)),
+			InsecureSkipVerify: o.TLS != nil && o.TLS.InsecureSkipVerify,
+		})
+
+		if o.Name == logging.OutputNameDefault || security.HasTLSCertAndKey(secret) {
```
jcantrill, line 99: Copy / Paste?   I don't believe we care about the name of the output in this case for splunk

internal/generator/vector/output/splunk/splunk.go
```diff
@@ -0,0 +1,129 @@
...
+			hasTLS = true
+			ca := CAFile{
+				CAFilePath: security.SecretPath(o.Secret.Name, constants.TrustedCABundleKey),
+			}
+			conf = append(conf, ca)
+		}
+		if !hasTLS {
+			return []Element{}
+		}
+	} else if secret != nil {
```
jcantrill, line 117: dont think this is required either

jcantrill (commented):
/hold

apis/logging/v1/output_types.go
alanconway: +1 we don't want to expose this now. We may offer other formats in future but we'll want to retain control of the splunk setting internally, not give it to the user.

internal/generator/vector/output/splunk/splunk.go
alanconway: +1 where does this special logic for default output come from? I don't think we should ever be configuring an output differently based on whether its called "default" or not. The operator is supposed to generate a default config that can be used normally.

internal/generator/vector/output/splunk/splunk.go
alanconway: We really need to centralize this TLS logic somewhere. 
@xperimental did you make any progress on that?

vparfonov: Look like a lot of duplication of TLS logic code, i did similar to Kafka config

internal/generator/vector/output/splunk/splunk.go
vparfonov: it was copy/past error, want to try something faster

apis/logging/v1/output_types.go
vparfonov: removed

apis/logging/v1/output_types.go
vparfonov: fixed

internal/generator/vector/output/splunk/splunk.go
xperimental: > did you make any progress on that?

Sadly, no. The plan was to move the logic into the `security` package, so that outputs using the standard TLS options can benefit from de-duplication.

The kafka output is not a good example, because it does not actually use the standard TLS configuration of vector. I did not notice that in the original implementation, but my fix (#1591) is still not merged. The `splink_hec` output does seem to use the standard TLS options, so you could start consolidating the TLS options in for example Elasticsearch and Splunk to get to a reusable package.

Another change i did in the linked PR is to remove the `hasTLS` flag in the code and not make TLS dependent on the existence of a "secret" (passphrase, CA, certificate) but instead only on the used URL. I think this makes it more transparent to the user.

apis/logging/v1/output_types.go
```diff
@@ -256,3 +259,12 @@ type GoogleCloudLogging struct {
 	//LogID is the log ID to which to publish logs. This identifies log stream.
 	LogID string `json:"logId,omitempty"`
 }
+
+// Splunk Deliver log data to Splunk’s HTTP Event Collector
+// Provides optional extra properties for `type: splunk_hec` ('splunk_hec_logs' after Vector 0.23
+type Splunk struct {
+	// Fields to be added to Splunk index. https://docs.splunk.com/Documentation/Splunk/8.0.0/Data/IFXandHEC
+	// Should be a valid JSON string
```
alanconway, line 267:
```suggestion
	// Should be a valid JSON object
```
I think "JSON sting" is confusing here, it's a string representation of a JSON object value not a JSON string value.

alanconway (changes requested):
One very minor change, then LGTM

apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -148,7 +148,7 @@ type OutputSpec struct {
 
 	// Type of output plugin.
 	//
-	// +kubebuilder:validation:Enum:=syslog;fluentdForward;elasticsearch;kafka;cloudwatch;loki;googleCloudLogging
+	// +kubebuilder:validation:Enum:=syslog;fluentdForward;elasticsearch;kafka;cloudwatch;loki;googleCloudLogging;splunk
```
jcantrill, line 151: @alanconway someplace in the api comments, I don't know if its here or in the Splunk spec, we should update to state what is supported for which collector:
* fluentd: syslog, forward, es, kafka, loki, cw
* vector: es, kafka, cw, loki, gcp, splunk

jcantrill: and a follow up task to where I introduce validation https://github.com/openshift/cluster-logging-operator/pull/1654  we should add validation to reject CLF with reason why

internal/generator/vector/output/splunk/splunk.go
```diff
@@ -0,0 +1,131 @@
...
+		TLSConf(o, secret),
+	)
+}
+
+func Output(o logging.OutputSpec, inputs []string, secret *corev1.Secret, op Options) Element {
+	return Splunk{
+		ComponentID:  strings.ToLower(vectorhelpers.Replacer.Replace(o.Name)),
+		Inputs:       vectorhelpers.MakeInputs(inputs...),
+		Endpoint:     o.URL,
+		DefaultToken: security.GetFromSecret(secret, "hecToken"),
```
jcantrill, line 82: @vparfonov we need to co-ordinate with @libander to document the name of this secret key

test/e2e/logforwarding/splunk/forward_to_splunk_test.go
```diff
@@ -0,0 +1,160 @@
...
+)
+
+func TestLogForwardingToSplunkWithVector(t *testing.T) {
+	_, err := e2e.KubeClient.CoreV1().Secrets(constants.OpenshiftNS).Create(context.TODO(), &secret, metav1.CreateOptions{})
+	if err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			sOpts := metav1.UpdateOptions{}
+			_, err := e2e.KubeClient.CoreV1().Secrets(constants.OpenshiftNS).Update(context.TODO(), &secret, sOpts)
+			if err != nil {
+				Fail(fmt.Sprintf("Unable to craete secret for Vector: %v", err))
```
jcantrill, line 79: typo: create

test/e2e/logforwarding/splunk/forward_to_splunk_test.go
```diff
@@ -0,0 +1,160 @@
...
+			}
+		} else {
+			fmt.Println(err)
+			Fail(fmt.Sprintf("Unable to craete secret for Vector: %v", err))
```
jcantrill, line 83: typo

test/e2e/logforwarding/splunk/forward_to_splunk_test.go
```diff
@@ -0,0 +1,160 @@
...
+	cl.Spec.Collection.CollectorSpec = loggingv1.CollectorSpec{}
+	clf := runtime.NewClusterLogForwarder()
+	clf.Spec = spec
+	testLogForwardingToSplunk(t, cl, clf)
+
+}
+
+func testLogForwardingToSplunk(t *testing.T, cl *loggingv1.ClusterLogging, clf *loggingv1.ClusterLogForwarder) {
+	c := client.ForTest(t)
+	defer framework.RunCleanupScript()
```
jcantrill, line 100: Can we defer "framework.Cleanup" which may also run this cleanup script?

test/framework/e2e/splunk.go
```diff
@@ -0,0 +1,374 @@
...
+	"net/url"
+
+	apps "k8s.io/api/apps/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	SplunkStandalone = "splunk-standalone"
+	Splunk           = "splunk"
+	splunkImage      = "splunk/splunk:9.0.0"
```
jcantrill, line 22: lets retag this image and push it to quay so we ensure we have a copy of it

apis/logging/v1/cluster_log_forwarder_types.go
alanconway: +1 How about on the collector field of the CL? "fluentd supports blah..., vector supports blah..." in one place so its easy to compare support and easy to update as we add more support to vector. I doubt we can use // kube: comments to validate this since it involves 2 separate resources (although I haven't taken time to dig into the kube valdation expression language that apparently has appeared.) in which case it'll have to be in code.

test/e2e/logforwarding/splunk/forward_to_splunk_test.go
vparfonov: done

test/e2e/logforwarding/splunk/forward_to_splunk_test.go
vparfonov: done

test/e2e/logforwarding/splunk/forward_to_splunk_test.go
vparfonov: done

test/framework/e2e/splunk.go
vparfonov: @jcantrill on which org, and who has credentials todo it?

jcantrill: looks like this is available to be updated https://quay.io/repository/openshift-logging/splunk?tab=tags

jcantrill (commented):
/approve

jcantrill (commented):
/hold
/retest

jcantrill (commented):
/lgtm

PR #1607, opened 2022-08-24
WIP: Fix test flakes

alanconway:
Flush out and fix test flakes.

test/functional/outputs/cloudwatch/forward_to_cloudwatch_test.go: Wait for logs to appear with Eventually.


jcantrill:
@alanconway do we believe the logs will eventually arrive?  The read call already has a read timeout of 10m that checks periodically.  I have added this PR to try and dump the logs on failure and I dont see they do. The only difference between now and this test passing previously (e.g. 2w ago) is that since I advanced the OCP version upon where we test from 4.9 to 4.11

alanconway:
/close
Fixed by other work.

PR #1620, opened 2022-09-01
LOG-2919: fix order of reconcilation for ServiceMonitor

jcantrill:
### Description
This PR:
* fixes ServiceMonitor order of reconciliation to try to Get before Create
* Refactors  ServiceMonitor code out of k8shandler/collection.go to decouple from ClusterLoggingRequest

### Links
* https://issues.redhat.com/browse/LOG-2919

### DIFF
--- a/internal/collector/service_monitor.go
+++ b/internal/collector/service_monitor.go
@@ -1,0 +1,72 @@
+package collector
+
+import (
+	"context"
+	"fmt"
+	log "github.com/ViaQ/logerr/v2/log/static"
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/openshift/cluster-logging-operator/internal/constants"
+	"github.com/openshift/cluster-logging-operator/internal/reconcile"
+	"github.com/openshift/cluster-logging-operator/internal/runtime"
+	"github.com/openshift/cluster-logging-operator/internal/utils"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	prometheusCAFile = "/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt"
+)
+
+func NewServiceMonitor(namespace, name string, owner metav1.OwnerReference) *monitoringv1.ServiceMonitor {
+	endpoints := []monitoringv1.Endpoint{}
+	for _, portName := range []string{MetricsPortName, ExporterPortName} {
+		endpoints = append(endpoints,
+			monitoringv1.Endpoint{
+				Port:   portName,
+				Path:   "/metrics",
+				Scheme: "https",
+				TLSConfig: &monitoringv1.TLSConfig{
+					CAFile:     prometheusCAFile,
+					ServerName: fmt.Sprintf("%s.%s.svc", name, namespace),
+				},
+			},
+		)
+	}
+	desired := runtime.NewServiceMonitor(namespace, name)
+	desired.Spec = monitoringv1.ServiceMonitorSpec{
+		JobLabel:  "monitor-collector",
+		Endpoints: endpoints,
+		Selector: metav1.LabelSelector{
+			MatchLabels: map[string]string{
+				"logging-infra": "support",
+			},
+		},
+		NamespaceSelector: monitoringv1.NamespaceSelector{
+			MatchNames: []string{namespace},
+		},
+	}
+
+	utils.AddOwnerRefToObject(desired, owner)
+
+	return desired
+}
+
+func ReconcileServiceMonitor(er record.EventRecorder, k8sClient client.Client, namespace, name string, owner metav1.OwnerReference) {
+	desired := NewServiceMonitor(namespace, name, owner)
+	if err := reconcile.ServiceMonitor(er, k8sClient, desired); err != nil {
+		log.Error(err, "collector.ReconcileServiceMonitor")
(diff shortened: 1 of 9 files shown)


jcantrill:
@xperimental  I updated this change to remove the async.  Thoughts?  Do you think this is better?

internal/collector/service_monitor.go
```diff
@@ -0,0 +1,75 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	prometheusCAFile = "/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt"
+)
+
+func NewServiceMonitor(namespace, name string, owner metav1.OwnerReference) *monitoringv1.ServiceMonitor {
+
```
xperimental, line 21:
```suggestion
```

internal/collector/service_monitor.go
```diff
@@ -0,0 +1,75 @@
...
+func ReconcileServiceMonitor(k8sClient client.Client, namespace, name string, owner metav1.OwnerReference) {
+	desired := NewServiceMonitor(namespace, name, owner)
+	if err := reconcile.ServiceMonitor(k8sClient, desired); err != nil {
+		log.Error(err, "collector.ReconcileServiceMonitor")
+	}
+}
+
+func RemoveServiceMonitor(k8sClient client.Client, namespace, name string) {
+	serviceMonitor := NewServiceMonitor(namespace, name, metav1.OwnerReference{})
```
xperimental, line 71: How about replacing this with an empty resource which just identifies itself with name and namespace?

```suggestion
	serviceMonitor := &monitoringv1.ServiceMonitor{
		ObjectMeta: metav1.ObjectMeta{
			Name:      name,
			Namespace: namespace,
		},
	}
```

internal/reconcile/service_monitor.go
```diff
@@ -0,0 +1,36 @@
...
+//if there is an issue creating or updating to the desired state
+func ServiceMonitor(k8Client client.Client, desired *monitoringv1.ServiceMonitor) error {
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		current := &monitoringv1.ServiceMonitor{}
+		key := client.ObjectKeyFromObject(desired)
+		if err := k8Client.Get(context.TODO(), key, current); err != nil {
+			if errors.IsNotFound(err) {
+				return k8Client.Create(context.TODO(), desired)
+			}
+			return fmt.Errorf("failed to get %v servicemonitor: %v", key, err)
```
xperimental, line 24: Use `%w` if you want to use the "error embedding" of fmt:

```suggestion
			return fmt.Errorf("failed to get %v servicemonitor: %w", key, err)
```

internal/runtime/monitoring.go
```diff
@@ -19,17 +15,6 @@ func NewServiceMonitor(serviceMonitorName, namespace string) *monitoringv1.Servi
 			Name:      serviceMonitorName,
 			Namespace: namespace,
 		},
+		Spec: monitoringv1.ServiceMonitorSpec{},
```
xperimental, line 18: I think the surrounding `NewServiceMonitor` function can be inlined into the function of the same name in `collector/service_monitor.go`.

internal/collector/service_monitor.go
```diff
@@ -0,0 +1,75 @@
...
+	endpoint := monitoringv1.Endpoint{
+		Port:   MetricsPortName,
+		Path:   "/metrics",
+		Scheme: "https",
+		TLSConfig: &monitoringv1.TLSConfig{
+			CAFile:     prometheusCAFile,
+			ServerName: fmt.Sprintf("%s.%s.svc", name, namespace),
+		},
+	}
+	logMetricExporterEndpoint := monitoringv1.Endpoint{
```
xperimental, line 33: `endpoint` and `logMetricExporterEndpoint` are mostly the same except for the port. Extract a function?

internal/reconcile/service_monitor.go
```diff
@@ -0,0 +1,36 @@
...
+	log "github.com/ViaQ/logerr/v2/log/static"
+	monitoringv1 "github.com/coreos/prometheus-operator/pkg/apis/monitoring/v1"
+	"github.com/openshift/cluster-logging-operator/internal/utils/comparators/servicemonitor"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/util/retry"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//ServiceMonitor reconciles a ServiceMonitor to the desired spec returning an error
+//if there is an issue creating or updating to the desired state
```
xperimental, line 15:
```suggestion
// ServiceMonitor reconciles a ServiceMonitor to the desired spec returning an error
// if there is an issue creating or updating to the desired state
```

xperimental (commented):
Yes, I think avoiding concurrency is a good thing for keeping things simple.

I still don't quite get the point of ignoring errors on this resource. Are there a lot of errors on this resource, so you get a lot of retries? I think I would still like to retry creating this resource if the first attempt fails. Maybe moving this resource to the end of the reconciliation loop is an option, so that the requests for all "necessary" resources are tried first?

internal/runtime/monitoring.go
jcantrill: This separation is more trying to establish "good" practice of... a factory to create or stub a SM for any component and one specifically for collection.  I can't anticipate an alternate need at the moment but I was really trying to keep this specific separation of concern

xperimental: I can see the point of keeping this convention, although I try to avoid introducing generalizations until I need them. Some of the "stub creators" are now only used by the delete-path anymore (not this one though), I think this is what originally tripped me up in the `runtime` package.

cahartma (commented):
/lgtm

PR #1623, opened 2022-09-02
all info metrics of Gauge type now got two values : 0 and 1, on chang…

pmoogi-redhat:
### Description
[LOG-2315]
All info metrics in CLO are implemented as Gauge type where values of those metrics were fixed to have a constant value '1'.
This was causing issue as when the labels attain a different value.  For each unique value of labels - prometheus creates a new time-series with update label-value pair.  It still keeps the older label-value pair alive with the older state of metric time-series.  To address this now all info type of clo metrics accommodate two values - 0 and 1.  When metric labels attain a different value then the previous state of metric with old label-value pair is reset to 0.

/cc @alanconway 
/assign @jcantrill 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-2315
- Enhancement proposal:

### DIFF
--- a/internal/telemetry/cloteleminfo.go
+++ b/internal/telemetry/cloteleminfo.go
@@ -2,6 +2,7 @@ package telemetry
 
 import (
 	log "github.com/ViaQ/logerr/v2/log/static"
+	v1 "github.com/openshift/cluster-logging-operator/apis/logging/v1"
 	"github.com/openshift/cluster-logging-operator/internal/utils"
 	"github.com/openshift/cluster-logging-operator/version"
 	"github.com/prometheus/client_golang/prometheus"
@@ -11,6 +12,23 @@ import (
 const (
 	IsPresent    = "1"
 	IsNotPresent = "0"
+
+	InputNameApplication    = v1.InputNameApplication
+	InputNameAudit          = v1.InputNameAudit
+	InputNameInfrastructure = v1.InputNameInfrastructure
+
+	OutputTypeDefault        = "default"
+	OutputTypeElasticsearch  = v1.OutputTypeElasticsearch
+	OutputTypeFluentdForward = v1.OutputTypeFluentdForward
+	OutputTypeSyslog         = v1.OutputTypeSyslog
+	OutputTypeKafka          = v1.OutputTypeKafka
+	OutputTypeLoki           = v1.OutputTypeLoki
+	OutputTypeCloudwatch     = v1.OutputTypeCloudwatch
+
+	ManagedStatus = "managedStatus"
+	HealthStatus  = "healthStatus"
+	VersionNo     = "version"
+	PipelineNo    = "pipelineInfo"
 )
 
 // placeholder for keeping clo info which will be used for clo metrics update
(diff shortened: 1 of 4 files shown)


internal/telemetry/cloteleminfo.go
```diff
@@ -147,6 +171,34 @@ func UpdateCLFMetrics() error {
...
+		"healthStatus": CLFInfo["healthStatus"],
+		"pipelineInfo": CLFInfo["pipelineInfo"]}).Set(0)
+
+	mCLFInputType.With(prometheus.Labels{
+		"application":    CLFInputType["application"],
+		"audit":          CLFInputType["audit"],
+		"infrastructure": CLFInputType["infrastructure"]}).Set(0)
+
+	mCLFOutputType.With(prometheus.Labels{
+		"default":        CLFOutputType["default"],
```
jcantrill, line 191: these all have constants defined in the API code.  we should use them

internal/telemetry/cloteleminfo.go
```diff
@@ -147,6 +171,34 @@ func UpdateCLFMetrics() error {
...
+	CLFInfo := Data.CLFInfo.M
+	CLFInputType := Data.CLFInputType.M
+	CLFOutputType := Data.CLFOutputType.M
+
+	mCLFInfo.With(prometheus.Labels{
+		"healthStatus": CLFInfo["healthStatus"],
+		"pipelineInfo": CLFInfo["pipelineInfo"]}).Set(0)
+
+	mCLFInputType.With(prometheus.Labels{
+		"application":    CLFInputType["application"],
```
jcantrill, line 186: these have constants define in input types, please use them

internal/telemetry/cloteleminfo.go
```diff
@@ -147,6 +171,34 @@ func UpdateCLFMetrics() error {
...
 }
+
+func ResetCLFMetrics() error {
+
+	CLFInfo := Data.CLFInfo.M
+	CLFInputType := Data.CLFInputType.M
+	CLFOutputType := Data.CLFOutputType.M
+
+	mCLFInfo.With(prometheus.Labels{
+		"healthStatus": CLFInfo["healthStatus"],
```
jcantrill, line 182: extract to constants

internal/telemetry/cloteleminfo.go
pmoogi-redhat: @jcantrill "infrastructure", "application", "audit" are the labels for this metrics. Similarly all supported output types also are the labels applied to other info metric. From readability point of view it is better to leave them as it is. They are not any other variables values here.  Please let me know if you still feel that we should replace them by the constants defined for API code or input types. thanks.

jcantrill: > From readability point of view it is better to leave them as it is
I would disagree.  Maintainability is more important to me.  It is common practice in any kind of coding to:
* refactor strings into constants
* reuse constants which provide values across the code base so they can all change as needed when requirements change

PR #1631, opened 2022-09-08
LOG-2620: containers violate new pod security restrictions

cahartma:
### Description
In 4.11+, a new SCC `restrictive-v2` will be added. If the pod/container can conform to it, the SCC will be applied to that pod/container. This will ensure that for 4.11+ workloads can still use the `runtime/default` `seccompProfile`. This field cannot be set in previous versions of OpenShift (<4.11) because there is no SCC which allows setting it.

Because the collector uses its own SCC `log-collector-scc`, we need to label our namespace as 'privileged'.  We are adding the  'seccompProfile' field to our custom scc, which I've verified does not affect previous versions (as it would if added directly to the manifest)

/cc @vimalk78 @syedriko 
/assign @jcantrill 

### Links
- https://issues.redhat.com/browse/LOG-2620


internal/k8shandler/collection.go
```diff
@@ -593,6 +599,29 @@ func (clusterRequest *ClusterLoggingRequest) createOrUpdateCollectorServiceAccou
...
 }
 
+func (clusterRequest *ClusterLoggingRequest) addSecurityLabelsToNamespace() error {
+	ns := &corev1.Namespace{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: clusterRequest.Cluster.Namespace,
+		},
+	}
+
+	if err := clusterRequest.Get(ns.Name, ns); err == nil {
```
vimalk78, line 609: better pattern is to check `err != nil` , write log and return. then continue with adding labels

internal/constants/constants.go
```diff
@@ -71,6 +71,10 @@ const (
 	LogStoreURL                = "https://" + ElasticsearchFQDN + ":" + ElasticsearchPort
 	MasterCASecretName         = "master-certs"
 	CollectorSecretName        = "collector"
+	PodSecurityLabelKey        = "pod-security.kubernetes.io/enforce"
+	PodSecurityLabelKey2       = "pod-security.kubernetes.io/audit"
+	PodSecurityLabelKey3       = "pod-security.kubernetes.io/warn"
```
vimalk78, line 76: rename variables to reflect label names. `PodSecurityLabelEnforce`, `PodSecurityLabelAudit`, `PodSecurityLabelWarn`

PR #1633, opened 2022-09-09
LOG-2720: Add clusterID to data model for vector

jcantrill:
### Description
This PR:
* add support to vector to include openshift.cluster_id in the log payload
* Enables some audit functional testing for vector

### Links
* https://issues.redhat.com/browse/LOG-2720

### DIFF
--- a/internal/generator/vector/normalize.go
+++ b/internal/generator/vector/normalize.go
@@ -11,6 +11,7 @@ import (
 )
 
 const (
+	ClusterID   = `.openshift.cluster_id = "${OPENSHIFT_CLUSTER_ID:-}"`
 	FixLogLevel = `
 if !exists(.level) {
   .level = "default"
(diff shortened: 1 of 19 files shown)


internal/generator/vector/normalize.go
```diff
@@ -11,7 +11,8 @@ import (
 )
 
 const (
-	FixLogLevel = `
+	ClusterIDFormat = `.openshift.cluster_id = "%s"`
```
vimalk78, line 14: Can we generate this line based on an env variable? 
```
	ClusterIDFormat = `.openshift.cluster_id = "${OPENSHIFT-CLUSTER-ID:-abc-123-mock-clusterid}"`
```
this way code changes are minimal, dont have to pass around clusterID in the generator functions. Also makes generated conf independent of a specific cluster.

jcantrill: Updated to use env var

test/helpers/types/types.go
```diff
@@ -179,6 +180,7 @@ type LinuxAuditLog struct {
 	OpenshiftLabels  OpenshiftMeta    `json:"openshift"`
 	Timing           `json:",inline"`
 	Level            string `json:"level,omitempty"`
+	WriteIndex       string `json:"write_index,omitempty"`
```
vimalk78, line 183: we dont need `write_index` now. it has been removed in a recent PR

PR #1636, opened 2022-09-12
LOG-2911: Add nodename label to vector metrics

vimalk78:
### Description
Metrics generated by Vector lack a label whose value is the name opf the kubernetes node on which the collector runs. best available now is hostname which is same as pod name. But podname is not suitable because it is transient. this name changes every time collector is re-deloyed.

This PR adds a new label to the emitted metrics whose value is same as the value of environment variable `${VECTOR_SELF_NODE_NAME}`

The new label will also help in determining correct values to raise an alert, and clear an alert.

metric after this change

```
{
  "name": "internal_metrics_cardinality_total",
  "namespace": "vector",
  "tags": {
    "host": "collector-7vkk5",
    "hostname": "ip-10-0-140-184.ec2.internal"
  },
  "timestamp": "2022-09-12T11:59:20.964479613Z",
  "kind": "absolute",
  "counter": {
    "value": 305.0
  }
}
```

/cc @cahartma @jcantrill 
/assign 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-2911

### DIFF
--- a/internal/generator/vector/conf_test.go
+++ b/internal/generator/vector/conf_test.go
@@ -388,9 +388,16 @@ key_file = "/var/run/ocp-collector/secrets/kafka-receiver-1/tls.key"
 crt_file = "/var/run/ocp-collector/secrets/kafka-receiver-1/tls.crt"
 ca_file = "/var/run/ocp-collector/secrets/kafka-receiver-1/ca-bundle.crt"
 
+[transforms.add_nodename_to_metric]
+type = "remap"
+inputs = ["internal_metrics"]
+source = '''
+.tags.hostname = get_env_var!("VECTOR_SELF_NODE_NAME")
+'''
+
 [sinks.prometheus_output]
 type = "prometheus_exporter"
-inputs = ["internal_metrics"]
+inputs = ["add_nodename_to_metric"]
 address = "0.0.0.0:24231"
 default_namespace = "collector"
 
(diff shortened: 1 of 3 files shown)


syedriko:
@vimalk78 I don't think we're going to find VECTOR_SELF_NODE_NAME env var to be set when it's CLO that deploys Vector. It's Vector's own DaemonSet in https://github.com/ViaQ/vector/blob/v0.21-rh/distribution/kubernetes/vector-agent/daemonset.yaml#L37 that sets VECTOR_SELF_NODE_NAME.

syedriko:
> @vimalk78 I don't think we're going to find VECTOR_SELF_NODE_NAME env var to be set when it's CLO that deploys Vector. It's Vector's own DaemonSet in https://github.com/ViaQ/vector/blob/v0.21-rh/distribution/kubernetes/vector-agent/daemonset.yaml#L37 that sets VECTOR_SELF_NODE_NAME.

It turns out, CLO sets VECTOR_SELF_NODE_NAME for Vector as well https://github.com/openshift/cluster-logging-operator/blob/b826121b5887cb6297d75f028ddc766bc8578c4b/internal/collector/vector/visitors.go#L19

/lgtm

vimalk78:
@syedriko i tested it before opening PR. Check the output example in the first message.

PR #1641, opened 2022-09-13
LOG-2918: Reconcile role/binding allowing reading application logs

xperimental:
### Description

This adds reconciliation for one additional `ClusterRole` and `ClusterRoleBinding` that allows any logged in user to view logs in the `application` tenant. Enforcing filtering by Namespace is done by the authorization component of LokiStack.

Additionally I noticed that the existing role/binding were not deleted by the operator when the CR was deleted, only when the configuration changed to a different log store. Because `ClusterRole` and `ClusterRoleBinding` are not namespaced we can not use the usual way of garbage collection by the Kubernetes apiserver using `OwnerReferences`. [Finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/) seem to be the concept that we need to enforce deletion of the non-namespaced resources that our operator manages connected to a namespaced resource.

I am open to splitting the Finalizer code from the role/role-binding code for separate review.

/cc @cahartma
/assign @alanconway

### Links

- JIRA: [LOG-2918](https://issues.redhat.com//browse/LOG-2918) [LOG-3020](https://issues.redhat.com//browse/LOG-3020)


periklis (approved):
/approve

PR #1642, opened 2022-09-13
fix reconciliation order of Daemonsets and trusted configmap

jcantrill:
### Description
This PR:
* fixes DaemonSet order of reconciliation to try to Get before Create
* fixes trusted ConfigMap order of reconciliation to try to Get before Create
* Adds Events with reasons for updates to the above objects

### Links
* https://issues.redhat.com/browse/LOG-2919


internal/utils/comparators/daemonsets/comparator.go
```diff
@@ -9,61 +9,61 @@ import (
 )
 
 //AreSame compares daemonset for equality and return true equal otherwise false
```
vparfonov, line 11: NIT: docs not updated, now func return also string value with short explanation, what actually not the same

PR #1654, opened 2022-09-21
LOG-2759: Remove message field when JSON parse is enabled

jcantrill:
### Description
This PR:
* Removes the message field when JSON parse is enabled
* Adds admission hook for CLF validation
* Adds validation to ensure structuredTypeKey or structuredTypeKey is defined when pipelines enable json and reference Elasticsearch
* Enables JSON parsing functional tests for all collector impls

/cc @alanconway 

### Links
* https://issues.redhat.com/browse/LOG-2759
* https://issues.redhat.com/browse/LOG-2862


Makefile
```diff
@@ -238,7 +238,7 @@ test-unit: test-forwarder-generator
 	RELATED_IMAGE_FLUENTD=$(IMAGE_LOGGING_FLUENTD) \
 	RELATED_IMAGE_LOG_FILE_METRIC_EXPORTER=$(IMAGE_LOGFILEMETRICEXPORTER) \
 	RELATED_IMAGE_LOGGING_CONSOLE_PLUGIN=$(IMAGE_LOGGING_CONSOLE_PLUGIN) \
-	go test -cover -race ./internal/... `go list ./test/... | grep -Ev 'test/(e2e|functional|client|helpers)'`
+	go test -cover -race ./apis/... ./internal/... ./internal/... `go list ./test/... | grep -Ev 'test/(e2e|functional|client|helpers)'`
```
vparfonov, line 241: `./internal/...` duplicate

alanconway, line 241: Looks like ./internal/... is listed twice.

internal/generator/fluentd/elements/stdout.go
```diff
@@ -7,3 +7,19 @@ const ToStdOut = `
...
+
+type StdOutFilter struct {
+	Pattern string
+}
+
+func (c StdOutFilter) Name() string {
+	return "filterToStdout"
+}
+
+func (c StdOutFilter) Template() string {
```
alanconway, line 19: Good idea for tests. 
@Pranjal-Gupta2 - this would be ideal for your rate limiting tests, you can attach the test to vector's stdout and get exact time of arrival of messages without loki or other store in the way. No need to poll for metrics, you can measure with time.Now() directly.  It can be a functional test that needs nothing but the vector collector.

alanconway: @jcantrill are  you planning to expose this as an output or is there some other way to enable it for tests?
We'd need it for vector also.

alanconway (changes requested):
Trivial type,  otherwise LGTM.

internal/generator/fluentd/elements/stdout.go
jcantrill: > @jcantrill are you planning to expose this as an output or is there some other way to enable it for tests? We'd need it for vector also.

I used this in testing an manually updated the config generation.  I had no intention to expose as output.  I think it would be clever is there were a way to expose it in a "dev" mode.  Maybe via golang tag? that makes it a viable output type?

internal/generator/vector/pipelines.go
```diff
@@ -33,6 +33,7 @@ func Pipelines(spec *logging.ClusterLogForwarderSpec, op generator.Options) []ge
 parsed, err = parse_json(.message)
 if err == null {
   .structured = parsed
+  del(.message)
```
alanconway, line 36: Can we defer JSON parsing till the end of the line (the output) so that the output can decide if it wants JSON parsing or not? We could mark the record here as .json_parse=true, and then in the output do the parse 
`if .json_parse && i_want_structured_data`

alanconway (changes requested):

internal/generator/vector/pipelines.go
jcantrill: This fix is for fluent only.  I have not investigated if we are breaking vector... but maybe i should :)

alanconway: Agreed that we shouldn't parse multiple times, so variation of this - parse at the pipeline and discard message or structured at the output?

cahartma (commented):
/lgtm

PR #1663, opened 2022-09-27
NO-JIRA: Remove flaky part of console test.

alanconway:
Failing in tests using cluster claims, probably based on bad assumptions.  
Keeping the initial part of the test that verifies the plugin is created.


jcantrill (commented):
/lgtm

PR #1668, opened 2022-10-03
LOG-2769: Support Elasticsearch 8.x as an output using vector

cahartma:
### Description
Rebase of previous work by Calvin L.

#### Summary:
Added optional Elasticsearch.Version field to types.
Added suppress_type_name = true -- if forwarding is greater than ES 7.x or is empty (zero value int)
Can forward both to default and external Elasticsearch store at the same time with a ClusterLogForwarder custom resource.
A version must now be specified in the output spec, if using ES v6.x or below

#### Details:
This PR addresses an issue with forwarding logs using vector to Elasticsearch 8.x (ES). Forwarding logs to ES 8.x produced the following error in the logs: Action/metadata line [1] contains an unknown parameter [_type].

To fix the error, in the vector.toml config file, [sinks - <ES>] section, suppress_type_name = true needs to be added for ES versions 7.x and higher. However, we cannot suppress the type name for ES 6.x, which is our internally managed ES.

To distinguish between ES versions, in the output_types.go file I added the Version field under Elasticsearch. This field is optional and will default to empty if not specified by the end user when defining a ClusterLogForwarder custom resource. The internal default store will be ES 6.x.

Using the new version field, the suppress_type_name = true will be added when the version is not specified, or is greater than ES 6.x (our default version).    In summary, a version must be specified, if using ES v6 or below.

### Links
- https://issues.redhat.com/browse/LOG-2769
- original PR https://github.com/openshift/cluster-logging-operator/pull/1543

### DIFF
--- a/apis/logging/v1/cluster_log_forwarder.go
+++ b/apis/logging/v1/cluster_log_forwarder.go
@@ -20,6 +20,8 @@ func IsInputTypeName(s string) bool { return ReservedInputNames.Has(s) }
 
 // Default log store output name.
 const OutputNameDefault = "default"
+const DefaultESVersion = 6
+const LatestESVersion = 8
 
 // IsReservedOutputName returns true if s is a reserved output name.
 func IsReservedOutputName(s string) bool { return s == OutputNameDefault }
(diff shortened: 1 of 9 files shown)


jcantrill (commented):
/lgtm

PR #1670, opened 2022-10-03
LOG-3093: Healthcheck fails for vector cloudwatch forwarding

cahartma:
### Description
When the vector collector pods start, there are healthchecks that run for every sink, including Cloudwatch.  The healthcheck for cw fails due to a validation error: Value '{{ group_name }}' at 'logGroupNamePrefix'  must satisfy regular expression pattern: [\.\-_/#A-Za-z0-9]

Since we use template syntax for this field, the healthcheck will always fail because it is run before any events actually flow through Vector.   We are disabling healthcheck for this sink

/cc @vimalk78 @jcantrill 

### Links
- https://issues.redhat.com/browse/LOG-3093


jcantrill (commented):
/appove

PR #1671, opened 2022-10-04
add wait and fix check DS availability

jcantrill:
### Description
This PR:
* fixes the check for DS availability
* Adds wait condition for logforwarding tests


alanconway (commented):
/lgtm

PR #1681, opened 2022-10-07
Upgrate to vector v0.23-rh, align use of tls.enabled with that version

syedriko:
### Description

Upgrate to vector v0.23-rh, align use of tls.enabled with that version.
In Vector 0.23, the only two components we care about that require tls.enable are the kafka and prometheus_exporter sinks.

/cc @vimalk78 
/assign @jcantrill

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ export LOGGING_VERSION?=5.6
 export NAMESPACE?=openshift-logging
 
 IMAGE_LOGGING_FLUENTD?=quay.io/openshift-logging/fluentd:1.14.6
-IMAGE_LOGGING_VECTOR?=quay.io/openshift-logging/vector:0.21-rh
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -22,7 +22,7 @@ export LOGGING_VERSION?=5.6
 export NAMESPACE?=openshift-logging
 
 IMAGE_LOGGING_FLUENTD?=quay.io/openshift-logging/fluentd:1.14.6
-IMAGE_LOGGING_VECTOR?=quay.io/openshift-logging/vector:0.21-rh
+IMAGE_LOGGING_VECTOR?=quay.io/openshift-logging/vector:0.23-rh
```
vimalk78, line 25: this change (among other vector version related changes) also coming in @vparfonov PR https://github.com/openshift/cluster-logging-operator/pull/1605

syedriko: that's alright, we can merge this one afterwards and drop this ^^^ change

PR #1682, opened 2022-10-07
LOG 2207: Add policies based log flow control in CLO

Pranjal-Gupta2:
### Description
"Flow control" refers to how the logging system behaves when logs are produced faster than they can be collected or forwarded. This PR enhances the API to let cluster administrators limit logging rates, or ignore some logs entirely. Logs may still be lost if the collector cannot keep up, but administrators have more control over what is lost, and more predictability of log rates.

**Control log rates and overflow policy at two points in the log forwarder:**

- Output: controlling the flow rate per destination to selected outputs.
  - Limit the rate of outbound logs to match output network and storage capacity.
  - Controls aggregated (per-destination) output rate.
- Input: Controlling log flow rates per container from selected containers.
  - Limit the rate of log collection for selected groups of containers per-container.
  - Controls individual (per-container) collection throttling.

**Note:**
- Limit is applied as number of records, not _bytes_
- This enhancement does not include a `block` policy, which would back-pressure containers that exceed rate limits, forcing them to block on stout/std err and slow down to keep within the rate limit.

**Example**:
**Set a per-container limit for containers with certain labels**
```
  inputs:
	- application:
          selector:
		matchLabels: { importance: low }
          limitPerContainer:
                policy: drop
                maxRecordsPerSecond: 10
  - application:
    selector:
	  	matchLabels: { importance: high }
    limitPerContainer:
                policy: drop
		maxRecordsPerSecond: 1000
```

/cc @jcantrill @vimalk78 @eranra 
/assign @alanconway 

### Links

- Depending on PR(s): NA
- Bugzilla: NA
- Github issue: NA
- JIRA: https://issues.redhat.com/browse/LOG-2207
- Enhancement proposal: https://issues.redhat.com/browse/LOG-1043


apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -130,6 +131,20 @@ type Application struct {
 	//
 	// +optional
 	Selector *LabelSelector `json:"selector,omitempty"`
+
+	// Flow control limit to be applied to the aggregated log
```
jcantrill, line 135: golang comment pattern is to start with the field or method name (e.g. GroupLimit ...)

apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -130,6 +131,20 @@ type Application struct {
...
+	//
+	// +optional
+	GroupLimit *LimitSpec `json:"limitGroup,omitempty"`
+
+	// Flow control limit to be applied to each container selected
+	// by this input. No container selected by this input can
+	// exceed this limit.
+	//
+	// +optional
+	ContainerLimit *LimitSpec `json:"limitPerContainer,omitempty"`
```
jcantrill, line 147: Same regarding comment here

apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -215,6 +230,13 @@ type OutputSpec struct {
 	//
 	// +optional
 	Secret *OutputSecretSpec `json:"secret,omitempty"`
+
+	// Flow control limit to be applied to the aggregated log
```
jcantrill, line 234: Same

apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -328,3 +350,26 @@ type ClusterLogForwarderList struct {
...
+	//
+	// +kubebuilder:validation:Enum:=drop;ignore
+	// +required
+	Policy PolicyType `json:"policy,omitempty"`
+
+	// MaxRecordsPerSecond is the maximum number of log records
+	// allowed per input/output in a pipeline
+	//
+	// +required
+	MaxRecordsPerSecond resource.Quantity `json:"maxRecordsPerSecond,omitempty"`
```
jcantrill, line 367: whould this be quantity?  not a long or the like?

jcantrill (commented):
/approve

jcantrill:
cc @alanconway @vimalk78

apis/logging/v1/cluster_log_forwarder_types.go
PranjalGupta2199: @alanconway Please check this comment

Pranjal-Gupta2: @alanconway Please check this comment

alanconway: Quantity isn't wrong: "Quantity is a fixed-point representation of a number" - it doesn't _have_ to be a byte quantity. However it almost always _is_ used for byte quantities so I think it might cause confusion here since  a byte rate would also make sense. I would go with int64.

internal/generator/vector/inputs.go
```diff
@@ -142,6 +147,8 @@ func UserDefinedAppRouting(spec *logging.ClusterLogForwarderSpec, o Options) map
 					}
 					if len(matchNS) != 0 || len(matchLabels) != 0 {
 						routeMap[input.Name] = Quote(AND(OR(matchNS...), AND(matchLabels...)))
+					} else {
+						routeMap[input.Name] = fmt.Sprintf(`'%s'`, "true")
```
alanconway, line 151: No need for printf
```suggestion
						routeMap[input.Name] = "'true'"
```

internal/generator/vector/outputs.go
```diff
@@ -43,19 +66,28 @@ func Outputs(clspec *logging.CollectionSpec, secrets map[string]*corev1.Secret,
...
-			outputs = generator.MergeElements(outputs, loki.Conf(o, inputs, secret, op))
-		case logging.OutputTypeElasticsearch:
-			outputs = generator.MergeElements(outputs, elasticsearch.Conf(o, inputs, secret, op))
-		case logging.OutputTypeCloudwatch:
-			outputs = generator.MergeElements(outputs, cloudwatch.Conf(o, inputs, secret, op))
-		case logging.OutputTypeGoogleCloudLogging:
-			outputs = generator.MergeElements(outputs, gcl.Conf(o, inputs, secret, op))
-		case logging.OutputTypeSplunk:
-			outputs = generator.MergeElements(outputs, splunk.Conf(o, inputs, secret, op))
+
```
alanconway, line 69: Normalize so Ignore is handled _exactly_ the same as MaxRecordsPerSecond == 0. 
It is logically equivalent so the same optimizations should apply either way.

```suggestion
if o.Limit.MaxRecordsPerSecond == 0 {
  o.Limit.Policy = logging.IgnorePolicy // Handle limit==0 as Ignore
}
if o.Limit.Policy == logging.Ignore {
  o.Limit.MaxRecordsPerSecond = 0 // make sure we ignore any max value given with Ignore
}

```

internal/generator/vector/pipelines.go
```diff
@@ -42,8 +70,24 @@ if .log_type == "application" {
...
+				} else if input.Application.GroupLimit != nil {
+					if input.Application.GroupLimit.Policy == logging.DropPolicy {
+						el = append(el, AddThrottle(input)...)
+						inputs = append(inputs, fmt.Sprintf(UserDefinedSourceThrottle, input.Name))
+					} else {
+						log.V(1).Info("Not adding source in pipeline due to ignore policy")
+					}
+				} else {
+					inputs = append(inputs, fmt.Sprintf(UserDefinedAppInput, inputName))
+				}
```
alanconway, line 90: There are possible nil pointer errors here an some duplication  (double check my logic in case I missed somethign)
```suggestion
			if input, ok := userDefined[inputName]; ok && input.Application != nil {
			  // determine if a throttle is needed
				if (((app.ContainerLimit != nil && app.ContainerLimit.Policy != logging.IgnorePolicy) || 
				     (app.GroupLimit != nil && app.GroupLimit.Policy != logging.IgnorePolicy))) {
						el = append(el, AddThrottle(input)...)
						inputs = append(inputs, fmt.Sprintf(UserDefinedSourceThrottle, input.Name))
				} else {
					inputs = append(inputs, fmt.Sprintf(UserDefinedAppInput, inputName))
				}
```

alanconway (changes requested):
A few minor changes but otherwise looks good.

internal/generator/vector/inputs.go
Pranjal-Gupta2: @alanconway Updated `maxRecordsPerSecond` to be `int64`. Do we need to make changes in the enhancement doc as well?

test/e2e/flowcontrol/application_test.go
```diff
@@ -0,0 +1,158 @@
...
+				} // 10 files and 100 group limit, so 10 lines per file
+
+				DeployLoggingComponents()
+				if WaitForMetricsToShow() == false {
+					Fail("Metrics not showing up in Prometheus")
+				}
+				// sleeping for 2 minutes to ensure rate of metrics is stable
+				time.Sleep(2 * time.Minute)
+
+				promQuery = fmt.Sprintf(SumMetric, VectorCompSentEvents)
```
alanconway, line 149: This line has no effect.

Pranjal-Gupta2: @alanconway Can you please explain why this line has no effect?

alanconway: I should have put the comment a line further down, so you can see the next line:
```
149				promQuery = fmt.Sprintf(SumMetric, VectorCompSentEvents)
150				promQuery = fmt.Sprintf(promQuery, FlowControlComponent)
```
Assigning twice to same variable so only second assignment has any effect.

apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -147,6 +147,20 @@ type Application struct {
 	//
 	// +optional
 	Selector *LabelSelector `json:"selector,omitempty"`
+
+	// Group limit applied to the aggregated log
+	// flow to this input. The total log flow from this input
+	// cannot exceed the limit.
+	//
+	// +optional
+	GroupLimit *LimitSpec `json:"groupLimit,omitempty"`
```
jcantrill, line 156: @Pranjal-Gupta2 @alanconway why do these only apply to application logs?  Why can they not apply to infrastructure container logs? or audit logs? or journal logs?  I believe especially infra journal logs, maybe  container logs are chatty

apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -352,3 +373,26 @@ type ClusterLogForwarderList struct {
 func init() {
 	SchemeBuilder.Register(&ClusterLogForwarder{}, &ClusterLogForwarderList{})
 }
+
+type LimitSpec struct {
+	// Policy is used to define the action taken
```
jcantrill, line 378: Given what i see in the code:
```

			if limit.Policy == loggingv1.IgnorePolicy {
				limit.MaxRecordsPerSecond = 0 // make sure we ignore any max value given with Ignore
				limit.Policy = loggingv1.DropPolicy
			}
```
The migration says that type "ignore" means to Ignore all values completely.  This makes sense since "ignore" is a variant of "drop" as evidenced by the migration.  This needs further clarification or different design in the struct IMO.  The notes here and usage of the API imply I can add a MaxRecordsPerSecond  and Ignore which is not the case.  We should either update the notes or fix the API in a way that allows defining these policy specific config bits with the policy to which it applies

test/functional/flowcontrol/utils.go
```diff
@@ -0,0 +1,32 @@
...
+var (
+	LokiNsQuery     = `{kubernetes_namespace_name=~"%s"}`
+	LokiAuditQuery  = `{log_type="audit"}`
+	AllLogs         = `.+`
+	MessageTemplate = "$(date -u +'%Y-%m-%dT%H:%M:%S.%N%:z') stdout F $msg "
+	FileTemplate    = "%s/%s_%s_%s/%s/0.log"
+	LogPath         = "/var/log/pods"
+)
+
+func WriteApplicationLogs(f *functional.CollectorFunctionalFramework, numOfLogs int) error {
```
jcantrill, line 20: What is special about these "write" functions that can not be replaced by those in test/framework/functional/write.go

Pranjal-Gupta2: This function writes `numOfLogs` lines in 1 sec. In the implementation, we sleep for `1/numOfLogs` secs after every echo. The other "write" functions don't do that, right? 
For these tests we want to maintain a rate at which logs are written so that validate the rate of logs flow through vector.

apis/logging/v1/cluster_log_forwarder_types.go
Pranjal-Gupta2: Yes, you can have `maxRecordsPerSecond` with IgnorePolicy in CLF spec but during migration it will normalize everything to drop policy. 
@alanconway Can you please take a look at the notes?

apis/logging/v1/cluster_log_forwarder_types.go
Pranjal-Gupta2: @jcantrill  For the first version, we just want to add flow control policies to the Application Logs.

alanconway: +1

apis/logging/v1/cluster_log_forwarder_types.go
alanconway: I've been back and forth on whether we should even have ignore.
Ignore is identical to drop+maxrecords=0; my original thinking was that completely ignoring a log stream is a special case that's worth highlighting as a policy.
However it seems more and more like the special case is just causing extra work & confusion.
@jcantrill I'm willing to drop Ignore if you think that would be clearer.

test/functional/flowcontrol/utils.go
alanconway: +1 this seems to be identical to 
```
func (f *CollectorFunctionalFramework) WritesApplicationLogs(numOfLogs int) error {
```

Pranjal-Gupta2: @alanconway But for this function we won't have a control on the log generation rate, right?

jcantrill: @Pranjal-Gupta2 so please add  a new signiture to `test/framework/functional/write.go` in the event we wish to reuse it

apis/logging/v1/cluster_log_forwarder_types.go
jcantrill: I like the idea of an "Ignore" policy for convenience but if they only way we get it is to have to "associate" a configuration property that does not apply to it, then I think it could lead to confusion. I don't know what a "better" design would look like.

Pranjal-Gupta2: @jcantrill Should we drop ignore policy or leave it as it is for now?

jcantrill:
@Pranjal-Gupta2 make sure to work with @syedriko if needed to ensure the vector changes are included with the image we are using for the 5.8 release

jcantrill:
/retest 

merged https://github.com/openshift/cluster-logging-operator/pull/2055 to hopfully resolve one flake

jcantrill:
@Pranjal-Gupta2 you may need to rebase on https://github.com/openshift/cluster-logging-operator/pull/2055  if you have not to take advantage of the changes.  I'm not certain they fix the issue but in the PR tests based without duplicating the issue at hand

PR #1686, opened 2022-10-10
LOG-3157: Do not use global variable when updating CR status

xperimental:
### Description

The `ClusterLoggingRequest` struct contains a `cluster` variable that is used as a "cache" for the current ClusterLogging CR that is reconciled. The variable is meant to be written only once during the start of reconciliation and only with a result from the `getClusterLogging` function, so that migrations are correctly applied. These migrations are currently only applied in-memory inside the CLO and never written back to the apiserver.

The `Update` and `UpdateStatus` functions write the updated resource back into the variable passed to them, which causes a resource without applied migrations to be in the variable after the call happened.

This PR changes the code for updating status information on the ClusterLogging CR to use an modified copy of the CR and not write the result back into the "global variable". It is a direct port of the [bugfix](https://github.com/openshift/cluster-logging-operator/pull/1685) for 5.5.

### Links

- JIRA: [LOG-3157](https://issues.redhat.com//browse/LOG-3157)

### DIFF
--- a/internal/k8shandler/clusterloggingrequest.go
+++ b/internal/k8shandler/clusterloggingrequest.go
@@ -59,7 +59,7 @@ func (clusterRequest *ClusterLoggingRequest) Update(object client.Object) (err e
 	return err
 }
 
-//Update the runtime Object status or return error
+// UpdateStatus modifies the status sub-resource or returns an error.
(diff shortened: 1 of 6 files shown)


xperimental (commented):
Left a few explanatory comments.

jcantrill (commented):
/lgtm

PR #1693, opened 2022-10-13
Fix issue with waiting on delete CL and CLF instance

vparfonov:

### Description
Sometimes got exceptions with `oc wait --for=delete` in case deleting going to fast command this PR replace it with periodic `oc get` command 

/cc 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:


jcantrill:
/approve

Looks like it needs to be modified even more as though it was deleted and by the time you get to the "assert" it never fails so that fails.  Maybe add a "loop" with the get and capture the retrurn code. Try for so many times until you get an error code and exit

cahartma (commented):
/lgtm

PR #1701, opened 2022-10-17
LOG-2789: fix collector deployment twice because of CA trustbundle

jcantrill:
### Description
This PR:
* removes controller to reconcile trusted CA configmap in favor of the watch that already exists for CL owned configmaps
* moves and refactors secret reconciliation test
* Addresses secret reconciliation as related to https://issues.redhat.com/browse/LOG-2919

### Links
* https://issues.redhat.com/browse/LOG-2789


internal/k8shandler/fluentd.go
```diff
@@ -1,10 +1,5 @@
 package k8shandler
 
-import (
```
vparfonov, line 3: Maybe we can logic from here to the forwarding.go and remove this?

PR #1707, opened 2022-10-18
LOG-3145:  dedot vector tag value

cahartma:
### Description
A customer who has switched collectors found that in vector the value of the "tag" field has a dot pre-pended.  In fluentd this field value was without the dot.  
This is the first I've heard of a customer actually using this field, as it is not intended to be part of our data model.   It's used to create stream names (based on the type of file) when using cloudwatch, then the tag field is deleted.  In this case, the customer is not using CW output, and thus the tag field remains.   As with a few other fields, this only remains for feature parity with fluentd.

/cc @vimalk78 @syedriko 
/assign @jcantrill 

### Links
- https://issues.redhat.com/browse/LOG-3145

### DIFF
--- a/internal/generator/vector/conf_test.go
+++ b/internal/generator/vector/conf_test.go
@@ -167,7 +167,7 @@ type = "remap"
 inputs = ["raw_journal_logs"]
 source = '''
   .openshift.cluster_id = "${OPENSHIFT_CLUSTER_ID:-}"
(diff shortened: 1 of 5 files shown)


cahartma:
Closed as will not be fixing the tag field.  It was never part of our model and is therefor a bug that it exists.

PR #1713, opened 2022-10-20
Update golangci-lint to v1.48

cahartma:
### Description
With 1.18 installed locally, golangci-lint currently ignores the project version and disables certain linters... causing an error.  Since v1.45.1, golangci-lint can detect the Go version used by a project and run the correct linters accordingly.   Version 1.48 was the latest that worked without larger dependency issues.
- https://github.com/golangci/golangci-lint/issues/2649

/cc @vimalk78 @syedriko 
/assign @jcantrill 

### Links
- Github issue: https://github.com/golangci/golangci-lint/issues/2649


cahartma:
/close
My local is no longer failing for v1.45.0 and this is proving to be a dependency headache.

PR #1725, opened 2022-10-25
LOG-3225: Update priority class value to be closer to system-cluster-…

jcantrill:
### Description
 Update priority class value to be closer to system-cluster-critical

### Links
* https://issues.redhat.com/browse/LOG-3225


syedriko:
@jcantrill I think it makes sense to run collector Pods with the system-cluster-critical priorityClass. With anything lower we're bound to be back here bumping numerical priority.

jcantrill:
> @jcantrill I think it makes sense to run collector Pods with the system-cluster-critical priorityClass. With anything lower we're bound to be back here bumping numerical priority.
@syedriko  this is the max value allowed for user workloads so there will be no need to bump.  I actually wondered if we could use `system-cluster-critical` but not certain if we fall into that category or if it is otherwise restricted to core components.

PR #1729, opened 2022-10-27
LOG-2906: Add ability forward logs to Splunk output

jcantrill:
Attempt to replace #1605 that is for some reason having test issues

cc @vparfonov


olm_deploy/scripts/env.sh
```diff
@@ -3,7 +3,7 @@ set -eou pipefail
 
 LOGGING_VERSION=${LOGGING_VERSION:-5.6}
 LOGGING_FLUENTD_VERSION=${LOGGING_FLUENTD_VERSION:-1.14.6}
-LOGGING_VECTOR_VERSION=${LOGGING_VECTOR_VERSION:-0.21-rh}
+LOGGING_VECTOR_VERSION=${LOGGING_VECTOR_VERSION:-v0.23-rh}
```
vimalk78, line 6: change this back to v0.21-rh

vparfonov: without `v` to 0.21-rh, right @vimalk78 ?

vimalk78: yes, without `v`

vparfonov:
need to squash commits

PR #1731, opened 2022-10-27
LOG-3228: fix applying OutputDefault to unmanaged Elasticsearch logStores

jcantrill:
### Description
This PR:
* fixes the issue where OutputDefaults are applied generically to all ES output types instead of only the managed Elasticsearch offering.

### Links
* https://issues.redhat.com/browse/LOG-3228

### DIFF
--- a/internal/k8shandler/forwarding.go
+++ b/internal/k8shandler/forwarding.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"github.com/openshift/cluster-logging-operator/internal/migrations"
 	"strings"
 
 	forwardergenerator "github.com/openshift/cluster-logging-operator/internal/generator/forwarder"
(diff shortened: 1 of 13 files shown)


internal/k8shandler/forwarding.go
```diff
@@ -40,7 +40,7 @@ func (clusterRequest *ClusterLoggingRequest) useOldRemoteSyslogPlugin() bool {
 
 func applyOutputDefaults(outputDefaults *logging.OutputDefaults, out logging.OutputSpec) logging.OutputSpec {
 
-	if outputDefaults != nil && outputDefaults.Elasticsearch != nil {
+	if outputDefaults != nil && outputDefaults.Elasticsearch != nil && out.Name == logging.OutputNameDefault {
 		// TODO: fix - outputDefaults should only apply to default output
```
vimalk78, line 44: nit: can remove this TODO

internal/k8shandler/forwarding_test.go
```diff
@@ -623,8 +623,7 @@ var _ = Describe("Normalizing forwarder", func() {
 					_, _ = request.generateCollectorConfig()
 
 					Expect(len(request.ForwarderSpec.Outputs) == 1).To(BeTrue())
-					Expect(request.ForwarderSpec.Outputs[0].Elasticsearch.StructuredTypeKey).To(Equal("kubernetes.labels.mylabel"))
```
vimalk78, line 626: can we enforce this at the API level that outputDefaults can be used only if there is an output named `default` ? if not, can we update this in status that "`outputDefaults` is redundant because there is no output named `default`" ?
otherwise how does user get any feedback ?

jcantrill (commented):
/hold

PR #1733, opened 2022-10-28
LOG-3311: Support Http output for vector

vimalk78:
### Description

Draft for review: API for `http` output

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-3311
- Enhancement proposal:


apis/logging/v1/output_types.go
```diff
@@ -277,3 +279,24 @@ type Splunk struct {
...
 	Fields []string `json:"fields,omitempty"`
 }
+
+// Http provided configuration for sending logs to a generic http endpoint
+type Http struct {
+	// Encoding to be used for sending logs.
+	//
+	// +kubebuilder:validation:Enum:=json
+	// +optional
+	Encoding string `json:"encoding,omitempty"`
```
alanconway, line 289: Don't need this yet - we can add it later if/when we have more than one encoding.

alanconway (approved):
Add a "method" and "maxBatch" as per https://docs.google.com/document/d/1VdeOohi70ZQdvJggxYZbxBp7vPTZ4XMVDOBh8ZYvtaw/edit#

apis/logging/v1/output_types.go
vimalk78: > Add a "method" and "maxBatch" as per https://docs.google.com/document/d/1VdeOohi70ZQdvJggxYZbxBp7vPTZ4XMVDOBh8ZYvtaw/edit#

Do we want `maxBatch` in terms of bytes, or number of events? vector supports 3 kinds
https://v0-23.vector.dev/docs/reference/configuration/sinks/http/#batch

alanconway: Bytes. The limits on loki or the like are in bytes and the size of events is not predictable.

jcantrill: @vimalk78 does fluent support "max" bytes in this context.  The choice will need to apply to both

test/framework/functional/output_http.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"strings"
+
+	log "github.com/ViaQ/logerr/v2/log/static"
+	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/internal/runtime"
+)
+
+const (
+	VectorImage        = "docker.io/timberio/vector:0.25.0-alpine"
```
alanconway, line 12: Prefer quay.io image if possible. Docker rate limits can cause failures if runing tests intensively.

alanconway (changes requested):
Nearly LGTM - can you add a functional test to forward to fluentd? The first use case for this is likely to provide a workaround for lack of fluent-forward to customers forwarding to fluentd.

test/framework/functional/output_http.go
vimalk78: updated to use quay image

vimalk78: @alanconway you mean sending to fluentd using http?

vimalk78: updated PR to add a functional test which sends logs from `vector` to `fluentd` using `http`

apis/logging/v1/output_types.go
vimalk78: `fluentd` doesn't seem to support any config parameters for batching.  but recommends putting multiple records in same message with json
https://docs.fluentd.org/input/http#handle-large-data-with-batch-mode

apis/logging/v1/output_types.go
```diff
@@ -282,3 +285,28 @@ type Splunk struct {
...
+	Timeout string `json:"timeout,omitempty"`
+
+	// Method specifies the Http method to be used for sending logs. If not set, 'POST' is used.
+	// +kubebuilder:validation:Enum:=GET;HEAD;POST;PUT;DELETE;OPTIONS;TRACE;PATCH
+	// +optional
+	Method string `json:"method,omitempty"`
+
+	// MaxBatch specifies maximum size of batch in bytes
+	// +optional
+	MaxBatch string `json:"maxBatch,omitempty"`
```
jcantrill, line 306: @vimalk78 this is the param we discussed that is not supported in both collectors?  If so please remove for the time being.  cc @alanconway

apis/logging/v1/output_types.go
```diff
@@ -282,3 +285,28 @@ type Splunk struct {
...
+
+	// Compression specifies the compression type to be used. If not set, defaults to 'none' which does not compress
+	// +kubebuilder:validation:Enum:=gzip;none
+	// +optional
+	Compression string `json:"compression,omitempty"`
```
jcantrill, line 311: same question here

test/framework/functional/output_http.go
```diff
@@ -0,0 +1,99 @@
...
+	log "github.com/ViaQ/logerr/v2/log/static"
+	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/internal/constants"
+	"github.com/openshift/cluster-logging-operator/internal/runtime"
+	"github.com/openshift/cluster-logging-operator/internal/utils"
+)
+
+const (
+	//VectorImage          = "quay.io/domino/vector:0.26.X-distroless-static"
+	VectorImage          = "docker.io/timberio/vector:0.25.0-alpine"
```
jcantrill, line 15: We should:
* move this to someplace where it is not buried here in the tests

@alanconway  any reason we should not use our collector image?

test/framework/functional/output_http.go
```diff
@@ -0,0 +1,99 @@
...
+	VectorHttpSourceConf = `
+[sources.my_source]
+type = "http"
+address = "127.0.0.1:8090"
+encoding = "json"
+
+[sinks.my_sink]
+inputs = ["my_source"]
+type = "file"
+path = "/shared/output.logs"
```
jcantrill, line 25: Does this path work on OCP 4.12?

test/framework/functional/output_http.go
```diff
@@ -0,0 +1,99 @@
...
+
+[sinks.my_sink.encoding]
+codec = "json"
+`
+
+	FluentdImage          = "quay.io/openshift-logging/fluentd"
```
jcantrill, line 31: Can we pull this from know constant we already maintain?

test/framework/functional/read.go
```diff
@@ -71,6 +71,25 @@ func (f *CollectorFunctionalFramework) ReadApplicationLogsFromKafka(topic string
 	return results, err
 }
 
+func (f *CollectorFunctionalFramework) ReadApplicationLogsFromVector(consumercontainername string) (results []string, err error) {
```
jcantrill, line 74: is the file format different enough we need a different function?  Is it not possible to have fluent and vector write out to the same file location and have a single function to read both?

test/functional/outputs/http/suite_test.go
```diff
@@ -0,0 +1,13 @@
+package http
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+func TestFunctionalKafkaOutput(t *testing.T) {
```
jcantrill, line 10: Change name to "TestSuite"

jcantrill (changes requested):
Please also update our docs with this PR https://github.com/openshift/cluster-logging-operator/blob/master/docs/features/collection.adoc#outputs

test/framework/functional/output_http.go
vimalk78: > any reason we should not use our collector image?

it is probably because we dont have `http` receiver in our image

apis/logging/v1/output_types.go
vimalk78: removed

apis/logging/v1/output_types.go
vimalk78: did not find any parameter for compression, have raised a query in fluentd forum

test/framework/functional/output_http.go
vimalk78: removed this variable

test/framework/functional/read.go
vimalk78: removed this function. thanks !!

test/functional/outputs/http/suite_test.go
vimalk78: changed.

test/framework/functional/output_http.go
vimalk78: but changed this, as per comment below

apis/logging/v1/output_types.go
vimalk78: https://github.com/fluent/fluentd/discussions/4008

test/framework/functional/output_http.go
jcantrill: Is enabling http as  a receiver in the image different then enabling it as a sender?  @alanconway do we want to enable http as a receiver in our image?

vimalk78: i  am mistaken. our image in quay has both http sink and http source enabled.

vimalk78: updated PR to use same image

vimalk78:
> Please also update our docs with this PR https://github.com/openshift/cluster-logging-operator/blob/master/docs/features/collection.adoc#outputs

@jcantrill updated `docs` for `http` output

apis/logging/v1/output_types.go
jcantrill: /hold

We need to either:
* remove  this from the API since it is not supported by both
* agree to qualifying how it applies to only one collector cc @alanconway

test/functional/outputs/http/forward_to_http_test.go
```diff
@@ -0,0 +1,88 @@
...
+		functional.NewClusterLogForwarderBuilder(framework.Forwarder).
+			FromInput(logging.InputNameApplication).
+			ToHttpOutput()
+	})
+
+	AfterEach(func() {
+		framework.Cleanup()
+	})
+
+	Context("Application Logs", func() {
```
jcantrill, line 36: Why do we need to write this test twice for different collectors?  Should we not be able to use the same test and change the "tag" to run

vimalk78: updated.

PR #1734, opened 2022-10-30
LOG-3211: refactor serviceaccount reconciliation

jcantrill:
### Description
This PR:
* Completes serviceaccount reconciliation that was missed

### Links
* https://issues.redhat.com/browse/LOG-3211


bundle/manifests/clusterlogging.clusterserviceversion.yaml
```diff
@@ -434,8 +434,6 @@ spec:
                     - ALL
               nodeSelector:
                 kubernetes.io/os: linux
-              securityContext:
```
cahartma, line 437: Even though RunAsUser is defined in our custom scc, we should still be adding the runAsNonRoot:true to our manifests.

config/manager/manager.yaml
```diff
@@ -16,8 +16,6 @@ spec:
       labels:
         name: cluster-logging-operator
     spec:
-      securityContext:
```
cahartma, line 19: Same, I believe is still needed.  Any specific reason you see for removing?

jcantrill: I had issues deploying the operator and this resolved it.  I don't see why the operator needs root access to its own file system which is what this says.  I'm open to discussion if you think otherwise.

bundle/manifests/clusterlogging.clusterserviceversion.yaml
jcantrill: Our custom SCC applies to the collector because those pods mount the host file system of the node.  This change is directed towards the operator.  Why does it need to run as root?

PR #1735, opened 2022-11-01
LOG-3194: Add pod security sync label

cahartma:
### Description
For some reason the olm operator has begun adding the label 'podSecurityLabelSync' = true to the openshift-logging namespace.   This results in our pod security labels being overwritten.  These new labels prevent our pods from starting.
This change adds the "false" sync label so that our privileged policy is maintained.  

Note:  This should not need to be done according to the IMPORTANT section in our docs, as well as several internal resources

/cc @syedriko @vimalk78 
/assign @jcantrill 

### Links
- JIRA:   https://issues.redhat.com/browse/LOG-3194
- OLM Controller change: https://issues.redhat.com//browse/OLM-2695
- Docs:  https://docs.openshift.com/container-platform/4.11/authentication/understanding-and-managing-pod-security-admission.html


jcantrill (commented):
/lgtm

PR #1751, opened 2022-11-10
LOG-3242: remove loki tag label from fluentd

cahartma:
### Description
'tag' is un-documented as a label key, nor is it part of viaq data model   It needs to be removed from Loki output.  Our Loki output needs to be aligned with our ES logstore output, which does not contain 'tag'. 

/cc @syedriko @vimalk78 
/assign  @jcantrill 

### Links
- JIRA task: https://issues.redhat.com/browse/LOG-3242
- Loki labelKeys: https://github.com/openshift/cluster-logging-operator/blob/master/apis/logging/v1/output_types.go#L233
- Viaq data model:  https://viaq.github.io/documentation/data_model/public/data_model.html

### DIFF
--- a/internal/generator/fluentd/output/loki/loki.go
+++ b/internal/generator/fluentd/output/loki/loki.go
@@ -21,7 +21,6 @@ import (
 
 const (
 	lokiLabelKubernetesHost = "kubernetes.host"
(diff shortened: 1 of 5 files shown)


jcantrill (commented):
/approve

PR #1754, opened 2022-11-14
LOG-1796:Add multi-line exception detection for Vector collector

vparfonov:

### Description

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-1796
- Enhancement proposal:

### DIFF
--- a/test/functional/normalization/multiline_exception_test.go
+++ b/test/functional/normalization/multiline_exception_test.go
@@ -1,6 +1,3 @@
-//go:build fluentd
-// +build fluentd
-
 package normalization
 
 import (
@@ -18,6 +15,7 @@ import (
 // Multiline Detect Exception test to verify proper re-assembly of
 // multi-line exceptions (e.g. java stacktrace)
 // https://issues.redhat.com/browse/LOG-1717
+// https://issues.redhat.com/browse/LOG-1796
 var _ = Describe("[Functional][Normalization] Multi-line exception detection", func() {
 	const (
 		timestamp = "2021-03-31T12:59:28.573159188+00:00"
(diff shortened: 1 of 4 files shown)


test/functional/normalization/multiline_exception_vector_test.go
```diff
@@ -0,0 +1,186 @@
+//go:build vector
```
jcantrill, line 1: It should be possible to modify and add to the existing test to support vector: https://github.com/openshift/cluster-logging-operator/blob/master/test/functional/normalization/multiline_exception_test.go#L1

Suggestions:
* Modify to use Elasticsearch as the output
* Remove tag hints at the op of the file

I'm otherwise impressed that vector is capable of supporting our needs without writing our own rust code

cc @alanconway

vparfonov: > It should be possible to modify and add to the existing test to support vector: [`master`/test/functional/normalization/multiline_exception_test.go#L1](https://github.com/openshift/cluster-logging-operator/blob/master/test/functional/normalization/multiline_exception_test.go?rgh-link-date=2022-11-15T19%3A03%3A31Z#L1)

I tried that but not work for me, got some problem with images. I will take a look one more time 

> 
> Suggestions:
> 
> * Modify to use Elasticsearch as the output
Maybe it was the problem because it uses fluentd as the output 

> * Remove tag hints at the op of the file
> 
> I'm otherwise impressed that vector is capable of supporting our needs without writing our own rust code
> 
> cc @alanconway

Makefile
```diff
@@ -47,7 +47,7 @@ export VERSION=$(LOGGING_VERSION).0
 export NAMESPACE?=openshift-logging
 
 IMAGE_LOGGING_FLUENTD?=quay.io/openshift-logging/fluentd:1.14.6
-IMAGE_LOGGING_VECTOR?=quay.io/openshift-logging/vector:0.21-rh
+IMAGE_LOGGING_VECTOR?=quay.io/vparfono/vector-0.21-reduce:latest
```
jcantrill, line 50: lets correct his now https://github.com/ViaQ/vector/pull/128 merged

olm_deploy/scripts/env.sh
```diff
@@ -10,7 +10,7 @@ LOGGING_IS=${LOGGING_IS:-openshift-logging}
 export IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY=${IMAGE_CLUSTER_LOGGING_OPERATOR_REGISTRY:-quay.io/${LOGGING_IS}/cluster-logging-operator-registry:${LOGGING_VERSION}}
 export IMAGE_CLUSTER_LOGGING_OPERATOR=${IMAGE_CLUSTER_LOGGING_OPERATOR:-quay.io/${LOGGING_IS}/cluster-logging-operator:${LOGGING_VERSION}}
 export IMAGE_LOGGING_FLUENTD=${IMAGE_LOGGING_FLUENTD:-quay.io/${LOGGING_IS}/fluentd:${LOGGING_FLUENTD_VERSION}}
-export IMAGE_LOGGING_VECTOR=${IMAGE_LOGGING_VECTOR:-quay.io/${LOGGING_IS}/vector:${LOGGING_VECTOR_VERSION}}
+export IMAGE_LOGGING_VECTOR=quay.io/vparfono/vector-0.21-reduce:latest
```
jcantrill, line 13: I assume this is a revert

test/functional/normalization/multiline_exception_vector_test.go
jcantrill: @vparfonov you will need to modify the test to read and write to elasticsearch since that is the only common output

jcantrill (changes requested):

test/functional/normalization/multiline_exception_vector_test.go
vparfonov: > @vparfonov you will need to modify the test to read and write to elasticsearch since that is the only common output

sorry, but i don't follow you

jcantrill: We should be able to use the existing test for both vector and fluentd.  It requires, however, to change the output type to Elasticsearch since this is the only one which is common to both which we can control in the functional framework

test/functional/normalization/multiline_exception_test.go
```diff
@@ -85,8 +159,12 @@ created by main.main
 			crioLine := functional.NewCRIOLogMessage(timestamp, line, false)
 			buffer = append(buffer, crioLine)
 		}
-		// Application log in namespace
-		Expect(framework.WriteMessagesToNamespace(strings.Join(buffer, "\n"), appNamespace, 1)).To(Succeed())
+
+		if testfw.LogCollectionType == logging.LogCollectionTypeVector {
+			Expect(framework.WriteMessagesToApplicationLog(strings.Join(buffer, "\n"), 1)).To(Succeed())
+		} else {
+			Expect(framework.WriteMessagesToNamespace(strings.Join(buffer, "\n"), appNamespace, 1)).To(Succeed())
```
jcantrill, line 166: This works for vector as long as it's the same ns as the test pod. And then maybe you could enable the other tests

jcantrill (approved):
/approve

test/functional/normalization/multiline_exception_test.go
vparfonov: > This works for vector as long as it's the same ns as the test pod. And then maybe you could enable the other tests

I checked, these tests not work for vector without modification related to the namespace name

PR #1757, opened 2022-11-17
LOG-3280: exclude log generated by loki

vparfonov:

### Description
This PR adds to exclude all logs produced by `loki*` containers

/cc 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-3280
- Enhancement proposal:


internal/generator/vector/helpers/helpers.go
```diff
@@ -30,3 +30,15 @@ func TrimSpaces(in []string) []string {
 func FormatComponentID(name string) string {
 	return strings.ToLower(Replacer.Replace(name))
 }
+
+func Unique(stringSlice []string) []string {
```
jcantrill, line 34: There is kube struct called "set" which can replace this.  I believe I saw its use in our api package for Reserved names

internal/generator/vector/sources.go
```diff
@@ -33,14 +34,18 @@ func Sources(spec *logging.ClusterLogForwarderSpec, namespace string, op generat
 }
 
 func LogSources(spec *logging.ClusterLogForwarderSpec, namespace string, op generator.Options) []generator.Element {
-	var el []generator.Element = make([]generator.Element, 0)
+	var el = make([]generator.Element, 0)
 	types := generator.GatherSources(spec, op)
+	var outputTypes []string
+	for _, output := range spec.Outputs {
+		outputTypes = append(outputTypes, output.Type)
```
jcantrill, line 41: Should this be name instead of type?  The exclusion method compares well known name to build the paths

jcantrill (changes requested):
These changes will need to also be applied to fluent since we still support fluent -> loki

internal/generator/vector/sources.go
vparfonov: name can be any string, but type is described and controlled by API
https://github.com/openshift/cluster-logging-operator/blob/master/apis/logging/v1/cluster_log_forwarder_types.go#L151

jcantrill: these are well known names by us; they will not change.  Loki, is the same unless we set something specific in cl.logStore

vparfonov: maybe we talk about different things by name i mean this  https://github.com/openshift/cluster-logging-operator/blob/master/internal/generator/vector/conf_test.go#L435

jcantrill: This confirms what I was asking: https://github.com/openshift/cluster-logging-operator/pull/1757/files#diff-e72ab332759bfa5c65f11425dfa95625f13ec0b617f5c4d3c23151c6ea0445f0R110-R113  It just so happens we chose an output type to be the same as the name (e.g. "Elasticsearch"), but we need to exclude by name an not type.  Instead of conditionally adding Loki, I would just add it to the list along with the others.  I do not think users can deploy Loki to openshift-logging and use it as the default using another name.  I'll ask @xperimental or @periklis to comment

jcantrill: @vparfonov conversation with @periklis suggested excluding all of `openshift-logging`.  This would simplify the change.  Need to be aware of we want to include the eventrouter so we will need to make sure we get the include/exclude correctly

internal/generator/fluentd/sources.go
```diff
@@ -100,9 +105,13 @@ func ContainerLogPaths() string {
...
-	for _, comp := range []string{constants.CollectorName, constants.ElasticsearchName, constants.KibanaName} {
+func ExcludeContainerPaths(namespace string, outputs []string) string {
+	var paths []string
+	toExclude := []string{constants.CollectorName, constants.ElasticsearchName, constants.KibanaName}
+	if slices.Contains(outputs, constants.LokiName) {
+		toExclude = append(toExclude, constants.LokiName)
+	}
+	for _, comp := range toExclude {
 		paths = append(paths, fmt.Sprintf("\"/var/log/pods/%s_%s-*/*/*.log\"", namespace, comp))
 	}
```
vparfonov, line 116: @jcantrill you mean something like this 
```suggestion
	paths = append(paths, fmt.Sprintf("\"/var/log/pods/%s_*/*/*.log\"", namespace))
```
exclude all logs from >` /var/log/pods/openshift-logging_*/*/*.log`?

jcantrill: yes, but we need to ensure we can still collect for the eventrouter.  It may be we simply have to add a dedicated in_tail config for the event router unless we can say "exclude all of openshift-logging except the eventrouter"

internal/generator/vector/sources.go
jcantrill: @vimalk78 does vector do exclusions by pod label? instead of path?

vparfonov: @jcantrill https://vector.dev/docs/reference/configuration/sources/kubernetes_logs/#container-exclusion

internal/generator/vector/conf_test.go
```diff
@@ -85,18 +85,25 @@ var _ = Describe("Testing Complete Config Generation", func() {
...
+exclude_paths_glob_patterns = ["/var/log/pods/openshift-logging_*/*/*.log", "/var/log/pods/*/*/*.gz", "/var/log/pods/*/*/*.tmp"]
 pod_annotation_fields.pod_labels = "kubernetes.labels"
 pod_annotation_fields.pod_namespace = "kubernetes.namespace_name"
 pod_annotation_fields.pod_annotations = "kubernetes.annotations"
 pod_annotation_fields.pod_uid = "kubernetes.pod_id"
 pod_annotation_fields.pod_node_name = "hostname"
 
+# Logs from Event Router
+[sources.raw_eventrouter_log]
+type = "file"
```
jcantrill, line 102: It a pod so it should still be kubernetes_logs

internal/generator/fluentd/conf_test.go
```diff
@@ -143,12 +143,12 @@ var _ = Describe("Testing Complete Config Generation", func() {
...
 </source>
 
-# Logs from containers (including openshift containers)
+# Logs from containers 
 <source>
   @type tail
   @id container-input
   path "/var/log/pods/*/*/*.log"
-  exclude_path ["/var/log/pods/openshift-logging_collector-*/*/*.log", "/var/log/pods/openshift-logging_elasticsearch-*/*/*.log", "/var/log/pods/openshift-logging_kibana-*/*/*.log", "/var/log/pods/*/*/*.gz", "/var/log/pods/*/*/*.tmp"]
+  exclude_path ["/var/log/pods/openshift-logging_*/*/*.log", "/var/log/pods/*/*/*.gz", "/var/log/pods/*/*/*.tmp"]
```
jcantrill, line 151: @vparfonov Rethinking the implementation since this might simplify this change .  The segment  before '*.log' is the container name.  Maybe we should just exclude any containers that match what they call the container in the deployment instead of trying to alter how we collect logs for the eventrouter

jcantrill (changes requested):

internal/generator/vector/conf_test.go
vparfonov: Above we excluded all logs from openshift-logging namespace and kubernetes_log don't have param for path so we can't use it here.

internal/generator/fluentd/conf_test.go
vparfonov: but it will be almost the same that we had before 
```suggestion
  exclude_path ["/var/log/pods/openshift-logging_*/collector/*.log", /var/log/pods/openshift-logging_*/elasticsearch /*.log", /var/log/pods/openshift-logging_*/kibana/*.log", /var/log/pods/openshift-logging_*/loki/*.log", "/var/log/pods/*/*/*.gz", "/var/log/pods/*/*/*.tmp"]
```

vparfonov: @jcantrill ˆˆˆ

jcantrill: correct it is almost the same.  Per my discussion with @periklis  please make the container segment "loki*"

PR #1759, opened 2022-11-18
LOG-3235: fix collector security context spec

jcantrill:
### Description
This PR:
* sets the log collector security context spec so the logging SCC is more likely to apply versus others
* adds reconcile logic for the collector SCC other then just "create"
* Add SCC comparator and tests

### Links
* https://issues.redhat.com/browse/LOG-3235


alanconway (approved):
/lgtm

alanconway (approved):
/lgtm

PR #1760, opened 2022-11-18
LOG-3302: fix collector security context spec

jcantrill:
### Description
This PR:
* sets the log collector security context spec so the logging SCC is more likely to apply versus others
* adds reconcile logic for the collector SCC other then just "create"
* Add SCC comparator and tests

### Links
* https://issues.redhat.com/browse/LOG-3302
* forward port of https://github.com/openshift/cluster-logging-operator/pull/1759

### DIFF
--- a/internal/utils/comparators/scc/comparator.go
+++ b/internal/utils/comparators/scc/comparator.go
@@ -10,6 +10,10 @@ import (
 func AreSame(current, desired security.SecurityContextConstraints) (bool, string) {
 	log.V(3).Info("Comparing SCC current to desired", "current", current, "desired", desired)
 
+	if same, property := samePriority(current.Priority, desired.Priority); !same {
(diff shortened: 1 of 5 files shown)


internal/utils/comparators/scc/comparator.go
```diff
@@ -80,3 +78,32 @@ func AreSame(current, desired security.SecurityContextConstraints) (bool, string
...
+
+func sameAllowPrivilegeEscalation(current, desired *bool) bool {
+	if (current != nil && desired == nil) ||
+		(current == nil && desired != nil) ||
+		(current != nil && desired != nil &&
+			*current != *desired) {
+		return false
+	}
+
+	if (current != nil && desired == nil) ||
```
vparfonov, line 102: copy-past error

PR #1769, opened 2022-12-01
fix e2e previous where pod wont start

jcantrill:
### Description
This PR:
* fixes e2e previous by using a common way to create the log generator pod


cahartma (commented):
/lgtm

PR #1782, opened 2022-12-08
LOG-3331: Disable log view plugin if cluster does not have console en…

jcantrill:
…abled

### Description
This PR:
* skips log view plugin deployment if the cluster does not have the console enabled

### Links
* https://issues.redhat.com/browse/LOG-3321

cc @periklis 

@anpingli please test this PR directly.  There is no way to mock the client to throw errors so I am unable to write a proper test

PR #1787, opened 2022-12-09
LOG-3407: Disable log view plugin if cluster does not have console en…

jcantrill:
…abled

### Description
This PR:
* skips log view plugin deployment if the cluster does not have the console enabled

### Links
* https://issues.redhat.com/browse/LOG-3407
* Backport of  https://github.com/openshift/cluster-logging-operator/pull/1782


alanconway (approved):
/lgtm

Do we have a watch so we will reconcile the console if it is re-enabled later?

PR #1794, opened 2022-12-15
LOG-3321: fix crash when console is not enabled

jcantrill:
### Description
This PR:
* fixes the crash by adding the console plugin to the watch for CL
* sets a sync period to 3 minutes which is the maximum time to notice changes in the console plugin

### Links
* https://issues.redhat.com/browse/LOG-3321

cc @alanconway


cahartma (commented):
/lgtm

PR #1798, opened 2022-12-16
LOG-3407: fix crash when console is not enabled

jcantrill:
### Description
This PR:

fixes the crash by adding the console plugin to the watch for CL
sets a sync period to 3 minutes which is the maximum time to notice changes in the console plugin

### Links
* https://issues.redhat.com/browse/LOG-3407
* backport of https://github.com/openshift/cluster-logging-operator/pull/1794

cc @alanconway

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,7 +147,7 @@ lint-dockerfile:
 .PHONY: fmt
 fmt:
 	@echo gofmt		# Show progress, real gofmt line is too long
-	find version test internal controllers apis -name '*.go' | xargs gofmt -s -l -w
+	find . version test internal controllers apis -name '*.go' | xargs gofmt -s -l -w
 
 # Do all code/CRD generation at once, with timestamp file to check out-of-date.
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -147,7 +147,7 @@ lint-dockerfile:
 .PHONY: fmt
 fmt:
 	@echo gofmt		# Show progress, real gofmt line is too long
-	find version test internal controllers apis -name '*.go' | xargs gofmt -s -l -w
+	find . version test internal controllers apis -name '*.go' | xargs gofmt -s -l -w
```
vimalk78, line 150: nit: if we have `.` why need other folder names?

vparfonov, line 150:
```suggestion
	find .  -name '*.go' | xargs gofmt -s -l -w
```

jcantrill: i think we only have folders to begin with because of maybe when we had the vendor dir?

PR #1799, opened 2022-12-16
LOG-3341: fix namespace_labels when writing to Elasticsearch

jcantrill:
### Description
This PR:
* replaces dots with underscores for namespace_labels when writing to Elasticsearch

### Links
* https://issues.redhat.com/browse/LOG-3341


alanconway (approved):
/lgtm
I think is a reasonable fix. We only stopped de-dotting the standard k8s labels because of an explicit request, and we still de-dot other labels. Reverting to de-dotting for namespace labels seems a safe fix.

PR #1800, opened 2022-12-20
LOG-3440: Update metrics endpoint for vector to use IPV6

vimalk78:
### Description
Updated vector conf generator to use `[::]:24231` as the metrics endpoint.
This address works for both IPV4 and IPV6 clusters.

/cc 
/assign 

/cherry-pick 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-3440

### DIFF
--- a/internal/generator/vector/conf_test.go
+++ b/internal/generator/vector/conf_test.go
@@ -2,9 +2,10 @@ package vector
 
 import (
 	"fmt"
+	"strings"
+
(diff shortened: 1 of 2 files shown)


jcantrill (commented):
/lgtm

jcantrill:
May be blocked by https://github.com/openshift/cluster-logging-operator/pull/1769

jcantrill:
override because we need this in 5.6 and the "e2e minus one succeeds"

PR #1801, opened 2022-12-20
LOG-3342: fix validation on empty spec

jcantrill:
### Description
This PR:
* fixes a merge issue where we always validate against an empty spec which causes logging to never deploy

### Links
* https://issues.redhat.com/browse/LOG-3342


jcantrill:
@anpingli this is the PR that once merged would allow release of 5.5

PR #1806, opened 2023-01-03
LOG-3302: Update scc to desired

jcantrill:
### Description
This PR:
* fixes updating by Updating the current SCC to desired

### Links
* https://issues.redhat.com/browse/LOG-3302

### DIFF
--- a/internal/reconcile/scc.go
+++ b/internal/reconcile/scc.go
@@ -14,7 +14,7 @@ import (
 func SecurityContextConstraints(k8Client client.Client, desired *security.SecurityContextConstraints) error {
 	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
 		current := &security.SecurityContextConstraints{}
-		key := client.ObjectKeyFromObject(desired)
+		key := client.ObjectKey{Name: desired.Name}
 		if err := k8Client.Get(context.TODO(), key, current); err != nil {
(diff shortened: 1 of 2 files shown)


internal/reconcile/scc.go
```diff
@@ -27,7 +27,37 @@ func SecurityContextConstraints(k8Client client.Client, desired *security.Securi
...
+	to.RunAsUser = from.RunAsUser
+	to.SupplementalGroups = from.SupplementalGroups
+	to.FSGroup = from.FSGroup
+	to.ReadOnlyRootFilesystem = from.ReadOnlyRootFilesystem
+	to.Users = from.Users
+	to.Groups = from.Groups
+	to.SeccompProfiles = from.SeccompProfiles
+	to.AllowedUnsafeSysctls = from.AllowedUnsafeSysctls
+	to.ForbiddenSysctls = from.ForbiddenSysctls
+
```
vimalk78, line 61: can we replace this block with `*to=*from` if all fields are to be copied?

jcantrill: SCCs don't have a "spec" field, they live off the root of the type. The only change we possibly could make is to use reflection over the fields

vimalk78: agree. a `*to=*from` would copy more fields than we need.

PR #1819, opened 2023-01-05
LOG-3447: fix logging-view-plugin deployment to enable features based…

jcantrill:
… upon OCP version

### Description
This PR:
* Enables logging-view-plugin features based upon the version of OCP

### Links
* https://issues.redhat.com/browse/LOG-3447

cc @alanconway @jgbernalp


internal/consoleplugin/reconciler.go
```diff
@@ -229,6 +230,8 @@ func (r *Reconciler) mutateDeployment() error {
 								Protocol:      "TCP",
 							},
 						},
+						Command: []string{"plugin-backend"},
+						Args:    []string{"-port", "9002", "-features", strings.Join(r.Config.Features, ",")},
```
jgbernalp, line 234: Nginx was replaced with the go backend. So here we need to add the certificate and key args that the new backend supports: https://github.com/openshift/logging-view-plugin/pull/55/commits/a39036acf1dad1cd0ec28bdd6ef055697db5b1db#diff-4d25889b839b563a7b3536730d4aa8bf08b34a5fa25702dc3bfcd9dbfdca444cR11, also the port should be `9443`

internal/consoleplugin/config.go
```diff
@@ -24,12 +25,13 @@ func (cf *Config) defaultMode() *int32 { return utils.GetInt32(420) }
 func (cf *Config) nginxPort() int32    { return 9443 }
```
jgbernalp, line 25:
```suggestion
func (cf *Config) pluginBackendPort() int32    { return 9443 }
```

alanconway (approved):
Looks good to me, will let @jcantrill give final /lgtm when he removes the /hold.

internal/consoleplugin/reconciler.go
```diff
@@ -229,6 +230,13 @@ func (r *Reconciler) mutateDeployment() error {
...
 							},
 						},
+						Command: []string{"plugin-backend"},
+						Args: []string{
+							"-port", "9443",
+							"-features", strings.Join(r.Config.Features, ","),
+							"-cert", "/var/serving-cert/tls.crt",
+							"-key", "/var/serving-cert/tls.key",
+						},
 						VolumeMounts: []corev1.VolumeMount{
```
jgbernalp, line 240: We probably don't need the nginx-config anymore

PR #1823, opened 2023-01-09
LOG-3477: fix degraded pipeline status

jcantrill:
### Description
This PR:
* fixes the degraded pipeline status when adding default to an existing CLF pipeline

### Links
* https://issues.redhat.com/browse/LOG-3477


alanconway (approved):
/lgtm

PR #1825, opened 2023-01-09
LOG-3474:  fix custom input in clf spec when using loki

cahartma:
### Description
This is the 5.5 fix.  Same general idea as the 5.6 fix: [LOG-3269](https://issues.redhat.com//browse/LOG-3269)
- Adds a name to custom inputs which will now prevent the invalid pipeline error.  
- Adds loki function for getting input type from name, in order for loki labels to work properly.

/cc @vimalk78 @syedriko 
/assign @jcantrill 

### Links
- https://issues.redhat.com/browse/LOG-3474


jcantrill (commented):
/approve
/lgtm

PR #1831, opened 2023-01-18
Addressed race condition during cert generation, fixed loki uid for vector CI

syedriko:
### Description
Two corrections for the functional and e2e tests:
- fixed a race condition during generation of certificates due to multiple instances of the certificate generation script running against the same directory
- Added a uid for the loki user to the security context of the loki receiver, this fixes the "container has runAsNonRoot and image has non-numeric user (loki), cannot verify user is non-root" issue happening in vector CI.

/cc @jcantrill 
/assign @cahartma

### DIFF
--- a/test/framework/e2e/framework.go
+++ b/test/framework/e2e/framework.go
@@ -4,14 +4,15 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"github.com/openshift/cluster-logging-operator/test/framework/e2e/certificates"
-	"github.com/openshift/cluster-logging-operator/test/runtime"
 	"math/rand"
 	"os"
 	"os/exec"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/openshift/cluster-logging-operator/test/framework/e2e/certificates"
+	"github.com/openshift/cluster-logging-operator/test/runtime"
 
 	"github.com/openshift/cluster-logging-operator/internal/constants"
 	"github.com/openshift/cluster-logging-operator/test/helpers"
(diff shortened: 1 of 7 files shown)


test/framework/e2e/framework.go
```diff
@@ -638,26 +639,19 @@ func (tc *E2ETestFramework) PodExec(namespace, name, container string, command [
...
-		"tls.key":       utils.GetWorkingDirFileContents("system.logging.fluentd.key"),
-		"tls.crt":       utils.GetWorkingDirFileContents("system.logging.fluentd.crt"),
-		"ca-bundle.crt": utils.GetWorkingDirFileContents("ca.crt"),
-		"ca.key":        utils.GetWorkingDirFileContents("ca.key"),
+		"tls.key":       utils.GetDirFileContents(certsDir, "system.logging.fluentd.key"),
+		"tls.crt":       utils.GetDirFileContents(certsDir, "system.logging.fluentd.crt"),
+		"ca-bundle.crt": utils.GetDirFileContents(certsDir, "ca.crt"),
+		"ca.key":        utils.GetDirFileContents(certsDir, "ca.key"),
 	}
+	os.RemoveAll(certsDir)
```
jcantrill, line 654: maybe this should be a 'defer' placed higher to ensure it always runs

test/framework/functional/framework.go
```diff
@@ -198,17 +200,19 @@ func (f *CollectorFunctionalFramework) DeployWithVisitors(visitors []runtime.Pod
 	}
 
 	log.V(2).Info("Generating Certificates")
+	certsDir := fmt.Sprintf("/tmp/clo-test-%d-%d", os.Getpid(), rand.Intn(10000)) //nolint:gosec
```
jcantrill, line 203: This code looks duplicated.  Any reason we can not collect into a common package under test/framework

jcantrill (changes requested):
/approve

test/framework/functional/framework.go
syedriko: @jcantrill Please take a look at the latest version

jcantrill (commented):
/hold cancel
/lgtm

PR #1832, opened 2023-01-18
LOG-1413: `http_proxy`, `https_proxy` and `no_proxy` variables keys in lowercase

gmeghnag:
Modified the `GetProxyEnvVars()` function to return an array of `v1.EnvVar` where `http_proxy`, `https_proxy` and `no_proxy` environment variables key are in lowercase.

### Description

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-1413
- Enhancement proposal:

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -469,7 +469,7 @@ func GetProxyEnvVars() []v1.EnvVar {
 				}
 			}
 			envVars = append(envVars, v1.EnvVar{
-				Name:  envvar,
+				Name:  strings.ToLower(envvar),
 				Value: value,
 			})
 		}


gmeghnag:
## Summary
This PR aims to fix the following warning that pop-up when the logs are sent to an external log aggregator through a proxy:
```
$ oc logs pod/collector-2hb7s -c collector --tail 1
/usr/share/ruby/net/http.rb:1122: warning: The environment variable HTTP_PROXY is discouraged.  Use http_proxy.
```

## Where does fluentd get the `HTTP_PROXY` variable from?
Fluentd take the `HTTP_PROXY`, `HTTPS_PROXY` and  `NO_PROXY` variables from the container exported variables, and these variables are injected into the `collector` pod definition by the `cluster-logging-operator`.

https://github.com/openshift/cluster-logging-operator/blob/2a03dabba673e1bfd851961f57c35edf93ffd898/internal/collector/collector.go#L189

## Where does `cluster-logging-operator` get the `HTTP_PROXY`, `HTTPS_PROXY` and  `NO_PROXY` variable from?
It takes these variables from `os.Getenv()`:

https://github.com/openshift/cluster-logging-operator/blob/2a03dabba673e1bfd851961f57c35edf93ffd898/internal/utils/utils.go#L462-L478

alanconway (approved):
/lgtm

internal/utils/utils.go
```diff
@@ -469,7 +469,7 @@ func GetProxyEnvVars() []v1.EnvVar {
 				}
 			}
 			envVars = append(envVars, v1.EnvVar{
-				Name:  envvar,
+				Name:  strings.ToLower(envvar),
```
jcantrill, line 472: This implies we are now setting the same env var twice which is unnecessary.  These vars come from the global proxy adding them to the operator.  Do we know if they ALWAYS populate the lower case version?  Is it possible to simply remove the uppercase version from the list we check?

jcantrill (commented):
/ok-to-test

internal/utils/utils.go
gmeghnag: The `cluster-logging-operator` container has these variables exported in uppercase only, so won't be any duplication:
```
$ oc get proxy/cluster -o yaml | yq '.spec'
httpProxy: http://username:password@X.X.X.X:8080
httpsProxy: http://username:password@X.X.X.X:8080
trustedCA:
  name: ""

$ oc get po -n openshift-logging                                                                            
NAME                                        READY   STATUS    RESTARTS   AGE
cluster-logging-operator-56b476f846-qxxzj   1/1     Running   0          47s

$ oc -n openshift-logging exec -ti cluster-logging-operator-56b476f846-qxxzj -- sh -c 'env | egrep -i proxy'
HTTP_PROXY=http://username:password@X.X.X.X:8080
NO_PROXY=.cluster.local,.eu-central-1.compute.internal,.svc,10.0.0.0/16,10.128.0.0/14,127.0.0.1,169.254.169.254,172.30.0.0/16,api-int.gmeghnag.emeashift.support,localhost
HTTPS_PROXY=http://username:password@X.X.X.X:8080

$ oc get csv -n openshift-logging
NAME                                     DISPLAY                            VERSION   REPLACES                                 PHASE
cluster-logging.5.5.5                    Red Hat OpenShift Logging          5.5.5                                              Succeeded
elasticsearch-operator.5.5.5             OpenShift Elasticsearch Operator   5.5.5                                              Succeeded
```

jcantrill: This still may not be the correct solution given this function is likely also used to deploy the vector container as the collector.  Does vector honor these lower case vars?  Does it expect uppercase?  There is more work to be done to ensure the functionality remains with no regression

gmeghnag: Right points to consider, I will inquire about them:
- How does [vector](https://github.com/vectordotdev/vector) get the proxy URL to use?
- If it takes the proxy URL from the environment variables does it support these variables to be in lowercase?

jcantrill: @gmeghnag are these questions directed at me because they read like the questions I asked of you

gmeghnag: @jcantrill No, the questions are just notes to remind me what information I need to get, I will try by the end of the week to get information on that.

gmeghnag: from [vector](https://github.com/vectordotdev/vector) issue [#6401](https://github.com/vectordotdev/vector/issues/6401#:~:text=Additionally%20we%20should%20read%20from%20the%20following%20env%20vars:), pull [#8205](https://github.com/vectordotdev/vector/pull/8205#discussion_r678568354) both the lowercase and uppercase form are supported.

jcantrill: @gmeghnag looks like we need a fix of unit tests and then we are ready to go

gmeghnag: Opened https://github.com/openshift/cluster-logging-operator/pull/1875 to fix unit tests.

jcantrill:
> @jcantrill The PR is updated with the modification made to `internal/collector/collector_test.go`, is there something more that we need to do before rerunning the tests?

test are run automatically by ci.  You can test yourself locally via `make test-unit`

PR #1841, opened 2023-01-27
LOG-3239: Add e2e test for vector http

vimalk78:
### Description
This PR adds a `e2e` test which sends logs from vector to a fluentd endpoint using http.

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:


alanconway (approved):
/lgtm
is there a vector e2e test?

internal/generator/vector/output/http/http.go
```diff
@@ -184,14 +185,19 @@ func Headers(o logging.OutputSpec) Element {
...
+		sortedKeys = append(sortedKeys, k)
+	}
+	sort.Strings(sortedKeys)
 	str := ""
 	skipComma := true
-	for k, v := range h {
+	for _, k := range sortedKeys {
 		if skipComma {
-			str += fmt.Sprintf("%q=%q", k, v)
+			str += fmt.Sprintf("%q=%q", k, h[k])
```
jcantrill, line 197: What about using `strings.Join` in multiple iterations so there is no need to "skip" the comma or not?

jcantrill (changes requested):
one nit but looks good

/hold

internal/generator/vector/output/http/http.go
vimalk78: updated.

PR #1848, opened 2023-02-07
LOG-3609: Fix collector restart after creation

vimalk78:
### Description
  Collector gets restarted if collector is started without waiting for waiting for trusted ca-bundle config map to get created.
  Modified the reconcile logic to wait for trusted ca-bundle to get created.

  Log Collector SCC desires 4 volumes `"configMap"`,`"emptyDir"`,`"projected"`,`"secret"`, but CRC container adds another volume for `"hostpath"`.
  The SCC comparator does a reflect comparison, which fails because there is one extra Volume.
  Fixed the SCC comparator to check for desired volumes only.

/cc 
/assign 

/cherry-pick 

### Links
- JIRA:  https://issues.redhat.com/browse/LOG-3609

### DIFF
--- a/internal/collector/service_account.go
+++ b/internal/collector/service_account.go
@@ -3,6 +3,7 @@ package collector
 import (
 	"context"
 	"fmt"
+
 	security "github.com/openshift/api/security/v1"
 	"github.com/openshift/cluster-logging-operator/internal/constants"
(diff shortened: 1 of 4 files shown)


jcantrill (commented):
/lgtm

jcantrill (commented):
/lgtm

PR #1849, opened 2023-02-07
[release-5.6] LOG-3437: fix for invalidate or migrate default output

cahartma:
### Description
The fix was created in migration and with minor updates to the logic surrounding 'default' outputRef for both ES and Loki.   Mainly adding a map to the migration and subsequent normalization (validate), indicating if the ES output was created by our default migration.  This allows us to handle verify with the context of migration.  This is the initial and least impactful change to a refactor of the migration and validation logic to come in: [LOG-3247](https://issues.redhat.com/browse/LOG-3247) when we remove degraded status.

Expected:  
If you use 'default' as an output, it will either correct the ES output, or if its Loki, it will inform the user the name is reserved.  Any other output named 'default' should now invalidate as expected. 
This should also fix the issue in [LOG-3559](https://issues.redhat.com//browse/LOG-3559) where collector restarts due to invalid pipelines when using default along with another outputRef (in same or separate pipeline) 

/cc @jcantrill
/assign @vimalk78 @syedriko

### Links
- https://issues.redhat.com/browse/LOG-3437
- https://issues.redhat.com/browse/LOG-3559
- (in progress) https://issues.redhat.com/browse/LOG-3247

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -138,6 +138,8 @@ const (
 	EventReasonGetObject            = "GetObject"
 	EventReasonRemoveObject         = "RemoveObject"
 	EventReasonUpdateObject         = "UpdateObject"
+
+	MigrateDefaultOutput = "default-replaced"
 )
 
 var ReconcileForGlobalProxyList = []string{CollectorTrustedCAName}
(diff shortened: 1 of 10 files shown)


jcantrill:
@cahartma looks good to me other then I have questions why there are 2 commits instead of one.  Please squash unless there is a valid reason to identify 2

jcantrill (commented):
/approve
/lgtm

PR #1850, opened 2023-02-08
LOG-3429: fix vector evaluation of journal priority to set the level

jcantrill:
### Description
This PR:
* fixes normalization of the log level for journal entries
* adds functional testing to verify log level from the journal priority
* add functionality to mock journal logs in the functional framework

### Links
* https://issues.redhat.com/browse/LOG-3429


test/framework/functional/framework.go
```diff
@@ -181,10 +181,8 @@ func (f *CollectorFunctionalFramework) DeployWithVisitors(visitors []runtime.Pod
 		if f.Conf, err = forwarder.Generate(logging.LogCollectionType(f.collector.String()), string(clfYaml), false, debugOutput, testClient); err != nil {
 			return err
 		}
-		//remove journal for now since we can not mimic it
-		re := regexp.MustCompile(`(?msU).*<source>.*type.*systemd.*</source>`)
-		f.Conf = string(re.ReplaceAll([]byte(f.Conf), []byte{}))
-
+		//mock sources to facilitate functional testing
+		f.Conf = f.collector.ModifyConfig(f.Conf)
```
vimalk78, line 185: does this mean every functional test will replace journal logs with fake journal logs?

jcantrill: Yes. This is a hook to allow the framework to modify a generated configuration.  This allows us to replace journal with something we can control as part of the framework.

test/functional/normalization/loglevel/journal_logs_test.go
```diff
@@ -0,0 +1,62 @@
...
+		framework *functional.CollectorFunctionalFramework
+	)
+
+	BeforeEach(func() {
+		framework = functional.NewCollectorFunctionalFrameworkUsingCollector(testfw.LogCollectionType)
+		functional.NewClusterLogForwarderBuilder(framework.Forwarder).
+			FromInput(logging.InputNameInfrastructure).
+			ToFluentForwardOutput().
+			FromInput(logging.InputNameAudit).
+			ToElasticSearchOutput()
```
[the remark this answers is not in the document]
jcantrill: This was added as part of another test to ensure we constructed pipelines correctly.  It may not be required here but it does offer a way to ensure we get configs "correct" when multiplexing across multiple pipelines

PR #1854, opened 2023-02-08
Extending to gather Lokistack instance CR

saurabhsadhale:
### Description
This PR will extend the current capabilities of the must-gather and capture LokiStack instance CR only if present. 

The current must-gather captures Clusterlogging and ClusterLogforwarder instances but not the LokiStack.

### DIFF
--- a/must-gather/collection-scripts/gather_logstore_resources
+++ b/must-gather/collection-scripts/gather_logstore_resources
@@ -17,6 +17,8 @@ NAMESPACE=${2:-openshift-logging}
 CLO_COLLECTION_PATH="$BASE_COLLECTION_PATH/cluster-logging"
 es_folder="$CLO_COLLECTION_PATH/es"
 
+loki_folder="$CLO_COLLECTION_PATH/loki"
+
 list_es_storage() {
   local pod=$1
   local mountPath=$(oc -n $NAMESPACE get pod $pod -o jsonpath='{.spec.containers[0].volumeMounts[?(@.name=="elasticsearch-storage")].mountPath}')
@@ -99,5 +101,14 @@ done
 
 log "-- Gather Elasticsearch CR"
 get_elasticsearch_cr
+
+# LokiStack resource gathering. if the condition is true only then loki directory will be created.
+data="$(oc -n $NAMESPACE get lokistack --ignore-not-found -o yaml)"
+if [ "$data" != "" ] ; then
+  log "Gathering Loki resources"
+  mkdir -p "$loki_folder"
+  echo "${data}" > "${loki_folder}/lokistack_instance.yaml"
+fi
+
 
 log "END gather_logstore_resources ..."


periklis (changes requested):
Let's fix the formatting here and I believe we need to put this better in the  `gather_logstore_resources` file.

saurabhsadhale:
@periklis  It is a little confusing to gauge where to add this snippet for collection of LokiStack instance CR. The main thing to note here is that the resources ( pod ,deployments,secrets, configmaps )  collection  is being done through the same script including the CRs ( ClusterLogging, ClusterLogForwarder ). 

Or if we could restructure the log collection ( Must-gather ) in a way,  where we collect CRs  ( Elasticsearch, LokiStack, ClusterLogging, ClusterLogForwarder ) only through a single script.  If that works ? 

Could you please guide me about the formatting ?

periklis:
> @periklis It is a little confusing to gauge where to add this snippet for collection of LokiStack instance CR. The main thing to note here is that the resources ( pod ,deployments,secrets, configmaps ) collection is being done through the same script including the CRs ( ClusterLogging, ClusterLogForwarder ).
> 
> Or if we could restructure the log collection ( Must-gather ) in a way, where we collect CRs ( Elasticsearch, LokiStack, ClusterLogging, ClusterLogForwarder ) only through a single script. If that works ?
> 
> Could you please guide me about the formatting ?

As mentioned above ` gather_logstore_resources` is good for the start to add you snippet. Formatting is a bit broken that the if block is intended by 2 more spaces than needed.
Your change:
```shell
data="$(oc -n $NAMESPACE get lokistack --ignore-not-found -o yaml)"
  if [ "$data" != "" ] ; then
    log "Gathering Loki resources"
    mkdir -p "$loki_folder"
    echo "${data}" > "${loki_folder}/lokistack_instance.yaml"
  fi
```
correct form:
```shell
data="$(oc -n $NAMESPACE get lokistack --ignore-not-found -o yaml)"
if [ "$data" != "" ] ; then
  log "Gathering Loki resources"
  mkdir -p "$loki_folder"
  echo "${data}" > "${loki_folder}/lokistack_instance.yaml"
fi
```

periklis (approved):
/lgtm

saurabhsadhale:
@periklis  there is one issue about this:

saurabhsadhale:
@periklis  there is one issue about this snippet being added in the gather_logstore_resources script. 

This code is present in the gather_logstore_resources script which will be triggered if the condition is satisfied in [this code here](https://github.com/openshift/cluster-logging-operator/blob/master/must-gather/collection-scripts/gather#L81-L86). 

Lokistack does not require elasticsearch which means that script of gather_logstore_resources ( where the code is currently added )  will never be triggered in case of Loki installations.

periklis:
@saurabhsadhale FYI I created a small patch to fix this in #1861

saurabhsadhale:
Thanks @periklis

PR #1857, opened 2023-02-09
LOG-3499: add /var/log/oauth-server/audit.log to the log collection path

vparfonov:

### Description
This PR:
- updates `fluentd` config to collect logs from `/var/log/oauth-server/audit.log`
- updates `vector` config to collect logs from `/var/log/oauth-server/audit.log`
- updates corresponding unit tests 

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-3499
- Enhancement proposal:

### DIFF
--- a/internal/generator/fluentd/conf_test.go
+++ b/internal/generator/fluentd/conf_test.go
@@ -202,7 +202,7 @@ var _ = Describe("Testing Complete Config Generation", func() {
   @type tail
   @id openshift-audit-input
   @label @INGRESS
-  path /var/log/oauth-apiserver/audit.log,/var/log/openshift-apiserver/audit.log
+  path /var/log/oauth-apiserver/audit.log,/var/log/openshift-apiserver/audit.log,/var/log/oauth-server/audit.log
   pos_file /var/lib/fluentd/pos/oauth-apiserver.audit.log
   follow_inodes true
   tag openshift-audit.log
--- a/internal/generator/fluentd/fluent_conf_test.go
+++ b/internal/generator/fluentd/fluent_conf_test.go
@@ -2831,7 +2831,7 @@ var _ = Describe("Generating fluentd config", func() {
   @type tail
   @id openshift-audit-input
   @label @INGRESS
-  path /var/log/oauth-apiserver/audit.log,/var/log/openshift-apiserver/audit.log
+  path /var/log/oauth-apiserver/audit.log,/var/log/openshift-apiserver/audit.log,/var/log/oauth-server/audit.log
   pos_file /var/lib/fluentd/pos/oauth-apiserver.audit.log
   follow_inodes true
   tag openshift-audit.log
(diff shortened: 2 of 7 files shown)


internal/generator/fluentd/conf_test.go
```diff
@@ -202,7 +202,7 @@ var _ = Describe("Testing Complete Config Generation", func() {
   @type tail
   @id openshift-audit-input
   @label @INGRESS
-  path /var/log/oauth-apiserver/audit.log,/var/log/openshift-apiserver/audit.log
+  path /var/log/oauth-apiserver/audit.log,/var/log/openshift-apiserver/audit.log,/var/log/auth-server/audit.log
```
vimalk78, line 205: it should be `/var/log/oauth-server/audit.log` as per Jira comment

vimalk78 (changes requested):
Check last comment in Jira task

jcantrill (commented):
/hold
/approve

vimalk78:
if possible in this PR (or in separate PR), 
- update `test/helpers/types/types.go` to add a struct for the structure of this new audit log
- update data model for vector audit logs in `internal/generator/vector/normalize.go`

internal/generator/fluentd/conf_test.go
vparfonov: eagle eye

vparfonov:
> if possible in this PR (or in separate PR),
> 
> * update `test/helpers/types/types.go` to add a struct for the structure of this new audit log
> * update data model for vector audit logs in `internal/generator/vector/normalize.go`

@vimalk78 What's so special in log from `oauth-server` that we might need a separate type for it? It still can be [OpenshiftAuditLog](https://github.com/vparfonov/cluster-logging-operator/blob/549d150a181e3e2ebd820bc78553d87209530bbb/test/helpers/types/types.go#L419), no?

vimalk78:
> @vimalk78 What's so special in log from `oauth-server` that we might need a separate type for it? It still can be [OpenshiftAuditLog](https://github.com/vparfonov/cluster-logging-operator/blob/549d150a181e3e2ebd820bc78553d87209530bbb/test/helpers/types/types.go#L419), no?

I am not sure about the format of these logs. if it is same as OpenshiftAuditLogs, yes we can create a type alias and use same.

jcantrill:
> > @vimalk78 What's so special in log from `oauth-server` that we might need a separate type for it? It still can be [OpenshiftAuditLog](https://github.com/vparfonov/cluster-logging-operator/blob/549d150a181e3e2ebd820bc78553d87209530bbb/test/helpers/types/types.go#L419), no?
> 
> I am not sure about the format of these logs. if it is same as OpenshiftAuditLogs, yes we can create a type alias and use same.

Per issue comment from @vparfonov its a  kube event

jcantrill (commented):
/lgtm
/hold cancel

PR #1870, opened 2023-02-15
LOG-3573: Add tlsSecurityProfile to CLF spec

jcantrill:
### Description
This PR:
* adds tlsSecurityProfile to the CLF spec
* adds secruityProfile to the output.tls spec
* updates the api documentation
* fixes manager.yaml to preserve seccompprofile on manifest generation

### Links
* https://issues.redhat.com/browse/LOG-3573

cc @syedriko


apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -236,6 +240,17 @@ type OutputTLSSpec struct {
 	//
 	// This option is *not* recommended for production configurations.
 	InsecureSkipVerify bool `json:"insecureSkipVerify,omitempty"`
+
+	// TLSSecurityProfile is the security profile to apply to the output connection
+	TLSSecurityProfile *TLSSecurityProfileSpec `json:"securityProfile,omitempty"`
+}
+
+// TLSSecurityProfileSpec sets the TLS security profile.
+type TLSSecurityProfileSpec struct {
```
syedriko, line 249: @jcantrill We don't need a new type here, config/v1/TLSSecurityProfile will do just fine

syedriko: I would rather name it "tlsSecurityProfile", to keep consistent with https://github.com/openshift/api/blob/master/config/v1/types_apiserver.go#L56

syedriko: @jcantrill We don't want a new type here, we actually need tlsSecurityProfile to be config/v1/TLSSecurityProfile https://github.com/openshift/api/blob/master/config/v1/types_tlssecurityprofile.go#L6. It's a bit more involved than just the name of a profile.

PR #1883, opened 2023-02-21
LOG-3247: Drop support of partially valid cluster log forwarder

Clee2691:
## Description

This PR is a combination of 2 tickets, `LOG-3247: Drop support of partially valid cluster log forwarder` and `LOG-3648: Refactor validation code`. 

### [LOG-3247](https://issues.redhat.com//browse/LOG-3247): Drop support of partially valid cluster log forwarder
`LOG-3247` removes the support for `degraded` or partially valid `ClusterLogForwarder` (CLF) instances. A `degraded` CLF involves one or more pipelines that have at least one valid input and output but could also contain invalid inputs and outputs. CLFs could still forward to valid outputs but the CLF had a condition status of `degraded`. This `degraded` status is now removed and all invalid events are streamed to the instance events.

A partially valid CLF is now completely **invalid**, with the `Ready` condition set to `false` and a message telling users that the CLF spec is invalid. Users can check events and/or the status in the `YAML` for specific errors. 

An invalid or partially valid CLF will now **NOT** reconcile the collectors.

### [LOG-3648](https://issues.redhat.com//browse/LOG-3648): Refactor validation code
`LOG-3648` decouples validating the CLF instance from generation of collectors and their configurations. Originally validating all inputs, outputs, and pipelines were performed in the `NormalizeForwarder()` function during config generation. Validating and all of the associated tests are now moved into the `validations/clusterlogforwarder` package and does not rely on the `ClusterLoggingRequest` object. I combined and removed another verification method defined in `generator.go` into the `validations/clusterlogforwarder` package.

This fix also removes mutation of the CLF spec during validations. Normalization of the CLF is removed and partially valid CLFs are rejected. All pipelines must now have a unique name and cannot be empty.

/cc @cahartma
/assign @jcantrill 

### JIRA 
- https://issues.redhat.com/browse/LOG-3247
- https://issues.redhat.com/browse/LOG-3648


jcantrill:
> A partially valid CLF is now completely valid
I need to review the code but I'm wondering how we arrived at this statement.  Partially valid CLF should be completely "invalid".  We want to reject them outright and force admins to correct them.  My expectation for the "upgrade" scenario is that any degraded deployments should continue to run because the CLF would now be rejected.

apis/logging/v1/cluster_log_forwarder.go
```diff
@@ -33,13 +33,17 @@ func IsOutputTypeName(s string) bool {
 }
 
 // Get all subordinate condition messages for condition of type "Ready" and False
+// A 'true' Ready condition with a message means some error with pipeline but it is still valid
```
jcantrill, line 36: Please investigate the kube api documentation for true/false as my recollection is that "true" means the condition applies and the message is additional information.  I can not speak to how on might further evaluate this statement in isolation without applying it to an example, but in general, if the condition is to indicate the pipeline is "degraded" then it should be 'true' plus some additional message to indicate which one.

apis/logging/v1/cluster_log_forwarder.go
```diff
@@ -33,13 +33,17 @@ func IsOutputTypeName(s string) bool {
...
 func (status ClusterLogForwarderStatus) GetReadyConditionMessages() []string {
 	var messages = []string{}
 	for _, nc := range []NamedConditions{status.Pipelines, status.Inputs, status.Outputs} {
 		for _, conds := range nc {
+			currCond := conds.GetCondition(ConditionReady)
 			if !conds.IsTrueFor(ConditionReady) {
-				currCond := conds.GetCondition(ConditionReady)
 				messages = append(messages, currCond.Message)
+				// If a pipeline is "degraded" then it should have an extra error message
```
jcantrill, line 44: In general, the existence of the message should be immaterial to the whether the condition applies or not.

apis/logging/v1/cluster_log_forwarder.go
```diff
@@ -58,16 +62,18 @@ func (status ClusterLogForwarderStatus) IsReady() bool {
 	return true
 }
 
-// IsDegraded returns true if any of the subordinate conditions are degraded.
-func (status ClusterLogForwarderStatus) IsDegraded() bool {
+// IsAllNotReady checks to see if all subordinate conditions are not ready
+func (status ClusterLogForwarderStatus) IsAllNotReady() bool {
```
jcantrill, line 66: Consider flipping the naming and meaning as it its possible to write an "if" as a double negative which is sometime challenging to wrap your head around:
```
if !IsAllNotReady {
}
```
vs
```
if !IsAllReady {
}
```

internal/constants/constants.go
```diff
@@ -137,6 +137,8 @@ const (
 	EventReasonGetObject            = "GetObject"
 	EventReasonRemoveObject         = "RemoveObject"
 	EventReasonUpdateObject         = "UpdateObject"
+
+	MigrateDefaultOutput = "default-replaced"
```
jcantrill, line 141: Looks like we maybe will have to rebase on @cahartma PR?

jcantrill (commented):
/hold

Clee2691:
> > A partially valid CLF is now completely valid
> > I need to review the code but I'm wondering how we arrived at this statement.  Partially valid CLF should be completely "invalid".  We want to reject them outright and force admins to correct them.  My expectation for the "upgrade" scenario is that any degraded deployments should continue to run because the CLF would now be rejected.

When users "upgrade" the CLF, wouldn't the existing spec be reconciled with the new code? If this fix was to reject degraded CLFs then existing deployments would be wiped out when upgraded. Am I thinking about this wrong?

internal/constants/constants.go
Clee2691: Yes will need to rebase

apis/logging/v1/cluster_log_forwarder.go
Clee2691: N/A "degraded" pipelines are invalid

apis/logging/v1/cluster_log_forwarder.go
Clee2691: N/A. "Degraded" piplines are invalid/ "degraded" is not removed.

apis/logging/v1/cluster_log_forwarder.go
Clee2691: This function is not needed anymore and was removed.

internal/k8shandler/forwarding.go
```diff
@@ -111,336 +88,17 @@ func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig(extras map[
 	return generatedConfig, err
 }
 
-func (clusterRequest *ClusterLoggingRequest) readClusterName() (string, error) {
```
jcantrill, line 114: Where did this go?

jcantrill (commented):
/approve

internal/k8shandler/forwarding.go
Clee2691: This is moved to `validate_clusterlogforwarderspec.go` at line 368. This was because this method was only ever used when verifying outputs. It was specifically for `CloudWatch`

internal/pkg/generator/forwarder/generator.go
```diff
@@ -61,9 +61,24 @@ func Generate(collectionType logging.LogCollectionType, clfYaml string, includeD
...
 	}
 
-	spec, status := clRequest.NormalizeForwarder(map[string]bool{})
-	log.V(2).Info("Normalization", "spec", spec)
-	log.V(2).Info("Normalization", "status", status)
+	// Added because this generator is used for tests and the tests assume a correct
+	// CLF spec.
+	// Originally, this generator normalized the forwarder spec only which dropped
+	// input specs with names that were reserved (application, audit, infrastructure).
+	// With the change in validation, CLFs are rejected outright with input specs name equal to reserved names.
```
jcantrill, line 68: Does looks to break input specs with a admin defined named where for instance you need or can "create" an input for selectively collecting a specific set of namespaces?

jcantrill: I'm not reading the code correctly; we are dropping inputs with reserved words.

PR #1888, opened 2023-02-27
LOG-3435: reduce logging verbosity of vector

vparfonov:
### Description
This PR:

- reduce logging verbosity of vector

/cc 
/assign @jcantril 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-3435
- Enhancement proposal:

### DIFF
--- a/internal/collector/vector/visitors.go
+++ b/internal/collector/vector/visitors.go
@@ -14,7 +14,7 @@ const (
 
 func CollectorVisitor(collectorContainer *corev1.Container, podSpec *corev1.PodSpec) {
 	collectorContainer.Env = append(collectorContainer.Env,
-		corev1.EnvVar{Name: "VECTOR_LOG", Value: "info"},
+		corev1.EnvVar{Name: "VECTOR_LOG", Value: "WARN"},
 		corev1.EnvVar{
 			Name: "VECTOR_SELF_NODE_NAME",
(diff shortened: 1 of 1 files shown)


internal/collector/vector/visitors.go
```diff
@@ -14,7 +14,7 @@ const (
 
 func CollectorVisitor(collectorContainer *corev1.Container, podSpec *corev1.PodSpec) {
 	collectorContainer.Env = append(collectorContainer.Env,
-		corev1.EnvVar{Name: "VECTOR_LOG", Value: "info"},
+		corev1.EnvVar{Name: "VECTOR_LOG", Value: "WARN"},
```
jcantrill, line 17: @vparfonov this code is common to fluentd.  can you also investigate:
* what is the impact on also changing fluentd
* does fluentd honor "WARN" or does it require "warn" ?

vparfonov: - It will not impact fluentd [see here](https://github.com/openshift/cluster-logging-operator/blob/master/internal/generator/fluentd/conf.go#L55)
- fluentd require "warn" in lowercase [fluentd/log-level](https://docs.fluentd.org/deployment/logging#log-level)

PR #1891, opened 2023-03-01
LOG-3567: Impl TLS Security Profile for vector

jcantrill:
### Description
This PR:
* Implements TLS Security Profile for vector sinks
* Settings are honored in priority of: output, clusterlogforwarder, cluster

### Links
* https://issues.redhat.com/browse/LOG-3567

/hold

### DIFF
--- a/docs/features/tls_security_profile.adoc
+++ b/docs/features/tls_security_profile.adoc
@@ -1,0 +1,49 @@
+== https://issues.redhat.com/browse/LOG-3270[LOG-3270]: OCP Cluster-wide Cryptographic Policies and Output Crypto Configuration
+
+This feature makes collector deployments comply with the OCP cryptographic policy that is defined for
+the https://docs.openshift.com/container-platform/4.6/rest_api/config_apis/apiserver-config-openshift-io-v1.html[cluster] config settings.
+Additionally, this feature enables the capability for administrators to define a TLS profile to be used when forwarding
+logs to different outputs.
+
+TLS profiles are uniquely configurable for each defined output or by defining a single profile to apply to all log forwarding outputs. Profiles
+are applied according to the following order of precedence:
+
+. Output Spec
+. Cluster Log Forwarder Spec
+. Cluster-wide Config
+
+IMPORTANT: This is a tech-preview feature for output configurations. This is enabled by adding the feature gate annotation to the ClusterLogForwarder
+
+.Enable TLS Security Profile feature
+[source]
+----
+apiVersion: "logging.openshift.io/v1"
+kind: ClusterLogForwarder
+metadata:
+  annotations:
+    logging.openshift.io/preview-tls-security-profile: enabled
+spec:
+  outputs:
+  - name: devel
+    type: elasticsearch
+  - name: production
+    type: kafka
+    tls:
+      securityProfile:
+        type: Modern
+  pipelines:
+   - name: application-logs
+     inputRefs:
+     - application
+     outputRefs:
+     - production
+   - name: devel-logs
+     inputRefs:
+     - application
+     outputRefs:
+     - devel
+  tlsSecurityProfile:
+    type: Intermediate
+----
+This configuration will enable the "Modern" profile for application-logs and "Intermediate" for devel-logs
+
--- a/internal/generator/tls.go
+++ b/internal/generator/tls.go
@@ -1,0 +1,38 @@
+package generator
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/internal/tls"
+	"strings"
+)
+
+// TLSProfileInfo returns the minTLSVersion, ciphers as a comma-separated list given the available TLSSecurityProfile
+func (op Options) TLSProfileInfo(clfProfile *configv1.TLSSecurityProfile, outputSpec logging.OutputSpec) (string, string) {
+	var tlsProfileSpec *configv1.TLSProfileSpec
+	if outputSpec.TLS != nil && outputSpec.TLS.TLSSecurityProfile != nil {
+		tlsProfileSpec = configv1.TLSProfiles[outputSpec.TLS.TLSSecurityProfile.Type]
+		if configv1.TLSProfileCustomType == outputSpec.TLS.TLSSecurityProfile.Type {
+			minTLSVersion := outputSpec.TLS.TLSSecurityProfile.Custom.MinTLSVersion
+			ciphers := outputSpec.TLS.TLSSecurityProfile.Custom.Ciphers
+			return string(minTLSVersion), strings.Join(ciphers, ",")
+		}
+	} else if clfProfile != nil {
+		tlsProfileSpec = configv1.TLSProfiles[clfProfile.Type]
+		if configv1.TLSProfileCustomType == clfProfile.Type {
+			minTLSVersion := clfProfile.Custom.MinTLSVersion
+			ciphers := clfProfile.Custom.Ciphers
+			return string(minTLSVersion), strings.Join(ciphers, ",")
+		}
+	} else if _, ok := op[ClusterTLSProfileSpec]; ok {
+		clusterSpec := op[ClusterTLSProfileSpec].(configv1.TLSProfileSpec)
+		tlsProfileSpec = &clusterSpec
+	}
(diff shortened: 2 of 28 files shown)


internal/tls/tls.go
```diff
@@ -48,7 +48,7 @@ func MinTLSVersion(profile configv1.TLSProfileSpec) string {
 	return string(profile.MinTLSVersion)
 }
 
-// getTLSProfileSpec returns TLSProfileSpec to be used for generating OPENSSL_CONF
+// GetTLSProfileSpec returns TLSProfileSpec to be used for generating OPENSSL_CONF
 func GetTLSProfileSpec(apiServerTLSProfile *configv1.TLSSecurityProfile) configv1.TLSProfileSpec {
```
syedriko, line 52: Should we rename this to GetClusterTLSProfileSpec()?

test/framework/functional/output_http.go
```diff
@@ -54,25 +69,62 @@ codec = "ndjson"
...
+		os.Exit(1)
+	}
+	return b.String()
+}
+
 func (f *CollectorFunctionalFramework) AddVectorHttpOutput(b *runtime.PodBuilder, output logging.OutputSpec) error {
+	return f.AddVectorHttpOutputWithConfig(b, output, "", nil)
+}
+
+func (f *CollectorFunctionalFramework) AddVectorHttpOutputWithConfig(b *runtime.PodBuilder, output logging.OutputSpec, toml configv1.TLSProfileType, secret *corev1.Secret) error {
```
syedriko, line 105: the toml name is confusing here, how about tlsProfile?

jcantrill: :+1: that is left over from a previous version

internal/generator/tls.go
```diff
@@ -0,0 +1,39 @@
...
+	for _, profile := range profiles {
+		if profile != nil {
+			if profile.Type == configv1.TLSProfileCustomType {
+				tlsProfileSpec = &profile.Custom.TLSProfileSpec
+			} else {
+				tlsProfileSpec = configv1.TLSProfiles[profile.Type]
+			}
+		}
+
+	}
```
syedriko, line 31: Just IMO, it'd be easier to follow if the logic flow was schematically like this:
`
if outputSpec.TLS != nil { 
     tlsProfileSpec = configv1.TLSProfiles[outputSpec.TLS.TLSSecurityProfile.Type]
} else if clfSpec != nil {
     tlsProfileSpec = configv1.TLSProfiles[clfSpec.Type]
} else { 
      tlsProfileSpec =  &op[ClusterTLSProfileSpec].(configv1.TLSProfileSpec)
}
`

internal/generator/tls.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	configv1 "github.com/openshift/api/config/v1"
+	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/internal/tls"
+	"strings"
+)
+
+// TLSProfileInfo returns the minTLSVersion, ciphers as a comma-separated list given the available TLSSecurityProfile
+func (op Options) TLSProfileInfo(clfSpec *configv1.TLSSecurityProfile, outputSpec logging.OutputSpec) (string, string) {
```
syedriko, line 11: Another thing I'm not sure about, why make TLSProfileInfo a method on Options when all we use Options for is to get cluster-scope TLS profile.

syedriko: Wait, but after calling this, we turn around and 
op[generator.MinTLSVersion] = outMinTlsVersion
op[generator.Ciphers] = outCiphers
May be we should divorce TLSProfileInfo() from Options and use its return values to populate the two options ^^^.

jcantrill: I have several motivations here:
* Profiles are or can be output specific so this is a convenient way to mod the generator behavior per output and consolidate the fall back logic
* Options are also the convenient way to pass around the feature gate to enable/disable the feature
* This puts the logic in a more singular place should we need to apply to the fluent impl

I like the idea of basically a factory that tells me how to configure the profile; it makes a decision like the following without polluting logic in several places with an if/else

```
		outMinTlsVersion, outCiphers := op.TLSProfileInfo(clfspec.TLSSecurityProfile, o)
		op[generator.MinTLSVersion] = outMinTlsVersion
		op[generator.Ciphers] = outCiphers
```

I'm open to an alternate impl if there is something cleaner

internal/generator/vector/conf_test.go
```diff
@@ -27,7 +27,7 @@ var _ = Describe("Testing Complete Config Generation", func() {
 		f = func(testcase testhelpers.ConfGenerateTest) {
 			g := generator.MakeGenerator()
 			if testcase.Options == nil {
-				testcase.Options = generator.Options{generator.TlsProfileSpec: tls.GetTLSProfileSpec(nil)}
+				testcase.Options = generator.Options{generator.ClusterTLSProfileSpec: tls.GetTLSProfileSpec(nil)}
```
alanconway, line 30: Why rename this? The string is still "tlsProfileSpec".

jcantrill: To make it more obvious which profile it represents.  the actual string is immaterial

docs/features/tls_security_profile.adoc
```diff
@@ -0,0 +1,49 @@
+== https://issues.redhat.com/browse/LOG-3270[LOG-3270]: OCP Cluster-wide Cryptographic Policies and Output Crypto Configuration
+
+This feature makes collector deployments comply with the OCP cryptographic policy that is defined for
+the https://docs.openshift.com/container-platform/4.6/rest_api/config_apis/apiserver-config-openshift-io-v1.html[cluster] config settings.
+Additionally, this feature enables the capability for administrators to define a TLS profile to be used when forwarding
```
syedriko, line 5: enables ~~the capability for~~ administrators to define

docs/features/tls_security_profile.adoc
```diff
@@ -0,0 +1,49 @@
...
+logs to different outputs.
+
+TLS profiles are uniquely configurable for each defined output or by defining a single profile to apply to all log forwarding outputs. Profiles
```
syedriko, line 8: TLS profiles are ~~uniquely~~ configurable for each ~~defined~~ output separately

docs/features/tls_security_profile.adoc
```diff
@@ -0,0 +1,49 @@
...
+are applied according to the following order of precedence:
+
+. Output Spec
+. Cluster Log Forwarder Spec
+. Cluster-wide Config
+
+IMPORTANT: This is a tech-preview feature for output configurations. This is enabled by adding the feature gate annotation to the ClusterLogForwarder
```
syedriko, line 15: It is enabled by the feature gate

docs/features/tls_security_profile.adoc
```diff
@@ -0,0 +1,49 @@
...
+     - production
+   - name: devel-logs
+     inputRefs:
+     - application
+     outputRefs:
+     - devel
+  tlsSecurityProfile:
+    type: Intermediate
+----
+This configuration will enable the "Modern" profile for application-logs and "Intermediate" for devel-logs
```
syedriko, line 48: Since just above we are taking about applying TLS profiles to outputs, it would be more correct to say here that the `Modern` profile applies to the `production` output and the `Intermediate` profile applies to the `devel` output, not pipelines.

internal/tls/profile.go
```diff
@@ -32,7 +32,7 @@ CipherSuites = {{CipherSuites}}
 
 func OpenSSLConf(k8client client.Client) string {
```
syedriko, line 33: We can remove this file, we're not generating OpenSSL config anymore.

internal/tls/tls.go
```diff
@@ -48,8 +48,8 @@ func MinTLSVersion(profile configv1.TLSProfileSpec) string {
 	return string(profile.MinTLSVersion)
 }
 
-// getTLSProfileSpec returns TLSProfileSpec to be used for generating OPENSSL_CONF
-func GetTLSProfileSpec(apiServerTLSProfile *configv1.TLSSecurityProfile) configv1.TLSProfileSpec {
+// GetClusterTLSProfileSpec returns TLSProfileSpec to be used for generating OPENSSL_CONF
```
syedriko, line 51: No reason to refer to OPENSSL_CONF anymore

PR #1893, opened 2023-03-03
Fix JSON parser of different log types

rchicoli:
### Description

I am running the EventRouter application inside the `openshift-logging` namespace, which has the "**log_type**" set to infrastructure.
By doing so, the JSON parser will ignore any messages coming from the infrastructure log type. This happens because the parser has a hardcoded condition to match only application logs.

#### How to reproduce this problem?

1. At first, make sure the JSON parser is enabled for the infrastructure:

```yaml
# ClusterLogForwarder.. minimized output
pipelines:
  - inputRefs:
    - infrastructure
    labels:
      logSource: ClusterLogForwarder
      logType: infrastructure
    name: infra-pipeline
    outputRefs:
    - etc
    parse: json # enabled <---
  
  - inputRefs:
    - application
    labels:
      logSource: ClusterLogForwarder
      logType: application
    name: app-pipeline
    outputRefs:
    - etc
    parse: json  # enabled <---
```

2. Then deploy the Eventrouter in the `openshift-logging` namespace, according to the documentation:

> You should always deploy the Event Router to the openshift-logging project to ensure it collects events from across the cluster

3. After running the Cluster Logging Operator, the auto-generated `vector.toml` will look like:

```toml
[transforms.infra-pipeline]
type = "remap"
inputs = ["infrastructure"]
source = '''
  .openshift.labels = {"logSource":"ClusterLogForwarder","logType":"infrastructure"}
 
  # <--- THE following condition will NEVER match
  if .log_type == "application" {
    parsed, err = parse_json(.message)
    if err == null {
      .structured = parsed
      del(.message)
    }
  }
'''

[transforms.app-pipeline]
type = "remap"
inputs = ["application"]
source = '''
  .openshift.labels = {"logSource":"ClusterLogForwarder","logType":"application"}

  if .log_type == "application" {
    parsed, err = parse_json(.message)
    if err == null {
      .structured = parsed
      del(.message)
    }
  }
'''
```

#### Code Changes

The easiest fix would be to remove the if condition, since the "transforms" applies only to specific "inputs".

/cc @cahartma
/assign @jcantrill

I am using Vector 5.6.2, but this problem should affect all versions.

/cherry-pick release-5.6

### Links

- Support Case: 03458554

### DIFF
--- a/internal/generator/vector/pipelines.go
+++ b/internal/generator/vector/pipelines.go
@@ -30,13 +30,11 @@ func Pipelines(spec *logging.ClusterLogForwarderSpec, op generator.Options) []ge
 		}
 		if p.Parse == ParseJson {
 			parse := `
-if .log_type == "application" {
   parsed, err = parse_json(.message)
   if err == null {
     .structured = parsed
     del(.message)
   }
-}
 `
 			vrls = append(vrls, parse)
 		}


jcantrill:
This hard-coding is intentional.  There is a larger desire to allow json parsing of infrastructure logs but we are not ready to embrace that change.

jcantrill:
@alanconway can you comment here?

alanconway:
I think it would be reasonable to lift the condition and give consistent treatment to JSON logs from infra components as we do for application, but as @jcantrill says, we didn't do that initially because we need to evaluate the impact and make sure it doesn't cause problems somewhere else.
@rchicoli can you raise a JIRA at https://issues.redhat.com/projects/LOG and link this patch? We need to schedule somebody on the logging team to do the investigation before we merge it.

ssonigra:
Hello @alanconway @jcantrill Raised a bug for testing and review of Team 
https://issues.redhat.com/browse/LOG-3857

rchicoli:
That is perfect. I was about to create the Jira task this afternoon, but it seems that @ssonigra did it already, thanks!

Additionally, I was wondering what could go wrong, so I did have a look at the `parse_json` source code:

https://github.com/vectordotdev/vector/blob/master/lib/vrl/stdlib/src/parse_json.rs#L10-L15
```rust
fn parse_json(value: Value) -> Resolved {
    let bytes = value.try_bytes()?;
    let value = serde_json::from_slice::<'_, Value>(&bytes)
        .map_err(|e| format!("unable to parse json: {e}"))?;
    Ok(value)
}
```
Besides more memory allocation and some processing time, it should work fine.
Thanks for taking this further and let me know, if I should update the code at some point.

PR #1898, opened 2023-03-08
LOG-3735: Add tls config for cloudwatch and google outputs

Clee2691:
### Description

This PR addresses the issue that `Amazon Cloudwatch` (CW) & `Google Cloud Logging` (GCL) are the only outputs that do not allow configuration of the TLS block in the vector config. This change allows addition of TLS settings to the mentioned outputs in the same way as other supported outputs (Elasticsearch, Kafka, etc.) CW and GCL outputs are also now configured with the `min_tls_setting` and the `ciphersuites` corresponding to the security profile.

In addition, this PR also refactors the way that the TLS settings for vector config are set. I've reduced redundancy of files and code to improve readability. 

/cc @cahartma
/assign @jcantrill 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-3735


internal/generator/vector/output/cloudwatch/ca_file.go
```diff
@@ -0,0 +1,17 @@
+package cloudwatch
```
jcantrill, line 1: Why do we have to duplicate the tls logic instead of reusing something that already exists?

Clee2691: I agree that we shouldn't but I kept it in line with the current implementations to all of the outputs. I can refactor the TLS logic for all of the outputs but the scope of this ticket is not that.

Clee2691: @jcantrill I've refactored TLS to remove redundancy in the code

PR #1899, opened 2023-03-08
LOG-3628: Add ocp-console plugin configuration

jcantrill:
### Description
This PR:
* updates the ocp console log plugin deployment to configure: loglimit, timeout

ref: https://issues.redhat.com/browse/LOG-3628

cc @jgbernalp @alanconway


internal/visualization/console/reconciler.go
```diff
@@ -236,13 +249,24 @@ func (r *Reconciler) mutateDeployment() error {
...
 							{
 								Name:      "serving-cert",
 								ReadOnly:  true,
 								MountPath: "/var/serving-cert",
 							},
+							{
+								Name:      "serving-cert",
+								ReadOnly:  true,
+								MountPath: "/var/serving-cert",
+							},
```
jgbernalp, line 264:
```suggestion
```

apis/logging/v1/clusterlogging_types.go
```diff
@@ -96,11 +96,20 @@ type ClusterLoggingStatus struct {
 
 // This is the struct that will contain information pertinent to Log visualization (Kibana)
 type VisualizationSpec struct {
+
 	// The type of Visualization to configure
+	// +kubebuilder:validation:Enum=ocp-console;kibana
```
xperimental, line 101: I think you also need to allow empty string as a value to not break existing resources.

internal/visualization/console/plugin.go
```diff
@@ -9,9 +9,9 @@ import (
 )
 
 // ReconcilePlugin reconciles the console plugin to expose log querying of storage
-func ReconcilePlugin(k8sClient client.Client, logStore *logging.LogStoreSpec, owner client.Object, clusterVersion string) error {
+func ReconcilePlugin(k8sClient client.Client, logStore *logging.LogStoreSpec, owner client.Object, clusterVersion string, consoleSpec *logging.OCPConsoleSpec) error {
 	lokiService := lokistack.LokiStackGatewayService(logStore)
-	r := NewReconciler(k8sClient, NewConfig(owner, lokiService, FeaturesForOCP(clusterVersion)))
+	r := NewReconciler(k8sClient, NewConfig(owner, lokiService, FeaturesForOCP(clusterVersion)), consoleSpec)
 	if logStore != nil && logStore.Type == logging.LogStoreTypeLokiStack {
```
xperimental, line 15: Should the visualization still also depend on the value of `logStore` now that there is an explicit `spec.visualization.type` value?

apis/logging/v1/clusterlogging_types.go
```diff
@@ -609,7 +633,8 @@ const (
 type VisualizationType string
 
 const (
-	VisualizationTypeKibana VisualizationType = "kibana"
+	VisualizationTypeKibana     VisualizationType = "kibana"
+	VisualizationTypeOCPConsole VisualizationType = "ocp-console"
```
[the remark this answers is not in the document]
jcantrill: Do we need a none if you do not specify visualization?

internal/visualization/console/plugin.go
jcantrill: yes it has to be since we only deploy the plugin when "loki" is being used for the store.  Similar to kibana, deploying the visualization does not make sense if you do not have a back-end to query.

internal/visualization/console/reconciler.go
```diff
@@ -189,7 +191,18 @@ func (r *Reconciler) mutateConfigMap() error {
...
+	}
+	if r.consoleSpec != nil {
+		config := []string{}
+		if r.consoleSpec.LogsLimit > 0 {
+			config = append(config, fmt.Sprintf("logsLimit: %d", r.consoleSpec.LogsLimit))
+		}
+		if r.consoleSpec.Timeout != "" {
+			config = append(config, fmt.Sprintf("timeout: %s", r.consoleSpec.Timeout))
+		}
+		o.Data["config.yaml"] = strings.Join(config, "\n")
```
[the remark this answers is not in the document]
jcantrill: serializing the original struct directly since it is the same values

apis/logging/v1/clusterlogging_types.go
```diff
@@ -96,11 +96,20 @@ type ClusterLoggingStatus struct {
...
+	// +kubebuilder:validation:Enum=ocp-console;kibana
 	Type VisualizationType `json:"type"`
 
 	// Specification of the Kibana Visualization component
+	// +nullable
+	// +optional
 	KibanaSpec `json:"kibana,omitempty"`
+
+	//OCPConsole is the specification for the OCP console plugin
+	// +nullable
```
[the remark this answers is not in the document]
jcantrill: :man_shrugging:

internal/visualization/console/reconciler.go
jcantrill: :man_shrugging:

jgbernalp: same volume added twice

apis/logging/v1/clusterlogging_types.go
xperimental: What about people that do not want either of the visualizations? No Kibana for ES and no OCP-console for Loki, because they want to use Loki and their own visualization.

internal/visualization/console/plugin.go
xperimental: Similar to the other thread: my idea was more going into the other direction: making the ocp-console an explicit opt-in for LokiStack instead of automatically activating it, just like Kibana was before with ES.

jcantrill: @alanconway this is your call about explicit opt-in.  My concern would be that now it is not 'opt-in' which is a change in behavior of what already exists for 5.6

apis/logging/v1/clusterlogging_types.go
jcantrill: @alanconway your call here.  I don't know we have this requirement and it would be a change in behavior from what exists in 5.6.   This may be a better topic or change for "logging 6.0"

alanconway: The whole visualization spec is optional, so people can just omit it. 
If it is present then the type is mandatory so I think this is ok.

apis/logging/v1/clusterlogging_types.go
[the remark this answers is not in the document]
alanconway: Again, visualization is optional so no need for a none value.

internal/visualization/console/plugin.go
alanconway: I'd prefer to leave it as an automatic default, so there are fewer ways to end  up at the "why can't I see any logs" support call.

alanconway (approved):
/lgtm

PR #1915, opened 2023-03-15
[release-5.6] Fix namespace metadata in gen-olm-artifacts.py

jacksgt:
Manual backport of https://github.com/openshift/cluster-logging-operator/pull/1900

/cc jcantrill
/assign vimalk78

### DIFF
--- a/hack/gen-olm-artifacts.py
+++ b/hack/gen-olm-artifacts.py
@@ -40,11 +40,11 @@ def generateNamespace(namespace):
       "name": namespace,
       "annotations": {
         "openshift.io/node-selector": ""
+      },
+      "labels": {
+        "openshift.io/cluster-logging": "true",
+        "openshift.io/cluster-monitoring": "true"
       }
-    },
-    "labels": {
-      "openshift.io/cluster-logging": "true",
-      "openshift.io/cluster-monitoring": "true"
     }
   }
   writeResource(ns)


jcantrill:
@jacksgt do you really need to port this back to another branch? I suspect whatever lives on master would also handle the other branches.  this isnt a script we even use any longer

jacksgt:
> do you really need to port this back to another branch? I suspect whatever lives on master would also handle the other branches.

I'm currently deploying 5.5, hence I thought it would be worthwhile to backport this to older branches.
Since I'm basically doing a `git checkout release-5.5 && python3 gen-olm-artifacts.py` each branch needs to have a functioning scripts.

> this isnt a script we even use any longer

Hmm, I wasn't aware of that! I was following the "Hacking" guide here: https://github.com/openshift/cluster-logging-operator/blob/master/docs/contributing/README.adoc#deploying-without-olm

Is there another way you can recommend for obtaining the "raw" manifests?

jcantrill:
> > do you really need to port this back to another branch? I suspect whatever lives on master would also handle the other branches.
> 
> I'm currently deploying 5.5, hence I thought it would be worthwhile to backport this to older branches. Since I'm basically doing a `git checkout release-5.5 && python3 gen-olm-artifacts.py` each branch needs to have a functioning scripts.
> 
> > this isnt a script we even use any longer
> 
> Hmm, I wasn't aware of that! I was following the "Hacking" guide here: https://github.com/openshift/cluster-logging-operator/blob/master/docs/contributing/README.adoc#deploying-without-olm
> 
> Is there another way you can recommend for obtaining the "raw" manifests?

This script was never anything more then a hack to do deployment with the same set of deployment artifacts you would get from OLM without relying on OLM.  I know of no other mechanism without defining your own helm chart or the like.

jacksgt:
> I know of no other mechanism without defining your own helm chart or the like.

I see. Would you say that it's a better approach to use kustomize to obtain the manifests?

```
kubectl kustomize config/overlays/stable/
```

PR #1924, opened 2023-03-20
LOG-3784: value provided in HTTP header 'Content-Type' for Fluentd HTTP plugin is automatically assigned to the 'content_type' parameter

vparfonov:
### Description

This PR allows users to overcome limitations in CLO API's to provide various content type for logs data output.
For the Fluentd HTTP plugin, users can now assign the content type for log forwarding seamlessly, by configuring the [HTTP header](https://docs.fluentd.org/output/http#headers) "Content-Type," the value provided is automatically assigned to the [`content_type`](https://docs.fluentd.org/output/http#content_type) parameter within the plugin.
For example:
```
kind: ClusterLogForwarder
metadata:
  name: instance
  namespace: openshift-logging
spec:
  outputs:
  - http:
      headers:
        k1: v1
        Content-Type: application/json
      method: POST
    name: my-http-storage
    type: http
    url: https://my-logstore.com/logs/app-logs
  pipelines:
  - inputRefs:
    - application
    name:  to-http
    outputRefs:
    - my-http-storage
```
will be:
```
<match **>
	@type http
	endpoint https://my-logstore.com/logs/app-logs
	http_method post
	content_type application/json
	headers {"Content-Type":"application/json","k1":"v1"}
	.......
  </match>
```
Also adds validation for Content-Type header in Http Output, valid types are: `application/json` and `application/x-ndjson`
other types will be rejected.

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-3784
- Enhancement proposal:

### DIFF
--- a/internal/generator/fluentd/output/http/http.go
+++ b/internal/generator/fluentd/output/http/http.go
@@ -25,6 +25,7 @@ type Http struct {
 	Headers        Element
 	SecurityConfig []Element
 	BufferConfig   []Element
+	ContentType    string
 
 	// Encoding is set by plugin according to
 }
@@ -38,7 +39,7 @@ func (h Http) Template() string {
 @type http
 endpoint {{.URI}}
 http_method {{.Method}}
-content_type "application/x-ndjson"
+content_type {{.ContentType}}
 {{kv  .Headers -}}
 {{compose .SecurityConfig}}
 {{compose .BufferConfig}}
@@ -69,6 +70,7 @@ func Output(bufspec *logging.FluentdBufferSpec, secret *corev1.Secret, o logging
 			URI:            o.URL,
 			Method:         Method(o.Http),
 			Headers:        Headers(o),
+			ContentType:    ContentType(o),
 			SecurityConfig: SecurityConfig(o, secret),
 			BufferConfig:   output.Buffer(output.NOKEYS, bufspec, storeID, &o),
 		},
@@ -80,6 +82,13 @@ func Headers(o logging.OutputSpec) Element {
 		return Nil
 	}
 	return KV("headers", utils.ToHeaderStr(o.Http.Headers, "%q:%q"))
+}
+
+func ContentType(o logging.OutputSpec) string {
+	if o.Http == nil || len(o.Http.Headers) == 0 || o.Http.Headers["Content-Type"] == "" {
+		return "application/x-ndjson"
+	}
+	return o.Http.Headers["Content-Type"]
 }
 
 func Method(h *logging.Http) string {
--- a/internal/generator/fluentd/output/http/http_test.go
+++ b/internal/generator/fluentd/output/http/http_test.go
@@ -64,7 +64,7 @@ var _ = Describe("Generate fluentd config", func() {
 	@type http
 	endpoint https://my-logstore.com/logs/app-logs
 	http_method post
-	content_type "application/x-ndjson"
+	content_type application/x-ndjson
 	<auth>
 	  method basic
 	  username "#{File.read('/var/run/ocp-collector/secrets/http-receiver/username') rescue nil}"
@@ -135,7 +135,7 @@ var _ = Describe("Generate fluentd config", func() {
 	@type http
 	endpoint https://my-logstore.com/logs/app-logs
 	http_method post
-	content_type "application/x-ndjson"
+	content_type application/x-ndjson
 	headers {"k1":"v1","k2":"v2"}
 	<auth>
 	  method basic
@@ -214,7 +214,7 @@ var _ = Describe("Generate fluentd config", func() {
 	@type http
 	endpoint https://my-logstore.com/logs/app-logs
 	http_method post
-	content_type "application/x-ndjson"
+	content_type application/x-ndjson
 	headers {"k1":"v1","k2":"v2"}
 	<auth>
 	  method basic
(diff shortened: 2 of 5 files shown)


alanconway:
I did a bit of digging: whether we use json or nd-json is linked to how we encode the message.
application-type/json - the content is a single JSON value, normally an array [] or an object {}
application-type/x-ndjson - the content is a stream of newline-delimited JSON values i.e. "{}\n{}\n"
So if we replace x-ndjson with json, we also must set the relevant fluetnd switchs to make sure it generates a single json object, I think there's a "use-array" or "no-array" switch to do that.

vparfonov:
> I did a bit of digging: whether we use json or nd-json is linked to how we encode the message.
> application-type/json - the content is a single JSON value, normally an array [] or an object {}
> application-type/x-ndjson - the content is a stream of newline-delimited JSON values i.e. "{}\n{}\n"
> So if we replace x-ndjson with json, we also must set the relevant fluetnd switchs to make sure it generates a single json object, I think there's a "use-array" or "no-array" switch to do that.

Yes exactly, you are right, but in the same way   we can do it with changing content-type, see here: 

- https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/out_http.rb#L116 
- https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/out_http.rb#L229

jcantrill:
@alanconway I think the real question is how to we support this in the API?  Do we want to support some key?  The issue is the receiver only supports one and not the other which means it is possible there are other "http" receivers that may have the same issue.  We can only resolve this issue by adding a control knob since we don't know

alanconway:
@jcantrill agreed. I think we need a content-type option, since that's a well-understood notion, and some enum of supported content types. We can take our encoding hint from the content type as @vparfonov points out.
@vparfonov is there a single choice that will solve the immediate problem so you can proceed? 
I'll go look at the API and figure out what we want to add there.

internal/generator/fluentd/output/http/http_test.go
```diff
@@ -228,5 +228,5 @@ func TestHeaders(t *testing.T) {
 
 func TestVectorConfGenerator(t *testing.T) {
```
alanconway, line 229:
```suggestion
func TestFluentdConfGenerator(t *testing.T) {
```

vparfonov:
> @vparfonov I agree with the content-type option. What information do you need - just a content-type string? I will think about where to put this in the API.

yes, content-type string should be enough

alanconway:
The "Content-Type" is already present as a header in the `outputs.http.headers` field, lets use that. 
Note HTTP header names are case-insensitive, so need to normalize case of header keys - probably the golang http library will do it for you.

vparfonov:
> The "Content-Type" is already present as a header in the `outputs.http.headers` field, lets use that. Note HTTP header names are case-insensitive, so need to normalize case of header keys - probably the golang http library will do it for you.
@alanconway 
I think it will not work because force set `Content-Type` header, see here: 
https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/out_http.rb#L229

alanconway:
@vparfonov true it won't be passed through automatically with the other headers.
What I mean is that we don't need a new contentType field in the CLF. 
You can read headers["Content-Type"] when generating the output configuration, and copy that value to the fluent/vector content-type configuration setting.

internal/generator/fluentd/output/http/http.go
```diff
@@ -82,6 +84,13 @@ func Headers(o logging.OutputSpec) Element {
 	return KV("headers", utils.ToHeaderStr(o.Http.Headers, "%q:%q"))
 }
 
+func ContentType(o logging.OutputSpec) string {
+	if o.Http == nil || len(o.Http.Headers) == 0 || o.Http.Headers["Content-Type"] == "" {
+		return "application/x-ndjson"
+	}
+	return o.Http.Headers["Content-Type"]
```
jcantrill, line 91: @alanconway we can not allow setting an arbitrary content type because users only control the receiver and have literally no control over how the payload is crafted, unless @vparfonov can identify fluent is able to handle them based on this value.  As an example, a user can set any of the following:
```
 application/java-archive
 application/EDI-X12   
 application/EDIFACT   
 application/javascript   
 application/octet-stream   
 application/ogg   
 application/pdf  
 application/xhtml+xml   
 application/x-shockwave-flash    
 application/json  
 application/ld+json  
 application/xml   
 application/zip  
 application/x-www-form-urlencoded 
 ```
 I'm pretty certain fluent doesn't speak "pdf" or "javascript".   Our config crafts and manipulates JSON.  The only reasonable options IMO are: JSON, NJSON.  Does fluent convert JSON to say CSV or tab delimited based upon content-type?  If it does not we should consider adding an enum.

Comment (vparfonov) [reply internal/generator/fluentd/output/http/http.go:91]: @jcantrill  Make sense, but i think checking will be done on plugin side according to this table https://docs.fluentd.org/output/http#content_type

Comment (vparfonov) [reply internal/generator/fluentd/output/http/http.go:91]: Oh, I see, you mean we will support only 2 options `application/json` and `application/x-ndjson`

Comment (jcantrill) [reply internal/generator/fluentd/output/http/http.go:91]: we do not set the formatter which dictates the content type.  The default is JSON.  I think it is reasonable that in the future we may allow other formats but we should:
* default one (e.g. json)
* allow an enum on the CR of: json,ndjson

I would check first to see if vector supports both json, ndjson before we make a final decision

Comment (vparfonov) [reply internal/generator/fluentd/output/http/http.go:91]: @jcantrill 
As I understand it, Vector defines Content-Type independently:
https://github.com/vectordotdev/vector/blob/master/src/sinks/http.rs#L339C29-L376

In same Vector allow overriding content-type header: https://github.com/vectordotdev/vector/pull/15625 

Vector also did something specific for Datadog output: https://github.com/vectordotdev/vector/blob/master/src/sinks/datadog/logs/sink.rs#L106

Comment (alanconway) [reply internal/generator/fluentd/output/http/http.go:91]: @jcantrill I take your point, but I think it makes sense to  use the content-type and reject any content type that we don't recognize or that is not appropriate in the context. If we add new API enum to indicate a content type, then we're likely going to end up having to extend it for one special case, then another, until it becomes a mini-content-type of its own.

Comment (jcantrill) [reply internal/generator/fluentd/output/http/http.go:91]: If we wish to accept 'content-type' then we should add a validation to support what you suggest.  It should only allow the 2 which are applicable

Comment (vparfonov) [reply internal/generator/fluentd/output/http/http.go:91]: agree, i will update

File: internal/validations/clusterlogforwarder/validate_http_content_type_headers.go [hunk 3]
```diff
@@ -0,0 +1,35 @@
...
+}
+
+// validateHttpContentTypeHeaders will validate Content-Type header in Http Output
+// valid content-type are: "application/json" and "application/x-ndjson"
+// was introduced in https://github.com/openshift/cluster-logging-operator/pull/1924
+// for https://issues.redhat.com/browse/LOG-3784
+func validateHttpContentTypeHeaders(clf v1.ClusterLogForwarder) error {
+	for i, output := range clf.Spec.Outputs {
+		_, output := i, output // Don't bind range variable.
+		if output.Type == v1.OutputTypeHttp && output.Http != nil && len(output.Http.Headers) != 0 && output.Http.Headers["Content-Type"] != "" {
```
jcantrill, line 23: I don't think you need to check the length.  I believe this will achieve the same:
```
if value, found := output.Http.Headers["Content-Type"]; found && value  != "" {
  //do something 
}
```

jcantrill (commented):
/approve
/lgtm

PR #1940, opened 2023-03-30
Update golangci-lint to latest version.

alanconway:
- Simplify .golangci.yaml, remove obsolete linters
- Fix minor style issues
- Update generated API docs.


alanconway:
Latest golangci-lint has gone bleeding edge and requires go 1.19 which is causing weird conflicts. 
Dropping this for now, may revisit later.

PR #1944, opened 2023-03-31
LOG-3830: Add common labels to CLF objects 

vparfonov:
### Description
This PR adds Kubernetes common labels to the deployed objects in our application. These labels will help with grouping and filtering resources in Kubernetes.

<table>
<thead>
<tr>
<th>

```
K8s Object
```

</th>
<th>

```
Name
```

</th>
<th>

```
Before the changes
```

</th>
<th>

```
With this PR
```

</th>
</tr>
</thead>
<tbody>
<tr>
	<td>

```
DaemonSet
```

</td>
	<td>

```
collector
```

</td>
	<td>

```
{
  "component": "collector",
  "logging-infra": "collector",
  "provider": "openshift"
}
```

</td>
<td>

```
{
  "app.kubernetes.io/component": "collector",
  "app.kubernetes.io/instance": "instance",
  "app.kubernetes.io/managed-by": "cluster-logging-operator",
  "app.kubernetes.io/name": "fluentd | vector",
  "app.kubernetes.io/part-of": "cluster-logging",
  "app.kubernetes.io/version": "5.7.0",

  "component": "collector",
  "logging-infra": "collector",
  "provider": "openshift"
}


```

</td>
</tr>
<tr>
	<td>

Pod

</td>
	<td>

```
collector-***
```

</td>
	<td>

```
{
  "component": "collector",
  "controller-revision-hash": "df5c5d69",
  "logging-infra": "collector",
  "pod-template-generation": "2",
  "provider": "openshift"
}
```

</td>
	<td>

```
{
  "app.kubernetes.io/component": "collector",
  "app.kubernetes.io/instance": "instance",
  "app.kubernetes.io/managed-by": "cluster-logging-operator",
  "app.kubernetes.io/name": "fluentd | vector",
  "app.kubernetes.io/part-of": "cluster-logging",
  "app.kubernetes.io/version": "5.7.0",
  "component": "collector",
  "controller-revision-hash": "6b58bcb6c6",
  "logging-infra": "collector",
  "pod-template-generation": "1",
  "provider": "openshift"
}
```

</td>
</tr>
<tr>
<td>

```
Service
```

</td>
	<td>

```
collector
```

</td>
	<td>

```
{
  "logging-infra": "support"
}
```

</td>
	<td>

```
{
  "app.kubernetes.io/component": "support",
  "app.kubernetes.io/instance": "instance",
  "app.kubernetes.io/managed-by": "cluster-logging-operator",
  "app.kubernetes.io/name": "fluentd | vector",
  "app.kubernetes.io/part-of": "cluster-logging",
  "app.kubernetes.io/version": "5.7.0",

  "logging-infra": "support"
}
```

</td>
</tr>
<tr>
	<td>

`Secret`

</td>
	<td>

collector-config (for Vector)

</td>
	<td>

```
{
  "pod-security.kubernetes.io/enforce": "privileged",
  "security.openshift.io/scc.podSecurityLabelSync": "false"
}
```

</td>
	<td>

```
{
  "app.kubernetes.io/component": "collector",
  "app.kubernetes.io/instance": "instance",
  "app.kubernetes.io/managed-by": "cluster-logging-operator",
  "app.kubernetes.io/name": "vector",
  "app.kubernetes.io/part-of": "cluster-logging",
  "app.kubernetes.io/version": "5.7.0",

  "pod-security.kubernetes.io/enforce": "privileged",
  "security.openshift.io/scc.podSecurityLabelSync": "false"
}
```

</td>
</tr>
<tr>
	<td>

```
ConfigMap
```

</td>
	<td>

```
collector (for Fluentd)
```

</td>
	<td>

```
{
  "pod-security.kubernetes.io/enforce": "privileged",
  "security.openshift.io/scc.podSecurityLabelSync": "false"
}
```

</td>
	<td>

```
{
  "app.kubernetes.io/component": "collector",
  "app.kubernetes.io/instance": "instance",
  "app.kubernetes.io/managed-by": "cluster-logging-operator",
  "app.kubernetes.io/name": "fluentd",
  "app.kubernetes.io/part-of": "cluster-logging",
  "app.kubernetes.io/version": "5.7.0",

  "pod-security.kubernetes.io/enforce": "privileged",
  "security.openshift.io/scc.podSecurityLabelSync": "false"
}
```

</td>
</tr>
<tbody>
</table>

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:

### DIFF
--- a/internal/collector/config.go
+++ b/internal/collector/config.go
@@ -25,6 +25,7 @@ func (f *Factory) ReconcileCollectorConfig(er record.EventRecorder, k8sClient cl
 				"run.sh":              fluentd.RunScript,
 				"cleanInValidJson.rb": fluentd.CleanInValidJson,
 			},
+			f.CommonLabelInitializer,
 		)
 		utils.AddOwnerRefToObject(collectorConfigMap, owner)
 		return reconcile.Configmap(k8sClient, k8sClient, collectorConfigMap)
@@ -34,7 +35,9 @@ func (f *Factory) ReconcileCollectorConfig(er record.EventRecorder, k8sClient cl
 			constants.CollectorConfigSecretName,
 			map[string][]byte{
 				"vector.toml": []byte(collectorConfig),
-			})
+			},
+			f.CommonLabelInitializer)
+
 		utils.AddOwnerRefToObject(secret, owner)
 		return reconcile.Secret(er, k8sClient, secret)
 	}
--- a/internal/runtime/core.go
+++ b/internal/runtime/core.go
@@ -3,6 +3,7 @@ package runtime
 import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 )
 
 // NewNamespace returns a corev1.Namespace with name.
@@ -13,12 +14,12 @@ func NewNamespace(name string) *corev1.Namespace {
 }
 
 // NewConfigMap returns a corev1.ConfigMap with namespace, name and data.
-func NewConfigMap(namespace, name string, data map[string]string) *corev1.ConfigMap {
+func NewConfigMap(namespace, name string, data map[string]string, visitors ...func(o runtime.Object)) *corev1.ConfigMap {
 	if data == nil {
 		data = map[string]string{}
 	}
 	cm := &corev1.ConfigMap{Data: data}
-	Initialize(cm, namespace, name)
+	Initialize(cm, namespace, name, visitors...)
 	return cm
 }
 
@@ -30,9 +31,9 @@ func NewPod(namespace, name string, containers ...corev1.Container) *corev1.Pod 
 }
 
 // NewService returns a corev1.Service with namespace and name.
-func NewService(namespace, name string) *corev1.Service {
+func NewService(namespace, name string, visitors ...func(o runtime.Object)) *corev1.Service {
 	svc := &corev1.Service{}
-	Initialize(svc, namespace, name)
+	Initialize(svc, namespace, name, visitors...)
 	return svc
 }
 
@@ -44,18 +45,18 @@ func NewServiceAccount(namespace, name string) *corev1.ServiceAccount {
 }
 
 // NewSecret returns a corev1.Secret with namespace and name.
-func NewSecret(namespace, name string, data map[string][]byte) *corev1.Secret {
+func NewSecret(namespace, name string, data map[string][]byte, visitors ...func(o runtime.Object)) *corev1.Secret {
 	if data == nil {
 		data = map[string][]byte{}
 	}
 	s := &corev1.Secret{Data: data}
-	Initialize(s, namespace, name)
+	Initialize(s, namespace, name, visitors...)
 	return s
 }
 
 // NewDaemonSet returns a daemon set.
-func NewDaemonSet(namespace, name string) *appsv1.DaemonSet {
+func NewDaemonSet(namespace, name string, visitors ...func(o runtime.Object)) *appsv1.DaemonSet {
 	ds := &appsv1.DaemonSet{}
-	Initialize(ds, namespace, name)
+	Initialize(ds, namespace, name, visitors...)
 	return ds
 }
--- a/version/version.go
+++ b/version/version.go
@@ -1,3 +1,3 @@
 package version
 
-var Version = "5.8"
+var Version = "5.8.0"
--- a/Makefile
+++ b/Makefile
@@ -193,7 +193,7 @@ generate: $(GEN_TIMESTAMP)
 $(GEN_TIMESTAMP): $(shell find apis -name '*.go')  $(OPERATOR_SDK) $(CONTROLLER_GEN) $(KUSTOMIZE) .target
 	@$(CONTROLLER_GEN) object paths="./apis/..."
 	@$(CONTROLLER_GEN) crd:crdVersions=v1 rbac:roleName=clusterlogging-operator paths="./..." output:crd:artifacts:config=config/crd/bases
-	echo -e "package version\n\nvar Version = \"$(or $(CI_CONTAINER_VERSION),$(LOGGING_VERSION))\"" > version/version.go
+	echo -e "package version\n\nvar Version = \"$(or $(CI_CONTAINER_VERSION),$(VERSION))\"" > version/version.go
 	@$(MAKE) fmt
 	@touch $@
 
--- a/internal/collector/collector.go
+++ b/internal/collector/collector.go
@@ -1,6 +1,7 @@
 package collector
 
 import (
+	"github.com/openshift/cluster-logging-operator/internal/runtime"
 	"path"
 	"strings"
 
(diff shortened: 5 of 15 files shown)


vparfonov:
@jcantrill 
Hey, quick question! Do we really need to keep the `component` and `login-infra` labels? And, could you explain to me what's the deal with the `provider` label and what value it should have?


https://github.com/openshift/cluster-logging-operator/blob/2cc353d3edc939f6add3893ab855e9b2cf48404a/internal/factory/daemonset.go#L14

internal/collector/config.go
```diff
@@ -14,27 +14,30 @@ import (
...
-			namespace,
+			instance.Namespace,
 			name,
 			map[string]string{
 				"fluent.conf":         collectorConfig,
 				"run.sh":              fluentd.RunScript,
 				"cleanInValidJson.rb": fluentd.CleanInValidJson,
 			},
 		)
+		utils.SetCommonLabels(collectorConfigMap, instance, constants.CollectorName)
```
jcantrill, line 29: I think there may be a more clever way to add these labels by modifying the [runtime.Initialize](https://github.com/openshift/cluster-logging-operator/blob/master/internal/runtime/runtime.go#L90)

```
func Initialize(o runtime.Object, namespace, name string, visitors ... func(meta metav1.Object)  {
	m := Meta(o)
	m.SetNamespace(namespace)
	m.SetName(name)
	o.GetObjectKind().SetGroupVersionKind(GroupVersionKind(o))
	m.SetLabels(map[string]string{
		"pod-security.kubernetes.io/enforce":             "privileged",
		"security.openshift.io/scc.podSecurityLabelSync": "false",
	})
        for _, v := range visitors {
             v(m)
        }
}
```

This will require modification to the other [factory](https://github.com/openshift/cluster-logging-operator/blob/master/internal/runtime/core.go) classes to take a range of visitors but it does not require changes necessarily to all the callers.

As a follow-up, we do not need to:
* we should be able move the parameters added to the reconcile method signiture to the collector factory struct and write a single "visitor" to add labels.  

Thoughts?

jcantrill (changes requested):

internal/collector/config.go
vparfonov: @jcantrill Your proposal seems to make sense and looks more consistent, but I have a question about it. If I can set the `instance name` and `kind` in the `factory struct`, it seems reasonable because those values are immutable. However, I'm unsure about how to pass the component name to the visitor. It seems like this could potentially disrupt the idea of having a single visitor. Could you please clarify this point for me?

jcantrill: I've been trying to use the [examples](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) with our usecase and I'm not exactly certain how this maps or what is useful.  I believe we have it mostly correct in our design but should it be for a CLF instance:
* name: vector|fluentd
* instance: clf.name
* version: clo version
* component: collector
* part-of: cluster-logging
* managed-by: cluster-logging-operator

I believe you could create the visitor when creating the factory and set the visitor as a field on the factory.

vparfonov: @jcantrill I've committed changes according to your comments, check please

internal/runtime/core.go
```diff
@@ -44,18 +47,63 @@ func NewServiceAccount(namespace, name string) *corev1.ServiceAccount {
...
-	Initialize(s, namespace, name)
+	Initialize(s, namespace, name, visitors...)
 	return s
 }
 
-// NewDaemonSet returns a daemon set.
-func NewDaemonSet(namespace, name string) *appsv1.DaemonSet {
-	ds := &appsv1.DaemonSet{}
-	Initialize(ds, namespace, name)
+func NewDaemonSet(daemonSetName, namespace, component string, podSpec corev1.PodSpec, visitors ...func(meta metav1.Object)) *appsv1.DaemonSet {
```
jcantrill, line 59: @alanconway how much do you care about this refactoring with the change in args? My initial understanding was that:

* runtime - Place for "generic" runtime factories
* collector/.... - place for collector specific factories

This change is combining the two to do away with one package that was taken over by 'runtime' .  @vparfonov seems like:

* the label changes should be pushed into the label initializer you introduced
* the DS specific template, scheduling, rollout strategy could be pushed into collector specific DS

internal/runtime/core.go
```diff
@@ -44,18 +47,63 @@ func NewServiceAccount(namespace, name string) *corev1.ServiceAccount {
...
+						Type:   intstr.String,
+						StrVal: "100%",
+					},
+				},
+			},
+		},
+	}
+	Initialize(ds, namespace, daemonSetName, visitors...)
+	//TODO: should we keep this labels?
+	utils.AddLabels(ds, labels)
```
jcantrill, line 103: This should not be needed if we encapsulate the changes into the visitors

internal/runtime/core.go
```diff
@@ -44,18 +47,63 @@ func NewServiceAccount(namespace, name string) *corev1.ServiceAccount {
...
+
+	dl := ds.GetLabels()
+	ds.Spec.Template.SetLabels(dl)
```
jcantrill, line 106: I would expect this to be encapsulated into a visitor as ell

internal/collector/daemonset.go
```diff
@@ -15,7 +15,7 @@ func (f *Factory) ReconcileDaemonset(er record.EventRecorder, k8sClient client.C
 	trustedCABundle, trustHash := GetTrustedCABundle(k8sClient, namespace, constants.CollectorTrustedCAName)
 	f.TrustedCAHash = trustHash
 	tlsProfile, _ := tls.FetchAPIServerTlsProfile(k8sClient)
-	desired := f.NewDaemonSet(namespace, name, trustedCABundle, tls.GetClusterTLSProfileSpec(tlsProfile))
+	desired := f.NewCollectorDaemonSet(namespace, name, trustedCABundle, tls.GetClusterTLSProfileSpec(tlsProfile))
```
jcantrill, line 18: Why does this name need to change?

version/version.go
```diff
@@ -1,3 +1,4 @@
 package version
 
 var Version = "5.7"
+var FullVersion = "5.7.0"
```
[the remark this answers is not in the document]
jcantrill: @alanconway ping...

internal/runtime/core.go
vparfonov: @jcantrill 
Transferring this operation to the visitor will require of identifying the type of K8s object, as the parameter involved is a `runtime.Object`, e.g:
```
gvk := runtime.GroupVersionKind(o)
if gvk.Group == "apps" && gvk.Kind == "DaemonSet" {
    ds := o.(*apps.DaemonSet)
    utils.SetCommonLabelsDaemonSet(ds)
} else {
    utils.SetCommonLabels(o)
}
```
This is the reason why I prefer not to move this logic to the visitor.

internal/collector/daemonset.go
vparfonov: rollback

version/version.go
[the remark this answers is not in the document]
alanconway: I don't think there's any problem with x.y.z version for telemetry. I can't be certain, but there's no strict schema for any of the telemetry properties. The telemetry platform deals with info from so many sources that I'd be very surprised if it had hard-coded assumptions about the number of components in the many version strings it sees.

internal/runtime/core.go
alanconway: I like the visitors... approach. It is a very common way to express "options" in Go. It doesn't change the generic nature of runtime, but it lets the caller inject additional setup on the fly. Not sure I understood your question @jcantrill

internal/runtime/core.go
alanconway: Looks like the context for these comments has changed, I don't see a `SetCommonLabelsDaemonSet` anymore.

PR #1954, opened 2023-04-14
LOG-3948: [Vector] No error message shown on collector pods when the syslog output url is wrong in CLF

syedriko:
### Description
Before this PR, when vector forwarded logs to a syslog endpoint, the scheme of the endpoint URL defaulted to 'tcp' (including 'tls') unless it was 'udp'. No error was produced for an invalid URL scheme.
With this PR, a URL scheme that is not 'tcp', 'tls' or 'udp', results in the status of the corresponding ClusterLogForwarder output set to False with an error message:

```
kind: ClusterLogForwarder
metadata:
  name: instance
  namespace: openshift-logging
spec:
  outputs:
  - name: rsyslog
    syslog:
      facility: local0
      rfc: RFC5424
      severity: informational
    type: syslog
    url: xyz://rsyslogd.syslog.svc.cluster.local:24224
...
status:
  outputs:
    rsyslog:
    - lastTransitionTime: "2023-04-14T21:27:26Z"
      message: 'invalid URL scheme: xyz'
      reason: Invalid
      status: "False"
      type: Ready
```

/cc @cahartma
/assign @jcantrill

### Links
- JIRA: https://issues.redhat.com/browse/LOG-3948

### DIFF
--- a/internal/generator/vector/output/syslog/syslog.go
+++ b/internal/generator/vector/output/syslog/syslog.go
@@ -18,7 +18,7 @@ import (
 const (
 	SyslogComponentType = `syslog`
 	TCP                 = `tcp`
-	UDP                 = `udp`
+	TLS                 = `tls`
 	RFC3164             = `rfc3164`
 	RFC5424             = `rfc5424`
 )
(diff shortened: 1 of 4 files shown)


jcantrill:
Please add a validation so users get an earlier failure then a pod collector not starting.  Ref: https://github.com/openshift/cluster-logging-operator/blob/master/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go#L255

PR #1955, opened 2023-04-19
LOG-3964: Update image to UBI9

jcantrill:
### Description
This PR:
* updates the base image to ubi9 for dev builds
* removes Dockerfile.in and linting of it since it is not being kept in sync


Dockerfile
```diff
@@ -29,11 +21,9 @@ COPY ${APP_DIR}/internal ./internal
 USER 0
 RUN make build
 
-#@follow_tag(registry-proxy.engineering.redhat.com/rh-osbs/openshift-ose-cli:v4.9)
-FROM quay.io/openshift/origin-cli:4.12 AS origincli
+FROM quay.io/openshift/origin-cli:4.13 AS origincli
 
-#@follow_tag(registry.redhat.io/ubi8:latest)
-FROM registry.redhat.io/ubi8:8.7
+FROM registry.redhat.io/ubi9/ubi:9.1
```
xperimental, line 26: We're already using 9.2 in our midstream builds. My suggestion would be to use "latest" as the tag to get the newest version:

```suggestion
FROM registry.redhat.io/ubi9/ubi:latest
```

Dockerfile
```diff
@@ -1,13 +1,5 @@
...
-#@follow_tag(registry-proxy.engineering.redhat.com/rh-osbs/openshift-golang-builder:rhel_8_golang_1.18)
-FROM registry.redhat.io/ubi8/go-toolset:1.18.9 AS builder
-
-ENV BUILD_VERSION=${CI_CONTAINER_VERSION}
-ENV OS_GIT_MAJOR=${CI_X_VERSION}
-ENV OS_GIT_MINOR=${CI_Y_VERSION}
-ENV OS_GIT_PATCH=${CI_Z_VERSION}
-ENV SOURCE_GIT_COMMIT=${CI_CLUSTER_LOGGING_OPERATOR_UPSTREAM_COMMIT}
-ENV SOURCE_GIT_URL=${CI_CLUSTER_LOGGING_OPERATOR_UPSTREAM_URL}
+FROM registry.redhat.io/ubi9/go-toolset:1.18.10-4 AS builder
```
xperimental, line 1: We're using Go 1.20.3 in midstream already. There does not seem to be a `go-toolset` image for that yet, so maybe `docker.io/library/golang` is an alternative?

xperimental (commented):
See inline comments, one additional comment:

What about the `Dockerfile.centos.patch` file? Is this still needed?

PR #1958, opened 2023-04-20
LOG-4204: Implement HTTP server API

alanconway:
API changes only, no implementation.


apis/logging/v1/input_source_types.go
```diff
@@ -0,0 +1,52 @@
+package v1
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// NOTE: The Enum validation on SourceSpec.Type must be updated if the list of types changes.
```
jcantrill, line 7: Don't we actually need type in the spec?  Will this break upgrades in future if we add others and existing deployments don't have "type" set?

apis/logging/v1/input_source_types.go
```diff
@@ -0,0 +1,52 @@
...
+
+// SourceSpec is a union of input source types.
+//
+// The fields of this struct define the set of known source types.
+type SourceSpec struct {
+	HTTP *HTTPSource `json:"http,omitempty"`
+}
+
+const (
+	FormatK8S  = "k8s"  // Log events in k8s list format, e.g. API audit log events.
```
jcantrill, line 17: maybe we should be more specific.  Isn't this actually a serialized kubernetes event as yaml?  If so then we should name it something like kubernetesEvent or the like

apis/logging/v1/input_source_types.go
```diff
@@ -0,0 +1,52 @@
...
+	FormatText = "text" // Newline-separated plain-text log records.
+)
+
+// HTTPSource receives encoded logs as a HTTP endpoint.
+type HTTPSource struct {
```
jcantrill, line 22: Do we need to be able to support auth of some kind (i.e. certificates, token).  If so maybe it goes up a level on source

apis/logging/v1/input_source_types.go
```diff
@@ -0,0 +1,52 @@
...
+	// the name must be a unique Service name in the ClusterLogForwarder's namespace.
+	//
+	// +required
+	Name string `json:"name"`
+
+	// Port that this source will listen on.
+	// Used to create a Service for this port.
+	//
+	// +required
+	Port corev1.ServicePort `json:"port"`
```
jcantrill, line 40: Do we really want this to be `corev1.ServicePort` ?  Does an admin need to really be capable of defining all the fields of ServicePort?  Aren't we in control meaning we will create the service that fronts the pod so we can only need them to find the outward facing service port and we control the rest

apis/logging/v1/input_source_types.go
alanconway: Since the spec is never empty, the type is redundant 
`{ http: { blah } }` vs. `{ type:http, http: {blah}}`
However if think we should  include a type for consistency with similar patterns elsewhere then I could be convinced.

apis/logging/v1/input_source_types.go
alanconway: Yes, this needs more thought. I was thinking of a generic k8s "list" format, using the standard k8s resource list YAML. The type of object in the list would be indicated by the standard meta GVK fields.
We could be more specific "Event.audit.k8s.io/v1` , but I'm not sure how specific we want to be at this point.

apis/logging/v1/input_source_types.go
alanconway: Yes - thats To Be Done, see FIXME comment below;

apis/logging/v1/input_source_types.go
alanconway: I was thinking that this _is_ the outward facing service port, we would create a service with this port for the user. We are in control of how that service is linked to the Pod. If you have an alternative in mind I'd like to see it, I may be off in the weeds here.

apis/logging/v1/input_source_types.go
jcantrill: Maybe the better question is to ask if we ever intend to support a different input source (YAGNI?)  ?  We mostly are defining this input for internal usage and we are dictating the transport and protocol.  I don't honestly see why we need to consider supporting multiple and it benefits us to simply state: this is what it is and if you want to talk to us this is how you do it.

apis/logging/v1/input_source_types.go
jcantrill: If we do not wish to be specific in this way and infer the content then the format maybe should be JSON (ndjson?) with the understanding we will infer what it is though I like the idea of being explict:

* kubernetes event json
* if the parse or deserialization fails then its the senders fault because they sent us the wrong thing

apis/logging/v1/input_source_types.go
jcantrill: My implication was that " corev1.ServicePort" is more then an integer for the port number which opens up the API for additional configuration.  Do we need all that config or can we default to someone providing a port number? Can we default a port number and make an optional override?

alanconway: willing to review, lets continue on https://github.com/openshift/cluster-logging-operator/pull/2155

alanconway:
Lets move the discussion to https://github.com/openshift/cluster-logging-operator/pull/2155
The API has already been absorbed into that PR which has changes from this original proposal. This PR won't be merged independently at this point.

apis/logging/v1/input_source_types.go
alanconway: Check https://github.com/openshift/cluster-logging-operator/pull/2155 which has some changes in this regard. For hypershift we need at least k8s list format (events are in array .items[]) For general use we should have ndjson and maybe generalize to allow the user to specify an array field name for list formats that are similar to k8s but not identical. My minimum goal at this point is to release a preview that will handle the hypershift case and then add further minor features on z-stream.

docs/features/logforwarding/sources/http-source.adoc
```diff
@@ -0,0 +1,23 @@
+=== Receiving logs over HTTP
```
abrennan89, line 1: @alanconway it looks like this never actually made it into the docs at https://github.com/openshift/cluster-logging-operator/tree/master/docs/features/logforwarding but seems like it would be good source material for https://issues.redhat.com/browse/OBSDOCS-212 - is the content here still accurate?

PR #1959, opened 2023-04-20
LOG-3982: ClusterLogForwarder API audit policy

alanconway:
[LOG-3982](https://issues.redhat.com//browse/LOG-3982): Kube audit policy API and implementation.

See upstream doc: docs/features/logforwarding/filters/api-audit-filter.adoc

- Introduced filters section in CLF.
- Added API Audit policy filter type.
- Policy-to-VRL generator in internal/generator/vector/apiaudit

This is a re-implementation of https://gitlab.cee.redhat.com:service/splunk-audit-exporter.git in Vector Remap Language.
The following features are not implemented:
  - Redactions (TODO)
  - Drop duplicate update events (TODO)
  - Drop events before time (CLO future).
  - Adding annotations (CLO future)
  - Metrics (Separate process)


apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -169,9 +168,14 @@ type Application struct {
...
-// Audit enables audit logs. Filtering may be added in future.
-// +docgen:ignore
-type Audit struct{}
+type Audit struct {
+	// APIPolicyRef is the name of an APIAuditLogPolicy resource in the same name-space.
+	APIPolicyRef string `json:"apiPolicyRef,omitempty"`
+
+	// EnableNode enables operating system audit logs from nodes.
+	// If not set, node audit logs are not forwarded.
+	EnableNode bool `json:"enableNode,omitempty"`
```
jcantrill, line 177: - Does this need to default to "true" to support upgrade?  
- This says "OS audit logs" which are... linuxd logs?  is there an implication here? How do kube logs fit into this?
- Can there be more then one like user defined app logs?  Should this have a name too?

config/crd/bases/logging.openshift.io_apiauditlogfilters.yaml
```diff
@@ -0,0 +1,192 @@
...
+spec:
+  group: logging.openshift.io
+  names:
+    categories:
+    - logging
+    kind: APIAuditLogFilter
+    listKind: APIAuditLogFilterList
+    plural: apiauditlogfilters
+    singular: apiauditlogfilter
+  scope: Namespaced
```
jcantrill, line 18: I'm playing @xperimental here... should this be cluster scoped?

internal/generator/vector/apiaudit/comaptibility_test.go
```diff
@@ -0,0 +1,56 @@
...
+	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
+)
+
+var _ = Describe("policy filter compatibility", func() {
+	It("generates the same output as git@gitlab.cee.redhat.com:service/splunk-audit-exporter.git", func() {
+		// Note: test data was generated as follows:
+		//	cd testdata; audit-exporter --follow=false --input audit_in.log  --policy example_policy.yaml  > audit_out_exporter.log
+
+		// Read events from audit_in.log
+		in, err := os.Open("testdata/audit_in.log")
```
jcantrill, line 20: You can maybe convert this to use "embed"

internal/generator/vector/apiaudit/helpers_test.go
```diff
@@ -0,0 +1,113 @@
...
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+
+	"github.com/openshift/cluster-logging-operator/test"
+	"k8s.io/apimachinery/pkg/runtime"
+	. "k8s.io/apiserver/pkg/apis/audit/v1"
+	"sigs.k8s.io/yaml"
+)
+
+const vectorImage = "quay.io/openshift-logging/vector:5.8.0"
```
jcantrill, line 20: There is a function in utils to get component images by name

internal/generator/vector/apiaudit/helpers_test.go
```diff
@@ -0,0 +1,113 @@
...
+// vectorCmd creates a temp dir for vector to use and returns the exec.Cmd
+func vectorCmd(p *AuditPolicy) *exec.Cmd {
+	vrl, err := PolicyToVRL(p)
+	Expect(err).NotTo(HaveOccurred(), "%#v", *p)
+	dir, err := os.MkdirTemp("", "apiaudit_test")
+	GinkgoT().Cleanup(func() { _ = os.RemoveAll(dir) })
+	Expect(err).NotTo(HaveOccurred())
+	Expect(os.WriteFile(filepath.Join(dir, "vector.toml"), []byte(vectorConf), 0666)).To(Succeed())
+	Expect(os.WriteFile(filepath.Join(dir, "audit_policy.vrl"), []byte(vrl), 0666)).To(Succeed())
+	return exec.Command("podman", "run", "-i", "--rm", "--env", "VECTOR_LOG=warn", "-v", dir+":/etc/vector:Z", vectorImage)
```
jcantrill, line 47: Do we really want to bind our unit tests to podman?  I'm fairly certain podman is not available on CI.  We may be better served either making some flavor of a functional test or figuring out how to prep CI to download vector  and run these.  Maybe  a special Docker image we build in CI which combines vector and adds the source?

apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -45,6 +45,11 @@ type ClusterLogForwarderSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Forwarder Outputs",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:forwarderOutputs"}
 	Outputs []OutputSpec `json:"outputs,omitempty"`
 
+	// Filters are applied to log records passing through a pipeline.
+	// There are different types of filter that can select and modify log records in different ways.
+	// See [FilterTypeSpec] for a list of filter types.
+	Filters []FilterSpec `json:"filters,omitempty"`
```
jcantrill, line 51: Maybe mention they are applied in order as defined

jcantrill, line 51: Audit says this enables filters.  Maybe mention they are only valid for pipelines that enable audit and they only apply to audit. To simplify coding do we maybe need to reject pipelines that enable filters and define more inputs then audit?

apis/logging/v1/filter_api_audit_types.go
```diff
@@ -0,0 +1,69 @@
...
+//
+// Policy event rules are the same format as the [Kube Audit Policy] with some minor extensions.
+// The extensions are described here, see the [Kube Audit Policy] for the standard rule behavior.
+// Rules are checked in order, checking stops at the first matching rule.
+//
+// The `level` of the rule determines how the event is treated:
+//   - None: the event is dropped.
+//   - Metadata: Audit metadata only, request and response objects are removed.
+//   - Request: Audit metadata and request are retained, response is removed.
+//   - RequestResponse: All data is retained, including the full response. This can lead to very large events.
```
jcantrill, line 17: Should this be: Full, Complete, All, Entire since its everything?

apis/logging/v1/cluster_log_forwarder_types.go
alanconway: Removed in latest version. Policy is a "filter", no changes to inputs. Policy filter can operate on any stream of logs, will only act on records that look like API audit events.

apis/logging/v1/filter_api_audit_types.go
```diff
@@ -0,0 +1,69 @@
...
+// If you want to disable these defaults, end your rules list with rule that has only a `level` field.
+// An empty rule matches any event, and prevents the defaults from taking effect.
+//
+// ## Omit Response Codes
+//
+// You can drop events based on the HTTP status code in the response. See the OmitResponseCodes field.
+//
+// [Kube Audit Policy]: https://kubernetes.io/docs/reference/config-api/apiserver-audit.v1/#audit-k8s-io-v1-Policy
+// [Kubernetes Auditing]: https://kubernetes.io/docs/tasks/debug/debug-cluster/audit/
+type APIAudit struct {
```
jcantrill, line 47: Is there an order of preference to apply the rules, stages, and code that needs to be documented here

config/crd/bases/logging.openshift.io_apiauditlogfilters.yaml
alanconway: Gone in latest version - policy is all inside the 'filters' section of  ClusterLogForwarder resource, no separate resource.

internal/generator/vector/apiaudit/helpers_test.go
alanconway: fixed

internal/generator/vector/apiaudit/helpers_test.go
alanconway: Will convert to functional unless I can think of a better solution, which I can't so far.

apis/logging/v1/cluster_log_forwarder_types.go
alanconway: Done.

alanconway: The filter actually can apply to logs from any source, it will ignore non-audit logs. It doesn't make sense to use on a pipeline unless it can receive audit logs but it's not destructive, and I think enforcing the restriction is more complexity.

apis/logging/v1/filter_api_audit_types.go
alanconway: Which bit? It says: _All_ data is retained, including the _full_ response

apis/logging/v1/filter_api_audit_types.go
alanconway: Done in latest version, also a docs/... entry for more overview.

apis/logging/v1/cluster_log_forwarder_types.go
jcantrill: Enforcing is a simple check on our validate code to reject but I defer to you here.  I'm looking to avoid the confusion of "why does my filter not applied to application log message"

alanconway: Gotcha, will review that and either enforce or document.

apis/logging/v1/filter_api_audit_types.go
jcantrill: I'm suggesting the enum type should maybe be one of: Full, Complete, All, Entire

alanconway: Gotcha -  the enum values come from the k8s audit policy spec so changing them would be confusing. I agree RequestResponse is a bit odd. It does emphasize that you will be including the response body, which is what really blows up the audit records.

alanconway:
This is nearly final:
- I refactored some test code to facilitate the audit tests, presently it's in a separate commit for review.
- I don't use podman, only a pod - check the approach. I create a single idle pod and then use it to run vector commands - its as fast as podman but only requires a cluster. They're not really functional tests, they  use a simple vector config to test the VRL in isolation, but they do require a cluster so maybe it's rude to have them in unit tests. We'll discuss.
- The validations are missing, the tests tied me up longer than expected. Should be trivial to finish.

docs/features/logforwarding/filters/api-audit-filter.adoc
```diff
@@ -0,0 +1,132 @@
+= API Audit Filter
+
+Each call to the Kube or Openshift API server generates an audit event containing  the request, response and details of who made the call.
```
syedriko, line 3: OpenShift

apis/logging/v1/filter_api_audit_types.go
```diff
@@ -0,0 +1,69 @@
...
+// Policy event rules are the same format as the [Kube Audit Policy] with some minor extensions.
+// The extensions are described here, see the [Kube Audit Policy] for the standard rule behavior.
+// Rules are checked in order, checking stops at the first matching rule.
+//
+// The `level` of the rule determines how the event is treated:
+//   - None: the event is dropped.
+//   - Metadata: Audit metadata only, request and response objects are removed.
+//   - Request: Audit metadata and request are retained, response is removed.
+//   - RequestResponse: All data is retained, including the full response. This can lead to very large events.
+//     For example the command `oc list -A pods` results in a response containing the YAML description of every pod in the cluster.
```
syedriko, line 18: audit response (or this reads like the response of the oc command, which is what's intended)

apis/logging/v1/filter_api_audit_types.go
```diff
@@ -0,0 +1,69 @@
...
+	//
+	// If Rules is empty or missing default rules apply, see [KubeAPIAudit]
+	Rules []auditv1.PolicyRule `json:"rules,omitempty"`
+
+	// OmitStages is a list of stages for which no events are created.
+	// Note that this can also be specified per rule in which case the union of both are omitted.
+	// +optional
+	OmitStages []auditv1.Stage `json:"omitStages,omitempty"`
+
+	// OmitResponseCodes is a list of response codes for which no events are created.
```
syedriko, line 61: response codes -> HTTP status codes

apis/logging/v1/filter_api_audit_types.go
```diff
@@ -0,0 +1,69 @@
...
+// [Kube Audit Policy]: https://kubernetes.io/docs/reference/config-api/apiserver-audit.v1/#audit-k8s-io-v1-Policy
+// [Kubernetes Auditing]: https://kubernetes.io/docs/tasks/debug/debug-cluster/audit/
+type KubeAPIAudit struct {
+
+	// Rules specify the audit Level a request should be recorded at.
+	// A request may match multiple rules, in which case the FIRST matching rule is used.
+	// PolicyRules are strictly ordered.
+	//
+	// If Rules is empty or missing default rules apply, see [KubeAPIAudit]
+	Rules []auditv1.PolicyRule `json:"rules,omitempty"`
```
syedriko, line 54: I'm not clear on this:
// +optional
... `json:"...omitempty"`

Do +optional and omitempty always go together?

internal/generator/vector/filter/apiaudit/filter.go
```diff
@@ -0,0 +1,53 @@
...
+
+//go:embed policy.vrl.tmpl
+var policyVRLTemplateStr string
+
+// This template compiles a Go k8s.io/apiserver/pkg/apis/audit/v1.Policy into a vector VRL remap program.
+// The VRL program filters JSON-serialized k8s.io/apiserver/pkg/apis/audit/v1.Event objects according to the policy.
+//
+// NOTES:
+// - Go templates and VRL both use "." to mean the current object, read carefully.
+// - Go templates and VRL both use {{foo}} for substitution, this can be quotwed as {{"{{"}}foo{{"}}"}} if you want VRL substitution.
```
syedriko, line 47: quotwed

internal/generator/vector/filter/apiaudit/policy.vrl.tmpl
```diff
@@ -0,0 +1,69 @@
...
+    # No rule matched, apply default rules for system events.
+    if match(username, r'system:.*') { # System events
+      readonly = r'get|list|watch|head|options'
+      if match(string!(.verb), readonly) {
+	.level = "None" # Drop read-only system events.
+      } else if ((int(.responseStatus.code) < 300 ?? true) && starts_with(username, "system:serviceaccount:"+namespace)) {
+	.level = "None" # Drop write events by service account for same namespace as resource or for non-namespaced resource.
+      }
+      if .level == "RequestResponse" {
+	.level = "Request" # Downgrade RequestResponse system events.
```
syedriko, line 44: alignment

test/functional/filters/apiaudit/vrl/comaptibility_test.go
```diff
@@ -0,0 +1,57 @@
+package apiaudit
```
syedriko, line 1: spelling of the file name

test/helpers/cmd/writer_test.go
```diff
@@ -0,0 +1,32 @@
...
+		out := &bytes.Buffer{}
+		cmd.Stdout = out
+		cmd.Stderr = test.Writer()
+		w, err := NewWriter(cmd)
+		ExpectOK(err)
+
+		ExpectOK(cmd.Start()) // cmd must be running to write to it.
+		_, err = fmt.Fprint(w, "hello world")
+		ExpectOK(err)
+		err = w.Close() // Close stdin causes cat to exit
```
syedriko, line 26: tee?

test/helpers/cmd/writer_test.go
```diff
@@ -0,0 +1,32 @@
...
+		ExpectOK(err)
+		Expect(0).To(Equal(w.ProcessState.ExitCode()))
```
syedriko, line 28: should here be a cmd.Wait()?

apis/logging/v1/filter_types.go
```diff
@@ -0,0 +1,18 @@
...
+// Filter type constants, must match JSON tags of FilterTypeSpec fields.
+const (
+	FilterKubeAPIAudit = "kubeAPIAudit"
+)
+
+// FilterTypeSpec is a union of filter specification types.
+// The fields of this struct define the set of known filter types.
+type FilterTypeSpec struct {
+	// +optional
+	*KubeAPIAudit `json:"kubeAPIAudit,omitempty"`
```
syedriko, line 15: I'm not certain, but don't we need named fields here to tell apart different filter types, once we have > 1 of them?

test/functional/filters/apiaudit/api_audit_filter_test.go
```diff
@@ -0,0 +1,91 @@
+package apiaudit
```
jcantrill, line 1: do we need to restrict this to vector only?

test/functional/filters/apiaudit/vrl/comaptibility_test.go
jcantrill, line 1: vector only?

apis/logging/v1/filter_api_audit_types.go
alanconway: Changed text to be clearer. The response body we are talking about actually is the literal output of the oc command if you use the '-o json' option, it gets included inside the audit event at the RequestResponse level.
The API servers don't really "log" in the normal sense, they dump a complete record of every request they receive and the response they send (by default.) Hence need for filtering...

apis/logging/v1/filter_api_audit_types.go
alanconway: Pretty much. +optional is a kubebuilder annotation, omitempty is for JSON serialization. AFAIK kubebuilder will infer optional if it sees omitempty, but the opposite is not true - the JSON serializer is not aware of kubebuilder comments, and will force a value into the serialized text if there is no omitempty. I think that omitempty by itself might be sufficient but I'm not certain if there are some tools that need the explicit optional - we seem to put them together most of the time.

internal/generator/vector/filter/apiaudit/filter.go
alanconway: nice typo if I do say so myself!

test/functional/filters/apiaudit/vrl/comaptibility_test.go
alanconway: good lord.

test/helpers/cmd/writer_test.go
alanconway: huh?

test/helpers/cmd/writer_test.go
alanconway: It is  included in w.Close().

apis/logging/v1/filter_types.go
alanconway: Nope, with "anonymous" fields like this the type name becomes the field names. Since all the options will be different types, they'll have different names. That said, there is no good reason for the un-named field here so maybe I should remove it by the law of least surprise.

docs/features/logforwarding/filters/api-audit-filter.adoc
```diff
@@ -0,0 +1,132 @@
...
+
+The OCP policy changes what events the API server _emits_, there can only be one such policy per cluster.
+Changing the policy may require permissions that are not available to the customer admin of a managed cluster.
+
+The ClusterLogForwarder filter changes what the log collector _forwards_.
+You can create multiple filters to send different summaries of the same audit stream to different places.
+For example, you could send a detailed stream to the local cluster log store, and a less detailed stream at a remote site.
+Changing the filter requires normal admin access to the ClusterLogForwarder.
+
+== Differences from splunk-audit-exporter
```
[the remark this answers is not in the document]
alanconway: +1 I can move the splunk part to the tests that are designed to check same behavior as the splunk exporter.

internal/generator/vector/filter/apiaudit/filter.go
```diff
@@ -0,0 +1,53 @@
...
+	"text/template"
+
+	loggingv1 "github.com/openshift/cluster-logging-operator/apis/logging/v1"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	auditv1 "k8s.io/apiserver/pkg/apis/audit/v1"
+	"k8s.io/client-go/kubernetes/scheme"
+)
+
+func init() {
+	utilruntime.Must(auditv1.AddToScheme(scheme.Scheme))
```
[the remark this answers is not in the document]
alanconway: I like having it close to the point of use so that unit tests can run, it's' more obvious why its needed etc.
We have some of these central  in main.go but we do have a couple of other examples that are close to the point of use. We also have a bunch of places where we have to add this in test code because it's not done at the point of use, which I don't like so much. There's no cost if we init the same types twice.

test/functional/filters/apiaudit/vrl/comaptibility_test.go
alanconway: The VRL tests don't use the collector, they start vector for themselves. We could make them vector only though, I guess they're running twice now...

docs/features/logforwarding/filters/api-audit-filter.adoc
```diff
@@ -0,0 +1,132 @@
+= API Audit Filter
```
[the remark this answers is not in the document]
alanconway: Should be, I'll look.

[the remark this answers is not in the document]
alanconway: Added to README

test/functional/filters/apiaudit/api_audit_filter_test.go
alanconway: Yes

test/helpers/cmd/writer_test.go
alanconway: Replaced with cat.

jcantrill:
@alanconway error in log:
```
"No image tag defined for component by environment variable","_error":{"msg":"Can't get component image"},"component":"consoleplugin","envVarName":"RELATED_IMAGE_LOGGING_CONSOLE_PLUGIN"}
```

apis/logging/v1/cluster_log_forwarder_types.go
jcantrill: I don't see mention of order

jcantrill: I don't see mention of applying to audit only though I defer to you if it is otherwise clear in the api

apis/logging/v1/filter_types.go
```diff
@@ -0,0 +1,13 @@
+package v1
+
+// NOTE: The Enum validation on FilterSpec.Type must be updated if the list of
+// known types changes.
+
+// FilterTypeSpec is a union of filter specification types.
+// The fields of this struct define the set of known filter types.
+type FilterTypeSpec struct {
```
jcantrill, line 8: We seem to be still missing "Type" to point us to which union is the one to review.  I see we have logic to check that only one is filled but that seems unnecessary if the spec tells us the type

alanconway: Fixed. Only the *target* tests are failing now, and the failures don't seem related to these changes.
Are those tests stable?

jcantrill:
@alanconway This likely is failing because of the verification https://github.com/alanconway/cluster-logging-operator/blob/aef37eb29ffb9b9bd877967851929cd6c537db2c/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go#  If I read the verification correctly it knows nothing about "filters" and subsequently fails

PR #1967, opened 2023-04-25
LOG-3985: make spec.visualization.kibana.replicas optional on Kibana spec type

vparfonov:
### Description

Currently, the `spec.visualization.kibana.replicas` field in the Kibana spec is mandatory, which means that if Kibana is not available, it can cause errors. 
With this PR,  `spec.visualization.kibana.replicas` field can be set to null or omit it entirely when Kibana is not needed. This will prevent the Kubernetes API server from raising errors when trying to validate the Kibana spec.

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com//browse/LOG-3985
- Enhancement proposal:

### DIFF
--- a/apis/logging/v1/clusterlogging_types.go
+++ b/apis/logging/v1/clusterlogging_types.go
@@ -131,8 +131,9 @@ type KibanaSpec struct {
 	Tolerations  []v1.Toleration   `json:"tolerations,omitempty"`
 
 	// Number of instances to deploy for a Kibana deployment
+	// +optional
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Kibana Size",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:podCount"}
-	Replicas *int32 `json:"replicas"`
+	Replicas *int32 `json:"replicas,omitempty"`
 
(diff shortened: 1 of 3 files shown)


apis/logging/v1/clusterlogging_types.go
```diff
@@ -131,6 +131,7 @@ type KibanaSpec struct {
 	Tolerations  []v1.Toleration   `json:"tolerations,omitempty"`
 
 	// Number of instances to deploy for a Kibana deployment
+	// +optional
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Kibana Size",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:podCount"}
 	Replicas *int32 `json:"replicas"`
```
alanconway, line 136: Needs omitempty to avoid serializing "replicas: nil". The default json package only looks at struct tags, it's not aware of the k8s comments like // +optional
```suggestion
	Replicas *int32 `json:"replicas,omitempty"`
```

alanconway (changes requested):

PR #1973, opened 2023-04-27
LOG-3985: Refactor Kibana declaration to use a pointer for improved nullability

vparfonov:
Signed-off-by: Vitalii Parfonov <vparfonov@redhat.com>

### Description
This PR refactors the way we declare the Kibana object in our codebase. Instead of using a direct declaration, we now use a pointer to the Kibana object.
The motivation for this change is to improve nullability and make it easier to omit the Kibana object when it is not needed. 
This also will prevent the Kubernetes API server from raising errors when trying to validate the Kibana spec.

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-3985
- Enhancement proposal:


jcantrill (commented):
/lgtm

PR #1981, opened 2023-04-30
LOG-4009: TLS configuration for multiple Kafka brokers is not created in Vector

syedriko:
### Description
Made the URL field mandatory for Kafka output, as per original design

/cc @Clee2691 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4009

### DIFF
--- a/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
+++ b/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
@@ -257,27 +257,50 @@ func verifyOutputURL(output *loggingv1.OutputSpec, conds loggingv1.NamedConditio
 		conds.Set(output.Name, c)
 		return false
 	}
-	if output.URL == "" {
-		// Some output types allow a missing URL
-		// TODO (alanconway) move output-specific valiation to the output implementation.
-		if output.Type == loggingv1.OutputTypeKafka || output.Type == loggingv1.OutputTypeCloudwatch ||
-			output.Type == loggingv1.OutputTypeGoogleCloudLogging || output.Type == loggingv1.OutputTypeLoki {
-			return true
-		} else {
-			return fail(CondInvalid("URL is required for output type %v", output.Type))
+
+	if output.Type == loggingv1.OutputTypeKafka {
+		brokerUrls := []string{}
+		if output.URL != "" {
+			brokerUrls = append(brokerUrls, output.URL)
 		}
-	}
-	u, err := url.Parse(output.URL)
-	if err != nil {
-		return fail(CondInvalid("invalid URL: %v", err))
-	}
-	if err := url.CheckAbsolute(u); err != nil {
-		return fail(CondInvalid("invalid URL: %v", err))
-	}
-	if output.Type == loggingv1.OutputTypeSyslog {
-		scheme := strings.ToLower(u.Scheme)
-		if !(scheme == `tcp` || scheme == `tls` || scheme == `udp`) {
-			return fail(CondInvalid("invalid URL scheme: %v", u.Scheme))
+		if output.Kafka != nil { // Add optional extra broker URLs.
+			brokerUrls = append(brokerUrls, output.Kafka.Brokers...)
+		}
+		if len(brokerUrls) == 0 {
+			return fail(CondInvalid("no broker URLs specified"))
+		}
+		for _, b := range brokerUrls {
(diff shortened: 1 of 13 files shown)


internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
```diff
@@ -260,7 +260,7 @@ func verifyOutputURL(output *loggingv1.OutputSpec, conds loggingv1.NamedConditio
 	if output.URL == "" {
 		// Some output types allow a missing URL
 		// TODO (alanconway) move output-specific valiation to the output implementation.
-		if output.Type == loggingv1.OutputTypeKafka || output.Type == loggingv1.OutputTypeCloudwatch ||
+		if output.Type == loggingv1.OutputTypeCloudwatch ||
```
jcantrill, line 263: @alanconway I don't believe this is the correct fix.  The original design essentially calls for a kafka configuration to have at least one broker entry instead of a URL.  Shouldn't the fix be to verify kafka has one broker

syedriko: This is what I went by: https://github.com/openshift/cluster-logging-operator/blob/eb22dc222ff11c7f368801f78ca43176c2b65596/apis/logging/v1/output_types.go#L174
// Brokers specifies the list of brokers
// to register in addition to the main output URL

jcantrill: I defer to @alanconway here to approve and release the hold

alanconway: The original design was that kafka could have a URL **or**  a broker entry, so the "kafka" part is only necessary if you need multiple target brokers. Here's the API doc, which could definitely be improved: 
```
// Brokers returns the list of broker endpoints of a kafka cluster.
// The list represents only the initial set used by the collector's kafka client for the
// first connention only. The collector's kafka client fetches constantly an updated list
// from kafka. These updates are not reconciled back to the collector configuration.
// The list of brokers are populated from the Kafka OutputSpec `Brokers` field, a list of
// valid URLs. If none provided the target URL from the OutputSpec is used as fallback.
// Finally, if neither approach works the current collector process will be terminated.
```

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
```diff
@@ -376,7 +376,7 @@ var _ = Describe("Validate clusterlogforwarderspec", func() {
 				},
 			}
 			verifyOutputs(cluster, client, forwarderSpec, clfStatus, extras)
-			Expect(clfStatus.Outputs["aKafka"]).To(HaveCondition("Ready", true, "", ""))
+			Expect(clfStatus.Outputs["aKafka"]).To(HaveCondition("Ready", false, "Invalid", "URL is required for output type kafka"))
```
alanconway, line 379: This check seems wrong - a URL is _not_ required for kafka, the brokers can be specified in the URL _or_ in the kafka.brokers field.

alanconway (commented):
I'm a little confused

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
alanconway: To state it more clearly: the total set of URLs for kafka should include all of
- outputs.url if present
- kafka.brokers if present
Either of outputs.url or kafka.brokers can be missing or empty, but not both.

syedriko: @alanconway The design doc (could you share a link?) sounds like `kafka.brokers` takes precedence over `outputs.url`. 
 I'm assuming we need to generate TLS config if we see at least one TLS URL in the mix?

alanconway: You're right about the API doc, but I am inclined to change the API doc to what I said above. If there's a brokers entry the URL is optional but I think it will be confusing if the URL is included but ignored. +1 on TLS configuration if there are any TLS URLs.

syedriko:
@alanconway Here's a new and impoved version. It does two things - validation looks at both the URL and brokers and the TLS config is generated if there's a TLS/HTTPS URL found anywhere in the mix.

PR #1984, opened 2023-05-02
WIP: [release-5.6] LOG-3249: Fix log parsing and index logic in fluentd

cahartma:
### Description
Two quick fix issues found with fluentd json parsing feature: 
1. Currently when a `structuredTypeKey` is not found, and a `structuredTypeName` is not specified, the log message is still parsed into structured object, but the index is not created AND the original message is not deleted.    This fixes the issue so that the messages are not parsed in that scenario.
2.  When the key `time` is included in the json message, fluentd was dropping that element from the new `structured` field.  This change will ensure the element remains and is now included in the `structured` elements

/cc @Clee2691 @syedriko 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-3249

### DIFF
--- a/internal/generator/fluentd/conf_test.go
+++ b/internal/generator/fluentd/conf_test.go
@@ -576,7 +576,7 @@ var _ = Describe("Testing Complete Config Generation", func() {
   <filter **>
     @type record_modifier
     <record>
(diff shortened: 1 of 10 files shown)


cahartma:
closing in favor 5.7.z version of fix.

PR #1986, opened 2023-05-03
[release-5.7] LOG-3749: Unability to configure nodePlacement and toleration for log…

Clee2691:
### Description
This PR addresses the inability to configure `nodeSelector` and `tolerations` for the `logging-view-plugin`.
This can now be done in the `CRD` under `visualization`.

This also changes the way that `nodeSelector` and `tolerations` are defined for `Kibana` as they are now defined under the same part of the `CRD`.

Example:
```
visualization:
  type: "ocp-console"
  nodeSelector:
    someKey: someVal
  tolerations:
  - key: "example.io/key"
    operator: "Exists"
    effect: "NoSchedule"
  ocpConsole:
```

One other change in this PR is the refactor of `migrate_collection.go` to `migrate_clusterlogging.go` to encapsulate migrating the visualization spec as well.

/cc @vparfonov  @cahartma
/assign @jcantrill

- JIRA: 
  - https://issues.redhat.com/browse/LOG-3749


jcantrill:
Please move these to be on the same level as "type" since they apply to any visualization. Those declared on kibana will have to take precedence for kibana

apis/logging/v1/clusterlogging_types.go
```diff
@@ -124,13 +136,6 @@ type KibanaSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Kibana Resource Requirements",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:resourceRequirements"}
 	Resources *v1.ResourceRequirements `json:"resources"`
 
-	// Define which Nodes the Pods are scheduled on.
-	//
-	// +nullable
-	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Kibana Node Selector",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:nodeSelector"}
-	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
```
jcantrill, line 131: These cannot be removed since it would result in a regression upon upgrade.  We have to:
* mark them as deprecated
* allow them to take precedence over those at the visualization level

The easiest solution for usage is to introduce a migration to migrate these fields into the new fields and then alter the remaining code to rely on the new fields... similar to what we did with the collector

Clee2691: Done

internal/k8shandler/reconciler.go
```diff
@@ -103,6 +103,12 @@ func Reconcile(cl *logging.ClusterLogging, requestClient client.Client, reader c
 			return clusterLoggingRequest.Cluster, fmt.Errorf("unable to create or update logstore for %q: %v", clusterLoggingRequest.Cluster.Name, err)
 		}
 
+		// Migrate visualization spec for Kibana
```
jcantrill, line 106: I don't see why we don't have something like https://github.com/openshift/cluster-logging-operator/blob/a9b7a1c4903704e0e073a6a61f8108f9a8606e9e/internal/migrations/migrate_clusterlogforwarder.go#L10 for ClusterLogging which migrates the collector (and visualization) at the controller level which is where I would have expected us to hook into migrating visualizatin

Clee2691: Actually looking into the migrations a bit more, there is a `migrate_collection.go` file that migrates ClusterLogging: https://github.com/openshift/cluster-logging-operator/blob/a9b7a1c4903704e0e073a6a61f8108f9a8606e9e/internal/migrations/migrate_collection.go#L10

I can move migrating visualization into there to reduce files and code.

This migration is still in the `k8shandler/reconciler.go`:
https://github.com/openshift/cluster-logging-operator/blob/a9b7a1c4903704e0e073a6a61f8108f9a8606e9e/internal/k8shandler/reconciler.go#L252

I don't see why we couldn't move it to the controller and wouldn't be opposed to the refactor.

jcantrill: migrate should be really early in the reconciliation which is what I hope we did for CLF.  and should actually introduce a new function: 
```
clusterLogging.Spec = migrations.MigrateClusterLogging
```

I think they exist where they do now because we do "gets" on the resources from one controller or the other which makes it difficult to locate that code only at the top of one controller

Clee2691: I think this should be captured in another story as it is deviating away from the heart of this specific issue. I propose to keep these changes as is and work on refactoring in another Jira story. What do you think?

jcantrill: In the interest of "minimizing" changes... lets simply introduce a "migrateClusterLoggingSpec" and encapsulate the collection and visualization in that .... and replace the places where we directly call "migrateCollectionSpec" with the new function.  This will suffice for me until we can otherwise refactor

Clee2691: Refactored `migrate_collection.go` to `migrate_clusterlogging.go` and encapsulated migrating visualization within it.

PR #1993, opened 2023-05-05
LOG-3837: Update skipRange to allow updating from 5.4 onwards

xperimental:
### Description

This changes the generated bundle to include a skipRange attribute that limits direct updates to only be able from 5.4 onwards.

### Links

- JIRA: https://issues.redhat.com/browse/LOG-3837


jcantrill (commented):
/lgtm

PR #2002, opened 2023-05-11
LOG-4086: remove default of collection.type

jcantrill:
…type

### Description
This PR:
* fixes the case where upon upgrade vector was rolling out even though fluentd was specified
* removes default of collection.type

### Links
* https://issues.redhat.com/browse/LOG-4086


hack/cr-deprecated.yaml
```diff
@@ -0,0 +1,19 @@
...
+  logStore:
+    type: "elasticsearch"
+    elasticsearch:
+      nodeCount: 1
+      resources:
+        requests:
+          memory: 2Gi
+      redundancyPolicy: "ZeroRedundancy"
+  collection:
+    type: vector
```
xperimental, line 16: I think this CR got mangled in the process:

```suggestion
```

internal/factory/daemonset.go
```diff
@@ -8,11 +8,12 @@ import (
...
-func NewDaemonSet(daemonsetName, namespace, loggingComponent, component string, podSpec core.PodSpec) *apps.DaemonSet {
+func NewDaemonSet(daemonsetName, namespace, loggingComponent, component, impl string, podSpec core.PodSpec) *apps.DaemonSet {
 	labels := map[string]string{
-		"provider":      "openshift",
-		"component":     component,
-		"logging-infra": loggingComponent,
+		"provider":       "openshift",
+		"component":      component,
+		"logging-infra":  loggingComponent,
+		"implementation": impl,
```
xperimental, line 16: This can not be part of the label-set used as the `Selector`, as that set can not be updated while the DaemonSet exists, making a change of implementation without completely deleting the DaemonSet impossible.

A possible solution would be to use the previous label-set as the selector and extend this with the new `implementation` label only for the `ObjectMeta.Labels` values.

PR #2011, opened 2023-05-20
LOG-4121: Vector fails to run when configuring syslog forwarding for audit log

syedriko:
### Description
Add a namespace for the transforms generated for the CLF pipelines to avoid user-specified pipeline names clashing with names of "reserved" transforms, such as "infrastructure", "application", "audit", "drop_journal_logs", etc.

/cc @Clee2691 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4121

### DIFF
--- a/internal/generator/vector/pipelines.go
+++ b/internal/generator/vector/pipelines.go
@@ -24,8 +24,7 @@ func Pipelines(spec *logging.ClusterLogForwarderSpec, op generator.Options) []ge
 	userDefined := spec.InputMap()
 	for i, p := range spec.Pipelines {
 		inputs := []string{}
-		// Sanitize pipeline name and persist it for config generation
-		p.Name = helpers.FormatComponentID(p.Name)
+		p.Name = helpers.FormatComponentID(p.Name) + "_user_defined"
 		spec.Pipelines[i].Name = p.Name
 
 		for _, inputName := range p.InputRefs {
--- a/internal/generator/vector/conf_test.go
+++ b/internal/generator/vector/conf_test.go
@@ -378,7 +378,7 @@ type = "route"
 inputs = ["application"]
 route.mytestapp = '.kubernetes.namespace_name == "test-ns"'
 
-[transforms.pipeline]
+[transforms.pipeline_user_defined]
 type = "remap"
 inputs = ["route_application_logs.mytestapp","infrastructure","audit"]
 source = '''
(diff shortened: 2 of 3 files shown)


internal/generator/vector/pipelines.go
```diff
@@ -24,8 +24,7 @@ func Pipelines(spec *logging.ClusterLogForwarderSpec, op generator.Options) []ge
 	userDefined := spec.InputMap()
 	for i, p := range spec.Pipelines {
 		inputs := []string{}
-		// Sanitize pipeline name and persist it for config generation
-		p.Name = helpers.FormatComponentID(p.Name)
+		p.Name = helpers.FormatComponentID(p.Name) + "_user_defined"
```
jcantrill, line 27: The conflict is the input name is the same as the pipeline and we use that name and pipeline name for routing.  @alanconway opinion here? Maybe we should do one or more of the following:

* preface our reserved names with an underscore
* add validation to not allow inputs/pipelines/output names to match reserved names
* add validation to not allow inputs/pipelines/output begin with underscore

syedriko: I don't think there's a need to change anything user-facing here. The entire issue is due to vector config generating [transform.<something>] with different <something>s clashing. A suffix added behind the scenes prevents the clash by separating user-defined names into their own namespace.

syedriko: There's a sister issue https://issues.redhat.com/browse/LOG-4120 where we used to allow anonymous pipelines and now don't. Customer sees this as a compatibility break, not without reason.

jcantrill: @syedriko I'm not debating about the need or how it was solved but the choice of the suffix.  It is unlikely someone may choose what you have here but wondering if we still would have a conflict

syedriko: @jcantrill It's impossible to have a conflict. We add the suffix after the user chose their identifier, even if they say foo_our_suffix, we will end up with foo_our_suffix_our_suffix. The suffix is there to make sure we move user-supplied name into their own scope, so they can't clash with "infrastructure", "application", "audit", "drop_journal_logs", etc.

syedriko: The suffix (or prefix) can be anything.

PR #2023, opened 2023-05-25
LOG-4095: fix loki labelKeys that contain slash

jcantrill:
### Description
This PR:
* fixes CLF loki label keys to replace slash with underscore

### Links
https://issues.redhat.com/browse/LOG-4095


alanconway (approved):
/lgtm
your secret is safe with me ;)

PR #2033, opened 2023-05-31
LOG-4091: fix input matchLabels with slash

jcantrill:
### Description
This PR:
* fixes input match labels that contain slash

### Links
* https://issues.redhat.com/browse/LOG-4091

### DIFF
--- a/internal/generator/vector/inputs.go
+++ b/internal/generator/vector/inputs.go
@@ -18,7 +18,7 @@ const (
 	NsDefault   = "default"
 
 	K8sNamespaceName = ".kubernetes.namespace_name"
-	K8sLabelKeyExpr  = ".kubernetes.labels.%s"
+	K8sLabelKeyExpr  = ".kubernetes.labels.%q"
 
 	InputContainerLogs   = "container_logs"
 	InputJournalLogs     = "journal_logs"
--- a/internal/generator/vector/sources_to_pipelines_test.go
+++ b/internal/generator/vector/sources_to_pipelines_test.go
@@ -228,7 +228,7 @@ source = '''
 [transforms.route_application_logs]
 type = "route"
 inputs = ["application"]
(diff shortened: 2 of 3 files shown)


internal/generator/vector/inputs.go
```diff
@@ -18,7 +18,7 @@ const (
 	NsDefault   = "default"
 
 	K8sNamespaceName = ".kubernetes.namespace_name"
-	K8sLabelKeyExpr  = ".kubernetes.labels.%s"
+	K8sLabelKeyExpr  = ".kubernetes.labels.%q"
```
alanconway, line 21: We have  a lot of different string literal quoting syntaxes to deal with, so we need to be clear which is which.
Suggest  at least a comment here that this is an attempt to generate a valid VRL path.
Here you are mixing
-  [Go interpreted string literal syntax from %q](https://go.dev/ref/spec#interpreted_string_lit)  
-  [VRL interpreted string literal syntax as part of a VRL path](https://vector.dev/docs/reference/vrl/expressions/#string) 
The quoting rules are _almost_ but not _exactly_ the same, so this will definitely work better than what we do now, but will crash vector if someone ever uses a label with "{{" in it, since Go won't quote that, and vector treats {{ as special. That _probably_ isn't a problem since "{" isn't legal in a DNS name, but at some point we should  sit down and analyze the formal specs for all the strings we deal with (URL escapes, DNS names, VRL, ruby, Golang, loki-label, non-standard or under-specified JSON paths...) and write explicit quote/unquote functions.

Hey - that sounds like a job for my new intern :)

alanconway (approved):
/lbtiub (Looks Better Than It Use To Be)
Definitely an improvement on current code, but raises questions about a more robust approch to "difficult" characters for the future.

/lgtm

internal/generator/vector/inputs.go
jcantrill: Kube should reject `if someone ever uses a label with "{{` since it is not a valid kubernetes label AFIK

PR #2037, opened 2023-06-02
LOG-4185: Always prefer collection.logs.* over collection.

jcantrill:
### Description
This PR:
* modifies migration to always prefer the deprecated collection.logs.* over collection.*
* This breaks any users that have both collection.logs.* over collection.* defined

### Links
* https://issues.redhat.com/browse/LOG-4185

cc @alanconway @libander @anpingli


cahartma (commented):
/lgtm

PR #2043, opened 2023-06-06
LOG-4031: Add feature doc for JSON parse and ES index creation

cahartma:
### Description
There has been confusion over the functionality of json parsing and ES index creation.    This document attempts to clear up those support requests.   Will also need to modify our downstream public docs to clear up some confusion over what is required or not.

/cc @syedriko @Clee2691 @vparfonov 
/assign @jcantrill 

### Links
- https://issues.redhat.com/browse/LOG-4031

### DIFF
--- a/docs/features/logforwarding/parse-json-structured-logs.adoc
+++ b/docs/features/logforwarding/parse-json-structured-logs.adoc
@@ -74,7 +74,8 @@ The new record with JSON parsing enabled replaces `message` field with a `struct
 
 === Output type: Elasticsearch
 
-Elasticsearch requires that JSON records with different structures or schemas *must* go to separate indices.
+Elasticsearch requires that keys must have the same types (e.g. object, int, string),
+thus JSON records with different structures or schemas *must* go to separate indices.
 This is a *required* part of the configuration.
 Type conflicts and cardinality problems can occur when a schema includes different field names and types.
 
--- a/docs/features/logforwarding/parse-json-structured-logs.adoc
+++ b/docs/features/logforwarding/parse-json-structured-logs.adoc
@@ -3,7 +3,15 @@
 
 Allows structured JSON log entries to be forwarded as JSON objects.
 
-=== ClusterLogForwarder configuration
+*WARNING: Advanced Usage with Potential Implications*
+
+This section describes an advanced usage of the API that can adversely impact cluster performance, stability, and resources.
+When using Elasticsearch, shard explosion is a possibility under certain conditions.
+
+Index creation and parsing with Elasticsearch should only be executed by experienced users who have assessed the shard explosion possibility.
+Please use caution and ensure a thorough understanding of the potential risks involved.
+
+=== ClusterLogForwarder Configuration
 
 *pipelines* field:
 
@@ -34,7 +42,7 @@ spec:
 * For each pipeline, if `parse: json` is set _and_ the log is valid JSON, the output record will include a `structured` field _equivalent_ to the JSON entry.
 
 
-=== Relationship between `structured` and `message` fields:
+==== Relationship between `structured` and `message` fields:
 
 * The original `message` field will be removed when the new `structured` field is included.
 * If there is no valid JSON data found, the `structured` field will not be present.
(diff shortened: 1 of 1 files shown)


docs/features/logforwarding/parse-json-structured-logs.adoc
```diff
@@ -0,0 +1,152 @@
...
+{"structured": {"level":"info","method":"get","source":"homepage"},
+ "more fields..."}
+----
+
+
+
+
+=== Output type: Elasticsearch
+
+With Elasticsearch, JSON records with different formats *must* go to different indices, otherwise type conflicts and cardinality problems can occur.
```
jcantrill, line 71: We should probably clarify 'different formats'.  The gist is like a database schema.  Maybe we can say something like same sets of keys for values of the same type???

docs/features/logforwarding/parse-json-structured-logs.adoc
```diff
@@ -0,0 +1,152 @@
...
+* Records with missing or empty `logFormat` label.
+* Records that could not be parsed as JSON,  _even if_ they have a `logFormat` label.
+
+==== Elasticsearch Implementation Details
+It is *important* not to overload elasticsearch with too many indices.
+
+Only use distinct structured types for distinct log _formats_, *not* for each application or namespace.
+For example, most Apache applications use the same JSON log format and should use the same structured type, for example "LogApache".
+
+It is up to the user to ensure that logs sent to an index are consistent, otherwise index explosion and type confusion are still possible.
```
jcantrill, line 148: what do we mean by "type confusion"

docs/features/logforwarding/parse-json-structured-logs.adoc
```diff
@@ -0,0 +1,152 @@
...
+
+Structured indices are created automatically by the managed default store.
+In order to forward to an external Elasticsearch instance, indices must be created in advance.
```
jcantrill, line 151: This may not be correct.  I believe ES has the ability to dynamically create indices.  The caveat here is internally we rely upon index rollover and writing to an alias.  Customers should configure their cluster to do the same which requires them to:
*  create the initial index
* define the alias
* configure rollover
* possibly include custom mappings

docs/features/logforwarding/parse-json-structured-logs.adoc
```diff
@@ -0,0 +1,152 @@
...
+----
+
+*Note*: In the example above, only _structured_ logs that contain a `logFormat` label go to the `google` or `apache` index.
+All others go to the default application index as _unstructured_ records, including:
+
+* Records with missing or empty `logFormat` label.
+* Records that could not be parsed as JSON,  _even if_ they have a `logFormat` label.
+
+==== Elasticsearch Implementation Details
+It is *important* not to overload elasticsearch with too many indices.
```
jcantrill, line 143: This needs to be some sort of BIG FAT WARNING NOTE.  Customers should only do this if they fully understand the implications.  This is an advanced usage of the API (one I have always had serious concerns about) and they really need to understand  the shard explosion possibility

docs/features/logforwarding/parse-json-structured-logs.adoc
cahartma: changed to:

> It is up to the user to ensure that logs sent to an index are consistent.  Otherwise, it may result in index explosion and data inconsistency.

docs/features/logforwarding/parse-json-structured-logs.adoc
cahartma: agreed

docs/features/logforwarding/parse-json-structured-logs.adoc
cahartma: This is good info, thanks.  Included

jcantrill (commented):
/approve

docs/features/logforwarding/parse-json-structured-logs.adoc
```diff
@@ -0,0 +1,166 @@
...
+[source,json]
+----
+{"structured": {"level":"info","method":"get","source":"homepage"},
+ "more fields..."}
+----
+
+
+=== Output type: Elasticsearch
+
+Elasticsearch requires that JSON records with different structures or schemas *must* go to separate indices.
```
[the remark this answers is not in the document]
cahartma: added a line to clarify what ES requires.  I think we can leave off those specific bullets for now.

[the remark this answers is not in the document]
cahartma: > Elasticsearch requires that keys must have the same types (e.g. object, int, string),
thus JSON records with different structures or schemas *must* go to separate indices.

alanconway (approved):
/lgtm
/unhold

cahartma:
I've really been wanting to try this...
/joke

PR #2067, opened 2023-06-27
LOG-4260: Add comparator function for secrets to compare data and labels during reconcile

vparfonov:
### Description
This PR, add the comparator function for Secrets to checking secrets data and labels.
Addresses the issue where the reconcile function for Secrets only checked the data of the Secret, but did not include label checks. The problem was observed after upgrading logging from version 5.7 to 5.8 common labels were not added

/cc 
/assign @cahartma 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-4260
- Enhancement proposal:


internal/utils/comparators/secrets/comparator.go
```diff
@@ -0,0 +1,35 @@
...
+	CompareLabels      ComparisonOption = 1
+)
+
+// AreSame compares secrets for equality and return true equal otherwise false.  This comparison
+// only compares the data of the secrets by default unless otherwise configured
+func AreSame(actual *corev1.Secret, desired *corev1.Secret, options ...ComparisonOption) bool {
+	log.V(5).Info("Compare secret", "actual", actual)
+	log.V(5).Info("Compare secret", "desired", desired)
+	log.V(5).Info("Compare secret", "options", options)
+	dataAreEqual := reflect.DeepEqual(actual.Data, desired.Data)
```
cahartma, line 22: nit, how about return here if false, saving the extra checks

internal/utils/comparators/secrets/comparator.go
```diff
@@ -0,0 +1,35 @@
...
+	annotationsAreEqual := true
+	for _, opt := range options {
+		switch opt {
+		case CompareAnnotations:
+			annotationsAreEqual = reflect.DeepEqual(actual.Annotations, desired.Annotations)
+		case CompareLabels:
+			labelsAreEqual = reflect.DeepEqual(actual.Labels, desired.Labels)
+		}
+	}
+	log.V(3).Info("Compare secrets", "dateAreEqual", dataAreEqual, "labelsAreEqual", labelsAreEqual, "annotationsAreEqual", annotationsAreEqual)
```
cahartma, line 33: nit continued.  I would either change this log level to 5, or change the others above to a 3 to match this?

vparfonov: Level 5 for more deep debugging, but 3 will show just result of comparing, so i think make sense to keep it

cahartma: That makes total sense.   For future reference, this package is one of the very few that uses level 5.  We currently mainly use 1-3 and 9     Going forward, we should discuss and decide on a strategy for setting log level to be consistent among all packages.

internal/reconcile/secret.go
```diff
@@ -28,12 +27,12 @@ func Secret(er record.EventRecorder, k8Client client.Client, desired *corev1.Sec
...
 			return fmt.Errorf("failed to get %v Secret: %w", key, err)
 		}
-		if reflect.DeepEqual(current.Data, desired.Data) {
-			// identical; no need to update.
-			log.V(3).Info("Secret are the same skipping update")
+		if secrets.AreSame(current, desired, opts...) {
 			return nil
+		} else {
+			current.Data = desired.Data
+			current.Labels = desired.Labels
```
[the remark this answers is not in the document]
vparfonov: @Clee2691 I've pushed changes

PR #2075, opened 2023-07-05
LOG-4302: skip checking URL on is secure if URL not provided in config

vparfonov:
### Description
This PR introduces a modification to the URL security check logic by skipping the check if a URL is not provided in the configuration, it because output URL is not necessary for certain output types, such as Amazon CloudWatch or Google Cloud Logging.

/cc @cahartma 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-4302
- Enhancement proposal:


internal/validations/clusterlogforwarder/validate_url_to_output_tls_config.go
```diff
@@ -12,12 +12,14 @@ import (
 func validateUrlAccordingToTls(clf v1.ClusterLogForwarder) error {
   for i, output := range clf.Spec.Outputs {
     _, output := i, output // Don't bind range variable.
-    u, _ := url.Parse(output.URL)
-    scheme := strings.ToLower(u.Scheme)
-    if !url.IsTLSScheme(scheme) && (output.TLS != nil && (output.TLS.InsecureSkipVerify || output.TLS.TLSSecurityProfile != nil)) {
-      log.V(3).Info("validateUrlAccordingToTls failed", "reason", "URL not secure but output has TLS configuration parameters",
-        "output URL", output.URL, "output Name", output.Name)
-      return fmt.Errorf("URL not secure: %v, but output %s has TLS configuration parameters", u, output.Name)
+    if output.URL != "" {  // some outputs not require to have output URL (e.g. Amazon CloudWatch or Google Cloud Logging) see verifyOutputURL()
```
jcantrill, line 15: We should check for the ones we know require URL and fail if it is missing.

jcantrill: Do we have any idea if these are settings we could set for CW or GCP based on those APIs? If not then :lgtm

PR #2078, opened 2023-07-07
Remove default application logs reader role binding

aminesnow:
### Description

Following the [Fine grained logs access enhancement proposal](https://github.com/openshift/enhancements/pull/1419), this PR removes the `logging-all-authenticated-application-logs-reader` ClusterRoleBinding that allows all authenticated users to view application logs in their namespaces.
This PR also introduces three ClusterRoles, one for each of the three log types (`application`, `infrastructure` and `audit`).

/cc @xperimental 
/assign @periklis

### Links

- JIRA: [LOG-4293](https://issues.redhat.com/browse/LOG-4293), [LOG-4294](https://issues.redhat.com/browse/LOG-4294)
- Enhancement proposal:  [Fine grained logs access enhancement proposal](https://github.com/openshift/enhancements/pull/1419)

### DIFF
--- a/internal/logstore/lokistack/logstore_lokistack.go
+++ b/internal/logstore/lokistack/logstore_lokistack.go
@@ -27,6 +27,10 @@ const (
 	lokiStackAppViewClusterRoleName   = "cluster-logging-application-view"
 	lokiStackInfraViewClusterRoleName = "cluster-logging-infrastructure-view"
 	lokiStackAuditViewClusterRoleName = "cluster-logging-audit-view"
+
+	applicationLogs    = "application"
+	infrastructureLogs = "infrastructure"
+	auditLogs          = "audit"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


internal/logstore/lokistack/logstore_lokistack.go
```diff
@@ -129,9 +138,9 @@ func newLokiStackWriterClusterRoleBinding() *rbacv1.ClusterRoleBinding {
 	)
 }
 
-func newLokiStackAppReaderClusterRole() *rbacv1.ClusterRole {
+func newLokiStackAppViewClusterRole() *rbacv1.ClusterRole {
```
xperimental, line 141: There are three functions only differing in two small places now. Should we replace this with a function that takes two parameters?

internal/logstore/lokistack/logstore_lokistack.go
```diff
@@ -24,8 +24,9 @@ const (
 	lokiStackWriterClusterRoleName        = "logging-collector-logs-writer"
 	lokiStackWriterClusterRoleBindingName = "logging-collector-logs-writer"
 
-	lokiStackAppReaderClusterRoleName        = "logging-application-logs-reader"
-	lokiStackAppReaderClusterRoleBindingName = "logging-all-authenticated-application-logs-reader"
+	lokiStackAppViewClusterRoleName   = "cluster-logging-application-view"
```
xperimental, line 27: This will break existing configurations depending on the name of the ClusterRole. Do we consider the previous name part of our "public API" of the cluster-logging-operator? If yes, we need to keep the old name. If no, we can change it, but need to be aware of that change.

The old ClusterRole/ClusterRoleBinding will stay in the cluster during an upgrade, so the functionality will not be impacted, this is something that will need to be documented though.

aminesnow: How would this break the existing configuration?

xperimental: I'm thinking of customers that have created their own RoleBindings/ClusterRoleBindings to our ClusterRole. That's why I'm asking if we're considering the role name part of our public API.

internal/logstore/lokistack/logstore_lokistack.go
xperimental: What impact are you referring to? As far as I can see the functions are only used in one place.

internal/logstore/lokistack/logstore_lokistack.go
aminesnow: IIUC the role will still be there after the update, so no breaking in behavior, but we will need to properly document this change.

jcantrill (commented):
/lgtm

PR #2080, opened 2023-07-07
LOG-4153: Add CLF permissions validation

Clee2691:
### Description
This PR implements `ServiceAccount` (SA) validation for custom `ClusterLogForwarders` (CLF) using the `SubjectAccessReview` API.

### Context:
Admins defining a custom CLF **not** named `instance` will have to provide a SA with the appropriate `ClusterRoleBindings`. 

### Features
1. This PR adds 3 `ClusterRoles` into the operator manifest corresponding to the supported inputs, `application`, `audit`, and `infrastructure` that admins can bind to a SA. 
2. This PR adds a new field in `clusterlogforwarder.Spec` for defining the SA to be used.
3. The operator can `create` `SubjectAccessReviews`

/cc 
/assign 

### Links
- JIRA: 
1. https://issues.redhat.com/browse/LOG-4153
2. https://issues.redhat.com/browse/LOG-4261
4. https://issues.redhat.com/browse/LOG-4262


internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account_test.go
```diff
@@ -0,0 +1,159 @@
...
+				Expect(validateServiceAccountPermissions(k8sClient, client, sets.NewString("application", "audit"), clfServiceAccount, constants.WatchNamespace)).ToNot(Succeed())
+			})
+		})
+	})
+
+	It("should not validate clusterlogforwarder named 'instance'", func() {
+		singletonClf := loggingv1.ClusterLogForwarder{
+			ObjectMeta: v1.ObjectMeta{
+				Name:      constants.SingletonName,
+				Namespace: constants.WatchNamespace,
```
jcantrill, line 129: openshift-logging NS

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account_test.go
```diff
@@ -0,0 +1,159 @@
...
+				Expect(validateServiceAccountPermissions(k8sClient, client, sets.NewString("application", "infrastructure"), clfServiceAccount, constants.WatchNamespace)).To(Succeed())
+			})
+
+			It("should return validation error if service account cannot collect specified inputs", func() {
+				Expect(validateServiceAccountPermissions(k8sClient, client, sets.NewString("application", "audit"), clfServiceAccount, constants.WatchNamespace)).ToNot(Succeed())
+			})
+		})
+	})
+
+	It("should not validate clusterlogforwarder named 'instance'", func() {
```
jcantrill, line 125: Maybe we should add a validation that rejects CLF(openshift-logging/instance) where the SA is defined since we provide one.... or there should be "migration" that defaults this field to our legacy name.  The migration may/should allow for a consistent reconciliation flow so we don't need to treat legacy as a special case

test/framework/functional/builders_test.go
```diff
@@ -46,6 +46,7 @@ pipelines:
   name: forward-pipeline
   outputRefs:
   - elasticsearch
+serviceAccount: ""
```
jcantrill, line 49: These may go away if you add 'omitempty'

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account_test.go
Clee2691: We can "migrate" the SA name to the legacy name for that field but I think we still need to skip validation here because we don't create the SA until we start reconciling. The first part in validating the SA is checking if it exists and for the legacy clf, the SA is non-existent until after validation (i.e, it would fail).

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
```diff
@@ -0,0 +1,103 @@
...
+	if osClient, err = openshiftclient.NewOpenShiftClient(); err != nil {
+		return err, nil
+	}
+
+	var serviceAccount *corev1.ServiceAccount
+	if serviceAccount, err = getServiceAccount(clf.Spec.ServiceAccount, clf.Namespace, k8sClient); err != nil {
+		return err, nil
+	}
+	// If SA present, validate permissions based off spec'd CLF inputs
+	clfInputs := gatherPipelineInputs(clf)
```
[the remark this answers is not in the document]
Clee2691: I was not able to find a function for this specific action

apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -50,6 +50,12 @@ type ClusterLogForwarderSpec struct {
 	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Forwarder Pipelines",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:forwarderPipelines"}
 	Pipelines []PipelineSpec `json:"pipelines,omitempty"`
 
+	// ServiceAccountName is the service account's name provided by the admin.
+	//
+	// +kubebuilder:validation:minLength:=1
```
jcantrill, line 55: We need to remove this validation to support the legacy case.  If not in this PR already, we should add our own validation for the non-legacy case to require the SA.. maybe even validate it exists

internal/migrations/migrate_clusterlogforwarder.go
```diff
@@ -7,8 +7,11 @@ import (
 	"github.com/openshift/cluster-logging-operator/internal/logstore/lokistack"
 )
 
-func MigrateClusterLogForwarderSpec(spec loggingv1.ClusterLogForwarderSpec, logStore *loggingv1.LogStoreSpec, extras map[string]bool, logstoreSecretName string) (loggingv1.ClusterLogForwarderSpec, map[string]bool) {
+func MigrateClusterLogForwarderSpec(name string, spec loggingv1.ClusterLogForwarderSpec, logStore *loggingv1.LogStoreSpec, extras map[string]bool, logstoreSecretName string) (loggingv1.ClusterLogForwarderSpec, map[string]bool) {
 	spec, extras = MigrateDefaultOutput(spec, logStore, extras, logstoreSecretName)
+	if name == constants.SingletonName {
```
jcantrill, line 12: Scope this to openshift-logging NS

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
```diff
@@ -0,0 +1,121 @@
...
+
+// ValidateServiceAccountPermissions validates a service account for permissions to collect
+// inputs specified by the CLF.
+// ie. collect-application-logs, collect-audit-logs, collect-infrastructure-logs
+func validateServiceAccountPermissions(k8sClient client.Client, inputs sets.String, serviceAccount *corev1.ServiceAccount, clfNamespace string) error {
+	var err error
+	var username = fmt.Sprintf("system:serviceaccount:%s:%s", serviceAccount.Namespace, serviceAccount.Name)
+
+	// Perform subject access reviews for each spec'd input
+	// var canPerform bool
```
jcantrill, line 66: Remove me

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account_test.go
```diff
@@ -0,0 +1,152 @@
...
+		It("should succeed when service account is found", func() {
+			providedSAName := "test-serviceAccount"
+			clfServiceAccount := &corev1.ServiceAccount{
+				ObjectMeta: v1.ObjectMeta{
+					Name:      providedSAName,
+					Namespace: customClf.Namespace,
+				},
+			}
+			customClf.Spec.ServiceAccountName = providedSAName
+			Expect(k8sClient.Create(context.TODO(), clfServiceAccount)).To(Succeed())
```
jcantrill, line 53: not necessarily needed to be change but FYI, you can use the builder for the fake client to add the object such that it will be returned without needing to "create it".  There are examples in the code base

jcantrill (changes requested):
/approve

apis/logging/v1/cluster_log_forwarder_types.go
Clee2691: I do validate that the SA exists for the non-legacy case. I removed this validation

PR #2097, opened 2023-07-17
LOG-3847: Refactor Collector Alerts to support M-CLF

Clee2691:
### Description
This PR consolidates the prometheus alerts for configured collectors (`fluentd`, `vector`). 

Two new `recording` rules are introduced to alias the metrics associated with the specific collector's implementation and the `alerting` rules use the recording rules to calculate the trigger.

The rules:
1.   Aliased name:  `collector:log_num_errors:sum_rate`
    - Expression: `sum by(pod, namespace, app_kubernetes_io_part_of)(rate(vector_component_errors_total[2m])) or sum by(pod, namespace, app_kubernetes_io_part_of)(rate(fluentd_output_status_num_errors[2m]))`
2.   Aliased name: `collector:received_events:sum_rate`
    - Expression: `sum by(pod, namespace, app_kubernetes_io_part_of)(rate(vector_component_received_events_total[2m])) or sum by(pod, namespace, app_kubernetes_io_part_of)(rate(fluentd_output_status_emit_records[2m]))`

There is only one instance of the prometheus alerting rules that can be reconciled only in the `openshift-logging` namespace.

This PR also fixes services and service monitors for multi `clusterlogforwarders` as well as the `logfilemetricexporter`. Each CLF instance has their own service and service monitor.
    
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-3847


internal/metrics/service_monitor.go
```diff
@@ -42,12 +42,17 @@ func NewServiceMonitor(namespace, name, portName string, owner metav1.OwnerRefer
 		Endpoints: endpoint,
 		Selector: metav1.LabelSelector{
 			MatchLabels: map[string]string{
-				"logging-infra": "support",
+				"logging-infra":            "support",
+				constants.LabelK8sInstance: name,
```
jcantrill, line 46: I wonder if we need another selector of similarity to the "instance" to further restrict it?  Maybe the comonent (e.g. logfilemetricexporter)?

internal/metrics/alerts/collector.go
```diff
@@ -0,0 +1,75 @@
...
+    "for": "1h"
+    "labels":
+      "service": "collector"
+      "severity": "Warning"
+      namespace: "openshift-logging"
+- "name": "logging_clusterlogging_telemetry.rules"
+  "rules":
+  - "expr": |
+      sum by(cluster)(log_collected_bytes_total)
+    "record": "cluster:log_collected_bytes_total:sum"
```
jcantrill, line 65: @xperimental what is the convention for namespacing here?  We should maybe consider how to make this "Red Hat logging specifc"

internal/metrics/alerts/collector.go
```diff
@@ -0,0 +1,75 @@
+package alerts
+
+const CollectorPrometheusAlert = `
+"groups":
+- "name": "logging_collector.alerts"
+  "rules":
+  - "alert": CollectorNodeDown
```
jcantrill, line 7: @xperimental  Do you think need to create one of these per deployed namespace?  What would that mean for the recording rules.  I assume if we have one set of alerts/rules per CLF > 0 then we need these expressions to somehow take into account all namespaces or be restricted to the namespaces of all CLF

internal/metrics/alerts/collector.go
```diff
@@ -0,0 +1,75 @@
...
+    "annotations":
+      "message": "Prometheus could not scrape {{ $labels.container }}-{{ $labels.app_kubernetes_io_name }} for more than 10m."
```
jcantrill, line 9: We should add namespace/daemonset name (from labels?) to this error

internal/metrics/alerts/collector.go
```diff
@@ -0,0 +1,75 @@
...
+      "summary": "Collector cannot be scraped"
+    "expr": |
+      up{app_kubernetes_io_component = "collector", container = "collector"} == 0 or absent(up{app_kubernetes_io_component="collector", container="collector"}) == 1
+    "for": "10m"
+    "labels":
+      "service": "collector"
+      "severity": "critical"
+      "namespace": "openshift-logging"
+  - "alert": "CollectorHighErrorRate"
```
jcantrill, line 18: This alert is fine as translated now but may change based on the investigation that @vparfonov has been doing

internal/metrics/alerts/collector.go
```diff
@@ -0,0 +1,75 @@
...
+    "annotations":
+      "message": |-
+        {{ $value }}% of records have resulted in an error by {{ $labels.app_kubernetes_io_name }} collector component.
```
jcantrill, line 21: namespace and instance name?

internal/k8shandler/collection.go
```diff
@@ -110,7 +110,8 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection() (err err
 		log.Error(err, "collector.ReconcileServiceMonitor")
 		return err
 	}
-	if err := collector.ReconcilePrometheusRule(clusterRequest.EventRecorder, clusterRequest.Client, collectionSpec.Type, clusterRequest.Forwarder.Namespace, clusterRequest.ResourceNames.CommonName, clusterRequest.ResourceOwner); err != nil {
+
+	if err := collector.ReconcilePrometheusRule(clusterRequest.EventRecorder, clusterRequest.Client, collectionSpec.Type, clusterRequest.Forwarder.Namespace, constants.CollectorName, clusterRequest.ResourceOwner); err != nil {
```
jcantrill, line 114: @xperimental I wonder if we need to change this log to something like:
* fetch all rules of interest to us
* if CLF > 0, restrict rules to 1
* remove if CLF =0

jcantrill (commented):
We need to have a larger discussion with what this should be for deployment (e.g. one vs. many).  I simply do not know and do not think I thought through all the usecases completely

vimalk78:
fyi, there is this style guide for alerts https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/alerting-consistency.md#style-guide

there are tests in origin CI, which checks alerts for these https://github.com/openshift/origin/blob/master/test/extended/prometheus/prometheus.go
but perhaps it doesn't deploy logging to be able to check logging alerts

xperimental (commented):
Overall the PR makes me think about the usage of the "app.kubernetes.io" and if we're applying them correctly. Ideally, we could structure all the recording / alerting rules, so that they filter and aggregate based on these labels, so that we can properly differentiate between the collectors of different "CLF" instances.

The current label setup & ServiceMonitor configuration does not seem to be geared towards this goal, but, for me, that would be the right direction.

internal/metrics/alerts/collector.go
```diff
@@ -0,0 +1,75 @@
...
+- "name": "logging_clusterlogging_telemetry.rules"
+  "rules":
+  - "expr": |
+      sum by(cluster)(log_collected_bytes_total)
+    "record": "cluster:log_collected_bytes_total:sum"
+  - "expr": |
+      sum by(cluster)(log_logged_bytes_total)
+    "record": "cluster:log_logged_bytes_total:sum"
+  - "expr": |
+      sum by(pod, instance)(rate(vector_component_errors_total[2m])) or sum by(pod, instance)(rate(fluentd_output_status_num_errors[2m]))
```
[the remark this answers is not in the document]
Clee2691: Under the `multi-clf` we will be able to have both fluetnd and vector at the same time. In my testing I found that using `or` produces both sets of metrics if they are available and one or the other if one isn't available. The value of `record` has to be unique as I tried putting it all under one umbrella. So these metrics do populate the way I have it.

internal/collector/prometheus_rule.go
```diff
@@ -1,26 +1,39 @@
...
-	alertRulesFile := alerts.FluentdPrometheusAlert
-	if collectorType == logging.LogCollectionTypeVector {
-		alertRulesFile = alerts.VectorPrometheusAlerts
+	if exists {
+		log.V(3).Info("prometheus alerting rules already exists")
+		return nil
 	}
 
-	spec, err := runtime.NewPrometheusRuleSpecFrom(alertRulesFile)
+	spec, err := runtime.NewPrometheusRuleSpecFrom(alerts.CollectorPrometheusAlert)
```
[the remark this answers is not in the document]
jcantrill: @xperimental I did not realize this was an option.  How would we manage this for upgrades? and do upgrades replace previous deployments?

internal/factory/service.go
```diff
@@ -1,16 +1,18 @@
...
 // NewService stubs an instance of a Service
-func NewService(serviceName string, namespace string, selectorComponent string, servicePorts []core.ServicePort, visitors ...func(o runtime.Object)) *core.Service {
+func NewService(serviceName string, namespace string, selectorComponent, instanceName string, servicePorts []core.ServicePort, visitors ...func(o runtime.Object)) *core.Service {
 	service := runtime.NewService(namespace, serviceName, visitors...)
 	selector := map[string]string{
-		"component": selectorComponent,
-		"provider":  "openshift",
+		"component":                selectorComponent,
+		"provider":                 "openshift",
+		constants.LabelK8sInstance: instanceName,
```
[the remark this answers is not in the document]
Clee2691: I think it would be redundant as the instance name is the name of the owner as that is the naming convention for multi-clfs.

internal/metrics/alerts/alerts.yaml
```diff
@@ -0,0 +1,67 @@
...
+      "service": "collector"
+      "severity": "critical"
+  - "alert": "CollectorHighErrorRate"
+    "annotations":
+      "message": |-
+        {{ $value }}% of records have resulted in an error by {{ $labels.namespace }}/{{ $labels.pod }} collector component.
+      "summary": "{{ $labels.namespace }}/{{ $labels.pod }} collector component errors are high"
+    "expr": |
+      100 * (
+          collector:num_errors:sum_rate{app_kubernetes_io_part_of = "cluster-logging"}
```
jcantrill, line 21: These need to match the recording rules

PR #2098, opened 2023-07-18
LOG-4084: Fix fluentd sts cloudwatch auth using multiple roles

cahartma:
### Description
Fix fluentd so it no longer uses ENV vars for sts credentials.
#### Notes:
A change to the fluentd code resulted in regression of multiple role_arn authentication, when forwarding to cloudwatch.    The change was made in logging 5.6, while implementing vector sts for cloudwatch.   This will need to be fixed forward in 5.7 and 5.8 as well.       A similar change will need to be made for vector, once we figure out an authentication workaround.

/cc @Clee2691 @syedriko @vparfonov 
/assign @jcantrill 

### Links
- https://issues.redhat.com/browse/LOG-4084


jcantrill (commented):
/lgtm

PR #2101, opened 2023-07-19
LOG-4371: Fix for fluentd sts cloudwatch using multiple roles

cahartma:
### Description
Fix forward from 5.6,5.7
Fix fluentd so it no longer uses ENV vars for sts credentials.

#### Notes:
A change to the fluentd code resulted in regression of multiple role_arn authentication, when forwarding to cloudwatch. The change was introduced in logging 5.6, while implementing vector sts for cloudwatch. A similar change will need to be made for vector, once we figure out an authentication workaround.

/cc @Clee2691 @syedriko @vparfonov
/assign @jcantrill

### Links
- v5.8:  https://issues.redhat.com/browse/LOG-4371
- v5.7:  https://issues.redhat.com/browse/LOG-4368
- v5.6:  https://issues.redhat.com/browse/LOG-4084

### DIFF
--- a/internal/collector/collector_test.go
+++ b/internal/collector/collector_test.go
@@ -302,7 +302,7 @@ var _ = Describe("Factory#CollectorResourceRequirements", func() {
 	})
 })
 
-var _ = Describe("Factory#NewPodSpec Add Cloudwatch Resources", func() {
+var _ = Describe("Factory#NewPodSpec Add Cloudwatch STS Resources", func() {
 	var (
 		factory   *Factory
 		pipelines = []logging.PipelineSpec{
@@ -336,16 +336,6 @@ var _ = Describe("Factory#NewPodSpec Add Cloudwatch Resources", func() {
 				},
 			},
 		}
-
-		verifyEnvVar = func(container v1.Container, name, value string) {
-			for _, elem := range container.Env {
-				if elem.Name == name {
-					Expect(elem.Value).To(Equal(value))
-					return
-				}
-			}
-			Fail(fmt.Sprintf("Expected collector to include env var '%s' with a value of '%s'", name, value))
-		}
 	)
 	Context("when collectorType is fluentd", func() {
 		BeforeEach(func() {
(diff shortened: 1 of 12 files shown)


internal/generator/fluentd/output/cloudwatch/cloudwatch.go
```diff
@@ -106,7 +107,9 @@ func SecurityConfig(o logging.OutputSpec, secret *corev1.Secret) Element {
 	// First check for credentials or role_arn key, indicating a sts-enabled authentication
 	if security.HasAwsRoleArnKey(secret) || security.HasAwsCredentialsKey(secret) {
 		return AWSKey{
-			KeyRoleArn: ParseRoleArn(secret),
+			KeyRoleArn:          ParseRoleArn(secret),
+			KeyRoleSessionName:  constants.AWSRoleSessionName,
+			KeyWebIdentityToken: path.Join(constants.AWSWebIdentityTokenMount, constants.AWSWebIdentityTokenFilePath),
```
jcantrill, line 112: Maybe I don't understand, but this identity token is constant across the configured outputs but its the 'arn' which is variable?

cahartma: Correct, the SA token is shared among all outputs, since it is not tied the role.   The latest/rotated token is used to obtain temporary credentials for each "role request", allowing the SA to assume that role during the request session.

internal/collector/collector_test.go
```diff
@@ -359,17 +368,17 @@ var _ = Describe("Factory#NewPodSpec Add Cloudwatch Resources", func() {
...
 					Outputs:   outputs,
 					Pipelines: pipelines,
 				}, "1234", "", tls.GetClusterTLSProfileSpec(nil))
 				collector := podSpec.Containers[0]
 
-				verifyEnvVar(collector, constants.AWSRegionEnvVarKey, outputs[0].OutputTypeSpec.Cloudwatch.Region)
-				verifyEnvVar(collector, constants.AWSRoleArnEnvVarKey, roleArn)
-				verifyEnvVar(collector, constants.AWSRoleSessionEnvVarKey, constants.AWSRoleSessionName)
-				verifyEnvVar(collector, constants.AWSWebIdentityTokenEnvVarKey, path.Join(constants.AWSWebIdentityTokenMount, constants.AWSWebIdentityTokenFilePath))
+				verifyNoLongerEnvVar(collector, constants.AWSRegionEnvVarKey)
```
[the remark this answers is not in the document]
cahartma: I will update to the env matcher.  
For an e2e, Moto does now include multi-account and sts support.   Since sts features are being expanded, it may be a good time create these tests.

PR #2102, opened 2023-07-20
LOG-3570: Remove tlsSecurityProfile feature gate

syedriko:
### Description

Remove tlsSecurityProfile feature gate

/cc @cahartma 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-3570


internal/generator/vector/outputs.go
```diff
@@ -68,15 +67,13 @@ func Outputs(clspec *logging.CollectionSpec, secrets map[string]*corev1.Secret,
...
-				op[generator.Ciphers] = ""
-			} else {
-				outMinTlsVersion, outCiphers := op.TLSProfileInfo(o, ",")
-				op[generator.MinTLSVersion] = outMinTlsVersion
-				op[generator.Ciphers] = outCiphers
-			}
+		if o.Name == logging.OutputNameDefault && o.Type == logging.OutputTypeElasticsearch {
+			op[generator.MinTLSVersion] = ""
+			op[generator.Ciphers] = ""
+		} else {
```
cahartma, line 73: Is this going to set tls for all other outputs?   Don't you still need to check for loki here?

syedriko: This is going to set TLS version and ciphersuites for the all the outputs but the default ES, if they enable TLS.
The line I removed said `if the feature gate flag is set or it's loki`, [docs/features/tls_security_profile.adoc] ^^^ had this `The feature gate is ignored for the default log storage of type loki.` Without feature gate, there's no need for loki exception.

jcantrill (commented):
/lgtm

PR #2054, opened 2023-06-13
LOG-1913: Modify functional-bench-marker to support vector

jcantrill:
### Description
This PR:
* Updates the functional benchmarking tool to support testing of vector

### Links
* https://issues.redhat.com/browse/LOG-1913

cc @syedriko @alanconway

PR #2113, opened 2023-07-27
LOG-1913: Modify functional-bench-marker to support vector

jcantrill:
### Description
Forward port of https://github.com/openshift/cluster-logging-operator/pull/2054

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -244,7 +244,7 @@ deploy-example: deploy
 	oc create -n $(NAMESPACE) -f hack/cr.yaml
 
 .PHONY: test-functional
-test-functional: test-functional-benchmarker test-functional-fluentd test-functional-vector
+test-functional: test-functional-fluentd test-functional-vector
 	RELATED_IMAGE_VECTOR=$(IMAGE_LOGGING_VECTOR) \
 	RELATED_IMAGE_FLUENTD=$(IMAGE_LOGGING_FLUENTD) \
 	RELATED_IMAGE_LOG_FILE_METRIC_EXPORTER=$(IMAGE_LOGFILEMETRICEXPORTER) \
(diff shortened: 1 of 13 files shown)


Makefile
```diff
@@ -271,11 +271,14 @@ test-forwarder-generator: bin/forwarder-generator
 	WATCH_NAMESPACE=openshift-logging bin/forwarder-generator --file hack/logforwarder.yaml --collector=fluentd > /dev/null
 	WATCH_NAMESPACE=openshift-logging bin/forwarder-generator --file hack/logforwarder.yaml --collector=vector > /dev/null
 
+test-functional-benchmarker-vector: bin/functional-benchmarker
+	@rm -rf /tmp/benchmark-test-test-vector
+	@out=$$(RELATED_IMAGE_VECTOR=$(IMAGE_LOGGING_VECTOR) bin/functional-benchmarker -image=$(IMAGE_LOGGING_VECTOR) --collector-impl=vector --artifact-dir=/tmp/benchmark-test-vector 2>&1); if [ "$$?" != "0" ] ; then echo "$$out"; exit 1; fi
```
syedriko, line 276: Can we stick to one dash or two dashes for consistency?

internal/cmd/functional-benchmarker/main.go
```diff
@@ -83,9 +81,15 @@ func RunBenchmark(artifactDir string, options config.Options) (*stats.ResourceMe
 	}
 	statistics := gatherStatistics(runner, options.Sample, options.MsgSize, startTime, endTime)
 
+	go func() { fetchContainerLogs(runner, artifactDir) }()
```
syedriko, line 84: This will run asynchronously with no synchronization and will most likely continue running after RunBenchmark() returns. Why not just call `fetchContainerLogs()` synchronously?

internal/cmd/functional-benchmarker/runners/cluster/cluster_runner.go
```diff
@@ -51,19 +54,34 @@ func (r *ClusterRunner) Pod() string {
...
+			spec.Application = &logging.Application{
+				Namespaces: []string{r.Namespace()},
+			}
+		}).
+		ToHttpOutput()
 
 	//modify config to only collect loader containers
 	r.framework.VisitConfig = func(conf string) string {
-		pattern := fmt.Sprintf("/var/log/pods/%s_*/loader-*/*.log", r.framework.Namespace)
+		//vector
```
syedriko, line 70: switch LogCollectionType ?

test/framework/functional/output_http.go
```diff
@@ -18,11 +18,12 @@ import (
 )
 
 const (
-	VectorHttpSourceConfTemplate = `
-[sources.my_source]
+	VectorHttpSourceConfTemplate = "" +
+		`[sources.my_source]
 type = "http"
```
syedriko, line 23: The type's been renamed to `http_server`

syedriko (approved):

internal/cmd/functional-benchmarker/main.go
jcantrill: I was trying to speed up the test runs and this change is the same that was part of the 5.7 merge and does not seem to have resulted in any issues

syedriko: That's why it's called a race condition - it doesn't cause any issues, until it does :) Would it make sense to keep main() from returning until fetchContainerLogs() is done?

jcantrill: Updated to remove go routine

PR #2114, opened 2023-08-01
LOG-4310: Creating CL with 'name != instance' gets error

Clee2691:
### Description
This PR addresses the issue of creating multiple `ClusterLogging` instances. This is the removal of the singleton patch.

/cc @cahartma @syedriko 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4310


cahartma:
Should we remove the patch file as well?

Clee2691:
> Should we remove the patch file as well?

Yes I think that is a good idea. I'll remove it.

jcantrill:
@Clee2691 do we still have the same for CLF?  If not, please feel free to release the hold

PR #2115, opened 2023-08-02
LOG-4379: LFME pods are not removed after CL and LFME instances are removed

syedriko:
### Description
Made LFME DS and Service owned by the CL

/cc @Clee2691 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4397


Clee2691:
This PR can be closed as it is fixed in here: [PR#2116](https://github.com/openshift/cluster-logging-operator/pull/2116)

PR #2116, opened 2023-08-04
LOG-4397: LFME pods are not removed

Clee2691:
### Description
This PR:
- Fixes LFME resource cleanup
- Remove creation of a "virtual" LFME resource when `ClusterLogging` CR is created without a LFME CR
- Remove dependence of LFME on a `ClusterLogging` instance

/cc @cahartma @syedriko 
/assign @jcantrill 
### Links
- JIRA: https://issues.redhat.com/browse/LOG-4397

### DIFF
--- a/internal/auth/service_account.go
+++ b/internal/auth/service_account.go
@@ -1,4 +1,4 @@
-package collector
+package auth
 
 import (
 	"context"
@@ -6,7 +6,6 @@ import (
 
 	"github.com/openshift/cluster-logging-operator/internal/constants"
 
-	security "github.com/openshift/api/security/v1"
 	"github.com/openshift/cluster-logging-operator/internal/factory"
 	"github.com/openshift/cluster-logging-operator/internal/reconcile"
 	"github.com/openshift/cluster-logging-operator/internal/runtime"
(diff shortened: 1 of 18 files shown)


internal/auth/service_account.go
```diff
@@ -17,7 +18,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const sccName = "log-collector-scc"
+const sccName = "logging-scc"
```
syedriko, line 21: It might be not worth the trouble renaming this SCC, unless we know the upgrade scenario deletes the old one and creates the new.

jcantrill: I mentioned to @Clee2691 we should create a dedicated SCC for LFME and give it a naming convention that some how maps to the CR name.  This gives us the freedom to deploy multiples in the future though I believe that to be unlikely

internal/k8shandler/securitycontextconstraint.go
```diff
@@ -0,0 +1,20 @@
+package k8shandler
```
jcantrill, line 1: Please move this into auth and out of 'k8shandler'

jcantrill (changes requested):
one change otherwise lgtm

internal/auth/service_account.go
Clee2691: Since they both use the same SCC I kept the naming but added code to remove the legacy SCC.

PR #2118, opened 2023-08-08
Fix Collector reconciliation

jcantrill:
### Description
This PR:
* refactors reconciliation of logging to a single method
* fixes collector.spec not applied for multi CLF
* fixes resources not being deleted for multi CLF
* moves some telemetry code into telemetry package

### Links
* https://issues.redhat.com/browse/LOG-4427
* https://issues.redhat.com/browse/LOG-4410

cc @anpingli It would be good to do some extra testing of telemetry to ensure I didn't break something

cc @Clee2691 @cahartma

### DIFF
--- a/controllers/clusterlogging/clusterlogging_controller.go
+++ b/controllers/clusterlogging/clusterlogging_controller.go
@@ -3,6 +3,7 @@ package clusterlogging
 import (
 	"context"
 
+	"github.com/openshift/cluster-logging-operator/internal/logstore/lokistack"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/source"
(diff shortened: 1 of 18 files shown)


controllers/clusterlogging/clusterlogging_controller.go
```diff
@@ -65,10 +67,15 @@ func (r *ReconcileClusterLogging) Reconcile(ctx context.Context, request ctrl.Re
 	telemetry.SetCLMetrics(0) // Cancel previous info metric
 	defer func() { telemetry.SetCLMetrics(1) }()
 
+	removeFinalizer := func(identifier string) error {
```
Clee2691, line 70: There are `finalizer` utility functions provided by the package `controllerutil`. It might be a better use than defining our own?
[Remove Finalizer](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.14.5/pkg/controller/controllerutil#RemoveFinalizer)

jcantrill: updated internal/k8shandler/finalizer.gogit

PR #2122, opened 2023-08-10
LOG-4231: Add open on every update fluentd tuning options

jcantrill:
### Description
This PR:

* adds a tuning option for fluentd to allow opening files on every update to mitigate file handles not being closed

### Links
* https://issues.redhat.com/browse/LOG-4241


jcantrill:
Closing.  We are choosing not to introduce tuning API changes in a released y-stream

PR #2125, opened 2023-08-12
LOG-4440: Prometheus rules are not reconciled for cl/clfs not in openshift-logging namespace

Clee2691:
### Description
This PR addresses the issue that `ClusterLogForwarders` created in a namespace other than `openshift-logging` would not reconcile the `Prometheus` alerting rules in the `openshift-logging` namespace.

The `Prometheus` alerting rules are statically deployed by the OLM and will be available in the `openshift-logging` namespace with or without any available collectors. 

/cc @syedriko @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4440


jcantrill:
@Clee2691  to summarize our discussion we desire:

* Move these rules to be statically deployed with the manifest
* Ensure we have proper documentation to identify the change
* Test rules do not fire when CLO is deployed but no collector deployment
* Test rules do not continue to fire when CLO is deployed and collector is  deployed and undeployed

This would allow us to remove the reconciliation and other associated code

internal/reconcile/prometheus_rule.go
```diff
@@ -36,7 +36,6 @@ func PrometheusRule(er record.EventRecorder, k8Client client.Client, desired *mo
 			return nil
```
jcantrill, line 36: Any reason to not completely remove this file?

jcantrill (commented):
/approve

internal/reconcile/prometheus_rule.go
Clee2691: Do we anticipate needing to create prometheus rules in the future? This seemed generic enough to keep for now for the future is my thinking.

PR #2127, opened 2023-08-15
LOG-4181: fix loki test flake caused by rate limiting

jcantrill:
### Description
This PR:
* Updates the loki test image to 2.8.4
* Modifies the ingestion rates and limits to address e2e loki test flake

### Links
* https://issues.redhat.com/browse/LOG-4181

### DIFF
--- a/test/helpers/loki/receiver.go
+++ b/test/helpers/loki/receiver.go
@@ -26,7 +26,7 @@ import (
 )
 
 const (
-	Image        = "grafana/loki:2.5.0"
+	Image        = "grafana/loki:2.8.4"
 	Port         = int32(3100)
 	lokiReceiver = "loki-receiver"
 )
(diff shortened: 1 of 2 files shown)


xperimental (commented):
The limits seem a bit excessive. This works, but might hide an issue down the road that we then only see when something really goes wrong and even hits those limits.

jcantrill:
bq. The limits seem a bit excessive. This works, but might hide an issue down the road that we then only see when something really goes wrong and even hits those limits.

These limits were added based upon what I saw in the error logs of the test so they are real numbers.  My concern more is to resolve the test flake and this test verifies we can write to Loki.  We make no changes to Loki config outside of using the defaults which are likely not ideal.  The test is intended to be an integration test that verifies connectivity between the collector and Loki.  While I agree it could be tweaked better, we would need to define additional goals of the test and bullet proof it accordingly

PR #2128, opened 2023-08-16
LOG-4413: Overrides the name of the log field used to retrieve the timestamp to send to Splunk HEC

vparfonov:
### Description
This PR overrides the default name of the log field to used to retrieve the timestamp to send to Splunk HEC to the `@timestamp`. 

https://vector.dev/docs/reference/configuration/sinks/splunk_hec_logs/#timestamp_key 

/cc 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-4413 
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/output/splunk/splunk_test.go
+++ b/internal/generator/vector/output/splunk/splunk_test.go
@@ -73,6 +73,7 @@ inputs = ["splunk_hec_dedot"]
 endpoint = "https://splunk-web:8088/endpoint"
 compression = "none"
 default_token = "` + hecToken + `"
+timestamp_key = "@timestamp"
 [sinks.splunk_hec.encoding]
 codec = "json"
 `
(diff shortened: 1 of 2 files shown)


internal/generator/vector/output/splunk/splunk_test.go
```diff
@@ -75,6 +75,17 @@ compression = "none"
...
 codec = "json"
+`
+		splunkSinkTimestampKey = splunkDedot + `
+[sinks.splunk_hec]
+type = "splunk_hec_logs"
+inputs = ["splunk_hec_dedot"]
+endpoint = "https://splunk-web:8088/endpoint"
+compression = "none"
+default_token = "` + hecToken + `"
+timestamp_key = "@timestamp"
```
jcantrill, line 86: In what cases do we need to override this value to something other then '@timestamp'.  Why can't we just set it to this value for all cases and not expand the API?

PR #2138, opened 2023-08-22
LOG-4446: CLO raises many error messages 'Could not find ServiceAccount token secret' after creating multiple-CLF

Clee2691:
### Description
This PR removes retrieval of the service account token for multi-CLF when reconciling the collection.

For the legacy case, when forwarding to the internal default lokistack store, migration of lokistack now includes adding the service account token as a secret.

/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4446

### DIFF
--- a/internal/k8s/loader/load.go
+++ b/internal/k8s/loader/load.go
@@ -3,6 +3,7 @@ package loader
 import (
 	"context"
 	"fmt"
+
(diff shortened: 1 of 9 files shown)


jcantrill (commented):
/approve

PR #2146, opened 2023-08-29
LOG-3949: increase max_read_bytes for container logs to release file …

jcantrill:
…handles sooner

### Description
This PR:
* Increases the max_read_bytes of container logs so vector will read rolled over files faster and release file descriptors sooner
* The value was increased to ~30% of the default log rotation settings or ~3mb

### Links
* https://issues.redhat.com/browse/LOG-3949

Gra
image

This graph was produced by testing loading services with high volume that produce rollover frequently.  My observation was a change from a gradual increase in FDs to a more horizontal profile.  I additionally `lsof` of deleted files and observed a change from a steady increase to  varying around 3 to 5 at any one time. 

This change looks to improve the overall behavior but will not resolve cases where the collector is unable to keep up with the incoming logs; it still has to be capable of pushing logs faster then reading them otherwise the FDs will continue to pile up while vector churns through the rotated files.

### DIFF
--- a/internal/generator/vector/conf_test.go
+++ b/internal/generator/vector/conf_test.go
@@ -98,6 +98,7 @@ expire_metrics_secs = 60
 # Logs from containers (including openshift containers)
 [sources.raw_container_logs]
 type = "kubernetes_logs"
+max_read_bytes = 3145728
 glob_minimum_cooldown_ms = 15000
 auto_partial_merge = true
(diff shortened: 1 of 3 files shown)


vparfonov:
/retitle LOG-3949: increase max_read_bytes for container logs to release file handles sooner

vparfonov:
@jcantrill 
May I have missed something in the description that states the value was increased by approximately 30%, and now it's `3145728`, but according to the Vector documentation, the default value is `2048`
https://vector.dev/docs/reference/configuration/sources/kubernetes_logs/#max_read_bytes

jcantrill:
>May I have missed something in the description that states the value was increased by approximately 30%, and now it's 3145728, but according to the Vector documentation, the default value is 2048
> ~30% of the default log rotation settings

should be ~30% of default CRIO log rotation settings not vectors default settings for that value

PR #2147, opened 2023-08-30
WIP: LOG-4476: Vector can't forward logs to default lokistack

syedriko:
### Description

Fix for [LOG-4476](https://issues.redhat.com//browse/LOG-4476) making sure the service CA bundle is configured for the default Loki outputs.

/cc @xperimental @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4476

### DIFF
--- a/internal/generator/vector/output/loki/loki.go
+++ b/internal/generator/vector/output/loki/loki.go
@@ -205,18 +205,10 @@ func TLSConf(o logging.OutputSpec, secret *corev1.Secret, op Options) []Element 
 	if o.Secret != nil || (o.TLS != nil && o.TLS.InsecureSkipVerify) {
 		if tlsConf := security.GenerateTLSConf(o, secret, op, false); tlsConf != nil {
 			tlsConf.NeedsEnabled = false
+			if strings.HasPrefix(o.Name, logging.OutputNameDefault) {
+				tlsConf.CAFilePath = `"/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"`
+			}
 			return []Element{tlsConf}
-		}
-
-	} else if secret != nil {
-		// Set CA from logcollector ServiceAccount for internal Loki
-		tlsConf := security.TLSConf{
-			ComponentID: strings.ToLower(vectorhelpers.Replacer.Replace(o.Name)),
-			CAFilePath:  `"/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"`,
-		}
-		tlsConf.SetTLSProfileFromOptions(op)
-		return []Element{
-			tlsConf,
 		}
 	}
 	return []Element{}


internal/generator/vector/output/loki/loki.go
```diff
@@ -205,19 +205,11 @@ func TLSConf(o logging.OutputSpec, secret *corev1.Secret, op Options) []Element
 	if o.Secret != nil || (o.TLS != nil && o.TLS.InsecureSkipVerify) {
 		if tlsConf := security.GenerateTLSConf(o, secret, op, false); tlsConf != nil {
 			tlsConf.NeedsEnabled = false
+			if strings.HasPrefix(o.Name, logging.OutputNameDefault) {
+				tlsConf.CAFilePath = `"/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"`
```
jcantrill, line 209: Why is the default output special?  Don't all pods get a CA mounted into them which would imply to me to either use the one mounted by default or use the one from a secret that is specified by the user

syedriko: I don't entirely like this fix either. We need to use InternalLogStoreSecret https://github.com/openshift/cluster-logging-operator/blob/e87beec8402beab9813eb5b81b37448aba5848aa/internal/k8s/loader/load.go#L59 for internal Loki.

Since https://github.com/openshift/cluster-logging-operator/commit/3928ce34acee8b387b8ba7ce109cc529ff6d95f5, we're feeding an SA Token secret to the internal Loki outputs and that throws off the TLS config. Before that, the hack at https://github.com/openshift/cluster-logging-operator/blob/e87beec8402beab9813eb5b81b37448aba5848aa/internal/generator/vector/output/loki/loki.go#L211 triggered because o.Secret was nil.

jcantrill: @syedriko is there a place to utilize "options" to do special handling where we could push treating "internal loki" into a migration or fix up the migration to do special handling?

xperimental: To me it looks like the underlying issue is that the code's assumption for "there is a logcollector secret" is broken. Not just this code-path but also in fluentd, possibly also for ES, because that's what the "special handling of default" is based on.

While I think the "special handling" is something that should be fixed (there seem to be a lot of special handling "flags" in the TLS configuration generation, some of which look unused on first glance), I think I'd rather keep the generation code as-is and fix the "there is a logcollector secret" assumption instead.

I'm not fully through debugging this, so this is a bit preliminary and my view might change :slightly_smiling_face:

xperimental: Wait, I think I spoke too soon. (like anticipated :wink: )

I think the issue is rather: the secret is incomplete (as opposed to "missing") at this point of the code. Maybe the "reading code" was accidentally removed.

xperimental: Ok, not quite correct again. The "new way" of feeding the "log collector token" like any other service-account token into the configuration generator actually works. The secret is there and has all the properties.

The issue why it does not work is that the file containing the CA bundle is called `service-ca.crt` in this case and not `ca-bundle.crt` like [this code](https://github.com/openshift/cluster-logging-operator/blob/354d4eefd0410abd2f62a678171d65766e297c59/internal/generator/vector/output/security/security.go#L140) assumes. I vaguely remembering having proposed this already, but I'm not 100% sure: should we use `service-ca.crt` as a fallback for the CA-bundle if `ca-bundle-crt` is not available? It looks like this should fix this issue (still in process of testing).

xperimental: Ok, I think I have come full-circle now that I reminded myself how ES works. The main difference is that ES has a "client-certificate authentication" Secret while LokiStack has a "Kubernetes ServiceAccount" Secret. I don't know if we use the ServiceAccount type of Secret anywhere else but for our default LokiStack.

If the default LokiStack is currently our only use of the ServiceAccount-Secret then explicitly setting the CA to `service-ca.crt` just for that seems like a good solution. I think I would put the "set the CAFilePath" line in this file though, I think I'd rather add it as another "special case" in [this function](https://github.com/openshift/cluster-logging-operator/blob/354d4eefd0410abd2f62a678171d65766e297c59/internal/generator/vector/output/security/security.go#L76) similar to what is done for ES. The main reason for this being that the TLS configuration is not modified "post creation" in that case and all the special cases are at the same location.

The removed `if secret != nil` code is apparently not used anymore and thus can stay removed.

jcantrill:
closing in favor of https://github.com/openshift/cluster-logging-operator/pull/2158

PR #2148, opened 2023-08-30
LOG-4481: make unique paths for Vector checkpoints file according to multiple-CLFs installation

vparfonov:
### Description
This PR:

- Adding a Run Script for Vector:
A run script for Vector is being added. This script will enable pre-run tasks, such as creating custom directories for persisting Vector state. This can be useful for setting up Vector environment or configuration before it starts processing data.

- Making Paths for Vector Checkpoints Unique:
To ensure uniqueness for Vector checkpoints files, the PR proposes adding the namespace name and the name of the CLF (Cloud Logging Framework) to the path. This means that the path to the checkpoints file will include both the namespace and CLF name, making it easier to manage and identify Vector instances for different namespaces and CLFs.

For example, the path to a Vector checkpoint file could look like this:` /var/lib/vector/openshift-logging/my-instance`, where:

`/var/lib/vector` is the base directory.
`openshift-logging` is the namespace name.
`my-instance` is the name of the CLF.

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com//browse/LOG-4481
- Enhancement proposal:

### DIFF
--- a/internal/collector/vector/run_script.go
+++ b/internal/collector/vector/run_script.go
@@ -1,0 +1,14 @@
+package vector
+
+// RunVectorScript is the run-vector.sh script for launching the Vector container process
+// will override content of /scripts/run-vector.sh in https://github.com/ViaQ/vector
+const RunVectorScript = `
+#!/bin/bash
+# The directory used for persisting Vector state, such as on-disk buffers, file checkpoints, and more.
+VECTOR_DATA_DIR=%s
+echo "Creating the directory used for persisting Vector state $VECTOR_DATA_DIR"
+mkdir -p $VECTOR_DATA_DIR
+echo "Starting Vector process..."
+exec "/usr/bin/vector"
+
+`
--- a/internal/factory/resource_names.go
+++ b/internal/factory/resource_names.go
@@ -16,6 +16,7 @@ type ForwarderResourceNames struct {
 	ServiceAccount                   string
 	InternalLogStoreSecret           string
 	ServiceAccountTokenSecret        string
+	ForwarderName                    string
 }
 
 func (f *ForwarderResourceNames) DaemonSetName() string {
@@ -34,6 +35,7 @@ func GenerateResourceNames(clf logging.ClusterLogForwarder) *ForwarderResourceNa
 		SecretMetrics:                    resBaseName + "-metrics",
 		ConfigMap:                        resBaseName + "-config",
 		MetadataReaderClusterRoleBinding: fmt.Sprintf("cluster-logging-%s-%s-metadata-reader", clf.Namespace, resBaseName),
+		ForwarderName:                    clf.Name,
 	}
 
 	if clf.Namespace == constants.OpenshiftNS && clf.Name == constants.SingletonName {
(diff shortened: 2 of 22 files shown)


internal/factory/resource_names.go
```diff
@@ -16,6 +16,7 @@ type ForwarderResourceNames struct {
 	ServiceAccount                   string
 	InternalLogStoreSecret           string
 	ServiceAccountTokenSecret        string
+	ForwarderName                    string
```
jcantrill, line 19: Is commonName the same as forwarder Name?

internal/generator/vector/global.go
```diff
@@ -1,19 +1,22 @@
...
+	"github.com/openshift/cluster-logging-operator/internal/collector/vector"
 	"github.com/openshift/cluster-logging-operator/internal/generator"
 )
 
-func Global() []generator.Element {
+func Global(name string) []generator.Element {
 	return []generator.Element{
 		GlobalOptions{
 			ExpireMetricsSecs: 60,
+			DataDir:           vector.DataPath + "/" + name,
```
jcantrill, line 12: path.join or better would be to use expose a function from vector that provides the same value as what is used to generate the container spec

test/framework/functional/vector/deploy.go
```diff
@@ -13,7 +14,7 @@ import (
 )
 
 const entrypointScript = `#!/bin/bash
-mkdir -p /var/lib/vector
+mkdir -p /var/lib/vector/%s
```
jcantrill, line 17: lets utilize the public constant from the vector package to form this path

internal/collector/vector/run_script.go
```diff
@@ -0,0 +1,14 @@
+package vector
+
+// RunVectorScript is the run-vector.sh script for launching the Vector container process
+// will override content of /scripts/run-vector.sh in https://github.com/ViaQ/vector
```
jcantrill, line 4: We do not need to override any script just the entrypoint of the container

internal/factory/resource_names.go
```diff
@@ -16,6 +16,8 @@ type ForwarderResourceNames struct {
...
+	Namespace                        string
```
jcantrill, line 20: Why do we need to carry the Namespace as part of the naming strategy since we already have it as part of the CLF?

test/framework/functional/vector/deploy.go
```diff
@@ -13,7 +15,7 @@ import (
 )
 
 const entrypointScript = `#!/bin/bash
```
jcantrill, line 17: Doesn't this get replaced by the content of run_script.go

vparfonov: no, in working environment launch script mount to the` /usr/bin/run-vector.sh`, in test environment` /opt/app-root/src/run.sh `

internal/factory/resource_names.go
vparfonov: We need it in CollectorVisitor https://github.com/openshift/cluster-logging-operator/blob/master/internal/collector/vector/visitors.go#L16

jcantrill: Can we pass the namespace to visitors.go without too much rework?  the name factory does not have need of NS to name resources

vparfonov: done

jcantrill (commented):
/lgtm

PR #2155, opened 2023-09-01
LOG-4048: Implement input as http server

syedriko:
### Description

An implementation of HTTP inputs for the vector log collector.

/cc @alanconway 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4048


apis/logging/v1/input_source_types.go
```diff
@@ -0,0 +1,52 @@
...
+	// or do we need vector to have certificates? Can the service be bypassed?
+
+	// FIXME What to do about service name clashes? Can't generate random name, service name is used to locate the source.
+
+	// Name of the source.
+	// A Service resource is generated for each source,
+	// the name must be a unique Service name in the ClusterLogForwarder's namespace.
+	//
+	// +required
+	Name string `json:"name"`
```
jcantrill, line 34: Why  do we need a "name" if the inputspec already has one that can be defined by the user.  This looks to be redundant

apis/logging/v1/input_source_types.go
```diff
@@ -0,0 +1,52 @@
...
+	// Port that this source will listen on.
+	// Used to create a Service for this port.
+	//
+	// +required
+	Port corev1.ServicePort `json:"port"`
+	// Format is the format of incoming log data.
+	//
+	// +kubebuilder:validation:Enum:=k8s;text
+	// +required
+	Format string `json:"format"`
```
jcantrill, line 45: Maybe we are limited by how we are integrating but could this not be determined from a ContentType header which will be more in line with typical http client/server interactions?

apis/logging/v1/cluster_log_forwarder_types.go
```diff
@@ -118,6 +118,9 @@ type InputSpec struct {
 	// +optional
 	// +docgen:ignore
 	Audit *Audit `json:"audit,omitempty"`
+
+	// Source to receive logs from non-cluster sources.
+	Source *SourceSpec `json:"source,omitempty"`
```
jcantrill, line 123: Is there an alternate name we can use, maybe "server" to make it more obvious the input is a receiver.   Source and Input are basically the same and the others denote essentially a well defined category of inputs (or sources)

jcantrill: I have many [comments](https://github.com/openshift/cluster-logging-operator/pull/1958) regarding the API that should be addressed IMO before this PR merges.  I will skip additional  API review

internal/generator/vector/inputs.go
```diff
@@ -153,6 +153,22 @@ func Inputs(spec *logging.ClusterLogForwarderSpec, o Options) []Element {
 		}
 	}
 
+	for _, input := range spec.Inputs {
+		if input.Source != nil && input.Source.HTTP != nil && input.Source.HTTP.LogType == `audit` {
```
jcantrill, line 157: instead of "audit" we should the well known types from the api dir

internal/generator/vector/normalize.go
```diff
@@ -95,6 +95,13 @@ func NormalizeLogs(spec *logging.ClusterLogForwarderSpec, op generator.Options)
 		el = append(el, NormalizeOpenshiftAuditLogs(RawOpenshiftAuditLogs, OpenshiftAuditLogs)...)
 		el = append(el, NormalizeOVNAuditLogs(RawOvnAuditLogs, OvnAuditLogs)...)
 	}
+
+	for _, input := range spec.Inputs {
+		if input.Source != nil && input.Source.HTTP != nil && input.Source.HTTP.LogType == `audit` {
+			el = append(el, NormalizeK8sAuditLogs(input.Source.HTTP.Name+`_items`, input.Source.HTTP.Name+`_normalized`)...)
```
jcantrill, line 101: Looks like we need a helper function to ensure "normalized" is consistent between the 2 files

internal/generator/vector/sources.go
```diff
@@ -118,6 +120,81 @@ func ExcludeContainerPaths(namespace string) string {
...
+	if _, ok := op[generator.ClusterTLSProfileSpec]; ok {
+		tlsProfileSpec := op[generator.ClusterTLSProfileSpec].(configv1.TLSProfileSpec)
+		minTlsVersion = tls.MinTLSVersion(tlsProfileSpec)
+		cipherSuites = strings.Join(tls.TLSCiphers(tlsProfileSpec), `,`)
+	}
+
+	el := []generator.Element{}
+	for _, input := range spec.Inputs {
+		if input.Source != nil {
+			if input.Source.HTTP != nil {
```
jcantrill, line 134: this can be combined with the previous line

internal/generator/vector/sources.go
```diff
@@ -118,6 +120,81 @@ func ExcludeContainerPaths(namespace string) string {
...
+	return "httpSource"
+}
+
+func (i HttpSource) Template() string {
+	return `
+{{define "` + i.Name() + `" -}}
+[sources.{{.ID}}]
+type = "http_server"
+address = "0.0.0.0:{{.Port}}"
+decoding.codec = "json"
```
jcantrill, line 168: What if the format is plain text?

internal/generator/vector/sources.go
```diff
@@ -118,6 +120,81 @@ func ExcludeContainerPaths(namespace string) string {
...
+
+[sources.{{.ID}}.tls]
+enabled = false
+key_file = "/etc/collector/{{.ID}}/tls.key"
```
jcantrill, line 172: We need to ensure this is still valid in context of multi CLF.  As long as we are setting up the volumes correctly for the DS this will be fine

test/functional/inputs/http/http_input_test.go
```diff
@@ -0,0 +1,137 @@
...
+				func() string { buf, _ := json.Marshal(auditRecord); return string(buf) }(),
+			)
+			Expect(err).To(BeNil(), "Expected no errors reading the logs")
+
+			raw, err := framework.ReadFileFrom("http", functional.ApplicationLogFile)
+			Expect(err).To(BeNil(), "Expected no errors reading the logs")
+			var logs []types.K8sAuditLog
+			err = types.ParseLogsFrom(utils.ToJsonLogs([]string{raw}), &logs, false)
+			Expect(err).To(BeNil(), "Expected no errors parsing the logs")
+			fmt.Fprintln(GinkgoWriter, raw)
```
jcantrill, line 97: Presumably we don't want this in the final merge

test/functional/inputs/http/http_input_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	Context("When sending an audit log record to an HTTP input", func() {
+		It("should be able to round trip it unharmed", func() {
+			Expect(framework.DeployWithVisitor(
+				func(b *runtime.PodBuilder) error {
+					return framework.AddVectorHttpOutput(b, framework.Forwarder.Spec.Outputs[0])
+				}),
+			).To(BeNil())
+
+			_, err := framework.RunCommand(constants.CollectorName, "curl", "-ksv", fmt.Sprintf("http://localhost:%d", servicePort), "-d",
```
jcantrill, line 87: looks like a method to be added to write.go and something with a convenience method to write to the named input

test/functional/inputs/http/http_input_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+	Context("When sending an array of audit log records to an HTTP input", func() {
+		It("should be able to round trip them unharmed", func() {
+			Expect(framework.DeployWithVisitor(
+				func(b *runtime.PodBuilder) error {
+					return framework.AddVectorHttpOutput(b, framework.Forwarder.Spec.Outputs[0])
+				}),
+			).To(BeNil())
+			type K8sAuditLogArray struct {
+				Items []types.K8sAuditLog `json:"items,omitempty"`
```
jcantrill, line 111: These types are ones we redefined in our repo but are these inputs not really audit logs from the core library?  It would be better to test those types

apis/logging/v1/input_source_types.go
syedriko: I was wondering myself. @alanconway, do we really need another Name here?

internal/collector/collector.go
```diff
@@ -196,13 +210,34 @@ func (f *Factory) NewCollectorContainer(secretNames []string, clusterID string)
 		{Name: f.ResourceNames.SecretMetrics, ReadOnly: true, MountPath: metricsVolumePath},
 		{Name: tmpVolumeName, MountPath: tmpPath},
 	}
+	for _, httpInput := range httpInputs {
+		collector.VolumeMounts = append(collector.VolumeMounts,
+			v1.VolumeMount{Name: httpInput, ReadOnly: true, MountPath: httpInputVolumePath + httpInput},
```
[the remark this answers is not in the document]
syedriko: I don't think this is affected by multiple collectors, This mount is for the secret with the HTTP input service's cert, it's not mounting a path from the host. Every collector is going to see its service secrets mounted at '/etc/collector/$input_name'.

internal/generator/vector/sources.go
syedriko: ^^^ https://github.com/openshift/cluster-logging-operator/pull/2155#discussion_r1313485608

test/functional/inputs/http/http_input_test.go
syedriko: I tried plugging in Event and EventList from k8s.io/apiserver/pkg/apis/audit into this test, but they lack JSON tags and don't marshal and unmarshal well. I tested the HTTP input with a live audit feed from a kube API server and it worked.

apis/logging/v1/input_source_types.go
syedriko: Take a look at https://vector.dev/docs/reference/configuration/sources/http_server/#examples. Even though they are sending Content-Type on the wire, they're still specifying encoding in the config.  We would need to change vector to take encoding from Content-Type.

internal/generator/vector/sources.go
syedriko: Only audit logs, as individual JSON-encoded events, one per line, and EventList JSON arrays have been tested so far, the code doesn't look at the format field. I need more clarity wrt the logType and format combinations.

test/functional/inputs/http/http_input_test.go
jcantrill: What is special about [this](https://github.com/openshift/eventrouter/blob/release-5.7/sinks/eventdata.go#L61) serializer which takes v1.Events and writes them to STDOUT as a json string rep of Event?  If we can not make it work the way we think we should at a minimum log an issue upstream to add the tags

internal/generator/vector/sources.go
jcantrill: I believe the API specifies "plan text" but given there is no support for it then we should remove that enum entry

apis/logging/v1/cluster_log_forwarder_types.go
alanconway: Replied on https://github.com/openshift/cluster-logging-operator/pull/1958 but I've closed that PR as it's not going to merge separately and this PR now has the "live" API design. We can continue discussion here.

apis/logging/v1/input_source_types.go
```diff
@@ -0,0 +1,52 @@
...
+	corev1 "k8s.io/api/core/v1"
+)
+
+// NOTE: The Enum validation on SourceSpec.Type must be updated if the list of types changes.
+
+// SourceSpec is a union of input source types.
+//
+// The fields of this struct define the set of known source types.
+type SourceSpec struct {
+	HTTP *HTTPSource `json:"http,omitempty"`
```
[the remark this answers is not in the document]
alanconway: Openstack already requested syslog server input, that was the original motivation here. We switched to HTTP first to resolve the hypershift issues but syslog remains on the backlog.

apis/logging/v1/input_source_types.go
alanconway: Also, not everything is fully described by content-type. There is an explict content-type for ndjson which tells us everything, but if the content type is just "json" then it could be a k8s event list or some other JSON format.

internal/generator/vector/inputs.go
alanconway: +1 you mean the constant `loggingv1.InputTypeAudit` right?

internal/generator/vector/sources.go
alanconway: +1, we can re-introduce later. It would be easy to implement but  I'm not sure if anybody will actually ship us logs as plain text over HTTP. All the HTTP logging formats I've seen so far wrap the original plain text message in JSON or protobuf.

apis/logging/v1/input_source_types.go
alanconway: Perhaps not, we need a name for the generated service but the name from the inputSpec should be enough.
I will review the API and see if we can simplify by making the source-input relationship 1-1.

test/functional/inputs/http/http_input_test.go
alanconway: +1,  we should use https://github.com/kubernetes/apiserver/blob/master/pkg/apis/audit/v1/types.go which is the official public API.
As far as I can tell /kubernetes/apiserver/blob/master/pkg/apis/audit/types.go is an internal detail of the kube-APIserver, which is just a copy (without tags!) of the public API. I imagine it's a historical remainder or a mistake by someone who was confused about go packages an thought they needed a copy in the same package.

test/functional/inputs/http/http_input_test.go
alanconway: On my audit branch I added a LogWriter and WriteLog to write.go, they  use a single "cat" command to pipe data directly to a log file. It is vastly faster than execing a a separate "echo" per line, and it keeps all the log generation in Go rather than quoted shell script parmeters, so avoids all the shell quoting madness:

https://github.com/alanconway/cluster-logging-operator/blob/api-audit-filter/test/framework/functional/write.go#L179

I'd suggest something similar here - i.e. rather than a curl command, set up a port-forward to the HTTP address and then use the Go http client to post directly.

alanconway:
@jcantrill I think the main issues left from your API comments are:
- name "source", I'm happy to change if we can decide on something clearer. "receiver"?
- use something simpler than service.Port - I'm open to  suggestion, but since we are going to generate a service, what is there in service.Port that is excess?

jcantrill:
> use something simpler than service.Port - I'm open to suggestion, but since we are going to generate a service, what is there in service.Port that is excess?

This is the struct for ServicePort:
```
struct ServicePort {
	Name string 
	Protocol Protocol
	AppProtocol *string 
	Port int32
	TargetPort intstr.IntOrString
	NodePort int32 
```
Services are designed to match a pod via a named port or specifically defined port but we control the service creation.  We control how it connects to our service.  The only thing we need is for the user to define on which port they want us to listen(or default one).  We do not need them to provide all the "wiring" between the fronting service and how ti connects to the pod.  I don't see the value in extended the API

apis/logging/v1/cluster_log_forwarder_types.go
jcantrill: Your other comment you suggested "receiver" which I think may be a more intuitive option.  "source" is very vector centric but may man comment is the others are already "sources"; they are specific sources so why is this new one not specific too?  It specifically "receives" or acts as a server?

apis/logging/v1/input_source_types.go
jcantrill: I'm not going to push the point if this is easier but we could define our own content types and what they represent.  I assume, however that our initial client won't be able to change so this is easier.

jcantrill, line 45: My previous comment from the other PR:
* we don't actually support "text" based on what I have seen in the rest of the impl so we should drop it or further define what "text" means.  I assume it would be a broad category
* Do we need to expand what k8s means?  I know its a list of "kubernetes events" but k8s is not very specific.  What if we get a list of pod yaml?  shoudnt we be more specific given we are only going to accept a list of v1.Events?

apis/logging/v1/input_source_types.go
```diff
@@ -0,0 +1,52 @@
...
+	//
+	// +kubebuilder:validation:Enum:=k8s;text
+	// +required
+	Format string `json:"format"`
+
+	// LogType indicates the type of logs received from this source.
+	//
+	// +kubebuilder:validation:Enum:=application;infrastructure;audit
+	// +required
+	LogType string `json:"logType"`
```
jcantrill, line 51: What does this tell us?  How we normalize the incoming logs?

internal/generator/vector/sources.go
jcantrill: in general, my opinion is we should drop any notion of "plain text" and only accept structured logs of a well defined format.  We are otherwise back to  3.x days before CRIO where we have to figure out a bunch of stuff and it is open ended

apis/logging/v1/input_source_types.go
syedriko: @jcantrill audit events

jcantrill: Aren't 'audit events' and corev1.Events the same?

syedriko: Nope, check https://kubernetes.io/docs/tasks/debug/debug-cluster/audit/  out. https://kubernetes.io/docs/reference/kubernetes-api/cluster-resources/event-v1/ vs https://kubernetes.io/docs/reference/config-api/apiserver-audit.v1/#audit-k8s-io-v1-Event.

test/functional/inputs/http/http_input_test.go
syedriko: @alanconway Done ^^^. Changes are in
test/client/client.go
test/framework/functional/read.go
test/framework/functional/write.go
test/functional/inputs/http/http_input_test.go

config/crd/bases/logging.openshift.io_clusterlogforwarders.yaml
```diff
@@ -114,6 +114,101 @@ spec:
...
+                                  description: The name of this port within the service.
+                                    This must be a DNS_LABEL. All ports within a ServiceSpec
+                                    must have unique names. When considering the endpoints
+                                    for a Service, this must match the 'name' field
+                                    in the EndpointPort. Optional if only one ServicePort
+                                    is defined on this service.
+                                  type: string
+                                nodePort:
+                                  description: 'The port on each node on which this
+                                    service is exposed when type is NodePort or LoadBalancer.  Usually
```
anpingli, line 165: How can we modify the type to NodePort?

config/crd/bases/logging.openshift.io_clusterlogforwarders.yaml
```diff
@@ -114,6 +114,101 @@ spec:
...
+                      description: Source to receive logs from non-cluster sources.
+                      properties:
+                        http:
+                          description: HTTPSource receives encoded logs as a HTTP
+                            endpoint.
+                          properties:
+                            format:
+                              description: Format is the format of incoming log data.
+                              enum:
+                              - k8s
```
anpingli, line 127: How to specify this value?    the incoming data format may be different. For example: infrastructure logs include container logs and journal logs. which may result in different format?

apis/logging/v1/input_source_types.go
alanconway: We could use extended content types like `application/json+k8s-event`. The potential problem in a HTTP setting is confusion between the content type on incoming HTTP headers `application/json` and how it matches to our "extended" type. If we do use content-type in future, I'd prefer it to  "expected HTTP content-type header of incoming requests" and keep a separate "format" to mean "extra information about how to interpret the HTTP body". This is a preview release though, so we can re-consider this issue before we GA.

apis/logging/v1/cluster_log_forwarder_types.go
alanconway: +1 lets change this.

internal/generator/vector/sources.go
alanconway: +1 we need to consider carefully the likely use cases and formats we will accept, hence leaving this till GA so we can think it through. There's limited value in accepting plain text logs over HTTP since  we have no information about where they came from, unlike when we are scraping from disk. For the audit case it is enough to configure just "log-type=audit" but when extending to other log types we may need to configure more "assumed source" information and/or parse it out of known log formats (leads back to your point)

alanconway:
> > use something simpler than service.Port - I'm open to suggestion, but since we are going to generate a service, what is there in service.Port that is excess?

@jcantrill +1 thanks, I understand now. We only need the "outside half" of the port.
Does this make sense?

struct ReceiverPort {
 	Name string 
 	Protocol Protocol
 	AppProtocol *string 
 	Port int32
}

apis/logging/v1/input_receiver_types.go
```diff
@@ -0,0 +1,28 @@
...
+// HTTPReceiver receives encoded logs as a HTTP endpoint.
+type HTTPReceiver struct {
+	// Port that this receiver will listen on.
+	// Used to create a Service for this port.
+	//
+	// +required
+	Port int32 `json:"port"`
+	// Format is the format of incoming log data.
+	//
+	// +kubebuilder:validation:Enum:=k8s_audit
```
alanconway, line 25: Change to `kubeAPIAudit` for consistency of caps and also naming of the API audit filter.
I'll LGTM this anyway, can fix in another merge

internal/collector/collector_test.go
```diff
@@ -355,7 +355,7 @@ var _ = Describe("Factory#NewPodSpec Add Cloudwatch STS Resources", func() {
 				podSpec := *factory.NewPodSpec(nil, logging.ClusterLogForwarderSpec{
 					Outputs:   outputs,
 					Pipelines: pipelines,
-				}, "1234", "", tls.GetClusterTLSProfileSpec(nil))
+				}, "1234", "", tls.GetClusterTLSProfileSpec(nil), []string{})
```
alanconway, line 358: Aside: it's safe to pass 'nil' for an empty slice, all the slice operators  (append, len  etc.) treat nil and [] as the same.
(note this does not apply to empty maps...)

alanconway (approved):
/lgtm

config/crd/bases/logging.openshift.io_clusterlogforwarders.yaml
alanconway: +1 - current preview is audit only for internal reasons. Other types  must be added in a later patch.

PR #2162, opened 2023-09-08
LOG-3856: Add standalone ocp console support for LokiStack migration

periklis:
### Description
Users migrating from Elasticsearch to LokiStack as the OpenShift Logging log store want to keep the flexibility to keep the existing stack while migrating to the new one. As per LokiStack installations being decoupled from ClusterLogging resources, this PR enables standalone OCP console installations via an ephemeral annotation `logging.openshift.io/ocp-console-migration-target` set to the LokiStack name w/o changing the `spec.logStore.Type` to `lokistack`. e.g.:

```yaml
kind: ClusterLogging
metadata:
  name: instance
  annotations:
    logging.openshift.io/ocp-console-migration-target: "lokistack-dev"
spec:
  managementState: Managed
  logStore:
    type: elasticsearch
    retentionPolicy:
      application:
        maxAge: 1d
      infra:
        maxAge: 7d
      audit:
        maxAge: 7d
    elasticsearch:
      nodeCount: 3
      storage:
        storageClassName: gp3-csi
        size: 20G
      resources:
        limits:
          memory: 2Gi
        requests:
          cpu: 200m
          memory: 2Gi
      proxy:
        resources:
          limits:
            memory: 256Mi
          requests:
            memory: 256Mi
      redundancyPolicy: SingleRedundancy
  visualization:
    type: kibana
    kibana:
      replicas: 1
  collection:
    logs:
      type: fluentd
      fluentd: { }
```

/cc @xperimental 
/assign @jcantrill

### DIFF
--- a/internal/constants/features.go
+++ b/internal/constants/features.go
@@ -12,4 +12,9 @@ const (
 	// AnnotationEnableSchema is the annotation to enable alternate output formats of logs.
 	// Currently only viaq & opentelemetry are supported
 	AnnotationEnableSchema = "logging.openshift.io/enableschema"
+
+	// AnnotationOCPConsoleMigrationTarget is to be used to enable the OCP Console for Logs
+	// without switching the default `logStore` to LokiStack. The value should be the
+	// LokiStack resource name representing the target store for the migration.
+	AnnotationOCPConsoleMigrationTarget = "logging.openshift.io/force-enable-ocp-console-target"
 )
--- a/internal/k8shandler/visualization.go
+++ b/internal/k8shandler/visualization.go
@@ -3,6 +3,7 @@ package k8shandler
 import (
 	"context"
 	"fmt"
+
 	log "github.com/ViaQ/logerr/v2/log/static"
 	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
 	"github.com/openshift/cluster-logging-operator/internal/constants"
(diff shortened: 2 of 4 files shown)


JoaoBraveCoding (commented):
from my side lgtm

internal/constants/features.go
```diff
@@ -7,5 +7,6 @@ const (
 	// +deprecated
 	UseOldRemoteSyslogPlugin = "clusterlogging.openshift.io/useoldremotesyslogplugin"
 
-	AnnotationDebugOutput = "logging.openshift.io/debug-output"
+	AnnotationDebugOutput               = "logging.openshift.io/debug-output"
+	AnnotationOCPConsoleMigrationTarget = "logging.openshift.io/ocp-console-migration-target"
```
xperimental, line 11: Does it make sense to rename this so it's more clear what it does? For example:

```suggestion
	AnnotationOCPConsoleMigrationTarget = "logging.openshift.io/force-enable-ocp-console-target"
```

xperimental (commented):
Enabling the ocp-console using the annotation seems to work, but it does not seem to get updated when the content of the annotation changes or if it is removed.

version/version.go
```diff
@@ -1,3 +1,3 @@
 package version
 
-var Version = "5.8.0"
+var Version = "5.8.0-preview.1"
```
jcantrill, line 3: Why are we updating this.  Otherwise LGTM with the aside that I thought @xperimental had questions about the naming of the annotation to enable.

periklis: @jcantrill Rename came from `make run-bundle` I believe. We renamed the annotation already.

PR #2142, opened 2023-08-25
WIP: CLO support for optional OTEL data model

jatinsu:
### Description
Allow the Cluster Logging Operator to support the Open Telemetry data model via a filter in the Pipelines.go file

/cc @alanconway 
/assign @alanconway

### Links
- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-4231
- Enhancement proposal: https://github.com/openshift/enhancements/pull/1442

PR #2163, opened 2023-09-11
LOG-4231: CLO operator support for optional OTEL data model

Clee2691:
## Description
This PR is a continuation of the effort by Jatin Suri to support transforming the log record formats to the OpenTelemetry semantic.

This feature is:
- Locked behind a feature-gate
- Applies only to `application` logs
- Can only be defined for `HTTP` outputs under the `HTTP` output spec

## Field Changes:
| Original| New	|
|---|---|
| @timestamp |	timeUnixNano |
| level	|severityText |
| N/A	|	severityNumber |
| file	|resources.logs.file.path |
| hostname	|resources.host.name |
| kubernetes.container_name	|resources.container.name |
| kubernetes.container_id	|resources.container.id |
| kubernetes.container_image	|resources.container.image.name |
| N/A	|resources.container.image.tag |
| kubernetes.pod_name	|resources.k8s.pod.name |
| kubernetes.pod_id	|resources.k8s.pod.uid |
| kubernetes.pod_ip	|resources.k8s.pod.ip |
| kubernetes.pod_owner	|resources.k8s.pod.owner |
| kubernetes.annotations	|resources.k8s.pod.annotations |
| kubernetes.labels	|resources.k8s.pod.labels |
| kubernetes.namespace_id	|resources.k8s.namespace.id |
| kubernetes.namespace_labels."kubernetes.io/metadata.name"	|resources.k8s.namespace.name |
| kubernetes.namespace_labels	|resources.k8s.namespace.labels |
| log_type	|resources.attributes.log_type |

/cc @alanconway @cahartma
/assign @jcantrill 

### Links
- Based on PR: https://github.com/openshift/cluster-logging-operator/pull/2142
- JIRA: https://issues.redhat.com/browse/LOG-4231
- Enhancement Proposal: https://github.com/openshift/enhancements/pull/1442

### DIFF
--- a/internal/generator/vector/normalize/schema/otel/transform.go
+++ b/internal/generator/vector/normalize/schema/otel/transform.go
@@ -1,0 +1,58 @@
+package otel
+
+import (
+	"fmt"
+	"strings"
+
+	. "github.com/openshift/cluster-logging-operator/internal/generator"
+	. "github.com/openshift/cluster-logging-operator/internal/generator/vector/elements"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
+)
+
+func ID(id1, id2 string) string {
+	return fmt.Sprintf("%s_%s", id1, id2)
+}
+
+func Transform(id string, inputs []string) Element {
+	return Remap{
+		Desc:        "Normalize log records to OTEL schema",
+		ComponentID: id,
+		Inputs:      helpers.MakeInputs(inputs...),
+		VRL: strings.TrimSpace(`
+# Tech preview, OTEL for application logs only
+if .log_type == "application" {
+	# Convert @timestamp to nano and delete @timestamp
+	.timeUnixNano = to_unix_timestamp(to_timestamp!(del(.@timestamp)))
+
+	.severityText = del(.level)
+
+	# Convert syslog severity keyword to number, default to 9 (unknown)
+	.severityNumber = to_syslog_severity(.severityText) ?? 9
+
+	# resources
+	.resources.logs.file.path = del(.file)
+	.resources.host.name= del(.hostname)
+	.resources.container.name = del(.kubernetes.container_name)
+	.resources.container.id = del(.kubernetes.container_id)
+  
+	# split image name and tag into separate fields
+	container_image_slice = split!(.kubernetes.container_image, ":", limit: 2)
+	if null != container_image_slice[0] { .resources.container.image.name = container_image_slice[0] }
+	if null != container_image_slice[1] { .resources.container.image.tag = container_image_slice[1] }
+	del(.kubernetes.container_image)
+	
+	# kuberenetes
+	.resources.k8s.pod.name = del(.kubernetes.pod_name)
+	.resources.k8s.pod.uid = del(.kubernetes.pod_id)
+	.resources.k8s.pod.ip = del(.kubernetes.pod_ip)
+	.resources.k8s.pod.owner = .kubernetes.pod_owner
+	.resources.k8s.pod.annotations = del(.kubernetes.annotations)
+	.resources.k8s.pod.labels = del(.kubernetes.labels)
+	.resources.k8s.namespace.id = del(.kubernetes.namespace_id)
+	.resources.k8s.namespace.name = .kubernetes.namespace_labels."kubernetes.io/metadata.name"
+	.resources.k8s.namespace.labels = del(.kubernetes.namespace_labels)
+	.resources.attributes.log_type = del(.log_type)
+}
+`),
+	}
+}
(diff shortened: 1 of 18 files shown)


internal/generator/vector/output/http/http.go
```diff
@@ -133,6 +134,19 @@ func Conf(o logging.OutputSpec, inputs []string, secret *corev1.Secret, op Optio
 	)
 }
 
+func Schema(o logging.OutputSpec, outputName, component string, inputs []string, op Options) []Element {
+	if op[constants.AnnotationEnableSchema] == "true" && o.Http.Schema == constants.OTELSchema {
```
alanconway, line 138: Need nil check in case o is not a HTTP output.

alanconway (approved):
This will work, restricts OTEL to the HTTP output which we will want to open up in the long run, but it's a good conservative choice for the preview given unresolved headaches.
Ultimately I want to move OTEL to a "filter" like the API audit stuff, but we should go ahead with this approach as the first cut since we may not have time to clean up all the loose ends after I merge the audit filter.

alanconway:
See comment about missing nil check - then I'll LGTM

internal/generator/vector/normalize/schema/otel/transform.go
```diff
@@ -0,0 +1,58 @@
...
+		VRL: strings.TrimSpace(`
+# Tech preview, OTEL for application logs only
+if .log_type == "application" {
+	# Convert @timestamp to nano and delete @timestamp
+	.timeUnixNano = to_unix_timestamp(to_timestamp!(del(.@timestamp)))
+
+	.severityText = del(.level)
+
+	# Convert syslog severity keyword to number, default to 9 (unknown)
+	.severityNumber = to_syslog_facility(.severityText) ?? 9
```
jcantrill, line 30: severity is not the same as facility (e.g. https://success.trendmicro.com/dcx/s/solution/TP000086250?language=en_US) .  Regardless, I think our fluent code has 0-9 but I believe 0-7 have always been the only ones available

internal/generator/vector/normalize/schema/otel/transform.go
```diff
@@ -0,0 +1,58 @@
...
+	
+	# kuberenetes
+	.resources.k8s.pod.name = del(.kubernetes.pod_name)
+	.resources.k8s.pod.uid = del(.kubernetes.pod_id)
+	.resources.k8s.pod.ip = del(.kubernetes.pod_ip)
+	.resources.k8s.pod.owner = .kubernetes.pod_owner
+	.resources.k8s.pod.annotations = del(.kubernetes.annotations)
+	.resources.k8s.pod.labels = del(.kubernetes.labels)
+	.resources.k8s.namespace.id = del(.kubernetes.namespace_id)
+	.resources.k8s.namespace.name = .kubernetes.namespace_labels."kubernetes.io/metadata.name"
```
jcantrill, line 52: if this is missing will this return empty string or null?

internal/generator/vector/output/http/http.go
```diff
@@ -133,6 +134,19 @@ func Conf(o logging.OutputSpec, inputs []string, secret *corev1.Secret, op Optio
 	)
 }
 
+func Schema(o logging.OutputSpec, outputName, component string, inputs []string, op Options) []Element {
+	if op[constants.AnnotationEnableSchema] == "true" && o.Http != nil && o.Http.Schema == constants.OTELSchema {
```
jcantrill, line 138: More nit then anything, I defer to you if you wish to change.  Options#Has(key) exists as a method and we explicitly give a value to Options meaning we don't care about the actual value.  It may read cleaner to adjust

test/helpers/types/types.go
```diff
@@ -29,6 +30,8 @@ const (
 type ContainerLog struct {
 	ViaQCommon `json:",inline,omitempty"`
 
+	OTELApplicationLog `json:",inline,omitempty"`
```
jcantrill, line 33: Do we need to add this into a different top level log type (e.g. OTELContainerLog, OTELLog)? I'm assuming OTEL format will likely be agnostic of Container versus other.  I'm thinking ahead of when we no longer normalize VIAQ at all and seems it would be more appropriate to have start the different normalization down their own path.  

Consider moving all "otel" struct to a separate file

jcantrill (changes requested):

internal/generator/vector/normalize/schema/otel/transform.go
Clee2691: It returns an empty string

test/helpers/types/types.go
Clee2691: Move to `test/helpers/schema/otel/types.go`.

jcantrill (approved):
/lgtm

We could actually do normalization tests in the same way as the others with templates.  Consider doing that in future so we can test "all" fields in lieu of verifying individually

PR #2164, opened 2023-09-13
LOG-4487: sort Map of Routes for predictable configuration generation during reconciliation

vparfonov:
### Description

In this PR, we aim to resolve an issue causing configuration discrepancies during reconciliation, ultimately leading to unnecessary collector pod restarts. The root of the problem lies in the behavior of the `Go range function`, which does not guarantee the same ordering in each call for the same map (see: https://go.dev/blog/maps). This unpredictability can result in occasional variations in configuration generation when processing the `Routes map` [(source),](https://github.com/openshift/cluster-logging-operator/pull/2164/files#diff-6b7baf71ad4874089f6aa160edf27cdd8914a23a5d8db2fad126e352db8459daR11) impacting the stability of the system. Although this change in ordering means nothing for `Vector`, it significantly influences MD5 hash calculations, that we use for detecting changes in configuration.

For example, if create an instance of CLF with this YAML:  
```
kind: ClusterLogForwarder
metadata:
  name: instance
  namespace: openshift-logging
spec:
  inputs:
  - application:
      containerLimit:
        maxRecordsPerSecond: 20
      namespaces:
      - test-1
    name: limited-rates-1
  - application:
      containerLimit:
        maxRecordsPerSecond: 20
      namespaces:
      - test-2
    name: limited-rates-2
  pipelines:
  - inputRefs:
    - limited-rates-1
    - limited-rates-2
    - infrastructure
    name: to-default
    outputRefs:
    - default 
```
Collector pods keep being recreated because of differences of md5 hash in env variable: `COLLECTOR_CONF_HASH`.
Here is diff of generated Vector config: 

```
201c201
< [transforms.source_throttle_limited-rates-2]
---
> [transforms.source_throttle_limited-rates-1]
203c203
< inputs = ["route_application_logs.limited-rates-2"]
---
> inputs = ["route_application_logs.limited-rates-1"]
209c209
< [transforms.source_throttle_limited-rates-1]
---
> [transforms.source_throttle_limited-rates-2]
211c211
< inputs = ["route_application_logs.limited-rates-1"]
---
> inputs = ["route_application_logs.limited-rates-2"]
```
and md5 hash:
```

< "collectorConfHash":"336e99a4eeadb2d0c1a2a65b51816c4b"
---
> "collectorConfHash":"02794c06255cf667f06bc8f1a86653bc"
```

To tackle this issue and ensure a consistent and predictable configuration generation process, we have introduced a solution to sort the Routes map before generating the configuration. By doing so, we establish a consistent ordering for the map's keys across multiple runs. This guarantees stability in configuration outputs and eliminates the source of discrepancies during reconciliation.

/cc @jcantrill @cahartma 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-4487
- Enhancement proposal:

### DIFF
--- a/internal/generator/generator.go
+++ b/internal/generator/generator.go
@@ -2,6 +2,7 @@ package generator
 
 import (
 	"bytes"
+	"sort"
 	"strings"
 	"text/template"
 )
@@ -64,6 +65,14 @@ func (g Generator) generate(es []Element) (string, error) {
 		"indent": indent,
 		"comma_separated": func(arr []string) string {
 			return strings.Join(arr, ", ")
+		},
+		"getSortedKeyFromMap": func(arg map[string]string) []string {
+			var keys []string
+			for key := range arg {
+				keys = append(keys, key)
+			}
+			sort.Strings(keys)
+			return keys
 		},
 	}
 	f["optional"] = f["kv"]
--- a/internal/generator/vector/elements/elements.go
+++ b/internal/generator/vector/elements/elements.go
@@ -23,8 +23,10 @@ func (r Route) Template() string {
 [transforms.{{.ComponentID}}]
 type = "route"
 inputs = {{.Inputs}}
-{{- range $route_name, $route_expr := .Routes}}
-route.{{$route_name}} = {{$route_expr}}
+{{- $values := .Routes -}}
+{{- $keys := getSortedKeyFromMap .Routes -}}
+{{ range $route_name := $keys}}
+route.{{$route_name}} = {{index $values $route_name}}
 {{- end}}
 {{end}}
 `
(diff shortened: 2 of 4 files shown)


jcantrill:
/approve

There is at least one sortedmap [impl](https://pkg.go.dev/github.com/umpc/go-sortedmap) which could simplify some of your changes but I will defer to you if you wish to modify the PR.  I'm otherwise good with this change as we have addressed similar issues in the past in the exact same way.  If you don't want to make any change, please release the hold

vparfonov:
> /approve
> 
> There is at least one sortedmap [impl](https://pkg.go.dev/github.com/umpc/go-sortedmap) which could simplify some of your changes but I will defer to you if you wish to modify the PR. I'm otherwise good with this change as we have addressed similar issues in the past in the exact same way. If you don't want to make any change, please release the hold

We do not have dependencies on this library, and I prefer not to add new dependencies for such minor changes.

PR #2170, opened 2023-09-21
LOG-4534 add multi toml and lfme to must-gather

cahartma:
### Description
Additional updates for must-gather fix original changes: https://github.com/openshift/cluster-logging-operator/pull/2143

This change adds additional vector toml files by appending the "name_" to the filename.   We are also now collecting the  "oc get lfme -o yaml" for each namespace that contains an instance.   They are being gathered in a file called "lfme.yaml"

/cc @syedriko @vparfonov 
/assign @jcantrill 

### Links
Original
- https://issues.redhat.com/browse/LOG-4404
More fixes:
- https://issues.redhat.com/browse/LOG-4534


jcantrill (commented):
/lgtm

PR #2171, opened 2023-09-21
LOG-4411: fix vector config for unreferenced output

jcantrill:
### Description
This PR:
* Adds CLF migration to drop outputs which are not referenced by pipelines to be consistent with previous release
* Adds validation to reject CLF that have unreferenced outputs
* Adds CLF conditions identifying when an output is not considered during reconciliation
* Adds matchers for conditions into the test package

### Links
* https://issues.redhat.com/browse/LOG-4411

### DIFF
--- a/test/matchers/condition.go
+++ b/test/matchers/condition.go
@@ -1,0 +1,35 @@
+package matchers
+
+import (
+	//"fmt"
+	"github.com/onsi/gomega/types"
+	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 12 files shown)


test/matchers/condition.go
```diff
@@ -0,0 +1,35 @@
+package matchers
+
+import (
+	//"fmt"
+	"github.com/onsi/gomega/types"
+	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
+	corev1 "k8s.io/api/core/v1"
+	//"k8s.io/utils/diff"
```
vparfonov, line 8: nit: comment code

PR #2172, opened 2023-09-22
LOG-4555: Show FluentD Buffer Usage in metrics dashboard instead of availability

bysnupy:
### Description

This PR:
* fixes that "FluentD Buffer Availability" can show us the buffer usage based on the most highest usage value.

If "fluentd_output_status_buffer_available_space_ratio" metrics returns 100 and 50, "max" aggregation operator(current) always show us 0%(100-100) buffer usage, not 50%(100-50) buffer usage. It's not helpful to detect collector pods in high buffer usage. So it should change it to "min" aggregation operator for showing us values based on the most highest usage of buffer.

### Links
N/A

### DIFF
--- a/files/dashboards/fluentd/openshift-logging-dashboard.json
+++ b/files/dashboards/fluentd/openshift-logging-dashboard.json
@@ -1400,7 +1400,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "100 - max by(plugin_id)(fluentd_output_status_buffer_available_space_ratio)",
+              "expr": "100 - min by(plugin_id)(fluentd_output_status_buffer_available_space_ratio)",
               "refId": "A"
             }
           ],
@@ -1408,7 +1408,7 @@
           "timeFrom": null,
           "timeRegions": [],
           "timeShift": null,
-          "title": "FluentD Buffer Availability",
+          "title": "FluentD Buffer Usage",
           "tooltip": {
             "shared": true,
             "sort": 0,


jcantrill:
@bysnupy please log a JIRA and associate it with this change so it can be properly tracked by errata

files/dashboards/fluentd/openshift-logging-dashboard.json
```diff
@@ -1400,7 +1400,7 @@
           "steppedLine": false,
           "targets": [
             {
-              "expr": "100 - max by(plugin_id)(fluentd_output_status_buffer_available_space_ratio)",
+              "expr": "100 - min by(plugin_id)(fluentd_output_status_buffer_available_space_ratio)",
```
xperimental, line 1403: I'm interpreting the title of the graph as "it should show the available buffer percentage". If that's the case in addition to the change from `max` to `min` we should also remove the `100 - `:

```suggestion
              "expr": "min by(plugin_id)(fluentd_output_status_buffer_available_space_ratio)",
```

If we want to keep showing the "usage" like we currently do, I'd suggest we change the name of the graph accordingly (`FluentD Buffer Usage`, for example). The change from `max` to `min` is still valid in that case.

bysnupy: I think "FluentD Buffer Usage" is better choice, because usage is intuitive than availability usually for users.

bysnupy:
FYI, I already opened the docs ticket about this here: https://github.com/openshift/openshift-docs/pull/65147

xperimental:
/retitle LOG-4555: Show FluentD Buffer Usage in metrics dashboard instead of availability

PR #2173, opened 2023-09-22
LOG-4362: remove GroupLimit from flow control

jcantrill:
### Description
This PR drops support of GroupLimit in the flow control epic by:
* disabling JSON serialization of the field
* removing it from the CRD and generated API docs

GroupLimit applies per instance of the collector so it can be misleading from an API perspective.  We agreed we need to revisit how to convey applying it to a group of pods that may be dispersed across several nodes.  Leaving the implementation to facilitate future corrections

### Links
* https://issues.redhat.com/browse/LOG-4362

cc @alanconway


alanconway (approved):
/lgtm

PR #2177, opened 2023-09-25
LOG-3651: Add ability to change the Splunk's index within the ClusterLogForwarder

vparfonov:
### Description
This PR adds the ability to change the Splunk's `index` within the `ClusterLogForwarder` for Vector collector.

Example of `ClusterLogForwarder` configuration: 
```
kind: ClusterLogForwarder
metadata:
  name: instance
  namespace: openshift-logging
spec:
  outputs:
    - name: splunk-receiver
      type: splunk
      splunk:
       index: "my_index"
      secret:
        name: splunk-secret
      url: 'http://example-splunk-hec-service:8088'
  pipelines:
    - name: my-logs
      inputRefs:
        - application
        - infrastructure
      outputRefs:
        - splunk-receiver 
```

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/output/splunk/splunk.go
+++ b/internal/generator/vector/output/splunk/splunk.go
@@ -27,6 +27,7 @@ type Splunk struct {
 	Inputs       string
 	Endpoint     string
 	DefaultToken string
+	Index        string
 }
 
 func (s Splunk) Name() string {
@@ -42,6 +43,9 @@ endpoint = "{{.Endpoint}}"
 compression = "none"
 default_token = "{{.DefaultToken}}"
 timestamp_key = "@timestamp"
+{{if .Index -}}
+index = "{{.Index}}"
+{{end -}} 
 {{end}}`
 }
 
@@ -82,12 +86,16 @@ func Conf(o logging.OutputSpec, inputs []string, secret *corev1.Secret, op Optio
 }
 
 func Output(o logging.OutputSpec, inputs []string, secret *corev1.Secret, op Options) Element {
-	return Splunk{
+	splunk := Splunk{
 		ComponentID:  vectorhelpers.FormatComponentID(o.Name),
 		Inputs:       vectorhelpers.MakeInputs(inputs...),
 		Endpoint:     o.URL,
 		DefaultToken: security.GetFromSecret(secret, constants.SplunkHECTokenKey),
 	}
+	if o.Splunk != nil && strings.TrimSpace(o.Splunk.Index) != "" {
+		splunk.Index = o.Splunk.Index
+	}
+	return splunk
 }
 
 func Encoding(o logging.OutputSpec) Element {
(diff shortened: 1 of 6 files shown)


internal/generator/vector/output/splunk/splunk.go
```diff
@@ -1,102 +1,110 @@
...
+  return `{{define "` + s.Name() + `" -}}
 [sinks.{{.ComponentID}}]
 type = "splunk_hec_logs"
 inputs = {{.Inputs}}
 endpoint = "{{.Endpoint}}"
 compression = "none"
 default_token = "{{.DefaultToken}}"
 timestamp_key = "@timestamp"
+{{if .Index -}}
+index = "{{.Index}}"
```
jcantrill, line 45: This will need to:
* support [templeting](https://vector.dev/docs/reference/configuration/sinks/splunk_hec_logs/#index) 
* need to understand what happens when the field is not found
* Update documentation with proper template example

jcantrill (commented):
Do we have a functional test for verification?

internal/generator/vector/output/splunk/splunk.go
vparfonov: Documentation updated https://github.com/openshift/cluster-logging-operator/pull/2177/files#diff-46e6d1852418f35b4db6dde8b446604e9d144d0cb4838d378ed8e2609156bca8R61

vparfonov: > This will need to:
> 
> * support [templeting](https://vector.dev/docs/reference/configuration/sinks/splunk_hec_logs/#index)
Template can be useful for separating events probably, but i not sure do we really need it.  
> * need to understand what happens when the field is not found
If you mean that will be if we try to send events to not existed index, here that i found:

_Splunk allows for automatic index creation. This means that if you send an event to an index that doesn't exist, Splunk will attempt to create the index for you. 
If automatic creation is disabled, the event might be rejected or placed in a default index, depending on Splunk configuration._

> * Update documentation with proper template example

Documentation updated https://github.com/openshift/cluster-logging-operator/pull/2177/files#diff-46e6d1852418f35b4db6dde8b446604e9d144d0cb4838d378ed8e2609156bca8R61

vparfonov:
> Do we have a functional test for verification?

No, only e2e test

internal/generator/vector/output/splunk/splunk.go
vparfonov: If will be used, not existed field in index template, we will get error in collector: 
`2023-09-26T10:43:55.867961Z ERROR sink{component_kind="sink" component_id=splunk component_type=splunk_hec_logs component_name=splunk}: vector::internal_events::template: Failed to render template for "index". error=Missing fields on event: [".kubernetes.namespacename"] error_type="template_failed" stage="processing" internal_log_rate_limit=true
`

PR #2183, opened 2023-09-29
Allow buffer and request tuning

jcantrill:
### Description
This PR:
* suggests a way to override buffer and request settings for vector collectors
* selectively allows tuning options

/hold

cc @alanconway @syedriko 

given
```
  outputs:
  - name: foo
    type: elasticsearch
    url: http://google
    tuning:
      buffer.max_events: "100"
      buffer.when_full: block
```

generates

```
[sinks.foo]
type = "elasticsearch"
inputs = ["foo_dedot_and_flatten"]
endpoints = ["http://google"]
bulk.index = "{{ write_index }}"
bulk.action = "create"
encoding.except_fields = ["write_index"]
request.timeout_secs = 2147483648
id_key = "_id"
api_version = "v6"

[sinks.foo.buffer]
max_events = 100
when_full = "block"


```


apis/logging/v1/output_tuning_type.go
```diff
@@ -0,0 +1,82 @@
...
+			errors = append(errors, fmt.Errorf("tuning key not allowed: %s", k))
+		}
+	}
+	return errors
+}
+
+func isValid(key string, value string) bool {
+	switch key {
+	case bufferWhenFull:
+		return sets.NewString("block", "drop_newest").Has(value)
```
syedriko, line 75: value == `block` || value == `drop_newest`

syedriko:
I didn't catch how the 'request.*' tunables end up in vector config.

syedriko:
@jcantrill We need a way to turn on allocation tracing (which I tack on as --allocation-tracing CLI param in vector run script), but that's not specific to an output, but affects vector as a whole.

jcantrill:
> I didn't catch how the 'request.*' tunables end up in vector config.

I didn't implement it because we will need some other minor alterations for the places where we do set these values

jcantrill:
> @jcantrill We need a way to turn on allocation tracing (which I tack on as --allocation-tracing CLI param in vector run script), but that's not specific to an output, but affects vector as a whole.

Can't we enable as an annotation that gets passed to the runscript as an arg to configure vector?... or is there an env var that enables it?

PR #2198, opened 2023-10-05
NO-JIRA: Makefile, fix missing references to $(OPERATOR_SDK)

alanconway:
/cc @cahartma


jcantrill:
Holding for code freeze so as to not trigger unnecessary build

PR #2203, opened 2023-10-09
LOG-4597: modify vector buffer and request to have mostly parity with…

jcantrill:
… fluentd

### Description
This PR:
* Adds buffer and request settings to all outputs for vector inorder to achieve  some parity to prior releases
* Modifies http api timeout to be an int instead of string of an int

### Links
* https://issues.redhat.com/browse/LOG-4597

cc @alanconway @syedriko

### DIFF
--- a/apis/logging/v1/output_types.go
+++ b/apis/logging/v1/output_types.go
@@ -286,7 +286,7 @@ type Http struct {
 
 	// Timeout specifies the Http request timeout in seconds. If not set, 10secs is used.
 	// +optional
-	Timeout string `json:"timeout,omitempty"`
+	Timeout int `json:"timeout,omitempty"`
 
 	// Method specifies the Http method to be used for sending logs. If not set, 'POST' is used.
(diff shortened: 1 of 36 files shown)


apis/logging/v1/output_types.go
```diff
@@ -286,7 +286,7 @@ type Http struct {
 
 	// Timeout specifies the Http request timeout in seconds. If not set, 10secs is used.
 	// +optional
-	Timeout string `json:"timeout,omitempty"`
+	Timeout int `json:"timeout,omitempty"`
```
alanconway, line 289: Should be `TimeoutSeconds` for clarity. IIRC this is in the k8s API guidelines for time variables. 
Not a blocker if it's too late to change.

alanconway (changes requested):

alanconway:
/hold
Suggested a name change, otherwise all looks good.

apis/logging/v1/output_types.go
jcantrill: The http type was released for 5.7 so there is not a place to change it.  I have concerns about changing the time as-is

cahartma: Can we release the hold?

PR #2211, opened 2023-10-13
Revert "LOG-4607: fix install mode to only AllNamespaces"

jcantrill:
This reverts commit b2384abf1c686ba7574c9fe337fcb356ef512ca5.

### Description
This PR reverts the prior change after testing that existing deployments can successfully upgrade to 5.8.  The implications here is existing deployments will need to adjust there operatorgroup to take full advantage of mCLF.  Additionally it may be new deployments must explicitly choose "all namespaces" or will need to edit the operatorgroup in the future if they do not

### Links
* https://issues.redhat.com/browse/LOG-4650


jcantrill:
cc @alanconway

PR #2215, opened 2023-10-17
LOG-4676: Modify receiverport to be a simple int instead of a struct

syedriko:
### Description
ReceiverPort is defined as a complex struct which mimics the need of defining a port for a generic service. The CLO controls all aspects of the receiver so it should need to have this complexity; users should only need to define potentially the port upon which to listen. Modify the API and code to convert 'receiverPort' into an int

/cc @alanconway 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4676


internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
```diff
@@ -201,10 +201,8 @@ func verifyInputs(spec *loggingv1.ClusterLogForwarderSpec, status *loggingv1.Clu
 			badInput("ReceiverSpec must define an HTTP receiver")
 		case isHTTPReceiver(input) && input.Receiver.HTTP.Format != loggingv1.FormatKubeAPIAudit:
 			badInput("invalid format specified for HTTP receiver")
-		case isHTTPReceiver(input) && isInvalidPort(input.Receiver.HTTP.ReceiverPort.Port):
+		case isHTTPReceiver(input) && isInvalidPort(input.Receiver.HTTP.Port):
```
jcantrill, line 204: is it going to be a problem to allow 1-1024?  I seem to recall those being reserved ports

jcantrill (commented):
looks good except one question.  also please run "make docs"

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
syedriko: Ran `make docs`, excluded ports [1,1023]

PR #2216, opened 2023-10-18
LOG-4612: Service/secret is not deleted when the inputs.receiver.http is removed

Clee2691:
### Description
This PR addresses removing stale `HTTP` input services when the `ClusterLogForwarder` (CLF) is modified. 

Originally when the `inputs.receiver.http` was removed from the CLF and updated, NOT deleted and recreated, any previously defined `HTTP` input services were not garbage collected. 

The approach was to label all `HTTP` input services with `http-input-service` and upon CLF reconciliation, check if the service is defined in the CLF as well as if the reconciling CLF owns the specific service. If the service is not defined, remove it.

/cc @syedriko @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4612

### DIFF
--- a/internal/k8shandler/collection.go
+++ b/internal/k8shandler/collection.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/openshift/cluster-logging-operator/internal/auth"
+	"github.com/openshift/cluster-logging-operator/internal/utils/sets"
 
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
@@ -109,14 +110,7 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection() (err err
 		return err
 	}
 
-	var httpInputs []string
-	for _, input := range clusterRequest.Forwarder.Spec.Inputs {
-		if input.Receiver != nil && input.Receiver.HTTP != nil {
-			httpInputs = append(httpInputs, input.Name)
-		}
-	}
-
-	if err := factory.ReconcileDaemonset(clusterRequest.EventRecorder, clusterRequest.Client, clusterRequest.Forwarder.Namespace, clusterRequest.ResourceOwner, httpInputs); err != nil {
+	if err := factory.ReconcileDaemonset(clusterRequest.EventRecorder, clusterRequest.Client, clusterRequest.Forwarder.Namespace, clusterRequest.ResourceOwner); err != nil {
 		log.Error(err, "collector.ReconcileDaemonset")
 		return err
 	}
@@ -140,7 +134,7 @@ func (clusterRequest *ClusterLoggingRequest) removeCollector() (err error) {
 			return err
 		}
 
-		if err = clusterRequest.RemoveInputServices(); err != nil {
+		if err = clusterRequest.RemoveInputServices([]metav1.OwnerReference{utils.AsOwner(clusterRequest.Forwarder)}, true); err != nil {
 			return
 		}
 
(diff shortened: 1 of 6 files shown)


internal/k8shandler/collection.go
```diff
@@ -93,6 +95,18 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection() (err err
 		return err
 	}
 
+	var httpInputs []string
```
syedriko, line 98: I would move this newly added code into ReconcileInputServices(), since what it is doing is reconciling input services and as it is, it breaks the pattern in this function of a sequence of Reconcile<T>.

internal/k8shandler/collection.go
```diff
@@ -199,17 +206,27 @@ func (clusterRequest *ClusterLoggingRequest) addSecurityLabelsToNamespace() erro
...
+func (clusterRequest *ClusterLoggingRequest) RemoveInputServices(httpInputs []string, currOwner []metav1.OwnerReference, removeAllServices bool) error {
 	if clusterRequest.Cluster.Spec.Collection.Type != logging.LogCollectionTypeVector {
 		return nil
 	}
 
-	for _, input := range clusterRequest.Forwarder.Spec.Inputs {
-		if input.Receiver != nil && input.Receiver.HTTP != nil {
-			if err := clusterRequest.RemoveInputService(input.Name); err != nil {
+	// Find all http input services by label
+	labelSelector, _ := labels.Parse("component=http-input-service")
```
syedriko, line 215:
```
labelSelector, _ := labels.Parse("component=http-input-service")
&client.ListOptions{LabelSelector: labelSelector}
```
can be lifted to the package level as they don't depend on anything inside this function.

internal/network/service.go
```diff
@@ -60,6 +60,8 @@ func ReconcileInputService(er record.EventRecorder, k8sClient client.Client, nam
 		constants.AnnotationServingCertSecretName: certSecretName,
 	}
 
+	desired.Labels["component"] = "http-input-service"
```
syedriko, line 63: It would be nice to make "component" and "http-input-service" constants and use them here and in labels.Parse().

internal/k8shandler/collection.go
syedriko, line 98:
```
httpInputs := sets.NewString()
...
httpInputs.Insert(input.Name)
```
? Make the collection of HTTP inputs a set from the start and avoid copying in RemoveInputServices()?

internal/k8shandler/collection.go
```diff
@@ -199,17 +206,27 @@ func (clusterRequest *ClusterLoggingRequest) addSecurityLabelsToNamespace() erro
...
+	httpServices := corev1.ServiceList{}
+	if err := clusterRequest.Client.List(context.TODO(), &httpServices, &client.ListOptions{LabelSelector: labelSelector}); err != nil {
```
syedriko, line 217: Not sure if it's possible, but can we include the owner reference into the query so we only get services owned by currOwner and with "component=http-input-service" label?

internal/k8shandler/collection.go
Clee2691: This wasn't new code, I just moved it around as it was later in this function. However, I have now moved this to where I think the appropriate location should be.

internal/network/service.go
Clee2691: Done

internal/k8shandler/collection.go
Clee2691: Done

internal/k8shandler/collection.go
Clee2691: Unfortunately it is not possible to query on `OwnerRefs`

internal/k8shandler/service.go
```diff
@@ -40,3 +43,13 @@ func (clusterRequest *ClusterLoggingRequest) RemoveInputService(serviceName stri
 
 	return nil
 }
+
+// GetServiceList returns a list of services based on a key/value label
+func (clusterRequest *ClusterLoggingRequest) GetServiceList(key string, val string) (*core.ServiceList, error) {
```
syedriko, line 48: Since we're only going to be looking at the services owned by a specific CLF, would it make sense to query here only for services in the same namespace as the CLF?

Clee2691: Yes that is a good idea so as to reduce the number of services listed. I will make the change.

jcantrill (commented):
/approve

jcantrill:
holding for 5.8 z-stream

PR #2218, opened 2023-10-18
LOG-4655:No error in CLF when the filterRefs refers to a non-existing filter

Clee2691:
### Description
This PR addresses the fact that no error is raised when a pipeline references a non-existing/ non-defined filter.

/cc @vparfonov @syedriko 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4655

### DIFF
--- a/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
+++ b/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
@@ -70,19 +70,26 @@ func verifyRefs(what, forwarderName string, status loggingv1.ClusterLogForwarder
 
 	msg := []string{}
 	for _, ref := range refs {
-		if what == "inputs" {
+		switch what {
+		case "inputs":
 			if loggingv1.ReservedInputNames.Has(ref) ||
 				(allowed.Has(ref) && status.Inputs[ref].IsTrueFor(loggingv1.ConditionReady)) {
 				good.Insert(ref)
 			} else {
 				bad.Insert(ref)
 			}
-		} else if what == "outputs" {
+		case "outputs":
 			// Check if custom CLF is forwarding to default store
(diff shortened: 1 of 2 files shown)


internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
```diff
@@ -87,6 +88,12 @@ func verifyRefs(what, forwarderName string, status loggingv1.ClusterLogForwarder
 			} else {
 				bad.Insert(ref)
 			}
+		case "filters":
+			if allowed.Has(ref) {
+				good.Insert(ref)
+			} else {
+				bad.Insert(ref)
+			}
 		}
```
syedriko, line 97: default:?

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
```diff
@@ -1173,6 +1173,36 @@ var _ = Describe("Validate clusterlogforwarderspec", func() {
 			Expect(clfStatus.Pipelines["test"]).To(HaveCondition("Ready", true, "", ""))
 		})
 
+		It("should fail with undefined filter in pipeline", func() {
+			forwarderSpec := &loggingv1.ClusterLogForwarderSpec{
+				Outputs: []loggingv1.OutputSpec{
```
syedriko, line 1178: const (
    pipelineName = "test"
    outputName = "my-output"
)

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
Clee2691: I don't believe we need a default case here. Originally it was an if/else if statement with no final else clause

PR #2219, opened 2023-10-19
LOG-4681: kubeAPIAudit policies don't work on audit logs from auditWebhook

alanconway:
The kubeapi-server webhook serializes audit Events without "apiVersion" or "kind" fields.
The audit filter was using these fields to identify API-audit log records.
Instead check for presence and type of "auditID" and "verb" which are always present.

Other changes:
- Don't overwrite "level" field for API audit events, users will expect this field.
- test/functionl/vector.go make vector the default collector default if no build tags are set.

### DIFF
--- a/test/functional/normalization/audit_logs_format_test.go
+++ b/test/functional/normalization/audit_logs_format_test.go
@@ -164,7 +164,7 @@ var _ = Describe("[Functional][LogForwarding][Normalization] message format test
 						Kind:             "Event",
 						Hostname:         framework.Pod.Spec.NodeName,
 						LogType:          "audit",
-						Level:            "default",
+						Level:            "Metadata",
 						Timestamp:        time.Time{},
 						PipelineMetadata: functional.TemplateForAnyPipelineMetadata,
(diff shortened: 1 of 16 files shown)


test/functional/normalization/audit_logs_format_test.go
```diff
@@ -164,7 +164,7 @@ var _ = Describe("[Functional][LogForwarding][Normalization] message format test
 						Kind:             "Event",
 						Hostname:         framework.Pod.Spec.NodeName,
 						LogType:          "audit",
-						Level:            "default",
+						Level:            "Metadata",
```
jcantrill, line 167: Why do we want this instead of some equivalent syslog severity?

jcantrill (commented):
/hold

test/functional/normalization/audit_logs_format_test.go
alanconway: kube-apiserver events use the field "level" to mean something different - how much of the event has been deleted (request body, response body etc.) there's no syslog-style severity, and no way to derive it.

PR #2220, opened 2023-10-19
LOG-4536: add request and buffer settings to address memory consumption

jcantrill:
### Description
* Adds buffer and request settings to all outputs for vector in order to achieve some parity to prior releases
* Modifies http api timeout to be an int instead of string of an int

### Links
* https://issues.redhat.com/browse/LOG-4536
* backport of #2203

### DIFF
--- a/internal/generator/helpers/optional_pair.go
+++ b/internal/generator/helpers/optional_pair.go
@@ -1,0 +1,29 @@
+package helpers
+
+import (
+	"fmt"
+	"reflect"
+)
+
+type OptionalPair struct {
+	key   string
+	Value interface{}
+}
+
+func NewOptionalPair(key string, value interface{}) OptionalPair {
+	return OptionalPair{
+		key,
+		value,
+	}
+}
+
+func (op OptionalPair) String() string {
+	if op.Value == nil {
+		return ""
+	}
+	format := "%s = %v"
+	if reflect.TypeOf(op.Value).Kind() == reflect.String {
+		format = "%s = %q"
+	}
+	return fmt.Sprintf(format, op.key, op.Value)
+}
--- a/internal/generator/vector/output/buffer.go
+++ b/internal/generator/vector/output/buffer.go
@@ -1,0 +1,24 @@
+package output
+
+type Buffer struct {
+	ComponentID string
+	WhenFull    string
+}
+
+func NewBuffer(id string) Buffer {
+	return Buffer{
+		ComponentID: id,
+		WhenFull:    "drop_newest",
+	}
+}
+
+func (b Buffer) Name() string {
+	return "buffer"
+}
+
+func (b Buffer) Template() string {
+	return `{{define "` + b.Name() + `" -}}
+[sinks.{{.ComponentID}}.buffer]
+when_full = "{{.WhenFull}}"
+{{end}}`
+}
--- a/internal/generator/vector/output/syslog/syslog.go
+++ b/internal/generator/vector/output/syslog/syslog.go
@@ -85,9 +85,10 @@ severity = "{{.Severity}}"
 }
 
 func Conf(o logging.OutputSpec, inputs []string, secret *corev1.Secret, op Options) []Element {
+	id := vectorhelpers.FormatComponentID(o.Name)
 	if genhelper.IsDebugOutput(op) {
 		return []Element{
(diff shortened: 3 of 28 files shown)


internal/generator/helpers/optional_pair.go
```diff
@@ -0,0 +1,29 @@
+package helpers
+
+import (
+	"fmt"
+	"reflect"
+)
+
+type OptionalPair struct {
+	key   string
+	Value interface{}
```
syedriko, line 10: lowercase v?

jcantrill: I remembered it is public because it is being set in various places after being "new'd".  We do not have need to expose the key but maybe the semantics should be to create SetValue but that seems unnecessary

internal/generator/vector/output/buffer.go
```diff
@@ -0,0 +1,24 @@
+package output
+
```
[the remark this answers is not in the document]
jcantrill: We should add this change when we actually allow buffer tuning

internal/generator/vector/output/syslog/syslog.go
```diff
@@ -85,9 +85,10 @@ severity = "{{.Severity}}"
 }
 
 func Conf(o logging.OutputSpec, inputs []string, secret *corev1.Secret, op Options) []Element {
+	id := vectorhelpers.FormatComponentID(o.Name)
 	if genhelper.IsDebugOutput(op) {
 		return []Element{
-			Debug(vectorhelpers.FormatComponentID(o.Name), vectorhelpers.MakeInputs(inputs...)),
+			Debug(id, vectorhelpers.MakeInputs(inputs...)),
```
syedriko, line 91: syslog is a socket sink, it supports buffer config https://vector.dev/docs/reference/configuration/sinks/socket/#buffer

syedriko:
@jcantrill I'm not sure `timeout_secs = 2147483648` is a good idea. A misbehaving server can block and not reply for a long time, essentially stopping our sink from draining events. `retry_attempts` will become irrelevant as we're going to be stuck at attempt 1. We won't entirely be back to square one as the sink buffer will drop_newest, but it's preferable to 'drop oldest', which we do by exhausting retry attempts.

jcantrill:
> @jcantrill I'm not sure `timeout_secs = 2147483648` is a good idea. A misbehaving server can block and not reply for a long time, essentially stopping our sink from draining events. `retry_attempts` will become irrelevant as we're going to be stuck at attempt 1. We won't entirely be back to square one as the sink buffer will drop_newest, but it's preferable to 'drop oldest', which we do by exhausting retry attempts.

These are all excellent comments that should have been addressed  in the original 5.8 PR that was released last week.  Can we merge this and write a followup JIRA to correct it in both streams is it makes sense.  The changes here are strickly to mimic the current fluent behavior as close as possible.  Honestly, I don't even know where timeout_secs comes from that you mention; I don't recall changing that value

syedriko:
> > @jcantrill I'm not sure `timeout_secs = 2147483648` is a good idea. A misbehaving server can block and not reply for a long time, essentially stopping our sink from draining events. `retry_attempts` will become irrelevant as we're going to be stuck at attempt 1. We won't entirely be back to square one as the sink buffer will drop_newest, but it's preferable to 'drop oldest', which we do by exhausting retry attempts.
> 
> These are all excellent comments that should have been addressed in the original 5.8 PR that was released last week. Can we merge this and write a followup JIRA to correct it in both streams is it makes sense. The changes here are strickly to mimic the current fluent behavior as close as possible. Honestly, I don't even know where timeout_secs comes from that you mention; I don't recall changing that value

2147483648 for the elastic timeout comes from
https://github.com/openshift/cluster-logging-operator/blob/38409de359bdda632734e26fc3d90664c5b4850f/internal/generator/vector/output/elasticsearch/elasticsearch.go#L230
, which in turn came from the fluentd side: https://github.com/openshift/cluster-logging-operator/blob/38409de359bdda632734e26fc3d90664c5b4850f/internal/generator/fluentd/output/elasticsearch/elasticsearch.go#L61.

In the fluentd elastic plugin, this timeout defaults to 5 seconds: https://github.com/uken/fluent-plugin-elasticsearch#request_timeout

jcantrill:
> For vector, 2147483648 seconds for the elastic timeout comes from

This value existed previously.  I have no idea why it is here but if we are changing then we should address it in a separate PR IMO

PR #2225, opened 2023-10-27
LOG-4757: update version to 5.9

jcantrill:
### Description
https://issues.redhat.com/browse/LOG-4713


Makefile
```diff
@@ -43,13 +43,13 @@ export CURRENT_BRANCH=$(shell git rev-parse --abbrev-ref HEAD;)
 export IMAGE_TAG?=127.0.0.1:5000/openshift/origin-$(OPERATOR_NAME):$(CURRENT_BRANCH)
 BUNDLE_TAG=$(error set OVERLAY to deploy or run a bundle)
 
-export LOGGING_VERSION?=5.8
+export LOGGING_VERSION?=5.9
```
xperimental, line 46: `OPENSHIFT_VERSIONS` in this file needs a bump to `v4.13-v4.17`. Looks fine otherwise :+1:

xperimental:
Hm. I think the e2e test might be right for a change ... There's no 5.9 version of EO (and as far as we currently know: there will not be one). I think we might need a change to the script that installs EO for the e2e test and make it use 5.8 even when on the 5.9 `master` branch.

jcantrill:
> Hm. I think the e2e test might be right for a change ... There's no 5.9 version of EO (and as far as we currently know: there will not be one). I think we might need a change to the script that installs EO for the e2e test and make it use 5.8 even when on the 5.9 `master` branch.

change made.  I came to the same conclusion

jcantrill:
Likely blocked by https://github.com/openshift/cluster-logging-operator/pull/2225

PR #2226, opened 2023-10-31
LOG-4741: cluster-logging-operator reconciling dozens of times

Clee2691:
### Description
This PR attempts to fix the `cluster logging operator` from reconciling dozens of times after a `ClusterLogForwarder` (CLF) was defined. After applying a CLF YAML, the resulting created daemonset's status was updating many time resulting in dozens of reconcile requests to the CLF. The proposed change is to ignore the daemonset's status changes and only reconcile on spec changes.

/cc @jlarriba @syedriko @cahartma 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4741

### DIFF
--- a/controllers/clusterlogging/clusterlogging_controller.go
+++ b/controllers/clusterlogging/clusterlogging_controller.go
@@ -5,7 +5,9 @@ import (
 
 	"github.com/openshift/cluster-logging-operator/internal/logstore/lokistack"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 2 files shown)


jcantrill:
@jlarriba please review and verify if this addresses your original findings.  @Clee2691 if this is present in 5.8 please cherrypick it pack

jlarriba (approved):

PR #2231, opened 2023-11-01
LOG-4780: use AWS token from secret in lieu of SA token

jcantrill:
### Description
This PR:
* uses the AWS token from the secret if specified

### Links
* https://issues.redhat.com/browse/LOG-4780

cc @cahartma

### DIFF
--- a/internal/generator/fluentd/output/security/security.go
+++ b/internal/generator/fluentd/output/security/security.go
@@ -64,6 +64,10 @@ func HasBearerTokenFileKey(secret *corev1.Secret) bool {
 	return HasKeys(secret, constants.BearerTokenFileKey)
 }
 
+func HasAWSWebIdentityTokenFilePath(secret *corev1.Secret) bool {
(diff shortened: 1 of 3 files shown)


internal/generator/fluentd/output/security/security.go
```diff
@@ -64,6 +64,10 @@ func HasBearerTokenFileKey(secret *corev1.Secret) bool {
 	return HasKeys(secret, constants.BearerTokenFileKey)
 }
 
+func HasAWSWebIdentityTokenFilePath(secret *corev1.Secret) bool {
+	return HasKeys(secret, constants.AWSWebIdentityTokenFilePath)
```
cahartma, line 68: This key won't be in the secret.   
Need to add and use a new constant AWSWebIdentityTokenKey = "web_identity_token_file"

cahartma: If they are adding it, then this works.

PR #2234, opened 2023-11-07
LOG-4786: Label 'openshift.sequence' is missing when forwarding logs over syslog with vector

syedriko:
### Description

For vector forwarding to a syslog destination, generate the dedot transform that adds openshift.sequence to the log message.

/cc @cahartma 
/assign @jcantrill 

/cherry-pick release-5.8

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4786

### DIFF
--- a/internal/generator/vector/output/syslog/syslog.go
+++ b/internal/generator/vector/output/syslog/syslog.go
@@ -11,6 +11,7 @@ import (
 	genhelper "github.com/openshift/cluster-logging-operator/internal/generator/helpers"
 	. "github.com/openshift/cluster-logging-operator/internal/generator/vector/elements"
 	vectorhelpers "github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/normalize"
 	"github.com/openshift/cluster-logging-operator/internal/generator/vector/output/security"
 	corev1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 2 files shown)


jcantrill:
@syedriko this will do too much since it dedots the labels which is not reported as an issue.  That would be a regression which I'm concerned would introduce another bug.  What does fluentd do?  Does it dedot labels? cc @alanconway

syedriko:
> @syedriko this will do too much since it dedots the labels which is not reported as an issue. That would be a regression which I'm concerned would introduce another bug. What does fluentd do? Does it dedot labels? cc @alanconway 

The transform that I added is exactly the same across the board, every output has it:
https://github.com/openshift/cluster-logging-operator/blob/d82c0ea89e177402d40312f4f009032f4088249b/internal/generator/vector/output/cloudwatch/cloudwatch.go#L86
https://github.com/openshift/cluster-logging-operator/blob/d82c0ea89e177402d40312f4f009032f4088249b/internal/generator/vector/output/gcl/gcl.go#L87
https://github.com/openshift/cluster-logging-operator/blob/d82c0ea89e177402d40312f4f009032f4088249b/internal/generator/vector/output/http/http.go#L126
https://github.com/openshift/cluster-logging-operator/blob/d82c0ea89e177402d40312f4f009032f4088249b/internal/generator/vector/output/kafka/kafka.go#L62
https://github.com/openshift/cluster-logging-operator/blob/d82c0ea89e177402d40312f4f009032f4088249b/internal/generator/vector/output/loki/loki.go#L122
https://github.com/openshift/cluster-logging-operator/blob/d82c0ea89e177402d40312f4f009032f4088249b/internal/generator/vector/output/splunk/splunk.go#L74

jcantrill:
Answering my original question of what does fluentd do... the answer is yes, it also dedots for syslog

/approve
/lgtm

PR #2235, opened 2023-11-07
LOG-3527: Collector to act as rsyslog server

jlarriba:
The included modifications generate an rsyslog server configuration for Collector. This configuration does not currently apply any transformations and send the logs succesfully to Loki, which stores them. 

It also applies the OutputTypeSpec to receivers so they can share common parameters between the receivers, defining a ReceiverTypeSpec.

/cc syedriko, aconway
/assign jcantrill

Links
JIRA: https://issues.redhat.com/browse/LOG-3527


apis/logging/v1/input_receiver_types.go
```diff
@@ -6,7 +6,17 @@ package v1
 //
 // The fields of this struct define the set of known Receiver types.
 type ReceiverSpec struct {
-	HTTP *HTTPReceiver `json:"http,omitempty"`
+	// Port the Receiver listen on.
```
syedriko, line 9: listens?

apis/logging/v1/input_receiver_types.go
```diff
@@ -6,7 +6,17 @@ package v1
...
+	// +required
+	Port int32 `json:"port"`
```
syedriko, line 11: @alanconway Can we have a receiver without a listen port? A messaging subscriber that's given a URL, like https://vector.dev/docs/reference/configuration/sources/amqp ? Or an fd https://vector.dev/docs/reference/configuration/sources/file_descriptor?

apis/logging/v1/input_receiver_types.go
```diff
@@ -15,14 +25,17 @@ const (
...
 	//
 	// +kubebuilder:validation:Enum:=kubeAPIAudit
 	// +required
 	Format string `json:"format"`
 }
+
+// SyslogReceiver receives logs from rsyslog
+type SyslogReceiver struct {
+	// The protocol of the connection the Receiver will listen on: tcp or upd
+	// +kubebuilder:default:=tcp
```
syedriko, line 38: // +kubebuilder:validation:Enum=tcp;udp

internal/generator/vector/normalize.go
```diff
@@ -124,6 +124,11 @@ func NormalizeLogs(spec *logging.ClusterLogForwarderSpec, op generator.Options)
 		if input.Receiver != nil && input.Receiver.HTTP != nil && input.Receiver.HTTP.Format == logging.FormatKubeAPIAudit {
 			el = append(el, NormalizeK8sAuditLogs(input.Name+`_items`, input.Name+`_normalized`)...)
 		}
+		// Any normalization currently in use by the system fails due to all of them (JournalLogs included depend on OpenShift labels we do not have)
```
syedriko, line 127: please remove if no longer needed

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
```diff
@@ -184,8 +184,14 @@ func verifyInputs(spec *loggingv1.ClusterLogForwarderSpec, status *loggingv1.Clu
 		validPort := func(port int32) bool {
 			return port == 0 || (port > 1023 && port < 65536)
 		}
+		receiverNonNil := func(input loggingv1.InputSpec) bool {
```
syedriko, line 187: isAReceiver?

test/e2e/logforwarding/fluent/fluent_test.go
```diff
@@ -92,7 +92,7 @@ var _ = Describe("[ClusterLogForwarder]", func() {
 			for _, name := range logTypes {
 				name := name // Don't bind to range variable
 				r := f.Receiver.Sources[name].TailReader()
-				Expect(r.ReadLine()).To(SatisfyAny(Equal(""), ContainSubstring(fmt.Sprintf(`"log-type":%q`, name))))
+				Expect(r.ReadLine()).To(SatisfyAny(Equal(""), ContainSubstring(fmt.Sprintf(`"log_type":%q`, name))))
```
syedriko, line 95: receivers are vector-only, we shouldn't need to change anything fluent-specific...

internal/collector/collector.go
```diff
@@ -229,8 +229,11 @@ func (f *Factory) ReconcileInputServices(er record.EventRecorder, k8sClient clie
 	}
 
 	for _, input := range f.ForwarderSpec.Inputs {
-		if input.Receiver != nil && input.Receiver.HTTP != nil {
-			listenPort := input.Receiver.HTTP.GetPort()
+		if input.Receiver != nil {
+			var listenPort int32
+			if input.Receiver != nil {
```
syedriko, line 234: that's a dup of the if above

syedriko:
@jlarriba please squash and rebase

test/e2e/logforwarding/fluent/fluent_test.go
jlarriba: It was failing on one of the e2e tests, but it was maybe something temporary, reversed the change and see what happens

jlarriba:
@syedriko squashed and rebased

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
```diff
@@ -184,8 +184,14 @@ func verifyInputs(spec *loggingv1.ClusterLogForwarderSpec, status *loggingv1.Clu
 		validPort := func(port int32) bool {
 			return port == 0 || (port > 1023 && port < 65536)
```
syedriko, line 185: Now with Port being required in the ReceiverSpec in the API and without HTTPReceiver.GetPort(), we need to disallow port 0 as a valid port. In the prior version, it was here as a placeholder to be substituted for the default 8443, but it doesn't make sense anymore. Deferring to @alanconway as an API concern.

apis/logging/v1/input_receiver_types.go
```diff
@@ -6,7 +6,17 @@ package v1
 //
 // The fields of this struct define the set of known Receiver types.
 type ReceiverSpec struct {
-	HTTP *HTTPReceiver `json:"http,omitempty"`
+	// Port the Receiver listens on.
+	// +required
```
jcantrill, line 10: This will be an API regression.  We previously identified it as optional with a default value and thus the reason "GetPort" exists.  This would break any existing deployments on upgrade

apis/logging/v1/input_receiver_types.go
```diff
@@ -6,7 +6,17 @@ package v1
 //
 // The fields of this struct define the set of known Receiver types.
 type ReceiverSpec struct {
```
jcantrill, line 8: This should have a "type" enum to match the pattern elsewhere to identify which type of input it is

internal/generator/vector/input/inputs.go
```diff
@@ -160,18 +160,30 @@ func Inputs(spec *logging.ClusterLogForwarderSpec, o Options) []Element {
...
-					ComponentID: input.Name + `_input`,
-					Inputs:      helpers.MakeInputs(input.Name + `_normalized`),
-					VRL: strings.Join(helpers.TrimSpaces([]string{
-						AddLogTypeAudit,
-						normalize.FixHostname,
-						normalize.FixTimestampField,
-					}), "\n"),
-				})
+		if input.Receiver != nil {
+			if input.Receiver.HTTP != nil && input.Receiver.HTTP.Format == logging.FormatKubeAPIAudit {
```
jcantrill, line 164: type field will allow use of switch

internal/generator/vector/output/loki/loki.go
```diff
@@ -114,6 +114,21 @@ func Conf(o logging.OutputSpec, inputs []string, secret *corev1.Secret, op Optio
 		}
 	}
 	componentID := fmt.Sprintf("%s_%s", id, "remap")
+	if strings.Contains(id, strings.ReplaceAll(fmt.Sprintf("%v", op["syslog"]), "-", "_")) {
```
jcantrill, line 117: type may help here as well, I don't think we should make decisions based upon this type of string evaluation

internal/logstore/lokistack/logstore_lokistack.go
```diff
@@ -242,12 +242,16 @@ func getInputTypeFromName(spec loggingv1.ClusterLogForwarderSpec, inputName stri
...
+		return input.Receiver != nil && input.Receiver.Syslog != nil
+	}
+
 	for _, input := range spec.Inputs {
 		if input.Name == inputName {
 			if input.Application != nil {
 				return loggingv1.InputNameApplication
 			}
-			if input.Infrastructure != nil {
+			if input.Infrastructure != nil || isSyslogReceiver(input) {
```
jcantrill, line 254: This smells wrong to me and outside the design of LogForwarding as we know it.  @alanconway there has to be a better way to identify the tenant.  Maybe This is where we need to allow tenant configuration..... and i see we are doing that below as well for http which smells wrong

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
jcantrill: This is a regression. I don't understand why we have to require the port.  Why can we not provide a validation that spins through the listeners and ensures they are unique from all other ports including the ones we reserve for metrics

Makefile
```diff
@@ -145,7 +145,6 @@ RUN_CMD?=go run
 
 .PHONY: run
 run:
-	@ls ./bundle/manifests/logging.openshift.io_*.yaml | xargs -n1 oc apply -f
```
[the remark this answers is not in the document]
jlarriba: For running locally, this is creating a lot of confusion, as "make generate" does not modify the resources in the bundle directory. If are developing locally (with make run), the modified crd resources are in config/crd/bases, not in the bundle. I removed this because every time I do make run, my CR changes would get removed. The solution would be to use "oc apply -f config/crd/bases" to replace this line. WDYT?

apis/logging/v1/input_receiver_types.go
jlarriba: That's true and my first attempt respected older deployments, but @alanconway requested the API to be changed to use ReceiverTypeSpec the same way OutputTypeSpec was working. 

Existing deployments that use HTTPReceiver will fail whether we make the receiver port as required or not because currently the port is a child of HTTP and now it is a child of Receiver. Deployments that do not make use of the HTTPReceiver will not need any changes.

internal/logstore/lokistack/logstore_lokistack.go
jlarriba: Indeed, I am here repeating the pattern established for HTTP Receiver, as @alanconway told me.

apis/logging/v1/input_receiver_types.go
alanconway: @jlarriba sorry I wasn't clear - I do want a top level receiver spec to contain future different types of receiver, but I think the port should still be inside the HTTPReceiver. Unlike "Output" I don't think there's any "universal" config we can put a the top level for receivers.  As @syedriko points out below, it's possible we may have client-based or file scraping receivers in future, so I don't want to bake in the assumption that all receivers listen on a port. The existing release is explicitly preview so its OK if we break compatibility to enable future expansion.

apis/logging/v1/input_receiver_types.go
alanconway: Yep, see above

apis/logging/v1/input_receiver_types.go
```diff
@@ -15,14 +25,18 @@ const (
...
 	// Format is the format of incoming log data.
 	//
 	// +kubebuilder:validation:Enum:=kubeAPIAudit
 	// +required
 	Format string `json:"format"`
 }
+
+// SyslogReceiver receives logs from rsyslog
+type SyslogReceiver struct {
+	// The protocol of the connection the Receiver will listen on: tcp or upd
```
alanconway, line 37: syslog receiver needs a port - it does have this in common with HTTP, but  we can't assume a port is a  common requirement for all future receivers.

internal/collector/collector.go
```diff
@@ -48,7 +48,7 @@ const (
 	logKubeapiserver                = "varlogkubeapiserver"
 	logKubeapiserverValue           = "/var/log/kube-apiserver"
 	metricsVolumePath               = "/etc/collector/metrics"
-	httpInputVolumePath             = "/etc/collector/"
+	receiverInputVolumePath         = "/etc/collector/"
```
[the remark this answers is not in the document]
alanconway: should this be '/etc/collector/receivers' or should we use a prefix/suffix to ensure uniqueness here?

apis/logging/v1/input_receiver_types.go
jlarriba: Working on it

internal/logstore/lokistack/logstore_lokistack.go
alanconway: Agreed, this does suck. We should configure log_type consistently, possibly this is one and only top-level configuration setting for ReceiverSpec? We have no other way to assign a log_type for receiver logs. We need to discuss this, I see two options:
1. Let user assign log_type, allow them to "impersonate" built-in types and/or allow them to invent new types and filter them for themselves.
2. Invent a fouth fixed type like "external" and stamp that on received logs. Let the user use other features (like attaching openshift.labels) to identify where the logs came from.
I'm leaning towarsd 1., since I think we want to de-emphasize log_type and focus on filtering/routing by other log properties, but I can also see arguments for 2.

alanconway (changes requested):
Very solid PR. API change requested is to move the location of "port". Sgreed with some minor changes suggested b y @jcantrill.  The open question is proper setting of log_type/tenant which I think we need to  hash out some more.

Makefile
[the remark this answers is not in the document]
jcantrill: Do CRD bases include everything needed for the CRD?  What about permissions that also exist in the bundle?  AFAIK the CRDs are not the only requirement to run the operator which means even running locally you may not be able to access other parts due to permission problems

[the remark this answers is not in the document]
jlarriba: Yeah, the CRD bases only include the crds itselves, thats right. But the manifests in the bundle do not change when make run/make generate happens, so this line always apply older resources into the cluster every time we do a make run. IMHO, this renders it useless for development.

internal/collector/collector.go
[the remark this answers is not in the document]
jlarriba: I was working with /etc/collector/receivers, but I am open to any other suggestions.

apis/logging/v1/input_receiver_types.go
jlarriba: OK, understood. The port will be part of the specific Receiver and there will only be one common field for all Receivers: "Type".

internal/generator/vector/output/loki/loki.go
jlarriba: This is much bigger than it looks like, let's discuss it in a different forum.

jlarriba: I forgot to highlight that the check we are doing is against an op property that we set up specifically for this usage in https://github.com/openshift/cluster-logging-operator/pull/2235/files#diff-0b3745488d65f7b49cae4e1d6f675a2250721a4b9fb055584bee8c6250ae4660R50

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
jlarriba: My new update leaves the Port not required.

Makefile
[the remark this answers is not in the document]
jlarriba: I have realized that we could just run "make bundle" which is as fast as make generate and will generate the correct resources in the bundle dir.

jlarriba:
With the latest changes, openstack syslog logs go through the journal transformations and succesfully go through the dedot function until they are correctly stored in Loki and shown in the console.

internal/logstore/lokistack/logstore_lokistack.go
jlarriba: As per our latest meeting, we are leaving the "isSyslogReceiver" here to be able to provide logging storage for Openstack until we deliver a better solution for version 6 of CLO.

jlarriba:
I am retesting because none of the issues discovered in the functional/e2e tests are related to my code.

/retest

jlarriba:
I think I have found the issue, working on it.

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
```diff
@@ -99,27 +98,101 @@ var _ = Describe("Validate clusterlogforwarderspec", func() {
...
 			}
-			checkPorts(8080, `no_such_format`, `invalid format specified for HTTP receiver`)
-			checkReceiver(&loggingv1.ReceiverSpec{}, `ReceiverSpec must define an HTTP receiver`, map[string]bool{constants.VectorName: true})
+			checkPortAndSyslogProtocol(10514, "http", `invalid protocol specified for Syslog receiver`)
+			checkReceiverType("wrong-receiver", `invalid Type specified for receiver`)
+			checkReceiver(&loggingv1.ReceiverSpec{}, `invalid ReceiverTypeSpec specified for receiver`, map[string]bool{constants.VectorName: true})
 			checkReceiver(&loggingv1.ReceiverSpec{}, `ReceiverSpecs are only supported for the vector log collector`, map[string]bool{})
 		})
 
+		It("should fail validation if both HTTP and Syslog receivers are defined", func() {
```
jcantrill, line 157: Does this matter if we are providing a "type" which is ... "use this one" ?

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
```diff
@@ -204,12 +204,22 @@ func verifyInputs(spec *loggingv1.ClusterLogForwarderSpec, status *loggingv1.Clu
 			badInput("inputspec cannot have a negative limit threshold")
 		case input.Receiver != nil && !extras[constants.VectorName]:
 			badInput("ReceiverSpecs are only supported for the vector log collector")
-		case input.Receiver != nil && input.Receiver.HTTP == nil:
-			badInput("ReceiverSpec must define an HTTP receiver")
-		case isHTTPReceiver(input) && input.Receiver.HTTP.Format != loggingv1.FormatKubeAPIAudit:
-			badInput("invalid format specified for HTTP receiver")
-		case isHTTPReceiver(input) && !validPort(input.Receiver.HTTP.Port):
+		case input.Receiver != nil && input.Receiver.ReceiverTypeSpec == nil:
```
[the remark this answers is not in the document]
jlarriba: Instead of moving them into a separate file, I moved them to a separate func in the same file: validateReceivers, which will get validated just after Inputs.

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
jlarriba: No it does not, removed.

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account_test.go
```diff
@@ -183,6 +186,38 @@ var _ = Describe("[internal][validations] validate clusterlogforwarder permissio
 			}
 			Expect(ValidateServiceAccount(customClf, k8sAuditClient, extras)).To(Succeed())
 		})
+
+		It("should pass validation if service account can collect external logs and there is a Syslog receiver", func() {
```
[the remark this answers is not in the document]
jlarriba: I dont really understand what you are trying to tell me here. In master, this very same validation but for HTTP receiver is still being done. What do you expect me to change here?

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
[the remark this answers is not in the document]
jlarriba: I have been trying nad this implies a lot of changes. Could we consider this validation re-org as candidate for its own PR?

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account_test.go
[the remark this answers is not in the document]
jcantrill: The text for the expectation is wrong (including the one for http).  We removed the actual RBAC check but not the check for existence of the SA (which really is not needed).  There is no change required.

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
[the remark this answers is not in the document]
jcantrill: This entire validation should be refactored.  I was thinking the 'receiver' validations could be moved out, but if the work is too invasive, please ignore this comment.

apis/logging/v1/cluster_log_forwarder.go
```diff
@@ -196,10 +195,19 @@ func (output *OutputSpec) GetMaxRecordsPerSecond() int64 {
...
+}
+
+func IsHttpReceiver(input *InputSpec) (ret bool) {
+	return input.Receiver != nil &&
+		input.Receiver.Type == ReceiverTypeHttp
+}
+
+func IsSyslogReceiver(input *InputSpec) (ret bool) {
+	return input.Receiver != nil &&
+		input.Receiver.Type == ReceiverTypeSyslog
```
syedriko, line 212: Don't need named return value in any of these funcs

apis/logging/v1/input_receiver_types.go
```diff
@@ -2,20 +2,37 @@ package v1
 
 // NOTE: The Enum validation on ReceiverSpec.Type must be updated if the list of types changes.
 
+// Receiver type constants, must match JSON tags of OutputTypeSpec fields.
+const (
+	ReceiverTypeHttp   = "http"
+	ReceiverTypeSyslog = "syslog"
+
```
syedriko, line 9: I think there should be validation code that rejects mismatched receiver types and receiver type specs, i.e. Type == ReceiverTypeHttp && Syslog != nil.

apis/logging/v1/input_receiver_types.go
```diff
@@ -2,20 +2,37 @@ package v1
...
+type ReceiverTypeSpec struct {
+	HTTP   *HTTPReceiver   `json:"http,omitempty"`
+	Syslog *SyslogReceiver `json:"syslog,omitempty"`
+}
 
 // HTTPReceiver receives encoded logs as a HTTP endpoint.
 type HTTPReceiver struct {
-	// Port the Service and the HTTP listener listen on.
+	// Port the Receiver listens on.
 	// +kubebuilder:default:=8443
```
syedriko, line 36: We should mention here and for syslog that we only consider ports [1024, 65535] to be valid
https://book.kubebuilder.io/reference/markers/crd-validation says there's `+kubebuilder:validation:Minimum and Maximum`

PR #2240, opened 2023-11-07
LOG-4779: Deploy collector as a deployment instead of daemonset

Clee2691:
## Description
This PR deals with the scenario in which a user wishes to specify the collector as an `HTTP` receiver as the sole input option.

Originally the collector was deployed as a `daemonset`. However, with this change, when all pipelines specified in the `ClusterLogForwarder` (CLF) reference an 'HTTP' receiver as their sole input reference, the collector will be deployed as a `deployment`.

This addresses the following:
- The HCP usecase that utilizes one collector per NS for collecting audit logs via a webhook
- Saves deployment resources
- Collector does not need to mount the host file system given it only receives logs via the receiver

### Extra
This also fixes a bug when defining `HTTP` input receivers in multi-CLF. Multiple CLFs with the same `HTTP` input receiver name will endlessly reconcile the same service because the names are the same. The fix for this was the make HTTP input service names unique with the format of `<CLF-NAME>-<input.Name>`.

/cc @syedriko @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4779


internal/collector/collector.go
```diff
@@ -127,36 +130,46 @@ func (f *Factory) NewDaemonSet(namespace, name string, trustedCABundle *v1.Confi
...
+func (f *Factory) NewDeployment(namespace, name string, trustedCABundle *v1.ConfigMap, tlsProfileSpec configv1.TLSProfileSpec, httpInputs []string) *apps.Deployment {
+	podSpec := f.NewPodSpec(trustedCABundle, f.ForwarderSpec, f.ClusterID, f.TrustedCAHash, tlsProfileSpec, httpInputs, namespace)
+	dpl := factory.NewDeployment(name, namespace, f.ResourceNames.CommonName, constants.CollectorName, string(f.CollectorSpec.Type), *podSpec, f.CommonLabelInitializer, f.PodLabelVisitor)
+	return dpl
+}
+
 func (f *Factory) NewPodSpec(trustedCABundle *v1.ConfigMap, forwarderSpec logging.ClusterLogForwarderSpec, clusterID, trustedCAHash string, tlsProfileSpec configv1.TLSProfileSpec, httpInputs []string, namespace string) *v1.PodSpec {
 
 	podSpec := &v1.PodSpec{
-		NodeSelector:                  utils.EnsureLinuxNodeSelector(f.NodeSelector()),
```
jcantrill, line 133: NodeSelector is still relevent in a deployment scenario too

internal/collector/collector.go
```diff
@@ -127,36 +130,46 @@ func (f *Factory) NewDaemonSet(namespace, name string, trustedCABundle *v1.Confi
...
 		PriorityClassName:             clusterLoggingPriorityClassName,
 		ServiceAccountName:            f.ResourceNames.ServiceAccount,
 		TerminationGracePeriodSeconds: utils.GetPtr[int64](10),
-		Tolerations:                   constants.DefaultTolerations(),
```
jcantrill, line 137: same with tolerations

internal/collector/collector.go
```diff
@@ -127,36 +130,46 @@ func (f *Factory) NewDaemonSet(namespace, name string, trustedCABundle *v1.Confi
...
-			{Name: logOauthserver, VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: logOauthserverValue}}},
-			{Name: logOpenshiftapiserver, VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: logOpenshiftapiserverValue}}},
-			{Name: logKubeapiserver, VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: logKubeapiserverValue}}},
 			{Name: f.ResourceNames.SecretMetrics, VolumeSource: v1.VolumeSource{Secret: &v1.SecretVolumeSource{SecretName: f.ResourceNames.SecretMetrics}}},
 			{Name: tmpVolumeName, VolumeSource: v1.VolumeSource{EmptyDir: &v1.EmptyDirVolumeSource{Medium: v1.StorageMediumMemory}}},
 		},
 	}
+
+	if f.isDaemonset {
+		podSpec.NodeSelector = utils.EnsureLinuxNodeSelector(f.NodeSelector())
```
jcantrill, line 152: which will change here

internal/collector/collector.go
```diff
@@ -127,36 +130,46 @@ func (f *Factory) NewDaemonSet(namespace, name string, trustedCABundle *v1.Confi
...
+		)
+	}
+
 	for _, httpInput := range httpInputs {
 		podSpec.Volumes = append(podSpec.Volumes,
 			v1.Volume{Name: httpInput, VolumeSource: v1.VolumeSource{Secret: &v1.SecretVolumeSource{SecretName: httpInput}}},
 		)
 	}
 
-	podSpec.Tolerations = append(podSpec.Tolerations, f.Tolerations()...)
```
jcantrill, line 158: and here

internal/collector/collector.go
```diff
@@ -223,15 +242,18 @@ func (f *Factory) NewCollectorContainer(secretNames []string, clusterID string,
...
 		return nil
 	}
 
 	for _, input := range f.ForwarderSpec.Inputs {
 		if input.Receiver != nil && input.Receiver.HTTP != nil {
 			listenPort := input.Receiver.HTTP.GetPort()
-			if err := network.ReconcileInputService(er, k8sClient, namespace, input.Name, selectorComponent, input.Name, listenPort, listenPort, owner, visitors); err != nil {
+			// Make input service name more unique
+			// Format: inputName-Namespace-CLFName
+			serviceName := fmt.Sprintf("%s-%s-%s", input.Name, namespace, name)
```
jcantrill, line 255: > Multiple CLFs with the same HTTP input receiver

I can see where "input.name" won't work in the same NS but we should be able to use "input.name" + clf.name for uniqueness.  Consider how you ref a service which has to be routable .  The full name would mean you would include the NS twice(e.g. input-namespace-clfname.namespace.svc.cluster.local); this seems clunky.  I would consider the pattern we use for other objects and suggest:

<clf.name>-<input.name> which should make it unique within the NS, unless I'm forgetting something

internal/generator/vector/sources.go
```diff
@@ -134,6 +134,8 @@ func HttpSources(spec *logging.ClusterLogForwarderSpec, op generator.Options) []
 		if input.Receiver != nil && input.Receiver.HTTP != nil {
 			el = append(el, HttpReceiver{
 				ID:            input.Name,
+				CLFName:       name,
```
jcantrill, line 137: Why can't we incroporate this into the ID instead of expanding the template

internal/k8shandler/collection.go
```diff
@@ -81,14 +81,14 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection() (err err
...
 
 	if err := network.ReconcileService(clusterRequest.EventRecorder, clusterRequest.Client, clusterRequest.Forwarder.Namespace, clusterRequest.ResourceNames.CommonName, constants.CollectorName, collector.MetricsPortName, clusterRequest.ResourceNames.SecretMetrics, collector.MetricsPort, clusterRequest.ResourceOwner, factory.CommonLabelInitializer); err != nil {
 		log.Error(err, "collector.ReconcileService")
 		return err
 	}
 
-	if err := factory.ReconcileInputServices(clusterRequest.EventRecorder, clusterRequest.Client, clusterRequest.Forwarder.Namespace, clusterRequest.ResourceNames.CommonName, utils.AsOwner(clusterRequest.Forwarder), factory.CommonLabelInitializer); err != nil {
+	if err := factory.ReconcileInputServices(clusterRequest.EventRecorder, clusterRequest.Client, clusterRequest.Forwarder.Namespace, clusterRequest.Forwarder.Name, clusterRequest.ResourceNames.CommonName, utils.AsOwner(clusterRequest.Forwarder), factory.CommonLabelInitializer); err != nil {
```
jcantrill, line 91: We introduced the Resourcenames to address a common place to build names based upon the CLF.name.  We should figure out how to make use of it given that was its purposes.  Shouldn't it be the "common name"

internal/k8shandler/collection.go
```diff
@@ -112,13 +112,21 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection() (err err
 	var httpInputs []string
 	for _, input := range clusterRequest.Forwarder.Spec.Inputs {
 		if input.Receiver != nil && input.Receiver.HTTP != nil {
-			httpInputs = append(httpInputs, input.Name)
+			serviceName := fmt.Sprintf("%s-%s-%s", input.Name, clusterRequest.Forwarder.Namespace, clusterRequest.Forwarder.Name)
```
jcantrill, line 115: we should bury the servicename or the base of the service name, or even a function on that struct that creates the servicename in ResourceNames

internal/k8shandler/collection.go
```diff
@@ -112,13 +112,21 @@ func (clusterRequest *ClusterLoggingRequest) CreateOrUpdateCollection() (err err
...
+			httpInputs = append(httpInputs, serviceName)
 		}
 	}
 
-	if err := factory.ReconcileDaemonset(clusterRequest.EventRecorder, clusterRequest.Client, clusterRequest.Forwarder.Namespace, clusterRequest.ResourceOwner, httpInputs); err != nil {
-		log.Error(err, "collector.ReconcileDaemonset")
-		return err
+	if clusterRequest.isDaemonset {
```
jcantrill, line 120: what if the someone edits the CLF to add more inputs?  do we lose ref to the previous deployment which may lead us to having multiple collectors running?

internal/k8shandler/collection_test.go
```diff
@@ -198,6 +226,8 @@ var _ = Describe("Reconciling", func() {
 					MountPath: constants.TrustedCABundleMountDir,
 				}
 			)
+
+			// Context("when collector is a daemonset", func() {
```
jcantrill, line 230: remove comment if not needed

internal/network/service.go
```diff
@@ -56,6 +56,10 @@ func ReconcileInputService(er record.EventRecorder, k8sClient client.Client, nam
 		visitors,
 	)
 
+	if !isDaemonset {
+		desired.Spec.Selector["collector-resource-type"] = "deployment"
```
jcantrill, line 60: maybe... 'collector-deployment-kind' ... and didn't I see a constant for the value

internal/utils/comparators/deployment/comparator.go
```diff
@@ -0,0 +1,80 @@
...
+import (
+	"reflect"
+
+	log "github.com/ViaQ/logerr/v2/log/static"
+	"github.com/openshift/cluster-logging-operator/internal/utils"
+	apps "k8s.io/api/apps/v1"
+)
+
+// AreSame compares deployments for equality and return true equal otherwise false
+func AreSame(current *apps.Deployment, desired *apps.Deployment) (bool, string) {
```
jcantrill, line 12: There is a lot of duplication in the meat of this function.  I believe we mostly are diffing the podspec or related template.  Can we refactor into maybe a pod comparitor or something

internal/utils/comparators/deployment/comparator.go
```diff
@@ -0,0 +1,80 @@
...
+		log.V(3).Info("Deployment image change", "Deployment", current.Name)
+		return false, "image"
+	}
+
+	if utils.AreResourcesDifferent(current, desired) {
+		log.V(3).Info("Deployment resource(s) change", "DeploymentName", current.Name)
+		return false, "resources"
+	}
+
+	if !utils.EnvValueEqual(current.Spec.Template.Spec.Containers[0].Env, desired.Spec.Template.Spec.Containers[0].Env) {
```
jcantrill, line 34: I'm surprised we did not spin through the containers here

internal/k8shandler/collection.go
Clee2691: When any input is added to a pipeline that isn't an HTTP receiver, it will reconcile into a daemonset and remove the deployment.

internal/utils/comparators/deployment/comparator.go
Clee2691: These comparators are only comparing PodSpecs so I refactored into `pod/comparator.go` I removed deployment and daemonset comparators.

internal/utils/comparators/deployment/comparator.go
Clee2691: Changed to loop through all containers to compare

internal/collector/collector_test.go
```diff
@@ -230,6 +231,229 @@ var _ = Describe("Factory#NewPodSpec", func() {
...
+	Describe("when creating of the collector container", func() {
+
+		It("should provide the pod IP as an environment var", func() {
+			Expect(collector.Env).To(IncludeEnvVar(v1.EnvVar{Name: "POD_IP",
+				ValueFrom: &v1.EnvVarSource{
+					FieldRef: &v1.ObjectFieldSelector{
+						APIVersion: "v1", FieldPath: "status.podIP"}}}))
+		})
+		It("should set a security context", func() {
+			Expect(collector.SecurityContext).To(Equal(&v1.SecurityContext{
```
jcantrill, line 272: We could also remove this context I'm betting all-to-gether.  This defines the cgroups for the containers which we do not need anything special in this scenario

internal/factory/deployment.go
```diff
@@ -0,0 +1,50 @@
...
+	"github.com/openshift/cluster-logging-operator/internal/constants"
+	"github.com/openshift/cluster-logging-operator/internal/runtime"
+	"github.com/openshift/cluster-logging-operator/internal/utils"
+	apps "k8s.io/api/apps/v1"
+	core "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+)
+
+// NewDeployment stubs an instance of a deployment
+func NewDeployment(deploymentName, namespace, loggingComponent, component, impl string, podSpec core.PodSpec, visitors ...func(o runtime.Object)) *apps.Deployment {
```
jcantrill, line 13: deploymentName and Namespace should be swapped for 2 reasons:

* prefer namespace/name in general for args
* to match the actual [caller](https://github.com/openshift/cluster-logging-operator/pull/2240/files#diff-f143e28681fca473048e426218fd5def9528a1dd9c5279be226665a3ab1a5601R23)

internal/k8shandler/collection.go
jcantrill: Please make sure to add this as a testcase to the issue we need to check

internal/reconcile/deployment.go
```diff
@@ -0,0 +1,58 @@
...
+		if err := k8Client.Get(context.TODO(), key, current); err != nil {
+			if errors.IsNotFound(err) {
+				reason = constants.EventReasonCreateObject
+				return k8Client.Create(context.TODO(), desired)
+			}
+			return fmt.Errorf("failed to get %v Deployment: %w", key, err)
+		}
+		same := false
+
+		if same, updateReason = pod.AreSame(&current.Spec.Template.Spec, &desired.Spec.Template.Spec, current.Name); same && util.HasSameOwner(current.OwnerReferences, desired.OwnerReferences) {
```
jcantrill, line 36: I do not think we can completely ONLY verify the pod template spec.  Don't we add labels to the Deployment/Daemonset which are interesting? Did we verify any specific deployment strategies ? 

If this results in no regression, please ignore my previous statement, though deployments do allow defining replicas which I presume we only ever want one

internal/utils/comparators/pod/comparator.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	"reflect"
+
+	log "github.com/ViaQ/logerr/v2/log/static"
+	"github.com/openshift/cluster-logging-operator/internal/utils"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// AreSame compares pods for equality and return true equal otherwise false
+func AreSame(current *corev1.PodSpec, desired *corev1.PodSpec, name string) (bool, string) {
```
jcantrill, line 12: I would remove "collector" from these given its a generic utility function

internal/reconcile/deployment.go
Clee2691: There was no code that specifically checked `deployment`/ `daemonset` specific parameters. We never compared labels on them, just the pods and we also never compared the strategies.

internal/factory/deployment.go
```diff
@@ -0,0 +1,50 @@
...
+	annotations := map[string]string{
+		"target.workload.openshift.io/management": `{"effect": "PreferredDuringScheduling"}`,
+	}
+
+	strategy := apps.DeploymentStrategy{
+		Type: apps.RollingUpdateDeploymentStrategyType,
+		RollingUpdate: &apps.RollingUpdateDeployment{
+			MaxUnavailable: &intstr.IntOrString{
+				Type:   intstr.String,
+				StrVal: "100%",
```
jcantrill, line 37: Having conversation the token does not immediately become unusable meaning we can have multiples running at the same time.  Kube docs say the default is 25% https://kubernetes.io/docs/concepts/workloads/controllers/deployment/ so I would try removing this value @wanghaoran1988 suggests we might start with 2 replicas

Clee2691:
Closing this PR in favor of: https://github.com/openshift/cluster-logging-operator/pull/2304

PR #2245, opened 2023-11-14
LOG-4813: Add matcher to help test conf generation

jcantrill:
### Description
This PR:

* adds test matchers to assist in testing conf generation and to make it possible for assertions to be consistent across the test base

### Links
* https://issues.redhat.com/browse/LOG-4813

cc @cahartma @syedriko

### DIFF
--- a/test/matchers/diff.go
+++ b/test/matchers/diff.go
@@ -39,7 +39,7 @@ type lineMatcher struct {
 }
 
 func (m *lineMatcher) Match(actual interface{}) (success bool, err error) {
-	m.diff = cmp.Diff(m.normalize(actual.(string)), m.normalize(m.expected.(string)))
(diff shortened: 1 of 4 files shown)


test/matchers/diff.go
```diff
@@ -51,10 +51,10 @@ func (m *lineMatcher) NegatedFailureMessage(actual interface{}) (message string)
 	return ("Expected differences but none found.")
 }
 
-func (m *lineMatcher) normalize(in string) []string {
+func normalize(in string, trim bool) []string {
 	out := []string{}
 	for _, line := range strings.Split(in, "\n") {
-		if m.trim {
+		if trim {
```
syedriko, line 57: Nothing to do with this PR, but I'm not quite seeing why we strings.TrimSpace(line) twice.

PR #2246, opened 2023-11-14
LOG-4743: Add constructor functions for source generator elements

jcantrill:
### Description
This PR:

* adds constructor functions for generator source elements

### Links
* https://issues.redhat.com/browse/LOG-4814

cc @syedriko @cahartma


internal/generator/vector/source/audit_logs.go
```diff
@@ -54,3 +54,39 @@ glob_minimum_cooldown_ms = 15000
 `
 
 type K8sAuditLog = framework.ConfLiteral
+
```
syedriko, line 57: The const strings above don't need to be exported anymore.

PR #2248, opened 2023-11-14
LOG-4815: relax the need for permissions on receiver input

jcantrill:
### Description
This PR:
* relaxes the need for RBAC for receiver inputs

### Links
* https://issues.redhat.com/browse/LOG-4815

cc @alanconway @syedriko

### DIFF
--- a/internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
+++ b/internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
@@ -44,8 +44,8 @@ func ValidateServiceAccount(clf loggingv1.ClusterLogForwarder, k8sClient client.
 		return err, nil
 	}
 	// If SA present, validate permissions based off spec'd CLF inputs
-	clfInputs := gatherPipelineInputs(clf)
-	if err = validateServiceAccountPermissions(k8sClient, clfInputs, serviceAccount, clf.Namespace, clf.Name); err != nil {
+	clfInputs, hasReceiverInputs := gatherPipelineInputs(clf)
+	if err = validateServiceAccountPermissions(k8sClient, clfInputs, hasReceiverInputs, serviceAccount, clf.Namespace, clf.Name); err != nil {
 		return err, nil
 	}
 	return nil, nil
@@ -67,7 +67,10 @@ func getServiceAccount(name, namespace string, k8sClient client.Client) (*corev1
 // ValidateServiceAccountPermissions validates a service account for permissions to collect
 // inputs specified by the CLF.
 // ie. collect-application-logs, collect-audit-logs, collect-infrastructure-logs
-func validateServiceAccountPermissions(k8sClient client.Client, inputs sets.String, serviceAccount *corev1.ServiceAccount, clfNamespace, name string) error {
+func validateServiceAccountPermissions(k8sClient client.Client, inputs sets.String, hasReceiverInputs bool, serviceAccount *corev1.ServiceAccount, clfNamespace, name string) error {
+	if inputs.Len() == 0 && hasReceiverInputs {
+		return nil
+	}
 	if inputs.Len() == 0 {
 		err := errors.NewValidationError("There is an error in the input permission validation; no inputs were found to evaluate")
 		log.Error(err, "Error while evaluating ClusterLogForwarder permissions", "namespace", clfNamespace, "name", name)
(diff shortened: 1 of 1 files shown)


internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
```diff
@@ -44,8 +44,8 @@ func ValidateServiceAccount(clf loggingv1.ClusterLogForwarder, k8sClient client.
 		return err, nil
 	}
 	// If SA present, validate permissions based off spec'd CLF inputs
-	clfInputs := gatherPipelineInputs(clf)
-	if err = validateServiceAccountPermissions(k8sClient, clfInputs, serviceAccount, clf.Namespace, clf.Name); err != nil {
+	clfInputs, noOfReceiverInputs := gatherPipelineInputs(clf)
```
syedriko, line 47: Would numOfReceiverInputs or countOfReceiverInputs read better?

syedriko: uint seems like a better type for a count

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
```diff
@@ -67,7 +67,10 @@ func getServiceAccount(name, namespace string, k8sClient client.Client) (*corev1
 // ValidateServiceAccountPermissions validates a service account for permissions to collect
 // inputs specified by the CLF.
 // ie. collect-application-logs, collect-audit-logs, collect-infrastructure-logs
-func validateServiceAccountPermissions(k8sClient client.Client, inputs sets.String, serviceAccount *corev1.ServiceAccount, clfNamespace, name string) error {
+func validateServiceAccountPermissions(k8sClient client.Client, inputs sets.String, noOfReceiverInputs int, serviceAccount *corev1.ServiceAccount, clfNamespace, name string) error {
+	if inputs.Len() == 0 && noOfReceiverInputs > 0 {
```
syedriko, line 71: `if inputs.Len() == 0 && noOfReceiverInputs > 0` can be evaluated without entering this function

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
```diff
@@ -112,16 +115,17 @@ func gatherPipelineInputs(clf loggingv1.ClusterLogForwarder) sets.String {
...
 	}
 
+	noOfReceivers := 0
 	for _, input := range clf.Spec.Inputs {
 		if inputRefs.Has(input.Name) && input.Application != nil {
 			inputTypes.Insert(loggingv1.InputNameApplication)
 		}
 		if input.Receiver != nil && input.Receiver.HTTP != nil {
-			inputTypes.Insert(loggingv1.InputNameAudit)
+			noOfReceivers += 1
```
syedriko, line 124: Are we counting all receivers or just the HTTP ones?

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
syedriko: On a second look, it can be a `hasReceiverInputs bool`

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
jcantrill: at this point we only have on type of receiver and we want to ensure http is configured.  Maybe it is good enough to simply check if receiver is non nil

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
jcantrill: I realized that as well but chose this implementation to defer the evaluation to one place instead of conditionals outside of the function and within.  I'm not hard pressed to leave it one way or the other if you think its better to not get this far

jcantrill: What if we have inputs in addition to receiver inputs?  Doesn't the way it is now allow consolidation of validation in one place?

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
syedriko: I'm wondering because of the syslog receiver that's coming up, how it should be treated.

jcantrill: I figured that was the source of your inquiry.  There should be validation that ensures one or the other type is populated.  This means, in theory, checking that receiver is non-nil should be sufficient

PR #2254, opened 2023-11-16
LOG-4823: Explicitly specify vector config file as /etc/vector/vector.toml

syedriko:
### Description
In preparation for upstream vector v0.34.0, explicitly specify vector config file as /etc/vector/vector.toml. The default config file from v0.34.0 on is /etc/vector/vector.yaml.

/cc @jcantrill 
/assign @cahartma 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4823


syedriko:
> If we mounted it incorrectly don't we also have to change the volume mounts?

It's mounted just fine, the difference is that vector.yaml is the new default

PR #2255, opened 2023-11-16
LOG-4792: Collecting `clusterlogging` and `clusterlogforwarder` with inspect

oarribas:
### Description
Get the `clusterlogging` and `clusterlogforwarder` with the `oc adm inspect` command to avoid the issue reported in [LOG-4792](https://issues.redhat.com//browse/LOG-4792).
Getting also the `customresourcedefinitions` to allow https://github.com/gmeghnag/omc/  to work with above (and other) resources (on testing, it adds only few MB to the full must-gather size for ~200 `customresourcedefinitions`).

I'm not sure if `clusterlogging` and `clusterlogforwarder` should be copied to the `clo` directory. Copying it for now for compatibility with current behavior:
https://github.com/oarribas/cluster-logging-operator/blob/oarribas-cl-and-clo/must-gather/collection-scripts/gather_cluster_logging_operator_resources#L42

/cc @periklis 

Not sure how to use the cherry pick feature to include this in 5.8.z (and if possible also in 5.7.z).

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4792

### DIFF
--- a/must-gather/collection-scripts/gather
+++ b/must-gather/collection-scripts/gather
@@ -49,6 +49,7 @@ cluster_resources+=(clusterrolebindings)
 cluster_resources+=(persistentvolumes)
 cluster_resources+=(clusterversion)
 cluster_resources+=(machineconfigpool)
+cluster_resources+=(customresourcedefinitions)
 
 log "-BEGIN inspecting CRs..." >> "${LOGFILE_PATH}"
 for cr in "${cluster_resources[@]}" ; do
@@ -67,7 +68,6 @@ resources+=(serviceaccounts)
 resources+=(events)
 # clusterlogging is only in the openshift-logging namespace, but it should not impact a lot to run the inspect in the other namespaces
 resources+=(clusterlogging)
-# with multi log forwarder, the clusterlogforwarder can be in custom namespaces, and they are not collected here
 resources+=(clusterlogforwarder)
 
 log "BEGIN inspecting namespaces ..." >> "${LOGFILE_PATH}"
--- a/must-gather/collection-scripts/gather
+++ b/must-gather/collection-scripts/gather
@@ -65,6 +65,10 @@ resources+=(rolebindings)
 resources+=(configmaps)
 resources+=(serviceaccounts)
 resources+=(events)
+# clusterlogging is only in the openshift-logging namespace, but it should not impact a lot to run the inspect in the other namespaces
+resources+=(clusterlogging)
+# with multi log forwarder, the clusterlogforwarder can be in custom namespaces, and they are not collected here
+resources+=(clusterlogforwarder)
 
 log "BEGIN inspecting namespaces ..." >> "${LOGFILE_PATH}"
 
@@ -121,7 +125,7 @@ if [ "$found_es" != "" ] || [ "$found_lokistack" != "" ] ; then
   if [ "$found" != "" ] ; then
     KUBECACHEDIR=${BASE_COLLECTION_PATH}/cache-dir ${SCRIPT_DIR}/gather_visualization_resources "$BASE_COLLECTION_PATH" >> "${LOGFILE_PATH}" 2>&1
   fi
-  log "BEGIN gathering CLO resources ..." >> "${LOGFILE_PATH}"
+  log "END gathering logstorage resources ..." >> "${LOGFILE_PATH}"
 else
   log "Skipping logstorage inspection.  No Elasticsearch deployment found" >> "${LOGFILE_PATH}" 2>&1
 fi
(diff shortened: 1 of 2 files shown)


jcantrill:
@oarribas you will likely need to manually backport to 5.7 given we made changes to support multi-CLF

jcantrill:
@oarribas please squash commits prior to merge

must-gather/collection-scripts/gather
```diff
@@ -65,6 +66,9 @@ resources+=(rolebindings)
 resources+=(configmaps)
 resources+=(serviceaccounts)
 resources+=(events)
+# clusterlogging is only in the openshift-logging namespace, but it should not impact a lot to run the inspect in the other namespaces
```
jcantrill, line 69: This statement is not correct.  With the introduction of multi CLF, users can add a CL resource to change resources and schedulting of their forwarder

must-gather/collection-scripts/gather_cluster_logging_operator_resources
```diff
@@ -38,16 +38,12 @@ if [ $NAMESPACE == "openshift-logging" ]; then
...
-for r in "clusterlogging" "clusterlogforwarder" ; do
-  names="$(oc -n $NAMESPACE get $r --ignore-not-found -o jsonpath='{.items[*].metadata.name}' | sort -u)"
-  for name in ${names[@]}; do
-    data="$(oc -n $NAMESPACE get $r ${name} --ignore-not-found -o yaml)"
-    if [ "$data" != "" ] ; then
-      echo $data > "${clo_folder}/${r}_${name}.yaml"
-    fi
-  done
-done
+log "Gathering data for 'clusterlogging' and 'clusterlogforwarder' from ${BASE_COLLECTION_PATH}/namespaces/openshift-logging/logging.openshift.io/"
```
jcantrill, line 41: This is incorrect as it breaks must-gather for the multi CLF scenario.  We should be getting these from ANY namespace that has them.  @cahartma didn't you already address this change?

jcantrill (commented):
/hold

acandelp:
Hello,

I would like to give my positive opinion on this PR. Practically every day we encounter clients who are not able to obtain a logging must-gather due to problems in their cluster and inspect in most cases does not cause problems when obtained. I think it would be very positive not to have to request the inspect on the one hand and the configuration files on the other

jcantrill:
> Hello,
> 
> I would like to give my positive opinion on this PR. Practically every day we encounter clients who are not able to obtain a logging must-gather due to problems in their cluster and inspect in most cases does not cause problems when obtained. I think it would be very positive not to have to request the inspect on the one hand and the configuration files on the other

I don't understand what you are saying here. What is it you are suggesting? How does the PR address your needs?

oarribas:
> @oarribas please squash commits prior to merge

@jcantrill , not sure if I can do it now, or if a new PR will be needed. I have seen [1], which seems that allow to squash the commits just before the merge, but I don't have that option. Is there any other way to do an squash directly to this PR before merging?


[1] https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/merging-a-pull-request

gmeghnag:
It could be super useful also to collect `installplan`, `subscription`, and `clusterserviceversion` via `oc adm inspect` command, so we can parse these resource either with `omc` in an easy way.

jcantrill:
@oarribas 

> Is there any other way to do an squash directly to this PR before merging?

```
git rebase -i HEAD~<NUMBER OF COMMITS>
```

jcantrill:
@gmeghnag 

> It could be super useful also to collect `installplan`, `subscription`, and `clusterserviceversion` via `oc adm inspect` command, so we can parse these resource either with `omc` in an easy way.

These are already gathered and found in the `<root>/registry***/cluster-logging/install`.  If you believe there is a better place or more info required, please submit a separate PR

oarribas:
@gmeghnag , yes, I think they can be collected also using `oc adm inspect`. I need to do some changes to this PR (probably creating a new one), and after that create a different one for `installplan`, `subscription`, and `clusterserviceversion` (if you don't do it before :) )

must-gather/collection-scripts/gather_cluster_logging_operator_resources
cahartma: I did and something is strange here.  Still trying to figure out what I'm looking at.  How is all that logic missing?

cahartma: We are passing the namespace into that script...    so it's collecting from all namespaces that contain a cl or clf
https://github.com/openshift/cluster-logging-operator/blob/master/must-gather/collection-scripts/gather#L29-L43

cahartma:
I suggest we create a new Jira and list out these other tasks.    This can be a powerful tool and we should continue to expand/cleanup its capabilities.

The issue described in this Jira is resolved by this [PR](https://github.com/openshift/cluster-logging-operator/pull/2256).  It is not part of the multi-clf logic and was a simple mistake on my part.

oarribas:
Replaced by #2257

PR #2261, opened 2023-11-20
LOG-4822: Loki output without URL should fail validation

vparfonov:
### Description
This PR fix validation for Loki output spec: should fail if URL not provided

/cc 
/assign @jcantrill 

/cherry-pick /release-5.8

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-4822
- Enhancement proposal:

### DIFF
--- a/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
+++ b/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
@@ -282,8 +282,7 @@ func verifyOutputURL(output *loggingv1.OutputSpec, conds loggingv1.NamedConditio
 		if output.URL == "" {
 			// Some output types allow a missing URL
 			// TODO (alanconway) move output-specific valiation to the output implementation.
-			if output.Type == loggingv1.OutputTypeCloudwatch ||
-				output.Type == loggingv1.OutputTypeGoogleCloudLogging || output.Type == loggingv1.OutputTypeLoki {
+			if output.Type == loggingv1.OutputTypeCloudwatch || output.Type == loggingv1.OutputTypeGoogleCloudLogging {
 				return true
 			} else {
 				return fail(conditions.CondInvalid("URL is required for output type %v", output.Type))
--- a/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
+++ b/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
@@ -1173,6 +1173,17 @@ func Test_verifyOutputURL(t *testing.T) {
 			},
 			want: false,
 		},
+		{
+			name: "With Loki output without URL should fail",
+			args: args{
+				output: &loggingv1.OutputSpec{
+					Name: "test-output",
+					Type: loggingv1.OutputTypeLoki,
+				},
+				conds: loggingv1.NamedConditions{},
+			},
+			want: false,
+		},
 	}
 	for _, tt := range tests {
 		tt := tt // Don't bind range variable.


internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go
```diff
@@ -330,7 +330,8 @@ func verifyOutputURL(output *loggingv1.OutputSpec, conds loggingv1.NamedConditio
 			// Some output types allow a missing URL
 			// TODO (alanconway) move output-specific valiation to the output implementation.
 			if output.Type == loggingv1.OutputTypeCloudwatch ||
-				output.Type == loggingv1.OutputTypeGoogleCloudLogging || output.Type == loggingv1.OutputTypeLoki {
+				output.Type == loggingv1.OutputTypeGoogleCloudLogging ||
+				(output.Type == loggingv1.OutputTypeLoki && output.Name == loggingv1.OutputNameDefault) {
```
jcantrill, line 334: @vparfonov This validates the opposite of what the observed issue.  Loki requires a URL; there is nothing in the client that crafts one.  I believe the issue is a result of a missing URL

vparfonov: > @vparfonov This validates the opposite of what the observed issue. Loki requires a URL; there is nothing in the client that crafts one. I believe the issue is a result of a missing URL

Yes, and URL will be added automatically for `default` output: https://github.com/openshift/cluster-logging-operator/blob/master/internal/logstore/lokistack/logstore_lokistack.go#L220

jcantrill: @vparfonov my apologize for the delay in response but this change will not work for loki regardless if it is the default or not.  We do not require the "default" loki to be named "default", only prefixed with it.  Please think about the reconciliation process:

* [migrate](https://github.com/openshift/cluster-logging-operator/blob/release-5.8/internal/migrations/clusterlogforwarder/migrate_clusterlogforwarder.go#L54)
* validate
* reconcile

This means the validate code should fail if  URL is not defined, which is what is reported. "Default" is a special case which we remove in reconciliation processing by "migrating" the CLF to have a proper definition for our internal loki

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
```diff
@@ -1210,6 +1210,57 @@ var _ = Describe("Validate clusterlogforwarderspec", func() {
...
+			clf := loggingv1.ClusterLogForwarder{}
+			clf.Spec = *forwarderSpec
+			clf.Name = constants.SingletonName
+			clf.Namespace = constants.OpenshiftNS
+
+			_, clfStatus = ValidateInputsOutputsPipelines(clf, client, extras)
+			Expect(clfStatus.Outputs["foo"]).To(HaveCondition(loggingv1.ConditionReady, false, loggingv1.ReasonInvalid, "URL is required for output type loki"))
+		})
+
+		It("should pass with Loki and without URL for default output", func() {
```
jcantrill, line 1238: This test should not be required or rewritten to confirm default loki does require URL

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
```diff
@@ -1173,6 +1173,17 @@ func Test_verifyOutputURL(t *testing.T) {
 			},
 			want: false,
 		},
+		{
+			name: "With Loki without URL and NOT default output",
```
jcantrill, line 1177: Nit. Update name.  It does not matter if it is the default output or not.  It will always need a URL.  Our migration of the default output should craft the URL so it should have one by the time it gets to the validation.

jcantrill: otherwise lgtm

jcantrill (commented):
/lgtm

PR #2262, opened 2023-11-20
LOG-4791: Fluentd collector Pods no longer picks up the log collector SAs Secret as a fallback

Clee2691:
### Description
This PR addresses forwarding to an internal `LokiStack` where in the legacy case, if the `pipelines.outputs` is not `default` it will not apply the `logcollector-token` secret's `token` and `ca.crt`. For a CLF named `instance` it will add the `logcollector` SA token to be used. 

/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4791


internal/k8shandler/forwarding.go
```diff
@@ -47,6 +48,14 @@ func (clusterRequest *ClusterLoggingRequest) generateCollectorConfig() (config s
 	op[generator.ClusterTLSProfileSpec] = tls.GetClusterTLSProfileSpec(tlsProfile)
 	EvaluateAnnotationsForEnabledCapabilities(clusterRequest.Forwarder, op)
 
+	// Use logcollector SA token/ca.crt for the legacy case
+	if clusterRequest.Forwarder.Spec.ServiceAccountName == constants.CollectorServiceAccountName {
```
jcantrill, line 52: I would move this logic into ClusterRequest#SetOutputSecret so that all the "retrieve secret and create map" logic is in one place

internal/k8shandler/forwarding.go
```diff
@@ -71,3 +80,21 @@ func (clusterRequest *ClusterLoggingRequest) SetOutputSecrets() {
 		clusterRequest.OutputSecrets[output.Name] = secret
 	}
 }
+
+func (clusterRequest *ClusterLoggingRequest) getLogCollectorServiceAccountTokenSecret() (*corev1.Secret, error) {
```
jcantrill, line 84: Can we use similar logic in the existing SetOutputSecrets function to just get the secret?  Do we need additional verification?  I believe there is an established pattern for SA token secrets upon which we can rely and may not require additional validation

PR #2263, opened 2023-11-20
LOG-4829: Refactor kubernetes_logs source to allow includes and excludes

jcantrill:
### Description
This PR:
* refactors kubernetes_logs source to allow configuration of includes and excludes

### Links
* https://issues.redhat.com/browse/LOG-4829


internal/generator/vector/source/kubernetes_logs.go
```diff
@@ -36,42 +42,94 @@ rotate_wait_ms = 5000
...
+	var previous byte
+	for i, _ := range entry {
+		current := entry[i]
+		if current == '*' && current == previous {
+		} else {
+			result = append(result, current)
+		}
+		previous = current
+	}
+	return string(result)
```
cahartma, line 134: A cleaner way perhaps:
```suggestion
func removeDuplicateWildcards(entry string) string {
	for strings.Contains(entry, "**") {
		entry = strings.ReplaceAll(entry, "**", "*")
	}
	return string(entry)
```

syedriko: Or cleaner yet
```
func removeDuplicateWildcards(entry string) string {
    return strings.ReplaceAll(entry, "**", "*")
}
```

internal/generator/vector/source/kubernetes_logs.go
```diff
@@ -36,42 +42,94 @@ rotate_wait_ms = 5000
...
+	return NewKubernetesLogs(id, "", joinContainerPathsForVector(excludes))
+}
+
+// NewKubernetesLogs element which always excludes temp and gzip files
+func NewKubernetesLogs(id, includes, excludes string) KubernetesLogs {
 	return KubernetesLogs{
 		ComponentID:  id,
 		Desc:         "Logs from containers (including openshift containers)",
-		ExcludePaths: ExcludeContainerPaths(namespace),
+		IncludePaths: includes,
```
cahartma, line 68: How does this get populated?  Is this a stub for future efforts?

internal/generator/vector/source/kubernetes_logs.go
```diff
@@ -36,42 +42,94 @@ rotate_wait_ms = 5000
...
+//	namespaces:
+//	  namespace:     /var/log/pods/namespace_*/*/*.log
+//	  **namespace:   /var/log/pods/*namespace_*/*/*.log
+//	  **name*pace**: /var/log/pods/*name*pace*/*/*.log
+//	  namespace**:   /var/log/pods/namespace*/*/*.log
+//	containers:
+//	  container:    /var/log/pods/*/container/*.log
+//	  *cont**iner*:    /var/log/pods/*/*cont*iner*/*.log
+//	  cont**iner*:    /var/log/pods/*/cont*iner*/*.log
+func ContainerPathGlobFrom(namespaces, containers []string, extensions ...string) string {
```
cahartma, line 95: No implementation I can see.  Is this for testing only, seems it could/should be incorporated.

internal/generator/vector/source/kubernetes_logs.go
jcantrill: It's here for future when we implement the input APIs and allow users to include or exclude namespaces or containers

internal/generator/vector/source/kubernetes_logs.go
jcantrill: I'm currently writing something now that utilizes this function to enable dedicated inputs to container source config

internal/generator/vector/source/kubernetes_logs.go
jcantrill: What if I have more then 2 '*'?  I don't really like spinning through it byte by byte but this is the only way I could think to ensure multiples are addressed.  Separately also thinking there is a validation to ensure it is a valid NS

syedriko: > What if I have more then 2 '*'? I don't really like spinning through it byte by byte but this is the only way I could think to ensure multiples are addressed. Separately also thinking there is a validation to ensure it is a valid NS

In that case
```
var consecutiveWildcards = regexp.MustCompile(`\*+`)
...
func collapseWildcards(entry string) string {
	return consecutiveWildcards.ReplaceAllString(entry, "*")
}
```

syedriko: Or we can rewrite the original to be explicit that it works with runes and not bytes
```
func collapseWildcards(entry string) string {
	var result strings.Builder
	var previous rune
	result.Grow(len(entry))
	for _, current := range entry {
		if current == '*' && previous == '*' {
		} else {
			result.WriteRune(current)
		}
		previous = current
	}
	return result.String()
}
```
https://go.dev/play/p/5b_KVnV0gPM

cahartma: Replace all works, then a simple for loop in case of more than 2 consecutive.    Explicit and efficient

PR #2272, opened 2023-11-28
LOG-4846: Adjust CLO to include the korrel8r proxy in the logging view plugin

alanconway:
Korrel8r proxy is enabled whenever Loki view is.
Points to fixed service korrel8r in namespace korrel8r.

Keeping it simple, the role of enabling korrel8r will likely move out of CLO,
and become part of the new observability operator in future.

/cc @jcantrill
/cc @xperimental
/cc @gbernal

### DIFF
--- a/internal/constants/features.go
+++ b/internal/constants/features.go
@@ -17,4 +17,8 @@ const (
 	// without switching the default `logStore` to LokiStack. The value should be the
 	// LokiStack resource name representing the target store for the migration.
 	AnnotationOCPConsoleMigrationTarget = "logging.openshift.io/force-enable-ocp-console-target"
+
+	// AnnotationPreviewKorrel8rConsole enables preview features in the console that use Korrel8r.
+	// Korrel8r must also be installed and running in the cluster for these features to work.
+	AnnotationPreviewKorrel8rConsole = "logging.openshift.io/preview-korrel8r-console"
 )
--- a/internal/visualization/console/plugin.go
+++ b/internal/visualization/console/plugin.go
@@ -7,19 +7,19 @@ import (
 	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
 	"github.com/openshift/cluster-logging-operator/internal/constants"
 	"github.com/openshift/cluster-logging-operator/internal/logstore/lokistack"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // ReconcilePlugin reconciles the console plugin to expose log querying of storage
 func ReconcilePlugin(k8sClient client.Client, cl *logging.ClusterLogging, owner client.Object, clusterVersion string) error {
 	lokiService := lokiServiceName(cl)
-
+	korrel8rNN := korrel8rNamespacedName(cl)
 	var consoleSpec *logging.OCPConsoleSpec
 	if cl != nil && cl.Spec.Visualization != nil {
 		consoleSpec = cl.Spec.Visualization.OCPConsole
 	}
-
-	r := NewReconciler(k8sClient, NewConfig(owner, lokiService, FeaturesForOCP(clusterVersion)), consoleSpec)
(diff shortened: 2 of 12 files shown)


jcantrill:
My only question is if this should be put behind a feature gate.  Will Korrel8r be GA at the time of 5.9 release? If it is going to move then seems to make more sense to call this an alpha type feature

internal/visualization/console/reconciler.go
```diff
@@ -161,16 +161,28 @@ func (r *Reconciler) mutateConsolePlugin() error {
...
+				Service: consolev1alpha1.ConsolePluginProxyServiceConfig{
+					Name:      r.LokiService,
+					Namespace: r.Namespace(),
+					Port:      r.LokiPort,
+				},
+			},
+			{
+				Type:      "Service",
+				Alias:     "korrel8r",
+				Authorize: true,
```
jgbernalp, line 178: This should be `false` as POST request will be blocked by the console proxy without a `X-CSRFToken` header, which is the current behavior.
```suggestion
				Authorize: false,
```

alanconway:
agreed on feature gate, will add one
/hold

internal/visualization/console/reconciler_test.go
```diff
@@ -92,16 +94,28 @@ func TestVerifyResources(t *testing.T) {
...
+							Name:      "someservice",
+							Namespace: "openshift-logging",
+							Port:      8080,
+						},
 					},
-				}}, cp.Spec.Proxy)
+					{
+						Type:      "Service",
+						Alias:     "korrel8r",
+						Authorize: true,
```
jgbernalp, line 111:
```suggestion
						Authorize: false,
```

internal/visualization/console/plugin.go
```diff
@@ -10,6 +10,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	korrel8rService   = "korrel8r"
+	korrel8rNamespace = "korrel8r"
```
jcantrill, line 15: will this indeed be the NS?  Is it something intended to live in openshift?  If it is an observability service, do we have a preferred NS prefix for the observability org?

alanconway:
@jcantrill added feature gate

internal/constants/features.go
```diff
@@ -17,4 +17,8 @@ const (
 	// without switching the default `logStore` to LokiStack. The value should be the
 	// LokiStack resource name representing the target store for the migration.
 	AnnotationOCPConsoleMigrationTarget = "logging.openshift.io/force-enable-ocp-console-target"
+
+	// AnnotationPreviewKorrel8rConsole enables preview features in the console that use Korrel8r.
+	// Korrel8r must also be installed and running in the cluster for these features to work.
+	AnnotationPreviewKorrel8rConsole = "logging.openshift.io/preview-korrel8r-consle"
```
jcantrill, line 23: (sp)... "console"

internal/visualization/console/plugin.go
jcantrill: @alanconway don't see this comment addressed..

jcantrill:
@alanconway one nit, one open question, one review from @jgbernalp ...

internal/visualization/console/plugin.go
alanconway: I don't know what the long term story will be. If you prefer 'openshift-korrel8r' as more consistent for now, I'll make the change. There isn't anything concrete from the observability side yet.

internal/constants/features.go
alanconway: Fixed

internal/visualization/console/plugin.go
alanconway: Going with korrel8r for now, bu taking note that we need to resolve the correct naming before tech preview.

internal/visualization/console/reconciler_test.go
alanconway: fixed

PR #2274, opened 2023-11-29
WIP: LOG-4445: Forwarder must use hosted cluster_id on HCP cluster.

alanconway:


internal/clusterinfo/clusterinfo.go
```diff
@@ -0,0 +1,66 @@
...
+// getHCPInfo returns ClusterInfo from a hypershift control plane if found, nil otherwise.
+func getHCPInfo(ctx context.Context, cfg *rest.Config, namespace string) *ClusterInfo {
+	// TODO: using dynamic client because of dependency problems with the hypershift API package:
+	// When https://issues.redhat.com/browse/HOSTEDCP-336 is fixed we can enable clusterinfo_static.go
+	dc, err := dynamic.NewForConfig(cfg)
+	if err != nil {
+		return nil
+	}
+	l, err := dc.Resource(hcpGVR).Namespace(namespace).List(ctx, metav1.ListOptions{})
+	if err != nil || len(l.Items) != 1 {
```
jcantrill, line 58: I expect it won't but do we care if it does not equal 1?  Don't we just car that it is greater then zero?

internal/clusterinfo/clusterinfo_static.go
```diff
@@ -0,0 +1,52 @@
+//go:build DISABLE
```
jcantrill, line 1: I know we prefer to use a static client in future but does it really matter to switch?  Looking at this code one particular concern I have is what is the behavior not running on an HCP cluster?  Won't the operator fail out right of the hypershift APIs are not available.  The init() function concerns me for nonHCP scenarios.  I'm betting the dynamic client is not subject to these same concerns

internal/clusterinfo/clusterinfo_static.go
```diff
@@ -0,0 +1,52 @@
...
+
+// GetClusterInfo gets cluster info for the cluster we are running in.
+//
+// If namespace contains a HostedControlPlane then return info for the *guest* cluster, not the host cluster.
+// We assume in this case that CLF is running on behalf of the guest cluster to collect API audit logs.
+func Get(c client.Reader, namespace string) (ClusterInfo, error) {
+	// Use HCP info if exactly one HCP is present in the namespace.
+	hcps := &hypershiftv1beta1.HostedControlPlaneList{}
+	err := c.List(context.Background(), hcps, client.InNamespace(namespace))
+	if err == nil && len(hcps.Items) == 1 {
```
jcantrill, line 37: I think we only care if the length is >0

internal/clusterinfo/clusterinfo_test.go
```diff
@@ -0,0 +1,52 @@
+package clusterinfo
```
jcantrill, line 1: Seems we could combine this test with the other since its supposed to be the same result with a different client.  Can we just add more entries and slightly modify it

main.go
```diff
@@ -116,13 +117,11 @@ func main() {
 		}
 	}()
 
-	clusterVersion, err := getClusterVersion(mgr.GetAPIReader())
+	info, err := clusterinfo.Get(context.Background(), mgr.GetAPIReader(), getOpenshiftNS())
```
jcantrill, line 120: Why do we need a NS here?  If it is for HCP, then this should be the NS of the CLF which means you need to fetch this on every reconcile or cache entries somewhere.  If this is for nonHCP, clusterinfo AFAIK is a clusterwide resource which has no need of a NS.

jcantrill:
@alanconway I would additionally rebase this against 5.8 since we are delivering their first

internal/clusterinfo/clusterinfo.go
alanconway: There should never be more than one, so it doesn't really matter. 
My thinking is if anything does not look as expected for the HCP case, then we fall back to the original behaviour. Multiple HCPs is not expected, and we don'tt know which one we should use.

internal/clusterinfo/clusterinfo_static.go
alanconway: +1. I used the static code & tests to give myself some confidence that I got the field names etc. correct in the dynamic code, but if we have to use the dynamic code then there's no real need to switch it unless/until we have to modify the logic in some way.

alanconway: Will clean up - static code is just a distraction now.

internal/clusterinfo/clusterinfo_test.go
alanconway: The static code & tests are uncompilable within the CLO module because of dependency clashes. I built them as a separate module to make sure the dynamic code was correct, but I'll remove now as they're a distraction.

main.go
alanconway: You are right - This is for HCP, we need to get it on every reconcile and cache it so it can be used for collector generation for that reconcile. I need to rethink this bit...

PR #2277, opened 2023-12-01
LOG-4445: Forwarder must use hosted cluster_id on HCP cluster.

alanconway:
NOTE: Using unstructured objects for HCP because of dependency problems with the hypershift API package.
When https://issues.redhat.com/browse/HOSTEDCP-336 is fixed, switch to static approach.
Code in ./_clusterinfo_static

/cc @jcantrill


alanconway:
/close
This is a rebase of https://github.com/openshift/cluster-logging-operator/pull/2274 which needs work. Will rebase again when complete.

PR #2280, opened 2023-12-06
LOG-4445: Forwarder must use hosted clusterID on HCP cluster.

alanconway:
NOTE: Using unstructured objects for HCP because of dependency problems with the hypershift API package.
See https://issues.redhat.com/browse/HOSTEDCP-336.


alanconway:
/assign @jcantrill 
Started a new PR - rewritten & simplified as per your comments on https://github.com/openshift/cluster-logging-operator/pull/2274

PR #2284, opened 2023-12-08
LOG-4852: Vector collector Pods no longer picks up the log collector SAs Secret as a fallback

Clee2691:
### Description
This PR addresses forwarding to an internal `LokiStack` where in the legacy case, if the `pipelines.outputs` is not default it will not apply the `logcollector-token` secret's `token` and `ca.crt`. For a CLF named instance it will add the `logcollector` service account token to be used.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4852

### DIFF
--- a/internal/generator/vector/output/loki/loki.go
+++ b/internal/generator/vector/output/loki/loki.go
@@ -2,8 +2,9 @@ package loki
 
 import (
 	"fmt"
-	"github.com/openshift/cluster-logging-operator/internal/generator/vector/output"
 	"strings"
+
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/output"
 
 	"github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
 	"github.com/openshift/cluster-logging-operator/internal/generator/vector/normalize"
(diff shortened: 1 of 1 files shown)


internal/generator/vector/output/loki/loki.go
```diff
@@ -206,26 +207,33 @@ func Tenant(l *logging.Loki) Element {
...
 	if o.Secret != nil || (o.TLS != nil && o.TLS.InsecureSkipVerify) {
 
 		if tlsConf := security.GenerateTLSConf(o, secret, op, false); tlsConf != nil {
 			tlsConf.NeedsEnabled = false
-			return []Element{tlsConf}
+			conf = append(conf, tlsConf)
+		}
+	} else if secret != nil {
+		// Use secret of logcollector service account as backup
+		tlsConf := security.TLSConf{
```
jcantrill, line 229: I think the logic is missing something here because it's not using any of the helper functions.  
* What if my secret defines a CA to use?  
* What if I define a secret ... but I don't get any TLS Profile options here

Following "GenerateTLSConf" leads me to think it still nees to utilize that function... and then check the CA afterwards and set it like this if not defined

jcantrill (commented):
/hold cancel

PR #2285, opened 2023-12-08
LOG-4604: Add infrastructure annotations

jcantrill:
### Description
This PR:
* adds infrastructure annotations

@cahartma is it correct that we can claim aws token support?  We can use the tokens as provided by the CloudCredentialsOperator?

@syedriko is it correct we can claim FIPS compliance based upon the following definition? 
```
Whether the opperator accepts the [FIPS-140](https://issues.redhat.com/browse/FIPS-140) configuration of the underlying platform and works on nodes that are booted into FIPS mode. In this mode, the operator and any workloads it manages (operands) are solely calling the RHEL cryptographic library submitted for [FIPS-140](https://issues.redhat.com/browse/FIPS-140) validation.
```

### Links
* https://issues.redhat.com/browse/LOG-4604

### DIFF
--- a/config/manifests/bases/clusterlogging.clusterserviceversion.yaml
+++ b/config/manifests/bases/clusterlogging.clusterserviceversion.yaml
@@ -9,6 +9,16 @@ metadata:
     containerImage: quay.io/openshift-logging/cluster-logging-operator:latest
     description: The Red Hat OpenShift Logging Operator for OCP provides a means for
       configuring and managing your aggregated logging stack.
+    features.operators.openshift.io/cnf: "false"
+    features.operators.openshift.io/cni: "false"
+    features.operators.openshift.io/csi: "false"
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "true"
+    features.operators.openshift.io/proxy-aware: "true"
+    features.operators.openshift.io/tls-profiles: "true"
+    features.operators.openshift.io/token-auth-aws: "true"
+    features.operators.openshift.io/token-auth-azure: "false"
+    features.operators.openshift.io/token-auth-gcp: "false"
     olm.skipRange: '>=5.7.0-0 <5.9.0'
     operatorframework.io/cluster-monitoring: "true"
     operatorframework.io/suggested-namespace: openshift-logging
--- a/bundle/manifests/clusterlogging.clusterserviceversion.yaml
+++ b/bundle/manifests/clusterlogging.clusterserviceversion.yaml
@@ -121,6 +121,16 @@ metadata:
     createdAt: "2024-01-03T18:36:22Z"
     description: The Red Hat OpenShift Logging Operator for OCP provides a means for
       configuring and managing your aggregated logging stack.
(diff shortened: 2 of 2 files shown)


config/manifests/bases/clusterlogging.clusterserviceversion.yaml
```diff
@@ -9,6 +9,16 @@ metadata:
     containerImage: quay.io/openshift-logging/cluster-logging-operator:latest
     description: The Red Hat OpenShift Logging Operator for OCP provides a means for
       configuring and managing your aggregated logging stack.
+    features.operators.openshift.io/cnf: "false"
+    features.operators.openshift.io/cni: "false"
+    features.operators.openshift.io/csi: "false"
+    features.operators.openshift.io/disconnected: "true"
+    features.operators.openshift.io/fips-compliant: "true"
```
anpingli, line 16: I don't think we can claim fips-compliant=true without fips scan. 

According the latest scan result. we have to fix the issue below
```
logging-view-plugin-container | /opt/app-root/plugin-backend | go binary is not CGO_ENABLED 
log-file-metric-exporter-container | /usr/local/bin/log-file-metric-exporter | go binary has no build tags set (should have strictfipsruntime) 
cluster-logging-operator-container | /usr/bin/cluster-logging-operator | go binary has no build tags set (should have strictfipsruntime) | 
```
The company only provide a tool to scan golang and java based image.  No idea how to scan fluentd? No idea how to scan vector?

jcantrill: @anpingli  I would like to emphasize our criteria for true/false is based upon the provided definition of the annotation:
```
Whether the opperator accepts the [FIPS-140](https://issues.redhat.com/browse/FIPS-140) configuration of the underlying platform and works on nodes that are booted into FIPS mode. In this mode, the operator and any workloads it manages (operands) are solely calling the RHEL cryptographic library submitted for [FIPS-140](https://issues.redhat.com/browse/FIPS-140) validation
```
If scanning is the only way to confirm :arrow_up: then we need the scan, but I do not believe this is what it says

jcantrill:
continue to hold pending larger conversation of what we can properly claim

config/manifests/bases/clusterlogging.clusterserviceversion.yaml
cahartma: My understanding is this text accurately represents "our" definition, and needs to be set to true in order for our operator to show in the filtered catalog list.

syedriko: I also think we need to set this to true and work on the scan in the background if we need to.

anpingli: I think we should use fips-compliant=false before the scan issue to be fixed.  https://issues.redhat.com/browse/LOG-4886

anpingli: The QE/ART/Security team is planning to scan the operator according to fips-compliant.  it may block the release if scan failed when fips-compliant=true

PR #2286, opened 2023-12-12
LOG-4881: Vector collector Pods no longer picks up the log collector SAs Secret as a fallback

Clee2691:
### Description
This is a manual forward port of [LOG-4852](https://github.com/openshift/cluster-logging-operator/pull/2284).

Original description:

This PR addresses forwarding to an internal LokiStack where in the legacy case, if the `pipelines.outputs` is not `default` it will not apply the `logcollector-token` secret's `token` and `ca.crt`. For a CLF named instance it will add the `logcollector` service account token to be used.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4881

### DIFF
--- a/internal/generator/vector/output/loki/loki.go
+++ b/internal/generator/vector/output/loki/loki.go
@@ -2,9 +2,10 @@ package loki
 
 import (
 	"fmt"
+	"strings"
+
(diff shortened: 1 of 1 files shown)


jcantrill (commented):
/approve
/lgtm

PR #2287, opened 2023-12-13
LOG-4848: Enable TLS on tcp connections for rsyslog server

jlarriba:
### Description
This change enables TLS on the Vector aggregator Syslog source using the the same secret key names and meanings as all the other TLS  outputs we support.

/cc @syedriko 
/assign @jcantrill

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4848

### DIFF
--- a/internal/generator/vector/input/viaq_receiver_http_audit.toml
+++ b/internal/generator/vector/input/viaq_receiver_http_audit.toml
@@ -5,8 +5,8 @@ decoding.codec = "json"
 
 [sources.input_myreceiver.tls]
 enabled = true
-key_file = "/etc/collector/receiver/input_myreceiver/tls.key"
-crt_file = "/etc/collector/receiver/input_myreceiver/tls.crt"
+key_file = "/etc/collector/receiver/myreceiver/tls.key"
+crt_file = "/etc/collector/receiver/myreceiver/tls.crt"
 
 [transforms.input_myreceiver_split]
 type = "remap"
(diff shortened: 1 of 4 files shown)


internal/collector/collector.go
```diff
@@ -212,7 +213,8 @@ func (f *Factory) NewCollectorContainer(secretNames []string, clusterID string,
 	}
 	for _, receiverInput := range receiverInputs {
 		collector.VolumeMounts = append(collector.VolumeMounts,
-			v1.VolumeMount{Name: receiverInput, ReadOnly: true, MountPath: path.Join(receiverInputVolumePath, receiverInput)},
+			// make receivers compatible with the next generation conf generation
+			v1.VolumeMount{Name: receiverInput, ReadOnly: true, MountPath: path.Join(receiverInputVolumePath, strings.ReplaceAll("input-" + receiverInput, "-", "_"))},
```
jcantrill, line 217: Why does the mount path need to change here? input names are unique.  If it is because of the change in the wiring, then we should consider:
* modifying the wiring to use the name (this is the option I prefer)
* use the helper method `helpers.MakeInputID` to craft this method

jcantrill (commented):
/hold
/approve

internal/collector/collector.go
jlarriba: Absolutely right, that was very bad.

I have changed the wiring for both http (which was also failing with the new conf) and syslog receivers to use the name instead of the ID when searching for the certs directory.

PR #2289, opened 2023-12-18
LOG-4898: Move LokiStack reader ClusterRoles to OLM

xperimental:
### Description

Currently the cluster-roles for authorizing read access to a LokiStack running in openshift-logging tenancy mode are only reconciled by the CLO when a LokiStack is selected as the default store. This introduces issues when customers are still in the process of migrating from an existing ES configuration to a LokiStack configuration.

This PR moves the ClusterRole definitions into the bundle deployed by the OLM on installation of the CLO, making them available as soon as the operator is installed into the cluster.

Questions for review:
- Does it make sense to also move the ClusterRole/Binding used for _writing_ to the LokiStack to the OLM bundle? This would eliminate the need to reconcile these cluster-scoped resources completely and also remove the need for a finalizer.
- The issue is currently opened against 5.8, so we need a cherry-pick to that release. Does it make sense to backport this to previous releases as well to have the same solution for all versions?

### Links

- JIRA: LOG-4898


periklis:
> Questions for review:
> 
>     * Does it make sense to also move the ClusterRole/Binding used for _writing_ to the LokiStack to the OLM bundle? This would eliminate the need to reconcile these cluster-scoped resources completely and also remove the need for a finalizer.
>

Yes I suggest to do this and eliminate the need for an extra reconciliation loop. In case of accidental deletion I assume OLM will reconcile them back from the bundle. But this needs extra verification. 

>     * The issue is currently opened against 5.8, so we need a cherry-pick to that release. Does it make sense to backport this to previous releases as well to have the same solution for all versions?
> 

If we are going to make them all static I suggest to backport to `release-5.7` and `release-5.6`

xperimental:
PR has been updated to also move the ClusterRole / ClusterRoleBinding for the collector into the bundle deployed on operator installation. I have updated the reconciliation loop for the ClusterLogging resource to check for the existence of the finalizer we have used so far and remove it, if it exists.

> In case of accidental deletion I assume OLM will reconcile them back from the bundle.

This does not seem to be the case. When I deleted the ClusterRole for the collector as a test it was only recreated when I re-installed the Cluster Logging Operator. This can be done without any downtime of the logging stack though.

PR #2291, opened 2023-12-20
LOG-3651: [Vector] Add ability to change the Splunk's index within the ClusterLogForwarder

Clee2691:
### Description
This PR adds the ability for users to configure `vector` to define an `index` for the `Splunk` output dynamically using a `meta-data` key such as `kubernetes.namespace_name` or a single static value such as `my-index`.

They are captured in new fields on the `ClusterLogForwarder` when specifying the Splunk output, named `indexKey` and `indexName`. Only one of the fields can be defined at once.

/cc @vparfonov @cahartma 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-3651

### DIFF
--- a/apis/logging/v1/output_types.go
+++ b/apis/logging/v1/output_types.go
@@ -276,6 +276,21 @@ type Splunk struct {
 	// Should be a valid JSON object
 	// +optional
 	Fields []string `json:"fields,omitempty"`
+
+	// IndexKey is a meta-data key field to use to send events to.
+	// For example: 'IndexKey: kubernetes.namespace_name` will use the kubernetes
+	// namespace as the index.
+	// If the IndexKey is not found, the default index defined within Splunk is used.
+	// Only one of IndexKey or IndexName can be defined.
+	// If IndexKey && IndexName are not specified, the default index defined within Splunk is used.
+	// +optional
+	IndexKey string `json:"indexKey,omitempty"`
+
+	// IndexName is the name of the index to send events to.
+	// Only one of IndexKey or IndexName can be defined.
+	// If IndexKey && IndexName are not specified, the default index defined within Splunk is used.
+	// +optional
+	IndexName string `json:"indexName,omitempty"`
 }
 
 // Http provided configuration for sending json encoded logs to a generic http endpoint.
--- a/internal/generator/vector/output/splunk/splunk.go
+++ b/internal/generator/vector/output/splunk/splunk.go
@@ -2,6 +2,8 @@ package splunk
 
 import (
 	"fmt"
+	"strings"
+
 	"github.com/openshift/cluster-logging-operator/internal/constants"
 	. "github.com/openshift/cluster-logging-operator/internal/generator/framework"
 	"github.com/openshift/cluster-logging-operator/internal/generator/vector/normalize"
@@ -23,6 +25,7 @@ type Splunk struct {
 	Inputs       string
 	Endpoint     string
 	DefaultToken string
+	Index        Element
 }
 
 func (s Splunk) Name() string {
@@ -37,13 +40,15 @@ inputs = {{.Inputs}}
 endpoint = "{{.Endpoint}}"
 compression = "none"
 default_token = "{{.DefaultToken}}"
+{{kv .Index -}}
 timestamp_key = "@timestamp"
 {{end}}`
 }
 
 type SplunkEncoding struct {
-	ComponentID string
-	Codec       string
+	ComponentID  string
+	Codec        string
+	ExceptFields Element
 }
 
 func (se SplunkEncoding) Name() string {
@@ -54,6 +59,7 @@ func (se SplunkEncoding) Template() string {
 	return `{{define "` + se.Name() + `" -}}
 [sinks.{{.ComponentID}}.encoding]
 codec = {{.Codec}}
+{{kv .ExceptFields -}}
 {{end}}`
 }
 
@@ -63,10 +69,20 @@ func New(id string, o logging.OutputSpec, inputs []string, secret *corev1.Secret
 			Debug(id, vectorhelpers.MakeInputs(inputs...)),
 		}
 	}
+
+	componentID := vectorhelpers.MakeID(id, "add_splunk_index")
 	dedottedID := vectorhelpers.MakeID(id, "dedot")
+
+	dedotInputs := inputs
+	indexRemapElement := SetSplunkIndexRemap(o.Splunk, componentID, inputs)
+	if len(indexRemapElement) != 0 {
+		dedotInputs = []string{componentID}
+	}
+
 	return MergeElements(
+		indexRemapElement,
 		[]Element{
-			normalize.DedotLabels(dedottedID, inputs),
+			normalize.DedotLabels(dedottedID, dedotInputs),
 			Output(id, o, []string{dedottedID}, secret, op),
 			Encoding(id, o),
 			common.NewBuffer(id),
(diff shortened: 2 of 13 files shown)


apis/logging/v1/output_types.go
```diff
@@ -276,6 +276,21 @@ type Splunk struct {
 	// Should be a valid JSON object
 	// +optional
 	Fields []string `json:"fields,omitempty"`
+
+	// IndexKey is a meta-data key field to use to send events to.
+	// For example: 'IndexKey: kubernetes.namespace_name` will use the kubernetes
+	// namespace as the index.
+	// If the IndexKey is not found, the `log_type` will be used as the index.
```
jcantrill, line 283: Based upon the vector's [statement](https://vector.dev/docs/reference/configuration/sinks/splunk_hec_logs/#index) regarding "default", we need to walk back our intent to use "log_type" here.  This would break in an upgrade scenario

apis/logging/v1/output_types.go
```diff
@@ -276,6 +276,21 @@ type Splunk struct {
 	// Should be a valid JSON object
 	// +optional
 	Fields []string `json:"fields,omitempty"`
+
+	// IndexKey is a meta-data key field to use to send events to.
+	// For example: 'IndexKey: kubernetes.namespace_name` will use the kubernetes
```
jcantrill, line 281: Ideally, I think we should restrict these to a limited field set (e.g. kubernetes.namespace_name, kubernetes.pod_name) but I do not find anyplace where we do verification or additionally make that limitation so I guess it is consistent

apis/logging/v1/output_types.go
```diff
@@ -276,6 +276,21 @@ type Splunk struct {
...
+	// Only one of IndexKey or IndexName can be defined.
+	// If IndexKey && IndexName are not specified, the default index defined within Splunk is used.
+	// +optional
+	IndexKey string `json:"indexKey,omitempty"`
+
+	// IndexName is the name of the index to send events to.
+	// Only one of IndexKey or IndexName can be defined.
+	// If IndexKey && IndexName are not specified, the default index defined within Splunk is used.
+	// +optional
+	IndexName string `json:"indexName,omitempty"`
```
jcantrill, line 293: @alanconway this is consistent with our Elasticsearch output but I'm not certain it is needed.  We could embrace the notion of using curly brackets to differentiate dynamic versus static like `{{kubernetes.namespace_name}}` and have a single field.  

Do we leave as-is for now and revisit in the next version of the API?

internal/generator/vector/output/splunk/splunk.go
```diff
@@ -63,10 +67,20 @@ func New(id string, o logging.OutputSpec, inputs []string, secret *corev1.Secret
 			Debug(id, vectorhelpers.MakeInputs(inputs...)),
 		}
 	}
+
+	componentID := vectorhelpers.MakeID(id, "add_splunk_index")
 	dedottedID := vectorhelpers.MakeID(id, "dedot")
+
+	dedotInputs := inputs
+	indexRemapElement := SetSplunkIndexRemap(o.Splunk, componentID, inputs)
```
jcantrill, line 75: If we are to use a remap we should [exclude](https://vector.dev/docs/reference/configuration/sinks/splunk_hec_logs/#encoding.except_fields) the field in the encoding.

Additionally, do we know the behavior if the field is missing? Does the log get dropped?  Does it just use "splunk default" behavior or exclude the index in the submission?

apis/logging/v1/output_types.go
Clee2691: Currently if no `index` is specified, 
> the default index defined within Splunk is used.

The `default` index in Splunk is `main`. What I can do here is fallback to `main` as the index if the specified key isn't found. Will that work?

internal/generator/vector/output/splunk/splunk.go
Clee2691: The only time a log get's indexed in splunk's `default` index is if the sink omits index.
I use remap so that I can check if an IndexKey is missing from the log, I can then default it to something else as I asked about above. Instead of defaulting to `log_type` we can default it to splunk's `default`

apis/logging/v1/output_types.go
jcantrill: @Clee2691 I would leave it unset completely since it appears to me you can define the default index on the splunk side.  Setting it to `main` on the client locks us in and does not meet backwards compatibility since we never set it previously

internal/generator/vector/output/splunk/splunk.go
jcantrill: The field which I suggested to omit is `write_index` which is the one created by this function.  Per my previous comment for backwards compatibility, I would set nothing

apis/logging/v1/output_types.go
alanconway: I like the idea of using {{}} instead of separate Key+Name fields, but we should make the same change to other places we have Key/Name pairs like this. We could make the change backwards compat now by having
- Index: index string, may include field templates such as {{kuberntes.namespace_name}}
- IndexName: literal index string
- IndexKey: name of a field to use as the index string. Note: `indexKey: "foo"` is equivalent to `index:{{foo}}`
and deprecate IndexName/IndexKey for removal in v2 APIs.

jcantrill: > we should make the same change to other places

I think Elasticsearch 'structuredKey' and 'structuredName' is the only other pairing.  We could introduce 'index' there with the same convention you identified, excluding internally managed ES

apis/logging/v1/output_types.go
Clee2691: After some testing, it seems that I can set the index as an empty string (`""`) which will allow logs to be sent to the default index. 

This is because if the key is not present in the log record and nothing is set, vector complains with:
```
Failed to render template for "index". error=Missing fields on event: ["write_index"] error_type="template_failed" stage="processing" internal_log_rate_limit=true
```
since I am using an artificial field, `.write_index`, to index logs.

apis/logging/v1/output_types.go
Clee2691: If using `Index`, if it includes a field that is not available what should the default behavior be?

For example,
I define this as an index: `{{kubernetes.field}}_customIndex`
Log 1 has `kubernetes.field = foo` but log 2 doesn't include `kubernetes.field`

Log 1's index will now be `foo_customIndex` but what should log 2 be?
1. Invalid? Write to `default` splunk index?
2. Omit only `kubernetes.field` and write to `_customIndex`?

jcantrill: > Log 1's index will now be foo_customIndex but what should log 2 be?
>
> Invalid? Write to default splunk index?
>2.  Omit only kubernetes.field and write to _customIndex?


@alanconway @cahartma I lean towards documenting absence implies `""` and the result is as suggested by 2.  It would be path of least surprise.  Additionally, IMO, defining a custom index in this way implies a schema; you know something about your log data and want it organized consistently.  This question is fundamentally the issue we have always had with log data not having any consistency.  

I can imagine this issue will pop up when:
* you send journal/audit logs through the same pipeline as container

Clee2691: As discussed, using an index with templating will be lined up for the next release

apis/logging/v1/output_types.go
jcantrill: Do we need to update the comment to be backwards compatible?

Clee2691: Yes good catch

PR #2300, opened 2024-01-09
LOG-4929: do not allow to choose protocol for syslog receiver

jlarriba:
### Description
QA found an issue when udp was selected and tls was automatically applied. tls cannot work on udp protocol, so this was causing confusion. After a conversation, we decided that the udp protocol case is not required at the moment because OpenStack (the first user of the syslog receiver) will always send the logs via tls-encrypted tcp.

This PR removes the option to configure a protocol for the syslog receiver, as it will always use tcp.

/cc @jcantrill 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4929

### DIFF
--- a/apis/logging/v1/input_receiver_types.go
+++ b/apis/logging/v1/input_receiver_types.go
@@ -54,10 +54,4 @@ type SyslogReceiver struct {
 	// +kubebuilder:validation:Maximum:=65535
 	// +optional
 	Port int32 `json:"port"`
-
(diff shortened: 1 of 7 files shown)


jcantrill:
@jlarriba please run `make docs` and it then :lgtm

jlarriba:
> @jlarriba please run `make docs` and it then :lgtm

`make docs` does not modify any file, so I think we are good to go

PR #2304, opened 2024-01-09
LOG-4779: Deploy collector as a deployment instead of daemonset

Clee2691:
### Description
This is the ported version of the PR originally for `release-5.8` (https://github.com/openshift/cluster-logging-operator/pull/2240) to `master` (release-5.9).

Added an `E2E` test.

#### Description from PR#2240
This PR deals with the scenario in which a user wishes to specify the collector as an HTTP receiver as the sole input option.

Originally the collector was deployed as a `daemonset`. However, with this change, when all pipelines specified in the `ClusterLogForwarder` (CLF) reference an 'HTTP' receiver as their sole input reference, the collector will be deployed as a `deployment`.

This addresses the following:

The HCP usecase that utilizes one collector per NS for collecting audit logs via a webhook
Saves deployment resources
Collector does not need to mount the host file system given it only receives logs via the receiver

#### Extra
This also fixes a bug when defining `HTTP` input receivers in multi-CLF. Multiple CLFs with the same `HTTP` input receiver name will endlessly reconcile the same service because the names are the same. The fix for this was the make HTTP input service names unique with the format of `<CLF-NAME>-<input.Name>`.
Related Jira: https://issues.redhat.com/browse/LOG-5009

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- Github Old PR: https://github.com/openshift/cluster-logging-operator/pull/2240
- JIRA: https://issues.redhat.com/browse/LOG-4779

### DIFF
--- a/internal/generator/vector/input/viaq.go
+++ b/internal/generator/vector/input/viaq.go
@@ -2,15 +2,17 @@ package input
 
 import (
 	"fmt"
+	"strings"
+
 	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
 	"github.com/openshift/cluster-logging-operator/internal/constants"
+	"github.com/openshift/cluster-logging-operator/internal/factory"
 	"github.com/openshift/cluster-logging-operator/internal/generator/framework"
 	"github.com/openshift/cluster-logging-operator/internal/generator/vector/elements"
 	"github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
 	"github.com/openshift/cluster-logging-operator/internal/generator/vector/normalize"
 	"github.com/openshift/cluster-logging-operator/internal/generator/vector/source"
 	"github.com/openshift/cluster-logging-operator/internal/utils/sets"
-	"strings"
 )
 
 const (
@@ -39,7 +41,7 @@ var (
 
 // NewViaQ creates an input adapter to generate config for ViaQ sources to collect logs excluding the
 // collector container logs from the namespace where the collector is deployed
-func NewViaQ(input logging.InputSpec, collectorNS string, op framework.Options) ([]framework.Element, []string) {
+func NewViaQ(input logging.InputSpec, collectorNS string, resNames *factory.ForwarderResourceNames, op framework.Options) ([]framework.Element, []string) {
 	els := []framework.Element{}
 	ids := []string{}
 	switch {
@@ -105,7 +107,7 @@ func NewViaQ(input logging.InputSpec, collectorNS string, op framework.Options) 
 				ids = append(ids, cids...)
 			}
 		} else if input.Receiver != nil {
-			els, ids = NewViaqReceiverSource(input, op)
+			els, ids = NewViaqReceiverSource(input, resNames, op)
 		}
 	}
 	els, ids = addLogType(input, els, ids)
@@ -125,6 +127,7 @@ func addLogType(spec logging.InputSpec, els []framework.Element, ids []string) (
 
 	if logType != "" {
 		id := helpers.MakeInputID(spec.Name, "viaq", "logtype")
+
 		remap := elements.Remap{
 			Desc:        `Set log_type`,
 			ComponentID: id,
--- a/internal/generator/vector/source/http_receiver.go
+++ b/internal/generator/vector/source/http_receiver.go
@@ -1,15 +1,16 @@
 package source
 
 import (
+	"strings"
+
 	configv1 "github.com/openshift/api/config/v1"
 	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
 	"github.com/openshift/cluster-logging-operator/internal/generator/framework"
 	"github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
 	"github.com/openshift/cluster-logging-operator/internal/tls"
-	"strings"
 )
 
-func NewHttpSource(id string, input logging.InputSpec, op framework.Options) framework.Element {
+func NewHttpSource(id, inputName string, input logging.InputSpec, op framework.Options) framework.Element {
 	var minTlsVersion, cipherSuites string
 	if _, ok := op[framework.ClusterTLSProfileSpec]; ok {
 		tlsProfileSpec := op[framework.ClusterTLSProfileSpec].(configv1.TLSProfileSpec)
@@ -18,7 +19,7 @@ func NewHttpSource(id string, input logging.InputSpec, op framework.Options) fra
 	}
 	return HttpReceiver{
 		ID:            id,
-		InputName:     input.Name,
+		InputName:     inputName,
 		ListenAddress: helpers.ListenOnAllLocalInterfacesAddress(),
 		ListenPort:    input.Receiver.HTTP.Port,
 		Format:        input.Receiver.HTTP.Format,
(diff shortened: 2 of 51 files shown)


internal/generator/vector/input/viaq.go
```diff
@@ -124,7 +126,14 @@ func addLogType(spec logging.InputSpec, els []framework.Element, ids []string) (
 	}
 
 	if logType != "" {
-		id := helpers.MakeInputID(spec.Name, "viaq", "logtype")
+		compId := spec.Name
```
jcantrill, line 129: Why did we need to make this change?  I don't see how the ID is no longer unique which is the primary issue

internal/generator/vector/source/http_receiver.go
```diff
@@ -1,15 +1,16 @@
...
 	configv1 "github.com/openshift/api/config/v1"
 	logging "github.com/openshift/cluster-logging-operator/apis/logging/v1"
 	"github.com/openshift/cluster-logging-operator/internal/generator/framework"
 	"github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
 	"github.com/openshift/cluster-logging-operator/internal/tls"
-	"strings"
 )
 
-func NewHttpSource(id string, input logging.InputSpec, op framework.Options) framework.Element {
+func NewHttpSource(id, inputName string, input logging.InputSpec, op framework.Options) framework.Element {
```
jcantrill, line 13: I don't recall since this is a subtle modification of my original code, but why would a "source" have an input name?  Looking at the template it should be the name of the secret associated with the input.  I presume the user defines the secret name.  But we should IMO:

* choose a better variable name
* confirm the mount path.  It may be we do some name mangling for the volume/volumeMounts

internal/reconcile/deployment.go
```diff
@@ -0,0 +1,58 @@
...
+		if err := k8Client.Get(context.TODO(), key, current); err != nil {
+			if errors.IsNotFound(err) {
+				reason = constants.EventReasonCreateObject
+				return k8Client.Create(context.TODO(), desired)
+			}
+			return fmt.Errorf("failed to get %v Deployment: %w", key, err)
+		}
+		same := false
+
+		if same, updateReason = pod.AreSame(&current.Spec.Template.Spec, &desired.Spec.Template.Spec, current.Name); same && util.HasSameOwner(current.OwnerReferences, desired.OwnerReferences) {
```
jcantrill, line 36: This is missing something as it only compares the template spec but Deployments and Daemonsets have other differences we may wish to control, specifically replicas and labels?  @alanconway can you weigh in here.  The way this is implemented admins control this field which is a departure from other objects, but "replicacount" would mean we would need to provide some other mechanism of control

internal/reconcile/daemon_sets.go
```diff
@@ -33,8 +33,8 @@ func DaemonSet(er record.EventRecorder, k8Client client.Client, desired *apps.Da
 		}
 		same := false
 
-		if same, updateReason = daemonsets.AreSame(current, desired); same && util.HasSameOwner(current.OwnerReferences, desired.OwnerReferences) {
-			log.V(3).Info("DaemonSet are the same skipping update")
+		if same, updateReason = pod.AreSame(&current.Spec.Template.Spec, &desired.Spec.Template.Spec, current.Name); same && util.HasSameOwner(current.OwnerReferences, desired.OwnerReferences) {
```
jcantrill, line 36: similar ask here as deployments.  We are now possibly missing items we wish to keep managed that are specific to a daemonset

internal/reconcile/service.go
```diff
@@ -33,10 +32,8 @@ func Service(er record.EventRecorder, k8Client client.Client, desired *corev1.Se
 		same := false
 
 		if same, updateReason = services.AreSame(current, desired); same {
-			if utils.HasSameOwner(current.OwnerReferences, desired.OwnerReferences) {
```
jcantrill, line 36: As this already being checked?

internal/collector/daemonset.go
```diff
@@ -25,7 +25,7 @@ func (f *Factory) ReconcileDaemonset(er record.EventRecorder, k8sClient client.C
 	var receiverInputs []string
 	for _, input := range f.ForwarderSpec.Inputs {
 		if logging.IsHttpReceiver(&input) || logging.IsSyslogReceiver(&input) {
-			receiverInputs = append(receiverInputs, input.Name)
+			receiverInputs = append(receiverInputs, f.ResourceNames.GenerateInputServiceName(input.Name))
```
jcantrill, line 28: I have another question further down but not certain why this change is needed?

internal/generator/vector/input/viaq.go
Clee2691: This change was made because of an issue with input services. Input service names were not unique and would clash with multiple forwarders which would cause an endless reconcile loop if 2 forwarders had the same http input name.

internal/generator/vector/source/http_receiver.go
Clee2691: This source is in reference to an HTTP receiver. The input name is the name used when defining the http receiver input. Something like this CLF config,
```
spec:
  inputs:
    - name: input-http1
      receiver:
        http: 
          format: kubeAPIAudit
          receiverPort: 
            name: httpserver1
            port: 443
            targetPort: 8081
```
would have an inputName of `input-http1`.

The mount paths are for the secret that the HTTP receiver service creates. Since I am now making unique names using `resources.GenerateInputServiceName()`, the paths need to reflect the new generated name.

internal/reconcile/deployment.go
Clee2691: There is no regression in what we had before. We were spinning through the pods and comparing them. There were no other specs specific to daemonsets that were compared.

See: https://github.com/openshift/cluster-logging-operator/pull/2240#discussion_r1385685305 also https://github.com/openshift/cluster-logging-operator/pull/2240#discussion_r1388502398

internal/reconcile/daemon_sets.go
Clee2691: See comment above

internal/reconcile/service.go
Clee2691: Indeed

internal/collector/daemonset.go
Clee2691: See above comments about receiver input names and services: https://github.com/openshift/cluster-logging-operator/pull/2304#discussion_r1459489323

internal/generator/vector/input/viaq.go
jcantrill: This is an issue across multiple CLF and needs to be addressed during:
* service creation
* maybe any secrets/configmaps we create

The actual config has a 1:1 relationship with the CLF and is only visible to the collector.  Am I missing why we need to propagate those changes to the config?

Clee2691: I see what you mean and I think you are correct here. I will make the appropriate changes.

internal/reconcile/deployment.go
alanconway: Since we are updating the labels below, I think we should be comparing them above - we should compare the same set of values that you want to update.

Clee2691: I've added back the daemonset comparator and added a deployment comparator. These compare the pod specs along with labels and owners.

PR #2306, opened 2024-01-15
LOG-4898: Deploy LokiStack read roles when console is active

xperimental:
### Description

Currently the cluster-roles for authorizing read access to a LokiStack running in openshift-logging tenancy mode are only reconciled by the CLO when a LokiStack is selected as the default store. This introduces issues when customers are still in the process of migrating from an existing ES configuration to a LokiStack configuration.

This PR changes the handling of the RBAC roles and bindings used for LokiStack support:

- It splits the roles into a "read" and a "write" group
  - "write" roles are deployed based on the setting of "default log storage", like all the roles used to be before this change
  - "read" roles are deployed based on whether the "logging console plugin" is deployed
- The finalizer used for safe-guarding removal of the roles is removed from the code
  - There is new code present that will remove the finalizer, if present, from the `ClusterLogging` instance

Questions for review:

- The removal of the finalizer is an additional change.
  - [x] Does this make sense to do?
  - [x] If yes, is it okay to do it within this PR or does it need another one?
- The issue is currently opened against 5.8, so we need a cherry-pick to that release.
  - ~~Does it make sense to backport this to previous releases as well to have the same solution for all versions?~~

### Links

- Supersedes: #2289 
- JIRA: [LOG-4898](https://issues.redhat.com//browse/LOG-4898)


xperimental:
Taking a look at the CI failures ... :mag: 

I also found an issue with the current code on deletion of a ES-based deployment with the console enabled: the "read roles" persist, because the cleanup code is not called.

Feel free to review the other parts / general idea already.

jcantrill:
> Questions for review:
>
>The removal of the finalizer is an additional change.
> Does this make sense to do?
> If yes, is it okay to do it within this PR or does it need another one?

I think its acceptable to do here in this PR.  My recollection is we generally had issues with it anyway so I don't see any reason to exclude it.  Without looking at the code I don't imagine there is any firm need to introduce an additional PR

> The issue is currently opened against 5.8, so we need a cherry-pick to that release. 
 Does it make sense to backport this to previous releases as well to have the same solution for all versions?

In the interest of backporting less I would say no.  We prefer users to upgrade.  Is there any functional change that is noticeable by a user that warrants the need to back port?

periklis (commented):
LGTM

Issue #2312, opened 2024-01-18
Nil pointer exception with clusterLogging.Spec.Collection empty and Forwarder enabled


andreaskaris:
Nil pointer exception with clusterLogging.Spec.Collection empty and Forwarder enabled.

The following combination throws a nil pointer exception:

ClusterLogging:
```
apiVersion: "logging.openshift.io/v1"
kind: "ClusterLogging"
metadata:
  name: "instance" 
  namespace: openshift-logging
spec:
  managementState: "Managed"
```

ClusterLogForwarder:
```
apiVersion: logging.openshift.io/v1
kind: ClusterLogForwarder
metadata:
  name: instance
  namespace: openshift-logging
spec:
  pipelines:
   - inputRefs:
     - audit
     outputRefs:
     - loki-external
     labels:
       job: openshift-audit
       clusterid: sno10
   - inputRefs:
     - application
     outputRefs:
     - loki-external
     labels:
       job: openshift-application
       clusterid: sno10
   - inputRefs:
     - infrastructure
     outputRefs:
     - loki-external
     labels:
       job: openshift-infrastructure
       clusterid: sno10
  outputs:
  - name: loki-external
    type: loki
    url: http://192.168.18.75:310
```

```
[akaris@workstation logging]$ oc apply -f configuring-cluster-logging-minimal-minimal.yaml
clusterlogging.logging.openshift.io/instance created
[akaris@workstation logging]$ oc apply -f configuring-log-forwarder.yaml
clusterlogforwarder.logging.openshift.io/instance created
[akaris@workstation logging]$ 
```

```
$ LOG_LEVEL=3 make run
(...)
{"_ts":"2024-01-18T11:18:45.623084752Z","_file:line":"internal/reconcile/rbac.go:37","_level":"3","_component":"cluster-logging-operator","_message":"Roles are the same skipping update"}
{"_ts":"2024-01-18T11:18:45.623126639Z","_file:line":"internal/reconcile/rbac.go:70","_level":"3","_component":"cluster-logging-operator","_message":"RoleBindings are the same skipping update"}
{"_ts":"2024-01-18T11:18:45.623159759Z","_file:line":"internal/k8shandler/clusterloggingrequest.go:107","_level":"3","_component":"cluster-logging-operator","_message":"Getting object","namespacedName":{"Namespace":"openshift-logging","Name":"logcollector-token"},"object":{"metadata":{"creationTimestamp":null}}}
[...]

PR #2313, opened 2024-01-18
Catch nil ptr exceptions with enabled Forwarder and empty Collection

andreaskaris:
When the Forwarder is enabled, make sure that
clusterLogging.Spec.Collection is non nil to avoid nil pointer exceptions from occurring.

Fixes https://github.com/openshift/cluster-logging-operator/issues/2312
Fixes https://github.com/openshift/cluster-logging-operator/issues/2314
Fixes https://github.com/openshift/cluster-logging-operator/issues/2315

### Links
* https://issues.redhat.com/browse/LOG-5006
* https://issues.redhat.com/browse/LOG-5007

### DIFF
--- a/internal/k8s/loader/load.go
+++ b/internal/k8s/loader/load.go
@@ -64,7 +64,8 @@ func FetchClusterLogForwarder(k8sClient client.Client, namespace, name string, i
 	forwarder.Spec, extras, migrationMessages = migrations.MigrateClusterLogForwarder(namespace, name, forwarder.Spec, fetchClusterLogging().Spec.LogStore, extras, internalLogStoreSecret, saTokenSecret)
 	setMigrationStatusConditions(&forwarder.Status, migrationMessages)
 
-	extras[constants.ClusterLoggingAvailable] = (fetchClusterLogging().Name != "")
+	clusterLogging := fetchClusterLogging()
+	extras[constants.ClusterLoggingAvailable] = (clusterLogging.Name != "" && clusterLogging.Spec.Collection != nil)
 	extras[constants.VectorName] = false
 	if fetchClusterLogging().Spec.Collection != nil {
 		extras[constants.VectorName] = fetchClusterLogging().Spec.Collection.Type == logging.LogCollectionTypeVector
--- a/internal/validations/clusterlogforwarder/validate_clusterlogging_dependency.go
+++ b/internal/validations/clusterlogforwarder/validate_clusterlogging_dependency.go
@@ -7,9 +7,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	ValidateClusterLoggingDependencyMSG = "is dependent on a ClusterLogging instance with a valid spec.collector configuration"
+)
+
 func ValidateClusterLoggingDependency(clf loggingv1.ClusterLogForwarder, k8sClient client.Client, extras map[string]bool) (error, *loggingv1.ClusterLogForwarderStatus) {
 	if clf.Name == constants.SingletonName && clf.Namespace == constants.OpenshiftNS && !extras[constants.ClusterLoggingAvailable] {
-		return errors.NewValidationError("ClusterLogForwarder '%s/%s' is dependent on a ClusterLogging instance", constants.OpenshiftNS, constants.SingletonName), nil
+		return errors.NewValidationError("ClusterLogForwarder '%s/%s' %s", constants.OpenshiftNS, constants.SingletonName, ValidateClusterLoggingDependencyMSG), nil
 	}
 	return nil, nil
 }
--- a/test/client/clf.go
+++ b/test/client/clf.go
@@ -2,9 +2,12 @@ package client
 
 import (
 	"fmt"
+	"strings"
 
 	loggingv1 "github.com/openshift/cluster-logging-operator/api/logging/v1"
+	clfvalidation "github.com/openshift/cluster-logging-operator/internal/validations/clusterlogforwarder"
 	"github.com/openshift/cluster-logging-operator/test"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/watch"
 )
 
@@ -19,4 +22,26 @@ func ClusterLogForwarderReady(e watch.Event) (bool, error) {
 	default:
 		return false, nil
 	}
+}
+
+// ClusterLogForwarderValidationFailure expects condition type "Validation" to be set on the ClusterLogForwarder
+// resource. If no such condition can be found, it returns false, and a nil error (so that c.WaitFor can wait until
+// the condition is set, or time out if the condition is never set). If the condition is set, we expect its message
+// to match clfvalidation.ValidateClusterLoggingDependencyMSG and we expect it to be "True". We also expect the "Ready"
+// condition to be "False". In that case, we return true and no error. In case of the contrary, we return false and an
+// error.
+func ClusterLogForwarderValidationFailure(e watch.Event) (bool, error) {
+	clf := e.Object.(*loggingv1.ClusterLogForwarder)
+	cond := clf.Status.Conditions
+
+	validationCondition := cond.GetCondition(loggingv1.ValidationCondition)
+	if validationCondition == nil {
+		return false, nil
+	}
+
+	if strings.Contains(validationCondition.Message, clfvalidation.ValidateClusterLoggingDependencyMSG) &&
+		validationCondition.Status == v1.ConditionTrue && cond.IsFalseFor(loggingv1.ConditionReady) {
+		return true, nil
+	}
+	return false, fmt.Errorf("ClusterLogForwarder unexpected condition: %v", test.YAMLString(clf.Status))
 }
--- a/Makefile
+++ b/Makefile
@@ -156,7 +156,7 @@ run:
 	WATCH_NAMESPACE=$(NAMESPACE) \
 	KUBERNETES_CONFIG=$(KUBECONFIG) \
 	WORKING_DIR=$(CURDIR)/tmp \
-	$(RUN_CMD) main.go
+	$(RUN_CMD) cmd/main.go
 
 .PHONY: run-debug
 run-debug:
--- a/api/logging/v1/cluster_log_forwarder.go
+++ b/api/logging/v1/cluster_log_forwarder.go
@@ -5,6 +5,7 @@ import (
 	"strings"
 
 	"github.com/openshift/cluster-logging-operator/internal/constants"
+	"github.com/openshift/cluster-logging-operator/internal/status"
 	"github.com/openshift/cluster-logging-operator/internal/utils/sets"
 )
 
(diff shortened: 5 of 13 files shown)


internal/k8s/loader/load.go
```diff
@@ -64,7 +64,8 @@ func FetchClusterLogForwarder(k8sClient client.Client, namespace, name string, i
 	forwarder.Spec, extras, migrationMessages = migrations.MigrateClusterLogForwarder(namespace, name, forwarder.Spec, fetchClusterLogging().Spec.LogStore, extras, internalLogStoreSecret, saTokenSecret)
 	setMigrationStatusConditions(&forwarder.Status, migrationMessages)
 
-	extras[constants.ClusterLoggingAvailable] = (fetchClusterLogging().Name != "")
+	clusterLogging := fetchClusterLogging()
+	extras[constants.ClusterLoggingAvailable] = (clusterLogging.Name != "" && clusterLogging.Spec.Collection != nil)
```
jcantrill, line 68: I can't wrap my head around this exactly but my expectation is:

* ns:openshift-logging, name:instance - Require collection stanza to be provided by admin
* ns:other, name:other - Optional collection stanza to be provided by admin

The general flow is: load, migrate, validate.  I believe this change still allows an admin to deploy a forwarder without explicitly specifying a ClusterLogging object but I want to be certain.  I believe we have a test to verify and if we don't I would expect this PR to add one.

controllers/forwarding/forwarding_controller.go
```diff
@@ -190,3 +190,18 @@ func (r *ReconcileForwarder) SetupWithManager(mgr ctrl.Manager) error {
 		Owns(&v1.ServiceMonitor{}).
 		Complete(r)
 }
+
+func SetForwardingControllerStatus(clf *logging.ClusterLogForwarder, newStatus *logging.ClusterLogForwarderStatus) {
```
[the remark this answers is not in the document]
andreaskaris: yeah nowhere, that was a leftover of an old revision, sorry

internal/k8s/loader/load.go
andreaskaris: > The general flow is: load, migrate, validate. I believe this change still allows an admin to deploy a forwarder without explicitly specifying a ClusterLogging object but I want to be certain.

When I create a ClusterLogForwarder without a ClusterLogging object (in my tests), the operator doesn't deploy any resources. I'll dig into this, but given that my exposure to / knowledge of the operator is a bit limited .. What's the expected outcome of just deploying the ClusterLogForwarder resource, without a ClusterLogging object?

apis/logging/v1/conditions.go
```diff
@@ -91,6 +91,26 @@ func (nc NamedConditions) IsAllReady() bool {
 	return true
 }
 
+// Synchronize synchronizes the current NamedCondition with a new NamedCondition.
+// This is not the same as simply replacing the NamedCondition: Conditions contain the LastTransitionTime
+// field which is left unmodified by Synchronize.
+func (nc NamedConditions) Synchronize(newNamedCondition NamedConditions) {
```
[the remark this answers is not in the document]
andreaskaris: will do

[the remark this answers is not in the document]
andreaskaris: same here - I'll add a unit test to make 100% sure (that's why the PR is still WIP btw)

apis/logging/v1/cluster_log_forwarder.go
```diff
@@ -66,7 +66,7 @@ func (status ClusterLogForwarderStatus) GetReadyConditionMessages() []string {
 }
 
 // IsReady returns true if all of the subordinate conditions are ready.
-func (status ClusterLogForwarderStatus) IsReady() bool {
+func (status *ClusterLogForwarderStatus) IsReady() bool {
```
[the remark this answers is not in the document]
andreaskaris: TL;DR: consistency, 1 interface for the type

The Synchronize method needs to be defined on a pointer receiver, so I switched the type of the other methods, too, to have one interface for the type (instead of 2 interfaces)

I learned that it's bad practice to define both pointer and value receivers for the same type and only ever use value receivers for small, immutable types:
https://go.dev/tour/methods/8
https://go.dev/doc/faq#different_method_sets

However, now after googling a bit more, it doesn't seem to be that black and white, after all. That said, there are lots of people advising to default to pointer receivers. Plus, converting the other 2 methods to pointers avoids us implicitly defining 2 interfaces for the same type.

apis/logging/v1/conditions.go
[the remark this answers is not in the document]
andreaskaris: done for unit test, I still have to add en E2E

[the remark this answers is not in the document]
andreaskaris: done for unit test, I still have to add en E2E

apis/logging/v1/cluster_log_forwarder.go
```diff
@@ -77,6 +77,24 @@ func (status ClusterLogForwarderStatus) IsReady() bool {
 	return true
 }
 
+// Synchronize synchronizes the current Status with a new Status.
+// This is not the same as simply replacing the Status: Conditions contain the LastTransitionTime
+// field which is left unmodified by Synchronize.
+func (status *ClusterLogForwarderStatus) Synchronize(newStatus *ClusterLogForwarderStatus) {
```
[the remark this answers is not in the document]
andreaskaris: done for unit test, I still have to add en E2E

apis/logging/v1/cluster_log_forwarder.go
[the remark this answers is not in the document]
andreaskaris: I went with your suggestion and left these 2 methods as they were

apis/logging/v1/cluster_log_forwarder.go
```diff
@@ -49,7 +49,7 @@ func IsFilterTypeName(s string) bool { return typeHasField(reflect.TypeOf(Filter
 
 // Get all subordinate condition messages for condition of type "Ready" and False
 // A 'true' Ready condition with a message means some error with pipeline but it is still valid
-func (status ClusterLogForwarderStatus) GetReadyConditionMessages() []string {
+func (status *ClusterLogForwarderStatus) GetReadyConditionMessages() []string {
```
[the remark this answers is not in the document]
andreaskaris: fixed, I went with your suggestion and left these 2 methods as they were

internal/k8s/loader/load.go
andreaskaris: I'll definitely add an E2E test for this

apis/logging/v1/clusterlogforwarder_types_test.go
```diff
@@ -151,3 +154,385 @@ var _ = Describe("ClusterLogForwarder", func() {
...
+
+		original.Synchronize(&syncTarget)
+		Expect(conditionEquals(original.Conditions, expected.Conditions)).NotTo(HaveOccurred())
+		Expect(namedConditionEquals(original.Inputs, expected.Inputs)).NotTo(HaveOccurred())
+	})
+})
+
+// namedConditionEquals verifies if 2 NamedConditions can be considered the same. It uses conditionEquals to check
+// for equality of conditions.
+func namedConditionEquals(got, expected NamedConditions) error {
```
jcantrill, line 484: There are condition matchers in /test/matchers.  Please review to see if they cover the needs here or can be adapted accordingly

apis/logging/v1/clusterlogforwarder_types_test.go
```diff
@@ -151,3 +154,385 @@ var _ = Describe("ClusterLogForwarder", func() {
   }`))
 	})
 })
+
+var _ = Describe("ClusterLogForwarderStatus", func() {
+	It("synchronizes conditions ClusterLogForwarderStatus correctly", func() {
```
jcantrill, line 159: Looks like we expect no update to occur so maybe:

```
Describe("ClusterLogForwarderStatus#Synchronize", func() {
	It("should not update any status when no conditions change", func() {
```

internal/k8s/loader/load.go
jcantrill: > When I create a ClusterLogForwarder without a ClusterLogging object (in my tests), the operator doesn't deploy any resources

This should only be the case for when ClusterLogForwarder(ns:openshift,name:instance) exists with no similarly matched NS/name ClusterLogging instances which is the legacy case. All other combinations CLF name/ns are supposed to deploy a collector without needing a ClusterLogging to define the collector. That is realized by [this](https://github.com/openshift/cluster-logging-operator/blob/master/controllers/forwarding/forwarding_controller.go#L147-L158) code.  This [e2e](https://github.com/openshift/cluster-logging-operator/blob/master/hack/testing-olm/test-1343-minimal-multi-clf.sh#L64) test should verify that behavior

apis/logging/v1/clusterlogforwarder_types_test.go
andreaskaris: done

apis/logging/v1/clusterlogforwarder_types_test.go
andreaskaris: see the comments a bit further down, it actually add and removes elements, all while also testing the timestamps

internal/k8s/loader/load.go
andreaskaris: Alright, got it:
```
LOG_LEVEL=10 make run NAMESPACE=""
```

```
#!/bin/bash

cat <<'EOF' | oc apply -f -
apiVersion: v1
kind: Namespace
metadata:
  name: log-forwarder
EOF

oc create sa log-forwarder -n log-forwarder
oc adm policy add-cluster-role-to-user collect-application-logs system:serviceaccount:log-forwarder:log-forwarder
oc adm policy add-cluster-role-to-user collect-infrastructure-logs system:serviceaccount:log-forwarder:log-forwarder
oc adm policy add-cluster-role-to-user collect-audit-logs system:serviceaccount:log-forwarder:log-forwarder

cat <<'EOF' | oc apply -f -
kind: ClusterLogForwarder
metadata:
  name: log-forwarder
  namespace: log-forwarder
spec:
  serviceAccountName: log-forwarder
  pipelines:
   - inputRefs:
     - audit
     outputRefs:
     - loki-external
     labels:
       job: openshift-audit
       clusterid: sno10
   - inputRefs:
     - application
     outputRefs:
     - loki-external
     labels:
       job: openshift-application
       clusterid: sno10
   - inputRefs:
     - infrastructure
     outputRefs:
     - loki-external
     labels:
       job: openshift-infrastructure
       clusterid: sno10
  outputs:
  - name: loki-external
    type: loki
    url: http://192.168.18.75:3100
EOF
```

That still deploys the forwarder.

andreaskaris: Interesting, so all 3 issues that I found affect the legacy implementation, including the infinite loop for the status

andreaskaris:
note: I'm still working on this, I got pulled into some partner cases and will resume work as soon as I get some time

andreaskaris:
@jcantrill  wdyt now? thanks!

internal/validations/clusterlogforwarder/validate_clusterlogging_dependency.go
```diff
@@ -9,7 +9,7 @@ import (
 
 func ValidateClusterLoggingDependency(clf loggingv1.ClusterLogForwarder, k8sClient client.Client, extras map[string]bool) (error, *loggingv1.ClusterLogForwarderStatus) {
 	if clf.Name == constants.SingletonName && clf.Namespace == constants.OpenshiftNS && !extras[constants.ClusterLoggingAvailable] {
-		return errors.NewValidationError("ClusterLogForwarder '%s/%s' is dependent on a ClusterLogging instance", constants.OpenshiftNS, constants.SingletonName), nil
+		return errors.NewValidationError("ClusterLogForwarder '%s/%s' is dependent on a ClusterLogging instance with a valid Collector configuration", constants.OpenshiftNS, constants.SingletonName), nil
```
jcantrill, line 12: lets say... with a valid spec.collector

maybe makes it more obvious where exactly to fix your CL

test/client/clf.go
```diff
@@ -2,12 +2,18 @@ package client
...
+	"strings"
 
 	loggingv1 "github.com/openshift/cluster-logging-operator/apis/logging/v1"
 	"github.com/openshift/cluster-logging-operator/test"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/watch"
 )
 
+const (
+	validationFailureMsg = "is dependent on a ClusterLogging instance with a valid Collector configuration"
```
jcantrill, line 14: Can this be moved into our validators as public constant and consumed here?

test/client/clf.go
```diff
@@ -20,3 +26,24 @@ func ClusterLogForwarderReady(e watch.Event) (bool, error) {
...
 	}
 }
+
+// ClusterLogForwarderValidationFailure expects condition type "Validation" to be set on the ClusterLogForwarder
+// resource. If no such condition can be found, it returns false, and a nil error (so that c.WaitFor can wait until
+// the condition is set, or time out if the condition is never set). If the condition is set, we expect its message
+// to match validationFailureMsg and we expect it to be "True". We also expect the "Ready" condition to be "False".
+// In that case, we return true and no error. In case of the contrary, we return false and an error.
+func ClusterLogForwarderValidationFailure(e watch.Event) (bool, error) {
+	clf := e.Object.(*loggingv1.ClusterLogForwarder)
```
jcantrill, line 36: Do we need a type check here? Is it possible to receive an object that is not a CLF?

jcantrill (commented):
small questions otherwise lgtm

test/client/clf.go
jcantrill: @andreaskaris I wasn't suggesting it does not need to be cast but that seems like this is unsafe and we need to type check.  The first line just assumes it is a CLF, but isn't there a chance it could be something else?

andreaskaris: ah!   `e.Object.(*loggingv1.ClusterLogForwarder)`  is a go type assertion though, so IMO it's safe?
[`https://go.dev/tour/methods/](https://go.dev/tour/methods/15) "If i does not hold a T, the statement will trigger a panic. "
As in, if someone passes something else, this will panic on purpose

Are we on the same page now, or am I still misunderstanding you? Anyway, I'm still polishing some of the other code, I'll move this back from draft to non-draft when ready

jcantrill:
@andreaskaris please squash otherwise lgtm

andreaskaris:
> @andreaskaris please squash otherwise lgtm

Done, I moved the tests to their respective commits

jcantrill:
Any idea if this issue effects earlier releases?  

Thank you for your cotribution

andreaskaris:
Yeah, I think that the nil pointer exception at least bit me on an earlier one like 5.8. I have a low prio task open to verify logging configuration for a partner, as soon as I get back to this, I'd be able to test

Also, if you look here: https://github.com/openshift/cluster-logging-operator/issues/2312#issuecomment-1931847559
There's someone else who complained about the nil pointer exception, for 5.8

PR #2327, opened 2024-01-28
First cut of v2 API, still some work to do: see doc.go for notes.

alanconway:
/cc @jcantrill
/cc @Clee2691


apis/logging/v2beta/doc.go
```diff
@@ -0,0 +1,53 @@
+// package v2beta custom resorces for the cluster logging operator.
+//
+// NOTE: This API is under construction and subject to change.
+//
+// NOTE: Doc comments are written using JSON names (lowerCase) rather than Go UpperCase stlye.
```
jcantrill, line 5: ..style (sp)

apis/logging/v2beta/doc.go
```diff
@@ -0,0 +1,53 @@
...
+
+   A cluster-scoped ClsterLogForwarder needs a namespace for:
+   - serviceaccount
+   - configmap
+   - secrets
+   - deployments, daemonset, pods etc.
+
+   The namespace field is pulled to the top of the CLF spec since it applies to more than the SA.
+
+   Namespace ownership/lifecycle? Can the CLF own the namepsace?
```
jcantrill, line 28: I don't see why we need to own the namespace

apis/logging/v2beta/doc.go
```diff
@@ -0,0 +1,53 @@
...
+   If the SA is provided by the user, then we don't want to delete the namespace (and therefore the SA) when the forwarder is deleted.
```
jcantrill, line 29: We are proposing the CLF to be a cluster wide resource which means we need to define a separate entity to be the owner of the various namespaced objects we create.  We had a similar issue in a previous release where kube does not like NS objects tied to clusterwide objects.

apis/logging/v2beta/doc.go
```diff
@@ -0,0 +1,53 @@
...
+
+   Given the tight association between CLF and this namespace, should we reconsider making it cluster-scoped?
```
jcantrill, line 31: Alternatively we could drop the notion of "ClusterLogForwarder" for "LogForwarder" but it seems awkward to continue to allow the namespace object access to cluster wide logs.  Is there any precedence for the construct we are proposing?

apis/logging/v2beta/doc.go
```diff
@@ -0,0 +1,53 @@
...
+
+   # Globs vs. regexps
+
+   Suggest using regexps anywhere we want to include a pattern, more flexible, avoid confusion over
+   where to use glob vs. regexp. (Need to verify this isn't a problem for implementation.)
```
jcantrill, line 36: This is mostly problematic in our input selectors for paths, etc.  I do not believe the vector impl supports regex which means we would need to make changes.  Would it not make sense to start with globs?  Aren't they mostly a subset of regex?

alanconway:
@jcantrill @Clee2691  @cahartma @periklis  I've incorporated the performance tuning stuff and ideas from recent discussion, I'm starting to feel good about this. I have not updated the perf tuning enhancments let's refine the ideas on this PR and I'll backport to the enhancement when we're done.

apis/logging/v2beta/output_types.go
```diff
@@ -0,0 +1,316 @@
...
+	LogID string `json:"logId,omitempty"`
+}
+
+// Splunk Deliver log data to Splunk’s HTTP Event Collector
+// Provides optional extra properties for `type: splunk_hec` ('splunk_hec_logs' after Vector 0.23
+type Splunk struct {
+	// Fields to be added to Splunk index. https://docs.splunk.com/Documentation/Splunk/8.0.0/Data/IFXandHEC
+	// Should be a valid JSON object
+	// +optional
+	Fields []string `json:"fields,omitempty"`
```
cahartma, line 299: I will be updating Fields description and type in the next day or so.   Here are the other fields that are missing:

	// IndexKey is a meta-data key field to use to send events to.
	// For example: 'IndexKey: kubernetes.namespace_name` will use the kubernetes
	// namespace as the index.
	// If the IndexKey is not found, the default index defined within Splunk is used.
	// Only one of IndexKey or IndexName can be defined.
	// If IndexKey && IndexName are not specified, the default index defined within Splunk is used.
	// +optional
	IndexKey string `json:"indexKey,omitempty

	// IndexName is the name of the index to send events to.
	// Only one of IndexKey or IndexName can be defined.
	// If IndexKey && IndexName are not specified, the default index defined within Splunk is used.
	// +optional
	IndexName string `json:"indexName,omitempty"

apis/logging/v2beta/doc.go
alanconway: Yep, I"m not clear on how this works.

apis/logging/v2beta/doc.go
alanconway: Neiter do I, I'm just wondering.

apis/logging/v2beta/doc.go
alanconway: Need to find out - there must be cluster-wide resources that create Pods etc. so there must be one somewhere.

apis/logging/v2beta/output_types.go
alanconway: The parts of the v2 interface that aren't changing since v1 will need to be re-synced before we ship. @jcantrill move fun with multiple API versions....

apis/logging/v2beta/cluster_log_forwarder_types.go
```diff
@@ -0,0 +1,384 @@
...
+
+// ClusterLogForwarderStatus defines the observed state of ClusterLogForwarder
+type ClusterLogForwarderStatus struct {
+	// Conditions of the log forwarder.
+	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="Forwarder Conditions",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:forwarderConditions"}
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
+
+	// Inputs maps input name to condition of the input.
+	//+operator-sdk:csv:customresourcedefinitions:type=status,displayName="Input Conditions",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:inputConditions"}
+	Inputs map[string]metav1.Condition `json:"inputs,omitempty"`
```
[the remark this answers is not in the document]
alanconway: I don't have a strong opinion either, I'd be happy to make that change if its simpler, more usable, or a more common pattern.

apis/logging/v2beta/cluster_log_forwarder_types.go
```diff
@@ -0,0 +1,384 @@
...
+
+// InputTuningSpec provides tuning parameters for an input.
+type InputTuningSpec struct {
+	// RateLimitByContainer maps container names to a rate limit in records-per-second for containers
+	// with that name.
+	RateLimitByContainer map[string]int `json:"rateLimitByContainer"`
+
+	// RateLimitByNodePath maps node log file path names to a rate limit in records-per-second-per node
+	// for node logs from that path.
+	RateLimitByNodePath map[string]int `json:"rateLimitByNodePath"`
```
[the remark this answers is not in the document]
alanconway: Agreed, and agreed we can drop it initially if we don't come up with a clean solution. I started to think about node logs and limiting and realized we also need some way to control them, but it can wait for a later version - it's a new feature.

apis/logging/v2beta/doc.go
alanconway: Functional subset but not upwards compatible e.g. `a.*` is regexp `a\..*`. Globs for paths is probably fine if that's what vector offers.

apis/logging/v2beta/cluster_log_forwarder_types.go
```diff
@@ -0,0 +1,384 @@
+/*
```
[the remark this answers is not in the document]
alanconway: Good question - that's what I originally called it, but I switched to "beta" because it seems like we won't have a long trial period before GA so this will be "almost final". I am OK with either.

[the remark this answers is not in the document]
periklis: @alanconway I do not recall the link, but you need to call it `v2beta1` then. K8S conventions require numbering for alpha and beta too.

jcantrill:
Replaced by https://github.com/openshift/cluster-logging-operator/pull/2451

PR #2332, opened 2024-01-31
LOG-4943: Enhance Infrastructure permissions to support application log collection

Clee2691:
### Description
This PR adds a permission validation for ClusterLogForwarderss that spec an application input with infrastructure namespaces.

Infrastructure namespaces are:

1. `default`
2. `openshift*`
3. `kube*`

In addition, since infra namespaces were excluded by default, this changes the behavior when a user includes an infra namespace by removing it from the exclude list.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4943


internal/generator/vector/input/viaq_application_with_infra_includes_excludes.toml
```diff
@@ -0,0 +1,78 @@
+# Logs from containers (including openshift containers)
+[sources.input_my_app_container]
+type = "kubernetes_logs"
+max_read_bytes = 3145728
+glob_minimum_cooldown_ms = 15000
+auto_partial_merge = true
+include_paths_glob_patterns = ["/var/log/pods/test-ns-foo_*/*/*.log", "/var/log/pods/openshift-logging_*/*/*.log", "/var/log/pods/kube-apiserver_*/*/*.log"]
+exclude_paths_glob_patterns = ["/var/log/pods/default_*/*/*.log", "/var/log/pods/test-ns1_*/*/*.log", "/var/log/pods/*/*/*.gz", "/var/log/pods/*/*/*.tmp"]
```
jcantrill, line 8: You mentioned to me at one point about the 'gz' and 'tmp' file here.  Given we explicitly say gather 'log' files I wonder if there is any need to actually add these excludes?  I can see deferring any change until after this PR lands

internal/generator/vector/input/viaq_test.go
```diff
@@ -79,6 +79,33 @@ var _ = Describe("inputs", func() {
...
+		Entry("with an application that specs infra namespaces and excludes infra namespaces", logging.InputSpec{
+			Name: "my-app",
+			Application: &logging.Application{
+				Namespaces:        []string{"test-ns-foo", "openshift*", "kube-apiserver"},
+				ExcludeNamespaces: []string{"test-ns1", "openshift-logging"},
+			},
+		},
+			"viaq_application_with_infra_includes_infra_excludes.toml",
+		),
+		Entry("with an application that specs specific infra namespace and excludes infra namespaces", logging.InputSpec{
```
jcantrill, line 100: This is an interesting test case because I think we will get this one alot:  "why do I not get logs from my openshift-logging" namespace..... its because the excludes take precedence to the includes

internal/validations/clusterlogforwarder/inputs/validate_test.go
```diff
@@ -409,53 +409,4 @@ var _ = Describe("#Validate", func() {
 			Expect(clfStatus.Inputs["custom-app-group-limit"]).To(HaveCondition(loggingv1.ValidationCondition, true, loggingv1.ValidationFailureReason, "cannot have a negative limit"))
 		})
 	})
-	Context("when validating application input types", func() {
```
jcantrill, line 412: Why are these test being removed? They are verifying admins improperly defining a namespace glob

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account_test.go
```diff
@@ -217,6 +218,84 @@ var _ = Describe("[internal][validations] validate clusterlogforwarder permissio
 			}
 			Expect(ValidateServiceAccount(customClf, k8sClient, extras)).To(Succeed())
 		})
+
+		Context("when evaluating custom application inputs that spec infrastructure namespaces", func() {
```
jcantrill, line 222: Where is our success test?

internal/validations/clusterlogforwarder/inputs/validate_test.go
Clee2691: I found that these tests were duplicated and were already defined here: https://github.com/openshift/cluster-logging-operator/blob/4023485a589db9952b8b39ea557ab3d63aa67b37/internal/validations/clusterlogforwarder/inputs/validate_test.go#L127
Just a little clean up.

internal/generator/vector/input/viaq.go
```diff
@@ -182,3 +195,37 @@ func NewViaqContainerSource(spec logging.InputSpec, namespace, includes, exclude
...
+				foundInfraNamespaces.Insert(fmt.Sprintf("%d", i))
+			}
+		}
+	}
+
+	infraNSSet := sets.NewString(infraNS...)
+	// Remove infra namespace depending on the capture group
+	for _, ns := range foundInfraNamespaces.List() {
+		switch ns {
+		case "1":
```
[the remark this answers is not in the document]
Clee2691: Changed with named capture groups. Good idea here to use it!

internal/generator/vector/input/viaq_application_with_infra_includes_excludes.toml
Clee2691: In my understanding, if `include_paths_glob_patterns` is defined, we don't really need to define any other excludes. 
In this specific `TOML`, it will include the 3 logs defined and then it will exclude from that include list if that makes sense.

From the [Vector docs](https://vector.dev/docs/reference/configuration/sources/file/#exclude)
> [excludes] Takes precedence over the include option. Note: The exclude patterns are applied after the attempt to glob everything in include. This means that all files are first matched by include and then filtered by the exclude patterns. This can be impactful if include contains directories with contents that are not accessible.

It doesn't change anything here if we keep them in.

internal/generator/vector/input/viaq_test.go
Clee2691: Basically from the [Vector docs](https://vector.dev/docs/reference/configuration/sources/file/#exclude)

> [excludes] Takes precedence over the include option. Note: The exclude patterns are applied after the attempt to glob everything in include. This means that all files are first matched by include and then filtered by the exclude patterns. This can be impactful if include contains directories with contents that are not accessible.

PR #2333, opened 2024-01-31
Revertlog5009

jcantrill:
### Description
revert #2332  to give more time for consumption, but I don't see an alternative


anpingli:
Confirm with SRE team, they haven't started to use httpserver. We can close this PR. and accept fix.

PR #2213, opened 2023-10-16
LOG-4606:Forwarding logs to the Azure Monitor Logs

vparfonov:
### Description

_**For Vector collector only**_

Azure Monitor Logs is a comprehensive service provided by Microsoft Azure that enables the collection, analysis, and actioning of telemetry data across various Azure and on-premises resources. It facilitates advanced analytics, offering profound insights into the performance and health of applications and workloads running on Azure.

In this update, we aim to enhance our product's capabilities by enabling log forwarding to Azure Monitor. By supporting this functionality, our users will be able to seamlessly integrate their application logs with Azure Monitor, leveraging its powerful analytics and monitoring capabilities. This integration will empower users to streamline their monitoring and troubleshooting processes, leading to better operational insights, faster issue resolution, and improved overall performance of their Azure-based applications and services.

This PR proposes an extension to the existing API by introducing a new output for log storage in Azure Monitor logs. 
Here is a new output type: 
```
type AzureMonitor struct {
	//CustomerId che unique identifier for the Log Analytics workspace.
	//https://learn.microsoft.com/en-us/azure/azure-monitor/logs/data-collector-api?tabs=powershell#request-uri-parameters
	CustomerId string `json:"customerId,omitempty"`

	//LogType the record type of the data that is being submitted.
	//Can only contain letters, numbers, and underscores (_), and may not exceed 100 characters.
	//https://learn.microsoft.com/en-us/azure/azure-monitor/logs/data-collector-api?tabs=powershell#request-headers
	LogType string `json:"logType,omitempty"`

	//AzureResourceId the Resource ID of the Azure resource the data should be associated with.
	//https://learn.microsoft.com/en-us/azure/azure-monitor/logs/data-collector-api?tabs=powershell#request-headers
	// +optional
	AzureResourceId string `json:"azureResourceId,omitempty"`

	//Host alternative host for dedicated Azure regions. (for now need only for Chine region)
	//https://docs.azure.cn/en-us/articles/guidance/developerdifferences#check-endpoints-in-azure
	// +optional
	Host string `json:"host,omitempty"`
}
```
The `CustomerId` and `LogType`  fields are required, while the `AzureResourceId`  and `Host`  field is optional.

- _CustomerId_ : The unique identifier for the Log Analytics workspace.
[...]

PR #2334, opened 2024-02-02
LOG-4606: Forwarding logs to the Azure Monitor Logs

vparfonov:

### Description
This pull request addresses adding functional tests for testing log forwarding to Azure Monitor Log.

**Changes:**
- Implemented functional tests using Mockoon, a tool for describing environments via JSON files.
- Mockoon has capabilities to define rules for checking HTTP requests and sending responses accordingly.
- Proposed rules to validate the `log-type` header and body fields in each record (`message` and `log_type`) to ensure that log records are properly formatted, e.g.:
```
"rules": [
            {
              "target": "header",
              "modifier": "log-type",
              "value": "myLogType",
              "invert": false,
              "operator": "equals"
            },
            {
              "target": "body",
              "modifier": "$.[0].message",
              "value": "This is my test message",
              "invert": false,
              "operator": "equals"
            },
            {
              "target": "body",
              "modifier": "$.[0].log_type",
              "value": "application",
              "invert": false,
              "operator": "equals"
            }
          ],
```
  
- Set up a mocking environment within a Pod and made it available via a Route to emulate the original API URI format
- Emulate the original API URI format (`https://<CustomerId>.<Host>/api/logs?api-version=2016-04-01`), including the `<CustomerID>` and `<Host>` components, to accurately replicate the production environment

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s): https://github.com/openshift/cluster-logging-operator/pull/2213
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5030
- Enhancement proposal:


test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
```diff
@@ -0,0 +1,141 @@
...
+		mockServer *azuremonitor.MockServer
+		secret     *v1.Secret
+		sharedKey  = rand.Word(16)
+		inputName  = "custom-app"
+	)
+
+	BeforeEach(func() {
+		framework = functional.NewCollectorFunctionalFrameworkUsingCollector(logging.LogCollectionTypeVector)
+		// Start mockServer server
+		mockServer = azuremonitor.NewMockServer(framework.Namespace, azuremonitor.Mockoon)
```
jcantrill, line 33: Why can this test not rely upon the output container being added to the pod of the functional test?  It is just a container and there are benefits to all the containers running in the same pod.  If we really can not us the functional framework we should move this test to rely upon the e2e framework but I suspect its possible with the functional one

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
```diff
@@ -0,0 +1,141 @@
...
+		)
+
+		framework.Secrets = append(framework.Secrets, secret)
+	})
+
+	AfterEach(func() {
+		framework.Cleanup()
+	})
+
+	It("should accept application logs", func() {
```
jcantrill, line 94: Is there no place for us to read the received logs from the mock to check them in stead of only checking the collector logs?

test/helpers/azuremonitor/azure-http-data-collector-api.json
```diff
@@ -0,0 +1,258 @@
+{
```
jcantrill, line 1: Maybe add a doc.go to this directory.  I assume this is a rule file for the mock.  It would be useful if you could highlight the relevent bits and how they work in providing a response

test/helpers/azuremonitor/mockServer.go
```diff
@@ -0,0 +1,154 @@
...
+	"golang.org/x/sync/errgroup"
+	corev1 "k8s.io/api/core/v1"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+//go:embed azure-http-data-collector-api.json
+var AzureHttpDataCollectorApi string
+
+const (
+	image         = "mockoon/cli:latest"
```
jcantrill, line 32: lets tag a version of this and push it to quay.io/openshift-logging so we can ensure it will not change from under us

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
vparfonov: no, we can only check logs on mock server side with rules and send back corresponding response

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
vparfonov: The problem is: i need to know public URL of Route, before deploy collector, that's why i did it in this way. 
See my comment here: [forward_to_azuremonitor_test.go#L43](https://github.com/vparfonov/cluster-logging-operator/blob/ca8bfbe8c047156105447c6dc92289afce3d625c/test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go#L43).
I will try to find way how do it in one pod 

@jcantrill

test/helpers/azuremonitor/azure-http-data-collector-api.json
vparfonov: added

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
jcantrill: We do we need a public URL?  I thought there was a place to set the host?  Can we not craft the URL from "localhost" and a port?

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
jcantrill: What about file logging? https://mockoon.com/docs/latest/logging-and-recording/requests-logging/#file-logging

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
jcantrill: Another thing we could do is to modify the framework to mount  the `/etc/host` file into the collector container.  This would provide a way to resolve a name without needing a route

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
vparfonov: Nice idea, i will try. I didn't think in this way. Thanks

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
vparfonov: Maybe we can do something with /etc/host, at the end URL must have format: https://<some-id>.<host>/

test/helpers/azuremonitor/mockServer.go
vparfonov: done

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
vparfonov: done

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
vparfonov: merged to one pod

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
```diff
@@ -0,0 +1,194 @@
...
+		framework  *functional.CollectorFunctionalFramework
+		mockServer *azuremonitor.MockoonServer
+		secret     *v1.Secret
+		sharedKey  = rand.Word(16)
+		inputName  = "custom-app"
+	)
+
+	BeforeEach(func() {
+		framework = functional.NewCollectorFunctionalFrameworkUsingCollector(logging.LogCollectionTypeVector)
+		mockServer = azuremonitor.NewMockoonServer(framework.Namespace, framework.Labels)
```
jcantrill, line 46: Looks like we are closer and the reason I suggested to modify `/etc/hosts` if needed to make this work.  With our functional framework it is only necessary to add a container to the existing pod.  You likely will not require any of the other existing support resources other then maybe configmaps.  Additionally this structure allows you to share a disk if needed between the collector and receiver.

One should be able to add a visitor that adds in the bits needed for the container

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
```diff
@@ -0,0 +1,194 @@
...
+		// HTTP Data Collector API URI	has format
+		// https://<CustomerId>.<Host>/api/logs?api-version=2016-04-01
+		// <CustomerID>	The unique identifier for the Log Analytics workspace.
+		// <Host> alternative host for dedicated Azure regions.
+		// We will split route host by dot (.) and use first part as customer id, second part as alternative host
+		// this allow as emulate forwarding
+		hostname, err := getServerHost()
+		Expect(err).To(BeNil())
+		subDomain := strings.SplitN(hostname, ".", 2)
+		framework.Forwarder.Spec.Outputs = append(framework.Forwarder.Spec.Outputs,
```
jcantrill, line 58: nit. There are "builder" functions that may make this easier to read but not necessary

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
```diff
@@ -0,0 +1,194 @@
...
+		Expect(framework.WriteMessagesToApplicationLog(applicationLogLine, 3)).To(BeNil())
+		time.Sleep(30 * time.Second)
+
+		//read log from collector container
+		collectorLog, err := framework.ReadCollectorLogs()
+		Expect(err).To(BeNil())
+		Expect(strings.Count(collectorLog, failedReason)).To(BeEquivalentTo(0))
+
+		//read log from mock server container and validate it
+		mockoonLogs := readLogFromMockoon(framework)
```
jcantrill, line 148: Can we filter out the server logs from the "received" logs here so the expectation is only looking for ingested log messages?

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
```diff
@@ -0,0 +1,194 @@
...
+		Expect(mockoonLogs).ToNot(BeNil())
+		Expect(len(mockoonLogs)).To(BeEquivalentTo(2))
+		Expect(mockoonLogs[0].Message).To(BeEquivalentTo(serverStarted))
+		Expect(mockoonLogs[1].RequestMethod).To(BeEquivalentTo(http.MethodPost))
```
jcantrill, line 152: Is this important for azure?  Do we allow users to alter the METHOD in our api?

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
```diff
@@ -0,0 +1,194 @@
...
+		Expect(mockoonLogs[1].Message).To(BeEquivalentTo(transactionRecorded))
```
jcantrill, line 153: Same here.  I'm not certain this is important to us

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
```diff
@@ -0,0 +1,194 @@
...
+		Expect(mockoonLogs[1].Transaction.Request.Body).ToNot(BeEmpty())
+		var appLogs []types.ApplicationLog
+		err = types.ParseLogsFrom(utils.ToJsonLogs([]string{mockoonLogs[1].Transaction.Request.Body}), &appLogs, false)
+		Expect(err).To(BeNil())
+		Expect(len(appLogs)).To(BeEquivalentTo(3))
+		for i := 0; i < 3; i++ {
+			validateAppLog(appLogs[i], message, framework)
```
jcantrill, line 160: Consider using templates like [these](https://github.com/openshift/cluster-logging-operator/blob/release-5.8/test/functional/normalization/message_format_test.go#L41) to validate the entire message format

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
```diff
@@ -0,0 +1,194 @@
...
+		Expect(mockoonLogs[1].Message).To(BeEquivalentTo(transactionRecorded))
+		Expect(mockoonLogs[1].Transaction.Request.Body).ToNot(BeEmpty())
+		var appLogs []types.ApplicationLog
+		err = types.ParseLogsFrom(utils.ToJsonLogs([]string{mockoonLogs[1].Transaction.Request.Body}), &appLogs, false)
+		Expect(err).To(BeNil())
+		Expect(len(appLogs)).To(BeEquivalentTo(1))
+		validateAppLog(appLogs[0], message, framework)
+	})
+
+	It("should accept batch application logs", func() {
```
jcantrill, line 134: Is this different from the other test?  Are we not always sending logs as a "batch"?  I think likely we can either drop the other test or this one unless there is compelling reason to keep both

test/helpers/azuremonitor/mockoonServer.go
```diff
@@ -0,0 +1,140 @@
...
+	container *v1.Container
+	route     *openshiftv1.Route
+	Timeout   time.Duration
+}
+
+// NewMockoonServer creates a MockoonServer
+// Server will be created using Mockoon tools, specific environment described in azure-http-data-collector-api.json
+func NewMockoonServer(ns string, labels map[string]string) *MockoonServer {
+	ms := &MockoonServer{
+		service: runtime.NewService(ns, Mockoon),
```
jcantrill, line 80: I think you will find moving these unnecessary by moving into container that runs within the context of the functional pod

test/helpers/azuremonitor/mockoonServer.go
```diff
@@ -0,0 +1,140 @@
...
+		route:   runtime.NewRoute(ns, Mockoon, Mockoon, Mockoon),
+		container: &v1.Container{
+			Name:  Mockoon,
+			Image: image,
+			Ports: []v1.ContainerPort{{Name: Mockoon, ContainerPort: port}},
+			Args: []string{
+				"--data=/data/azure-http-data-collector-api.json",
+				"--log-transaction",
+			},
+			SecurityContext: &v1.SecurityContext{
```
jcantrill, line 90: Do you need a special context for this test if we write to ephemeral storage?

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
vparfonov: METHOD is not configured parameter, so 'yes' we can drop this check

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
vparfonov: looks like not, CLI except this parameter: 
`-t, --log-transaction       Log the full HTTP transaction (request and response)`

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
vparfonov: done

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
vparfonov: done

test/functional/outputs/azuremonitor/forward_to_azuremonitor_test.go
jcantrill: My statement of "filter" was more in the direction of we control what we return from the "read" function.  Within that function we can return only the bits that are of interest to us.  The following lines already do some of this logic by separating receiver logs from the logs we are interested.

jcantrill (commented):
This is much closer to what I would expect.  Good work

PR #2337, opened 2024-02-05
LOG-5002: input.receiver.syslog should be an optional field

jlarriba:
### Description
This is a change mandated by concision in the yaml, as we had a lot of "syslog" mandatory definitions. Due to the limitations on kubebuilder:default (it does not work on fields whose "parent" is empty), we had to go back to the original "GetPort()" funcs.

The change was requested for syslog receiver, but the necessary changes were also implemented in http receiver, as it would not have made any sense to have "syslog" optional but "http" mandatory.

I have also used this change to assign the IsSyslogReceiver, IsHttpReceiver and IsAuditHttpReceiver funcs to the Receiver struct, as it they "belong" naturally to a Receiver.

/cc @jcantrill @anpingli 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5002

### DIFF
--- a/apis/logging/v1/cluster_log_forwarder.go
+++ b/apis/logging/v1/cluster_log_forwarder.go
@@ -4,6 +4,7 @@ import (
 	"reflect"
 	"strings"
 
+	"github.com/openshift/cluster-logging-operator/internal/constants"
 	"github.com/openshift/cluster-logging-operator/internal/utils/sets"
 )
 
@@ -196,19 +197,42 @@ func (output *OutputSpec) GetMaxRecordsPerSecond() int64 {
 	return output.Limit.MaxRecordsPerSecond
 }
 
-func IsAuditHttpReceiver(input *InputSpec) bool {
-	return input.Receiver != nil &&
-		input.Receiver.HTTP != nil &&
-		input.Receiver.Type == ReceiverTypeHttp &&
-		input.Receiver.HTTP.Format == FormatKubeAPIAudit
+func (receiver *ReceiverSpec) IsAuditHttpReceiver() bool {
+	return receiver != nil &&
+		receiver.Type == ReceiverTypeHttp &&
+		receiver.GetHTTPFormat() == FormatKubeAPIAudit
 }
 
-func IsHttpReceiver(input *InputSpec) bool {
-	return input.Receiver != nil &&
-		input.Receiver.Type == ReceiverTypeHttp
+func (receiver *ReceiverSpec) IsHttpReceiver() bool {
+	return receiver != nil &&
+		receiver.Type == ReceiverTypeHttp
 }
 
-func IsSyslogReceiver(input *InputSpec) bool {
-	return input.Receiver != nil &&
-		input.Receiver.Type == ReceiverTypeSyslog
+func (receiver *ReceiverSpec) IsSyslogReceiver() bool {
+	return receiver != nil &&
+		receiver.Type == ReceiverTypeSyslog
+}
+
+func (receiver *ReceiverSpec) GetSyslogPort() (ret int32) {
+	ret = constants.SyslogReceiverPort
+	if receiver != nil && receiver.ReceiverTypeSpec != nil && receiver.Syslog != nil && receiver.Syslog.Port != 0 {
+		ret = receiver.Syslog.Port
+	}
+	return
+}
+
+func (receiver *ReceiverSpec) GetHTTPPort() (ret int32) {
+	ret = constants.HTTPReceiverPort
+	if receiver != nil && receiver.ReceiverTypeSpec != nil && receiver.HTTP != nil && receiver.HTTP.Port != 0 {
+		ret = receiver.HTTP.Port
+	}
+	return
+}
+
+func (receiver *ReceiverSpec) GetHTTPFormat() (ret string) {
+	ret = constants.HTTPFormat
+	if receiver != nil && receiver.ReceiverTypeSpec != nil && receiver.HTTP != nil && receiver.HTTP.Format != FormatKubeAPIAudit {
+		ret = receiver.HTTP.Format
+	}
+	return
 }
--- a/internal/collector/collector.go
+++ b/internal/collector/collector.go
@@ -249,12 +249,12 @@ func (f *Factory) ReconcileInputServices(er record.EventRecorder, k8sClient clie
 	for _, input := range f.ForwarderSpec.Inputs {
 		var listenPort int32
 		serviceName := f.ResourceNames.GenerateInputServiceName(input.Name)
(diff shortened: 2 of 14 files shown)


apis/logging/v1/cluster_log_forwarder.go
```diff
@@ -196,19 +197,42 @@ func (output *OutputSpec) GetMaxRecordsPerSecond() int64 {
...
-		input.Receiver.Type == ReceiverTypeHttp
+func (receiver *ReceiverSpec) IsHttpReceiver() bool {
+	return receiver != nil &&
+		receiver.Type == ReceiverTypeHttp
 }
 
-func IsSyslogReceiver(input *InputSpec) bool {
-	return input.Receiver != nil &&
-		input.Receiver.Type == ReceiverTypeSyslog
+func (receiver *ReceiverSpec) IsSyslogReceiver() bool {
```
jcantrill, line 211: Probably a nit but do we really need these or the port functions? We already have type which can be used to get the port from the correct subfield

internal/collector/collector.go
```diff
@@ -249,12 +249,12 @@ func (f *Factory) ReconcileInputServices(er record.EventRecorder, k8sClient clie
 	for _, input := range f.ForwarderSpec.Inputs {
 		var listenPort int32
 		serviceName := f.ResourceNames.GenerateInputServiceName(input.Name)
-		if input.Receiver != nil && input.Receiver.ReceiverTypeSpec != nil {
-			if logging.IsHttpReceiver(&input) {
-				listenPort = input.Receiver.HTTP.Port
+		if input.Receiver != nil {
```
jcantrill, line 252: Following up on the need for the functions... this is prime usage for a 'switch' on the type

internal/collector/daemonset.go
```diff
@@ -24,7 +23,7 @@ func (f *Factory) ReconcileDaemonset(er record.EventRecorder, k8sClient client.C
 
 	var receiverInputs []string
 	for _, input := range f.ForwarderSpec.Inputs {
-		if logging.IsHttpReceiver(&input) || logging.IsSyslogReceiver(&input) {
+		if input.Receiver.IsHttpReceiver() || input.Receiver.IsSyslogReceiver() {
```
jcantrill, line 26: here, we only have 2 kinds of receivers and you must specify a non nil receiver and type or atleast the sub spec, do we really need the function check too.  We can enforce spec requirement in our internal/validators.... if receiver is non nil then you must specify one of the other

internal/collector/collector.go
jlarriba: It definitively could, but we have changed so many times how do we know whether is something an HttpReceiver or a SyslogReceiver that I want to be a little defensive here. I think it is manageable with two receivers without a switch. May be a good candidate for a switch if the number of receiver types grow.

internal/collector/daemonset.go
jlarriba: That would check a different thing. Here, we want to make sure that the Receiver is http or syslog. Other receiver types might do something different. I feel comfortable with this specificity.

jlarriba: If we would like to be less specific, this could be replaced with `input.Receiver != nil` because our current validators+migration combination make sure that whether we have a Receiver, we will have a receiverType. I prefer the way it is now, though.

jcantrill: I think its reasonable to only check for nil.  I presume if we implement another receiver type the behavior would be the same: we want to identify it as a receiver input

internal/collector/deployment.go
```diff
@@ -24,7 +23,7 @@ func (f *Factory) ReconcileDeployment(er record.EventRecorder, k8sClient client.
 
 	var receiverInputs []string
 	for _, input := range f.ForwarderSpec.Inputs {
-		if logging.IsHttpReceiver(&input) || logging.IsSyslogReceiver(&input) {
+		if input.Receiver.IsHttpReceiver() || input.Receiver.IsSyslogReceiver() {
```
jcantrill, line 26: same here, I think only need to check non-nil

jcantrill (commented):
nit questions, but generally lgtm

internal/validations/clusterlogforwarder/inputs/validate.go
```diff
@@ -49,21 +49,17 @@ func Verify(inputs []loggingv1.InputSpec, status *loggingv1.ClusterLogForwarderS
 		case !validAudit(input, status, extras):
 		case input.Receiver != nil && !extras[constants.VectorName]:
 			badInput("ReceiverSpecs are only supported for the vector log collector")
-		case input.Receiver != nil && input.Receiver.ReceiverTypeSpec == nil:
-			badInput("invalid ReceiverTypeSpec specified for receiver")
 		case input.Receiver != nil && input.Receiver.Type != loggingv1.ReceiverTypeHttp && input.Receiver.Type != loggingv1.ReceiverTypeSyslog:
 			badInput("invalid Type specified for receiver")
-		case input.Receiver != nil && input.Receiver.Type == loggingv1.ReceiverTypeHttp && input.Receiver.Syslog != nil:
+		case input.Receiver != nil && input.Receiver.Type == loggingv1.ReceiverTypeHttp && input.Receiver.ReceiverTypeSpec != nil && input.Receiver.Syslog != nil:
```
Clee2691, line 54: Can you use the above `isReceiverSpecDefined()` to simplify it a bit along with `isHttpReceiver()`?
```suggestion
		case isReceiverSpecDefined(input) && input.Receiver.IsHttpReceiver() && input.Receiver.Syslog != nil:
```

internal/validations/clusterlogforwarder/inputs/validate.go
```diff
@@ -49,21 +49,17 @@ func Verify(inputs []loggingv1.InputSpec, status *loggingv1.ClusterLogForwarderS
...
 			badInput("mismatched Type specified for receiver, specified HTTP and have Syslog")
-		case input.Receiver != nil && input.Receiver.Type == loggingv1.ReceiverTypeSyslog && input.Receiver.HTTP != nil:
+		case isReceiverSpecDefined(input) && input.Receiver != nil && input.Receiver.Type == loggingv1.ReceiverTypeSyslog && input.Receiver.ReceiverTypeSpec != nil && input.Receiver.HTTP != nil:
```
Clee2691, line 56: You are already checking if `input.Receiver != nil && .ReceiverTypeSpec != nil` with `isReceiverSpecDefined()`. Can also use the defined `isSyslogReceiver()`
```suggestion
		case isReceiverSpecDefined(input) && input.Receiver.IsSyslogReceiver() && input.Receiver.HTTP != nil:
```

PR #2338, opened 2024-02-05
LOG-5053: make  spec.inputs[0].receiver.type optional to prevent failing on update to the CLO 5.9

vparfonov:
### Description
This PR makes `spec.inputs[0].receiver.type` an optional field instead of a required one. This change is necessary to prevent potential failures during updates from version 5.8.x to version 5.9. Additionally, it introduces a migration feature that sets `receiver.type` according to the given configuration if it is not already set, allowing for `http` or `syslog` configurations.

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5053
- Enhancement proposal:

### DIFF
--- a/internal/migrations/clusterlogforwarder/migrate_inputs.go
+++ b/internal/migrations/clusterlogforwarder/migrate_inputs.go
@@ -6,6 +6,19 @@ import (
 )
 
 func MigrateInputs(namespace, name string, spec loggingv1.ClusterLogForwarderSpec, logStore *loggingv1.LogStoreSpec, extras map[string]bool, logstoreSecretName, saTokenSecret string) (loggingv1.ClusterLogForwarderSpec, map[string]bool, []loggingv1.Condition) {
+	for i, input := range spec.Inputs {
+		if input.Receiver != nil {
+			if input.Receiver.HTTP != nil && input.Receiver.Type == "" {
+				input.Receiver.Type = loggingv1.ReceiverTypeHttp
+				spec.Inputs[i] = input
+			}
+			if input.Receiver.Syslog != nil && input.Receiver.Type == "" {
+				input.Receiver.Type = loggingv1.ReceiverTypeSyslog
+				spec.Inputs[i] = input
+			}
+		}
+	}
+
 	inputs := map[string]loggingv1.InputSpec{}
 	for _, p := range spec.Pipelines {
 		for _, i := range p.InputRefs {
--- a/apis/logging/v1/input_receiver_types.go
+++ b/apis/logging/v1/input_receiver_types.go
@@ -16,9 +16,8 @@ const (
 type ReceiverSpec struct {
 
 	// Type of Receiver plugin.
-	//
 	// +kubebuilder:validation:Enum:=http;syslog
-	// +required
+	// +optional
 	Type string `json:"type"`
 
 	// The ReceiverTypeSpec that handles particular parameters
--- a/bundle/manifests/logging.openshift.io_clusterlogforwarders.yaml
+++ b/bundle/manifests/logging.openshift.io_clusterlogforwarders.yaml
@@ -424,8 +424,6 @@ spec:
                           - http
                           - syslog
                           type: string
-                      required:
-                      - type
                       type: object
                   required:
                   - name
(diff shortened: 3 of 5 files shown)


jlarriba:
At the same time I was submitting this PR: https://github.com/openshift/cluster-logging-operator/pull/2337 which makes input.receiver.syslog and input.receiver.http optional. This could lead to allow the user to set up a valid empty receiver which just would do nothing.

Also, the current code works on the idea that a Type would always exist, so, for example https://github.com/openshift/cluster-logging-operator/blob/master/apis/logging/v1/cluster_log_forwarder.go#L206 would not return a correct result. In the current status of this PR, the http and syslog receivers would not work if the user decides not to configure a ReceiverType.

jcantrill:
@jlarriba seems we should close this in favor of #2337 and ensure we link that issue against the other PR

jlarriba:
I think that this PR is still valid as it provides extra stuff, like migrations that will help customers to upgrade to 5.9. Those migrations will help current customers to have a valid environment. 

However, the customer could create an invalid configuration. If he does not declare a Type on his Receivers, the operator will fail to create the correct configuration from the "valid" clf.

Either we introduce the changes that remove our dependency on the Type here or we close this and move the changes to #2337 to merge everything together. The problem is that I dont think it is a good idea to make both Type and input.receiver.syslog/input.receiver.http as optional, as that would make checks too complicated for no reason.

My preferred way to go forward would be to contribute changes to this PR that check either on type and syslog/http and remove PR #2337, which will need to wait until an API break is reasonable.

WDYT?

internal/migrations/clusterlogforwarder/migrate_inputs.go
```diff
@@ -6,6 +6,21 @@ import (
 )
 
 func MigrateInputs(namespace, name string, spec loggingv1.ClusterLogForwarderSpec, logStore *loggingv1.LogStoreSpec, extras map[string]bool, logstoreSecretName, saTokenSecret string) (loggingv1.ClusterLogForwarderSpec, map[string]bool, []loggingv1.Condition) {
+	for i, input := range spec.Inputs {
+		if input.Receiver != nil {
+			if input.Receiver.HTTP != nil && len(input.Receiver.Type) == 0 {
```
jcantrill, line 11: strings are initialized so you can just compare this to `""`

internal/migrations/clusterlogforwarder/migrate_inputs.go
```diff
@@ -6,6 +6,21 @@ import (
...
+				input.Receiver.Type = loggingv1.ReceiverTypeHttp
+				spec.Inputs[i] = input
+				extras[constants.MigrateInputApplication] = true
```
jcantrill, line 14: Do we need to set this here?  We did this in other places because otherwise it would fail a validation but I'm not certain its required

internal/migrations/clusterlogforwarder/migrate_inputs.go
vparfonov: done

internal/migrations/clusterlogforwarder/migrate_inputs.go
vparfonov: removed

internal/migrations/clusterlogforwarder/migrate_inputs.go
```diff
@@ -6,6 +6,21 @@ import (
 )
 
 func MigrateInputs(namespace, name string, spec loggingv1.ClusterLogForwarderSpec, logStore *loggingv1.LogStoreSpec, extras map[string]bool, logstoreSecretName, saTokenSecret string) (loggingv1.ClusterLogForwarderSpec, map[string]bool, []loggingv1.Condition) {
+	for i, input := range spec.Inputs {
+		if input.Receiver != nil {
+			if input.Receiver.HTTP != nil && input.Receiver.Type == "" {
+				input.Receiver.Type = loggingv1.ReceiverTypeHttp
+				spec.Inputs[i] = input
+				//extras[constants.MigrateInputApplication] = true
```
jcantrill, line 14: We can delete these comments

PR #2341, opened 2024-02-07
LOG-5028: Replace lua transforms with VRL

jcantrill:
### Description
This PR:
* Replaces all lua transforms with VRL equivalent
* Adds linter to check use of lua transforms

### Links
https://issues.redhat.com/browse/LOG-5028

### DIFF
--- a/internal/generator/vector/normalize/dedot_labels.go
+++ b/internal/generator/vector/normalize/dedot_labels.go
@@ -2,64 +2,38 @@ package normalize
 
 import (
 	. "github.com/openshift/cluster-logging-operator/internal/generator/framework"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/elements"
 	"github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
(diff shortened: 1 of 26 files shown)


internal/generator/vector/normalize/dedot_labels.go
```diff
@@ -2,64 +2,38 @@ package normalize
...
 import (
 	. "github.com/openshift/cluster-logging-operator/internal/generator/framework"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/elements"
 	"github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
+	"strings"
 )
 
-// Dedotting namespace labels
-// Replaces '.' and '/' with '_'
+const (
```
vparfonov, line 10: nit: what about using go:embed feature and put this script to dedot.vrl file

jcantrill: :shrug: We could, seems like an extra layer here

PR #2344, opened 2024-02-07
LOG-5047: Fix Vector collector crash for invalid splunk.IndexKey symbols

vparfonov:
### Description
This PR addressing to fix collector crashes when the `splunk.IndexKey` parameter contains symbols not allowed by Vector's template syntax, which adheres to VRL path expressions. 
Implements stricter validation for `splunk.IndexKey` to ensure it conforms to VRL path rules:

- only alphanumeric characters and underscores `(a-zA-Z0-9_)` are allowed.
- segments containing other characters must be quoted (e.g., `kubernetes.labels."test-logging"`).

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5047
- Enhancement proposal:

### DIFF
--- a/internal/validations/clusterlogforwarder/outputs/splunk.go
+++ b/internal/validations/clusterlogforwarder/outputs/splunk.go
@@ -1,0 +1,46 @@
+package outputs
+
+import (
+	log "github.com/ViaQ/logerr/v2/log/static"
+	loggingv1 "github.com/openshift/cluster-logging-operator/apis/logging/v1"
+	"github.com/openshift/cluster-logging-operator/internal/constants"
+	"github.com/openshift/cluster-logging-operator/internal/status"
+	"github.com/openshift/cluster-logging-operator/internal/validations/clusterlogforwarder/conditions"
+	"k8s.io/api/core/v1"
+	"regexp"
+)
+
+const pattern = `^([a-zA-Z0-9_]+|"(?:\\.|[^"\\])*")(?:\.([a-zA-Z0-9_]+|"(?:\\.|[^"\\])*"))*$`
+
+func VerifySplunk(name string, splunk *loggingv1.Splunk) (bool, status.Condition) {
+	if splunk != nil {
+		if splunk.IndexKey != "" && splunk.IndexName != "" {
+			log.V(3).Info("verifyOutputsFailed", "reason", "splunk output allows only one of indexKey or indexName, not both.")
+			return false, conditions.CondInvalid("output %q: Only one of indexKey or indexName can be set, not both.", name)
+		}
+
+		if splunk.IndexKey != "" && !isIndexKeyMatch(splunk.IndexKey) {
+			return false, conditions.CondInvalid("output %q: IndexKey can only contain letters, numbers, and underscores (a-zA-Z0-9_). "+
+				"Segments that contain characters outside of this range must be quoted.", name)
+		}
+	}
+	return true, conditions.CondReady
+}
+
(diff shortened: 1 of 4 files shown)


internal/validations/clusterlogforwarder/outputs/splunk_test.go
```diff
@@ -0,0 +1,89 @@
...
+		})
+
+		It("should fail if no IndexKey and ", func() {
+			splunk := &loggingv1.Splunk{
+				IndexKey:  "kubernetes.labels.test_logging",
+				IndexName: "application",
+			}
+			valid, st := VerifySplunk("splunk", splunk)
+			Expect(valid).To(BeFalse())
+			Expect(map[string]status.Condition{"azureMonitor": st}).To(HaveCondition("Ready", false, loggingv1.ReasonInvalid, "output \"splunk\": Only one of indexKey or indexName can be set, not both."))
```
jcantrill, line 35: @Clee2691 is this consistent with the epic?

internal/validations/clusterlogforwarder/outputs/splunk.go
```diff
@@ -0,0 +1,46 @@
...
+const pattern = `^([a-zA-Z0-9_]+|"(?:\\.|[^"\\])*")(?:\.([a-zA-Z0-9_]+|"(?:\\.|[^"\\])*"))*$`
+
+func VerifySplunk(name string, splunk *loggingv1.Splunk) (bool, status.Condition) {
+	if splunk != nil {
+		if splunk.IndexKey != "" && splunk.IndexName != "" {
+			log.V(3).Info("verifyOutputsFailed", "reason", "splunk output allows only one of indexKey or indexName, not both.")
+			return false, conditions.CondInvalid("output %q: Only one of indexKey or indexName can be set, not both.", name)
+		}
+
+		if splunk.IndexKey != "" && !isIndexKeyMatch(splunk.IndexKey) {
```
jcantrill, line 22: @Clee2691 is this consistent with the epic and how we treet the other outputs?  I don't recall off hand but want to ensure we are not rejecting here, and accepting and fixing for others

jcantrill (commented):
/hold
/approve

internal/validations/clusterlogforwarder/outputs/splunk_test.go
Clee2691: Yes this is consistent. We only allow one of `IndexKey` or`IndexName` not both.

internal/validations/clusterlogforwarder/outputs/splunk.go
Clee2691: For something like `Loki` or `Elasticsearch`, we haven't verified `indexkeys` and didn't reject them outright. Instead we send them to the default. But I can see actually this should have a been a problem for other outputs that rely on `indexKey`. It's the fact that certain path segments must be quoted. Maybe we just didn't check with invalid VRL path expressions.

If we don't want to reject the config outright I suggest this solution:
1. Split the indexkey by (`.`) to form an array of path segments
2. Quote each segment in the array
3. Pass that array into VRL and use the function `get()` to determine if that path array produces a value
4. If value is `null` then send to default `Splunk` index otherwise send to the index specified by the value.

In this way, we don't have to verify the `indexKey` outright and can maintain the default behavior.

vparfonov: I had this in mind, but I thought it would be better to inform the user about potential errors in the configuration that may lead to unexpected behavior. Especially because the fixes will be very easy and clearly explained in the Vector documentation

PR #2348, opened 2024-02-09
LOG-3883: Implement pruning fields from log messages

Clee2691:
### Description
This PR adds the prune filter feature. This feature allows admins to prune fields from log records to reduce the overall size of a record.

This feature extends the filter API by adding the fields `prune` along with `in` and `notIn`.  

The fields, `in` and `notIn`, expect an array of dot-delimited path expressions  which should contain alphanumeric and underscores. Anything outside of that will have to be quoted. 
- Ex1 : `.kubernetes."foo-bar/baz"`
- Ex2: `.log_type`

An example prune filter spec is as follows:
```
spec:
  filters:
    - name: my-prune
      type: prune
      prune:
        in: [.kubernetes.namespace_name,.foo."test-dashes/slashes"]
        notIn: [.log_type,.message,.kubernetes,."@timestamp",.openshift]
```
/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-3883
- Enhancement proposal: https://github.com/openshift/enhancements/blob/a6a1feb9cceb0b61960bcf00f292cb0d04ee3753/enhancements/cluster-logging/content-filter.md#prune-filters


internal/generator/vector/filter/prune/prune.vrl.tmpl
```diff
@@ -0,0 +1,22 @@
+{{- if .NotIn}}
+notIn = {{.NotIn}}
+
+# Prune keys not in `notIn` list
+new_object = {}
+for_each(notIn) -> |_index, pathSeg| {
+    val = get(., pathSeg) ?? null
```
vparfonov, line 7: can we use `exist` function instead of `get` https://vector.dev/docs/reference/vrl/functions/#exists ?

Clee2691: I am using `get` here because I broke up the field path into an array of path segments. The `get` function expects an array instead of a path. I could not use something like:
```
paths = [.log, .message]
```
and iterate through that because `for_each` would only extract the values and not preserve the path for the `exists` function.

PR #2352, opened 2024-02-12
LOG-5055: Add deprecation alerts

Clee2691:
### Description
This PR introduces three new Prometheus alerts to notify about the deprecation of `fluentd`, `elasticsearch`, and `kibana`.

/cc @cahartma @vparfonov @xperimental 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5055

### DIFF
--- a/bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
+++ b/bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
@@ -61,6 +61,48 @@ spec:
       labels:
         service: collector
         severity: Warning
+    - alert: ElasticsearchDeprecation
+      annotations:
+        message: The OpenShift Elasticsearch Operator is deprecated and is planned
+          to be removed in a future release. Red Hat provides bug fixes and support
+          for this feature during the current release lifecycle, but this feature
+          no longer receives enhancements. As an alternative to using the OpenShift
+          Elasticsearch Operator to manage the default log storage, you can use the
+          Loki Operator.
+        summary: Detected Elasticsearch as the in-cluster storage which is deprecated
+          and will be removed in a future release.
+      expr: |
+        sum(kube_pod_labels{namespace="openshift-logging",label_component='elasticsearch'}) > 0
+      for: 5m
+      labels:
+        service: storage
+        severity: Warning
+    - alert: FluentdDeprecation
+      annotations:
+        message: Fluentd is deprecated and is planned to be removed in a future release.
+          Red Hat provides bug fixes and support for this feature during the current
+          release lifecycle, but this feature no longer receives enhancements. As
+          an alternative to Fluentd, you can use Vector instead.
+        summary: Detected Fluentd as the collector which is deprecated and will be
+          removed in a future release.
+      expr: |
+        sum(kube_pod_labels{namespace="openshift-logging", label_implementation='fluentd', label_app_kubernetes_io_managed_by="cluster-logging-operator"}) > 0
+      for: 5m
+      labels:
+        service: collector
+        severity: Warning
+    - alert: KibanaDeprecation
+      annotations:
+        message: The Kibana web console is now deprecated and is planned to be removed
+          in a future logging release.
+        summary: Detected Kibana as the visualization which is deprecated and will
+          be removed in a future release.
+      expr: |
+        sum(kube_pod_labels{namespace="openshift-logging",label_component='kibana'}) > 0
+      for: 5m
+      labels:
+        service: visualization
+        severity: Warning
   - name: logging_clusterlogging_telemetry.rules
     rules:
     - expr: |
--- a/config/prometheus/collector_alerts.yaml
+++ b/config/prometheus/collector_alerts.yaml
@@ -55,6 +55,36 @@ spec:
       labels:
         service: collector
         severity: Warning
+    - alert: ElasticsearchDeprecation
+      annotations:
+        message: "The OpenShift Elasticsearch Operator is deprecated and is planned to be removed in a future release. Red Hat provides bug fixes and support for this feature during the current release lifecycle, but this feature no longer receives enhancements. As an alternative to using the OpenShift Elasticsearch Operator to manage the default log storage, you can use the Loki Operator."
+        summary: "Detected Elasticsearch as the in-cluster storage which is deprecated and will be removed in a future release."
+      expr: |
+        sum(kube_pod_labels{namespace="openshift-logging",label_component='elasticsearch'}) > 0
+      for: 5m
+      labels:
+        service: storage
+        severity: Warning
+    - alert: FluentdDeprecation
+      annotations:
+        message: "Fluentd is deprecated and is planned to be removed in a future release. Red Hat provides bug fixes and support for this feature during the current release lifecycle, but this feature no longer receives enhancements. As an alternative to Fluentd, you can use Vector instead."
+        summary: "Detected Fluentd as the collector which is deprecated and will be removed in a future release."
+      expr: |
+        sum(kube_pod_labels{namespace="openshift-logging", label_implementation='fluentd', label_app_kubernetes_io_managed_by="cluster-logging-operator"}) > 0
+      for: 5m
+      labels:
+        service: collector
+        severity: Warning
+    - alert: KibanaDeprecation
+      annotations:
+        message: "The Kibana web console is now deprecated and is planned to be removed in a future logging release."
+        summary: "Detected Kibana as the visualization which is deprecated and will be removed in a future release."
+      expr: |
+        sum(kube_pod_labels{namespace="openshift-logging",label_component='kibana'}) > 0
+      for: 5m
+      labels:
+        service: visualization
+        severity: Warning
   - name: logging_clusterlogging_telemetry.rules
     rules:
     - expr: |


bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -61,6 +61,48 @@ spec:
...
         severity: Warning
+    - alert: ElasticsearchDeprecation
+      annotations:
+        message: The OpenShift Elasticsearch Operator is deprecated and is planned
+          to be removed in a future release. Red Hat provides bug fixes and support
+          for this feature during the current release lifecycle, but this feature
+          no longer receives enhancements. As an alternative to using the OpenShift
+          Elasticsearch Operator to manage the default log storage, you can use the
+          Loki Operator.
+        summary: Detected Elasticsearch as the on cluster storage which is being deprecated
```
jcantrill, line 72: ...which is deprecated and will be removed in a future release

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -61,6 +61,48 @@ spec:
...
+          in a future release.
+      expr: |
+        sum(min_over_time(container_cpu_usage_seconds_total{container=~'elasticsearch'}[5m])) > 0
```
jcantrill, line 75: We need to restrict this to "openshift-logging" NS since AFAIK tracing could still be relying upon 'elasticsearch'.  We may want to also see if there is a reliable label on the pod in lieu of container given it's possible maybe to have an Elastic deployed operator managing an elasticsearch cluster

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -61,6 +61,48 @@ spec:
...
+      labels:
+        service: storage
+        severity: Warning
+    - alert: FluentdDeprecation
+      annotations:
+        message: Fluentd is deprecated and is planned to be removed in a future release.
+          Red Hat provides bug fixes and support for this feature during the current
+          release lifecycle, but this feature no longer receives enhancements. As
+          an alternative to Fluentd, you can use Vector instead.
+        summary: Detected Fluentd as the collector which is being deprecated in a
```
jcantrill, line 86: is deprecated...

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -61,6 +61,48 @@ spec:
...
+          future release.
+      expr: |
+        sum(min_over_time(container_cpu_usage_seconds_total{image=~'.+fluentd.+'}[5m])) > 0
```
jcantrill, line 89: Same here regarding NS.  My recollection is for mCLF we only allow fluentd as part of a legacy deployment so we can restrict it there.  We should also have consistent labels going back to 5.5 that would allow us to further narrow this to our pods

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -61,6 +61,48 @@ spec:
...
+      for: 10m
+      labels:
+        service: collector
+        severity: Warning
+    - alert: KibanaDeprecation
+      annotations:
+        message: The Kibana web console is now deprecated and is planned to be removed
+          in a future logging release.
+        summary: Detected Kibana as the visualization which is being deprecated in
```
jcantrill, line 98: ..is deprecated

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -61,6 +61,48 @@ spec:
...
+          a future release
+      expr: |
+        sum(min_over_time(container_cpu_usage_seconds_total{container=~'kibana'}[5m])) > 0
```
jcantrill, line 101: same comment here about NS and possible use of labels

jcantrill (commented):
/hold

config/prometheus/collector_alerts.yaml
```diff
@@ -55,6 +55,36 @@ spec:
       labels:
         service: collector
         severity: Warning
+    - alert: ElasticsearchDeprecation
+      annotations:
+        message: "The OpenShift Elasticsearch Operator is deprecated and is planned to be removed in a future release. Red Hat provides bug fixes and support for this feature during the current release lifecycle, but this feature no longer receives enhancements. As an alternative to using the OpenShift Elasticsearch Operator to manage the default log storage, you can use the Loki Operator."
+        summary: "Detected Elasticsearch as the on cluster storage which is being deprecated in a future release."
+      expr: |
+        sum(min_over_time(container_cpu_usage_seconds_total{container=~'elasticsearch'}[5m])) > 0
```
xperimental, line 63: This query seems to work, but is a bit "unspecific" in my opinion. We're not really interested in the CPU usage, only the existence of the elasticsearch pods. Also we're not interested in `elasticsearch` containers that belong to other applications (possibly of the customer).

I would suggest adding matchers to "drill down" into the pod that we want to monitor and also use a metric that just gives us "existence", for example one of the metrics provided by kube-state-metrics:

```suggestion
        sum(kube_pod_labels{namespace="openshift-logging", label_component="elasticsearch"}) > 0
```

Alternatively counting containers called "elasticsearch" similar to the previous query:

```suggestion
        sum(kube_pod_container_info{namespace="openshift-logging", container="elasticsearch"}) > 0
```

This suggestion also avoids the "time-based aggregation" and "range vector" which might cause flapping of the alert if the CPU usage is variable (which is probably no issue for ES but might be for Kibana). This should also mean that you can reduce the `for` time to five minutes or even less, because we do not have to wait for the result to "settle" because there is no time component involved.

config/prometheus/collector_alerts.yaml
```diff
@@ -55,6 +55,36 @@ spec:
...
+      for: 10m
+      labels:
+        service: collector
+        severity: Warning
+    - alert: KibanaDeprecation
+      annotations:
+        message: "The Kibana web console is now deprecated and is planned to be removed in a future logging release."
+        summary: "Detected Kibana as the visualization which is being deprecated in a future release"
+      expr: |
+        sum(min_over_time(container_cpu_usage_seconds_total{container=~'kibana'}[5m])) > 0
```
xperimental, line 83: See `elasticsearch` for the longer explanation:

```suggestion
        sum(kube_pod_labels{namespace="openshift-logging", label_component="kibana"}) > 0
```

config/prometheus/collector_alerts.yaml
```diff
@@ -55,6 +55,36 @@ spec:
...
+      for: 10m
+      labels:
+        service: storage
+        severity: Warning
+    - alert: FluentdDeprecation
+      annotations:
+        message: "Fluentd is deprecated and is planned to be removed in a future release. Red Hat provides bug fixes and support for this feature during the current release lifecycle, but this feature no longer receives enhancements. As an alternative to Fluentd, you can use Vector instead."
+        summary: "Detected Fluentd as the collector which is being deprecated in a future release."
+      expr: |
+        sum(min_over_time(container_cpu_usage_seconds_total{image=~'.+fluentd.+'}[5m])) > 0
```
xperimental, line 73: See `elasticsearch` for the longer explanation:

```suggestion
        sum(kube_pod_labels{namespace="openshift-logging", label_app_kubernetes_io_component="collector", label_app_kubernetes_io_name="fluentd"}) > 0
```

config/prometheus/collector_alerts.yaml
```diff
@@ -55,6 +55,36 @@ spec:
       labels:
         service: collector
         severity: Warning
+    - alert: ElasticsearchDeprecation
+      annotations:
+        message: "The OpenShift Elasticsearch Operator is deprecated and is planned to be removed in a future release. Red Hat provides bug fixes and support for this feature during the current release lifecycle, but this feature no longer receives enhancements. As an alternative to using the OpenShift Elasticsearch Operator to manage the default log storage, you can use the Loki Operator."
+        summary: "Detected Elasticsearch as the on cluster storage which is being deprecated in a future release."
```
xperimental, line 61: Aren't all these components already deprecated? I think we're better suited with "going to be removed in a future release".

edit: it seems Jeff had a similar thought.

xperimental (commented):
fyi: I just ran the queries on a test-cluster using Elasticsearch, did not actually apply the alerts, so I did not see how the text looks in the UI.

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
Clee2691: I've used metrics suggested by @xperimental below

config/prometheus/collector_alerts.yaml
Clee2691: Incorporated the suggested metrics for the alerts. Thanks!

config/prometheus/collector_alerts.yaml
```diff
@@ -55,6 +55,36 @@ spec:
       labels:
         service: collector
         severity: Warning
+    - alert: ElasticsearchDeprecation
+      annotations:
+        message: "The OpenShift Elasticsearch Operator is deprecated and is planned to be removed in a future release. Red Hat provides bug fixes and support for this feature during the current release lifecycle, but this feature no longer receives enhancements. As an alternative to using the OpenShift Elasticsearch Operator to manage the default log storage, you can use the Loki Operator."
+        summary: "Detected Elasticsearch as the on cluster storage which is deprecated and will be removed in a future release."
```
xperimental, line 61: The `on` tripped me up on first read, I think it needs a `-`:

```suggestion
        summary: "Detected Elasticsearch as the on-cluster storage which is deprecated and will be removed in a future release."
```

Also, isn't it "in-cluster"?

```suggestion
        summary: "Detected Elasticsearch as the in-cluster storage which is deprecated and will be removed in a future release."
```

jcantrill:
/hold

pending PM review

PR #2359, opened 2024-02-16
LOG-5049: Deprecated non-working splunk fields and add validation

cahartma:
### Description
After discussion about implementation difficulties, we have deprecated splunk.Fields.   They have indexKey or indexName to use instead.  The functionality has never been implemented in 2 years since the splunk forwarding was added.   

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-5049
- https://issues.redhat.com/browse/LOG-4558

### DIFF
--- a/internal/validations/clusterlogforwarder/outputs/splunk.go
+++ b/internal/validations/clusterlogforwarder/outputs/splunk.go
@@ -23,6 +23,10 @@ func VerifySplunk(name string, splunk *loggingv1.Splunk) (bool, status.Condition
 			return false, conditions.CondInvalid("output %q: IndexKey can only contain letters, numbers, and underscores (a-zA-Z0-9_). "+
 				"Segments that contain characters outside of this range must be quoted.", name)
 		}
+
+		if splunk.Fields != nil {
+			return true, conditions.CondDegraded(loggingv1.ReasonUnused, "Warning: Support for 'fields' is not implemented and deprecated for output %q", name)
+		}
 	}
 	return true, conditions.CondReady
 }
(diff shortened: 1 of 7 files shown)


apis/logging/v1/output_types.go
```diff
@@ -294,6 +289,11 @@ type Splunk struct {
 	// If IndexKey && IndexName are not specified, the default index defined within Splunk is used.
 	// +optional
 	IndexName string `json:"indexName,omitempty"`
+
+	// Deprecated. Fields to be added to Splunk index. Use indexKey or indexName instead.
```
jcantrill, line 293: I don't think there is a relation between "Fields" and index key/name. The former is meta that is indexed where the latter provides the "root" name to store the data.  I would drop the last sentance

internal/validations/clusterlogforwarder/outputs/splunk.go
```diff
@@ -23,6 +23,10 @@ func VerifySplunk(name string, splunk *loggingv1.Splunk) (bool, status.Condition
 			return false, conditions.CondInvalid("output %q: IndexKey can only contain letters, numbers, and underscores (a-zA-Z0-9_). "+
 				"Segments that contain characters outside of this range must be quoted.", name)
 		}
+
+		if splunk.Fields != nil {
+			return false, conditions.CondInvalid("Warning: 'fields' is deprecated for output %q.  Use indexKey or indexName instead.", name)
```
jcantrill, line 28: Warning: Support for 'fields' is not implemented and deprecated.

jcantrill (commented):
/hold

internal/validations/clusterlogforwarder/outputs/splunk_test.go
```diff
@@ -43,6 +43,15 @@ var _ = Describe("Validate Splunk:", func() {
 			Expect(valid).To(BeFalse())
 			Expect(map[string]status.Condition{"splunk": st}).To(HaveCondition("Ready", false, loggingv1.ReasonInvalid, "output \"splunk\": IndexKey can only contain letters, numbers, and underscores \\(a-zA-Z0-9_\\). Segments that contain characters outside of this range must be quoted."))
 		})
+
+		It("should fail if Fields is spec'd", func() {
```
jcantrill, line 47: Can we pass validations but provide the warning.  My concern here is the upgrade scenario.  What will happen if this is defined and we fail it

cahartma: How do you feel about passing the validation, but setting condition of Degraded=True?

PR #2364, opened 2024-02-19
LOG-5117: Pruning field doesn't work when the specified field contains non-alphanumeric and non-underscores.

Clee2691:
### Description
This PR fixes the issue that quoted fields containing a `.` wouldn't be pruned .
Example:
```
 .foo.bar."baz.test.field-111/222"
```

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5117


cahartma (commented):
/lgtm

PR #2365, opened 2024-02-19
LOG-5210: Combine container and namespace includes/excludes

jcantrill:
### Description
This PR:
* Combines namespace includes with container includes for a given application input
* Combines namespace excludes with container excludes for a given application input
* Sorts includes and exclude paths for consistency

This change allows a user to spec including or excluding specific container logs from specific namespaces

### Links
* https://issues.redhat.com/browse/LOG-5120


jcantrill:
cc @cahartma @Clee2691

PR #2370, opened 2024-02-20
LOG-5121: Prevent LogType names from starting with underscores (_) for Azure log forwarder

vparfonov:
### Description
This PR implements a rule to prevent LogType names from starting with an underscore (_). 
`_invalidName` : not valid name

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5121
- Enhancement proposal:

### DIFF
--- a/internal/validations/clusterlogforwarder/outputs/azuremonitor_test.go
+++ b/internal/validations/clusterlogforwarder/outputs/azuremonitor_test.go
@@ -41,7 +41,7 @@ var _ = Describe("Validate AzureMonitor:", func() {
 			}
 			valid, st := VerifyAzureMonitorLog("azureMonitor", azureMonitor)
 			Expect(valid).To(BeFalse())
-			Expect(map[string]status.Condition{"azureMonitor": st}).To(HaveCondition("Ready", false, loggingv1.ReasonInvalid, "output \"azureMonitor\": LogType can only contain letters, numbers, and underscores \\(_\\), and may not exceed 100 characters."))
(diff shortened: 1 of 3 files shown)


internal/validations/clusterlogforwarder/outputs/azuremonitor_test.go
```diff
@@ -41,7 +41,7 @@ var _ = Describe("Validate AzureMonitor:", func() {
 			}
 			valid, st := VerifyAzureMonitorLog("azureMonitor", azureMonitor)
 			Expect(valid).To(BeFalse())
-			Expect(map[string]status.Condition{"azureMonitor": st}).To(HaveCondition("Ready", false, loggingv1.ReasonInvalid, "output \"azureMonitor\": LogType can only contain letters, numbers, and underscores \\(_\\), and may not exceed 100 characters."))
+			Expect(map[string]status.Condition{"azureMonitor": st}).To(HaveCondition("Ready", false, loggingv1.ReasonInvalid, "output \"azureMonitor\": LogType names must start with a letter/number, contain only letters/numbers/underscores \\(_\\), and be between 1-100 characters."))
```
jcantrill, line 44: no need to change... but you could also use back ticks instead of quotes for the string and then there would be no need to escape the internal quotes

jcantrill (approved):
/approve
/lgtm

PR #2371, opened 2024-02-21
LOG-5125: Disable API and feature gate for OTEL

Clee2691:
### Description
This PR disables the API for the `OTEL` schema option under the `http` sink for vector. It also disables the feature gate that was required to enable it.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5125

### DIFF
--- a/api/logging/v1/output_types.go
+++ b/api/logging/v1/output_types.go
@@ -320,7 +320,8 @@ type Http struct {
 	// +kubebuilder:validation:Enum:=opentelemetry;viaq
 	// +kubebuilder:default:viaq
 	// +optional
-	Schema string `json:"schema,omitempty"`
+	// +operator-sdk:csv:customresourcedefinitions:xDescriptors={"urn:alm:descriptor:com.tectonic.ui:hidden"}
+	Schema string `json:"-"`
 }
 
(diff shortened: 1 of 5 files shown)


api/logging/v1/output_types.go
```diff
@@ -320,7 +320,8 @@ type Http struct {
 	// +kubebuilder:validation:Enum:=opentelemetry;viaq
 	// +kubebuilder:default:viaq
 	// +optional
-	Schema string `json:"schema,omitempty"`
+	// +operator-sdk:csv:customresourcedefinitions:xDescriptors={"urn:alm:descriptor:com.tectonic.ui:hidden"}
```
cahartma, line 323: Curious, why not remove this as well?

jcantrill: Attempting to not remove anything so we can reuse it when available.  This should make it so its not public though thinking about it, did this already exist in 5.8?

Clee2691: Yes this field did exist in 5.8. I am not inclined to remove this because of what you stated.

PR #2380, opened 2024-03-01
LOG-5143: The tune option can not be appled when output name is default.

Clee2691:
### Description
This PR fixes the issue where the default outputs (Elasticsearch & LokiStack) did not take into account the tuning parameters if a user defined them in the outputs.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5143

### DIFF
--- a/internal/migrations/clusterlogforwarder/migrate_clusterlogforwarder_test.go
+++ b/internal/migrations/clusterlogforwarder/migrate_clusterlogforwarder_test.go
@@ -2,6 +2,7 @@ package clusterlogforwarder
 
 import (
 	"fmt"
+
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	logging "github.com/openshift/cluster-logging-operator/api/logging/v1"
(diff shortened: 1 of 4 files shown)


internal/migrations/clusterlogforwarder/migrate_clusterlogforwarder_test.go
```diff
@@ -288,91 +360,28 @@ var _ = Describe("MigrateClusterLogForwarderSpec", func() {
 					Expect(forwarderSpec).To(Equal(exp), fmt.Sprintf("Exp. default output because of pipeline %v and ElasticsearchSpec %v", pipelines, esSpec))
 					Expect(extras).To(Equal(map[string]bool{constants.MigrateDefaultOutput: true}))
 				})
-			})
-
-		})
-		Context("when a pipeline references 'default'", func() {
```
Clee2691, line 294: Duplicate tests here, so I removed them. These tests are already defined [here](https://github.com/openshift/cluster-logging-operator/blob/e6229cb1b4ff6143312331e6d53bfaf59eebfe95/internal/migrations/clusterlogforwarder/migrate_clusterlogforwarder_test.go#L281)

PR #2389, opened 2024-03-07
Updates to API template to meet docs standards

abrennan89:
### Description

This PR is to update the API doc templates to help remove roadblocks to publishing our documentation in the OCP docs set.
I have updated things like headings, metadata, and formatting for now as a first step to minimize additional work for tech writers, but there is more to do and will likely require additional follow up PRs.

This change should also be updated in older branches, but I can do that manually and regenerate the API docs as I go, rather than auto-CPing.

/cc @alanconway 
/assign @jcantrill 

### Links
- https://issues.redhat.com/browse/OBSDOCS-845


jcantrill (commented):
/lgtm

PR #2394, opened 2024-03-11
LOG-5062: align validation with documentation

vparfonov:
### Description
This PR address to align validation feature with current documentation. During validation for `ClusterLogging` we will check is `ClusterLogForwarder`  in the same namespace named the same exist if not, exception will throw. 

/cc @cahartma @Clee2691 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5062
- Enhancement proposal:

### DIFF
--- a/internal/validations/clusterlogging/validate_clusterlogging_spec.go
+++ b/internal/validations/clusterlogging/validate_clusterlogging_spec.go
@@ -1,12 +1,18 @@
 package clusterlogging
 
 import (
+	"context"
+	"fmt"
+
 	v1 "github.com/openshift/cluster-logging-operator/api/logging/v1"
 	"github.com/openshift/cluster-logging-operator/internal/constants"
 	"github.com/openshift/cluster-logging-operator/internal/validations/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func validateClusterLoggingSpec(cl v1.ClusterLogging) error {
+func validateClusterLoggingSpec(cl v1.ClusterLogging, k8sClient client.Client) error {
 
 	if cl.Namespace == constants.OpenshiftNS && cl.Name == constants.SingletonName {
 		if cl.Spec.Collection != nil && !cl.Spec.Collection.Type.IsSupportedCollector() {
@@ -15,6 +21,7 @@ func validateClusterLoggingSpec(cl v1.ClusterLogging) error {
 		return nil
 	}
 	spec := cl.Spec
+
 	if spec.Forwarder != nil || spec.LogStore != nil || spec.Curation != nil || spec.Visualization != nil {
 		return errors.NewValidationError("Only spec.collection is allowed when using multiple instances of ClusterLogForwarder: %s/%s", cl.Namespace, cl.Name)
 	}
@@ -25,4 +32,31 @@ func validateClusterLoggingSpec(cl v1.ClusterLogging) error {
 		return errors.NewValidationError("Only vector collector impl is supported when using multiple instances of ClusterLogForwarder: %s/%s", cl.Namespace, cl.Name)
 	}
 	return nil
+}
+
+//CL(openshift-logging/instance) only - This is a valid LEGACY use case
+//CL(openshift-logging/instance) & CLF(openshift-logging/instance) - This is a valid LEGACY usecase
+//CL(ANY_NS/OTHER) - This is an invalid use case
+//CLF(ANY_NS/ANY_NAME) - This is a valid mCLF use case
+//CL(ANY_NS/ANY_NAME) && CLF(ANY_NS/ANY_NAME) - This is a valid mCLF use case
+func validateSetup(cl v1.ClusterLogging, k8sClient client.Client) error {
+	key := types.NamespacedName{Name: cl.Name, Namespace: cl.Namespace}
+	clf := &v1.ClusterLogForwarder{}
+
+	// Check if ClusterLogForwarder exists
+	err := k8sClient.Get(context.TODO(), key, clf)
+	isCLFNotFound := apierrors.IsNotFound(err)
+	if isCLFNotFound {
+		// Determine if it's a legacy or mCLF instance
+		isLegacy := cl.Namespace == constants.OpenshiftNS && cl.Name == constants.SingletonName
+
+		if isLegacy {
+			return nil // Legacy ClusterLogging with no ClusterLogForwarder, valid use case
+		} else {
+			msg := fmt.Sprintf("ClusterLogging instance requires to have a ClusterLogForwarder deployed in the same namespace and named the same, if not eqauls %s/%s", constants.OpenshiftNS, constants.SingletonName)
+			return errors.NewValidationError(msg)
+		}
+	}
+
+	return nil // Legacy ClusterLogging with existing ClusterLogForwarder or mCLF ClusterLogForwarder, valid use case
 }
--- a/internal/validations/clusterlogging/validate_clusterlogging_spec_test.go
+++ b/internal/validations/clusterlogging/validate_clusterlogging_spec_test.go
@@ -5,13 +5,16 @@ import (
 	. "github.com/onsi/gomega"
 	logging "github.com/openshift/cluster-logging-operator/api/logging/v1"
 	"github.com/openshift/cluster-logging-operator/test/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
 var _ = Describe("[internal][validations] ClusterLogging", func() {
 
 	Context("#validateClusterLoggingSpec", func() {
 		var (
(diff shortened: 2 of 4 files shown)


internal/validations/clusterlogging/validate_clusterlogging_spec.go
```diff
@@ -15,8 +21,16 @@ func validateClusterLoggingSpec(cl v1.ClusterLogging) error {
 		return nil
 	}
 	spec := cl.Spec
+
 	if spec.Forwarder != nil || spec.LogStore != nil || spec.Curation != nil || spec.Visualization != nil {
-		return errors.NewValidationError("Only spec.collection is allowed when using multiple instances of ClusterLogForwarder: %s/%s", cl.Namespace, cl.Name)
+		key := types.NamespacedName{Name: cl.Name, Namespace: cl.Namespace}
+		clf := runtime.NewClusterLogForwarder(cl.Namespace, cl.Name)
```
jcantrill, line 27: I don't think this needs to be anything other then "&logging.ClusterLogForwarder{}" since the "get" will fill in all the pieces if it is found using the key

internal/validations/clusterlogging/validate_clusterlogging_spec.go
```diff
@@ -15,8 +21,16 @@ func validateClusterLoggingSpec(cl v1.ClusterLogging) error {
 		return nil
 	}
 	spec := cl.Spec
+
 	if spec.Forwarder != nil || spec.LogStore != nil || spec.Curation != nil || spec.Visualization != nil {
```
jcantrill, line 25: I think we are loosing something here in validation of ClusterLogging.  We have 2 scenarios:
* legacy: you must define CLF and it must be named openshift-logging/instance and it can have any of these fields
* mCLF: it must ns/name match the CLF and ONLY define spec.collection

This change appears to break the legacy scenario. 

I believe the load of CL needs to occur before this IF check and then based upon the result, we check and validate to see which of the scenarios is valid

internal/validations/clusterlogging/validate_clusterlogging_spec.go
```diff
@@ -15,8 +21,16 @@ func validateClusterLoggingSpec(cl v1.ClusterLogging) error {
...
-		return errors.NewValidationError("Only spec.collection is allowed when using multiple instances of ClusterLogForwarder: %s/%s", cl.Namespace, cl.Name)
+		key := types.NamespacedName{Name: cl.Name, Namespace: cl.Namespace}
+		clf := runtime.NewClusterLogForwarder(cl.Namespace, cl.Name)
+		if err := k8sClient.Get(context.TODO(), key, clf); err != nil {
+			if apierrors.IsNotFound(err) {
+				return errors.NewValidationError("ClusterLogForwarder: %s/%s instance not found, ClusterLogging object in %q not named %q also "+
```
jcantrill, line 30: This should change based upon my first comment.  As-is the message is really long and difficult to understand

internal/validations/clusterlogging/validate_clusterlogging_spec_test.go
```diff
@@ -52,28 +57,78 @@ var _ = Describe("[internal][validations] ClusterLogging", func() {
...
+			It("should pass when collection.type is spec'd", func() {
+				cl = runtime.NewClusterLogging()
+				cl.Namespace = constants.OpenshiftNS
+				cl.Name = "mycollector"
+				cl.Spec.Collection.Type = constants.VectorName
+				err := validateClusterLoggingSpec(*cl, k8sClient)
+				Expect(err).To(Succeed())
+			})
+
+			It("should pass when anything but collection is spec'd and CLF with same name/namespace exist", func() {
```
jcantrill, line 91: This use-case is incorrect.  This is not allowed by design

internal/validations/clusterlogging/validate_clusterlogging_spec.go
vparfonov: I don't think that it will break the legacy scenario because we check this above:
```
  if cl.Namespace == constants.OpenshiftNS && cl.Name == constants.SingletonName {
		if cl.Spec.Collection != nil && !cl.Spec.Collection.Type.IsSupportedCollector() {
			return errors.NewValidationError("Collector implementation is not supported: %q", cl.Spec.Collection.Type)
		}
		return nil
	}
```

internal/validations/clusterlogging/validate_clusterlogging_spec.go
```diff
@@ -15,6 +21,16 @@ func validateClusterLoggingSpec(cl v1.ClusterLogging) error {
 		return nil
 	}
 	spec := cl.Spec
+	key := types.NamespacedName{Name: cl.Name, Namespace: cl.Namespace}
+	clf := &v1.ClusterLogForwarder{}
+	if err := k8sClient.Get(context.TODO(), key, clf); err != nil {
+		if apierrors.IsNotFound(err) {
+			return errors.NewValidationError("ClusterLogging: %s/%s instance requires to have a ClusterLogForwarder deployed in the same namespace named the same", cl.Namespace, cl.Name)
```
jcantrill, line 28: We should consider a custom error message depending upon the usecase; and CLF is not always required

1. CL(openshift-logging/instance) only - This is a valid LEGACY usecase
1. CL(openshift-logging/instance) & CLF(openshift-logging/instance)   - This is a valid LEGACY usecase
1. CL(ANY_NS/OTHER) - This is an invalid usecase
1. CLF(ANY_NS/ANY_NAME) - This is a valid mCLF usecase
1. CL(ANY_NS/ANY_NAME) && CLF(ANY_NS/ANY_NAME) - This is a valid mCLF usecase

ANY_NAME is any name other then "instance" in "openshift-logging"

It looks like this still does not support the expected outcome for #1.  It may make more sense to refactor the method to try to load CLF first and then use a switch depending upon the outcome and ns/name combination:

```
clf, err := k8sClient.Get(NS, Name)
switch: {
case legacy CL only:
case legacy CL and CLF:
case mclf CL only:
case mclf CLF only:
case mCLF CLF and CL:
}
```

internal/validations/clusterlogging/validate_clusterlogging_spec_test.go
```diff
@@ -52,31 +57,85 @@ var _ = Describe("[internal][validations] ClusterLogging", func() {
...
+			Expect(err).To(Succeed())
+		})
+
+		It("should pass when CL(openshift-logging/instance) & CLF(openshift-logging/instance) - This is a valid LEGACY use case", func() {
+			k8sClient = fake.NewClientBuilder().WithRuntimeObjects(clf).Build()
+			err := validateSetup(*cl, k8sClient)
+			Expect(err).To(Succeed())
+		})
+
+		It("should fail when CL(ANY_NS/instance) - This is an invalid use case", func() {
```
jcantrill, line 112: We should explicitly add another test  for CL & CLF  ANY_NS/instance.  This is a valid mCLF case.  I know you have test on line 124 which should cover this but lets add one anyway

internal/validations/clusterlogging/validate_clusterlogging_spec.go
```diff
@@ -26,3 +33,30 @@ func validateClusterLoggingSpec(cl v1.ClusterLogging) error {
 	}
 	return nil
 }
+
+//CL(openshift-logging/instance) only - This is a valid LEGACY use case
+//CL(openshift-logging/instance) & CLF(openshift-logging/instance) - This is a valid LEGACY usecase
+//CL(ANY_NS/OTHER) - This is an invalid use case
+//CLF(ANY_NS/ANY_NAME) - This is a valid mCLF use case
+//CL(ANY_NS/ANY_NAME) && CLF(ANY_NS/ANY_NAME) - This is a valid mCLF use case
+func validateSetup(cl v1.ClusterLogging, k8sClient client.Client) error {
```
jcantrill, line 42: where is this function tied into the entire validation cycle?

vparfonov: added here: https://github.com/openshift/cluster-logging-operator/pull/2394/files#diff-d4e83260483dc0e217046eeaa5c5951c2ee39e2989b1fb09faef6cfd5c452465R20 before spec validation

internal/validations/clusterlogging/validate_clusterlogging_spec_test.go
```diff
@@ -52,31 +57,93 @@ var _ = Describe("[internal][validations] ClusterLogging", func() {
...
+		})
+
+		It("should fail when CL(ANY_NS/OTHER_NAME) - This is an invalid use case", func() {
+			cl.Name = otherName
+			cl.Namespace = otherNamespace
+			err := validateSetup(*cl, k8sClient)
+			Expect(err).To(Not(Succeed()))
+		})
+
+		It("should pass when CL(ANY_NS/ANY_NAME) && CLF(ANY_NS/ANY_NAME) - This is an invalid use case", func() {
```
jcantrill, line 139: nit.  this is a valid usecase

jcantrill:
@vparfonov one nit but we will keep this on hold until after the release

jcantrill:
Manually merging. job was retired after PR was created

PR #2399, opened 2024-03-12
LOG-5132: Identify fields that should be omitted from pruning

Clee2691:
### Description
This PR ensures a fixed set of fields are not pruned from any log records and are required fields.
These fields are:
1. `.log_type`
2. `.message`
3. `.hostname` [^1]

[^1]: Only applies if a pipeline references `GoogleCloudLogging` as a destination, the `.hostname` field cannot be pruned from the log record.

> [!NOTE]  
> See upstream docs and API docs for more information.

This PR also refactors the `CloudWatch` functional test to make it easier to deploy.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5132

### DIFF
--- a/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
+++ b/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
@@ -986,6 +986,125 @@ var _ = Describe("Validate clusterlogforwarderspec", func() {
 			conds := clfStatus.Pipelines["someDefinedPipeline"]
 			Expect(conds).To(HaveCondition(loggingv1.ValidationCondition, true, loggingv1.ValidationFailureReason, "invalid: custom ClusterLogForwarders cannot forward to the `default` log store, unrecognized outputs: .?default.?"))
 		})
+
+		DescribeTable("gcl and prune filter", func(pruneSpec loggingv1.PruneFilterSpec) {
+			gclOutput := loggingv1.OutputSpec{
+
+				Name: "gcl-out",
+				Type: "googleCloudLogging",
+				OutputTypeSpec: loggingv1.OutputTypeSpec{
+					GoogleCloudLogging: &loggingv1.GoogleCloudLogging{BillingAccountID: "billingAccountID"},
+				},
+			}
+
+			pruneHost := loggingv1.FilterSpec{
+				Name: "prune",
+				Type: loggingv1.FilterPrune,
+				FilterTypeSpec: loggingv1.FilterTypeSpec{
+					PruneFilterSpec: &pruneSpec,
+				},
+			}
+
+			forwarderSpec.Outputs = []loggingv1.OutputSpec{gclOutput}
+			forwarderSpec.Filters = []loggingv1.FilterSpec{pruneHost}
+
+			forwarderSpec.Pipelines = []loggingv1.PipelineSpec{
+				{
+					Name:       "gclPruneHost",
+					OutputRefs: []string{"gcl-out"},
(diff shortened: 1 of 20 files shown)


internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
```diff
@@ -986,6 +986,125 @@ var _ = Describe("Validate clusterlogforwarderspec", func() {
...
+					FilterRefs: []string{"prune"},
+				},
+			}
+			clfStatus.Outputs.Set("gcl-out", condReady)
+
+			verifyPipelines("custom-clf-name", &forwarderSpec, clfStatus)
+			Expect(clfStatus.Pipelines).To(HaveLen(1))
+			Expect(clfStatus.Pipelines).To(HaveKey("gclPruneHost"))
+			conds := clfStatus.Pipelines["gclPruneHost"]
+			Expect(conds).To(HaveCondition(loggingv1.ConditionReady, false, loggingv1.ReasonInvalid, "googleCloudLogging cannot prune `.hostname` field.+"))
```
jcantrill, line 1025: Shouln't this HaveCondition of ConditionTrue ? since the condition exists (i.e. we have a failure)?  There was a bug in or condition creation that was not seting "ConditionTrue" correctly but my belief was that was since corrected

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
```diff
@@ -986,6 +986,125 @@ var _ = Describe("Validate clusterlogforwarderspec", func() {
...
+					OutputRefs: []string{"gcl-out"},
+					InputRefs:  []string{loggingv1.InputNameApplication},
+					FilterRefs: []string{"prune"},
+				},
+			}
+
+			clfStatus.Outputs.Set("gcl-out", condReady)
+
+			verifyPipelines("custom-clf-name", &forwarderSpec, clfStatus)
+			Expect(clfStatus.Pipelines["gclPruneHost"]).To(HaveCondition("Ready", true, "", ""))
```
jcantrill, line 1067: This is the contrary to the other comment.  I think this should be "ConditionFalse" because the condition is "there is an invalid prune filter" and it does not apply: ref https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ConditionStatus

test/framework/functional/output_cloudwatch.go
```diff
@@ -2,13 +2,99 @@ package functional
...
+)
+
+var (
+	cwlClient *cwl.Client
+	service   *v1.Service
+	route     *openshiftv1.Route
+)
+
+func (f *CollectorFunctionalFramework) AddCloudWatchOutput(b *runtime.PodBuilder, output logging.OutputSpec) error {
+	if err := f.createCloudWatchService(); err != nil {
```
jcantrill, line 33: This is a candidate for refactoring even more since @vparfonov  had similar changes for azure.  Functional tests run in the same container and have no need for routes or  services.  For Azure I think we did some hackery to modify the host file.  I don't see any reason the mock CW should run otherwise.

internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec_test.go
Clee2691: The condition here is `Type: ConditionReady` so I believe the condition should be false since it should be failing validation and thus is not in a ready state. If it was a `ValidationCondition` it would be `True`

PR #2404, opened 2024-03-13
LOG-5186: Refactor input filter for includes and excludes

cahartma:
### Description
Some local tests are still failing so doubtful this will pass all of CI yet, but need to get this moving, so here goes.     Please review and comment!

/cc @Clee2691 @vparfonov
/assign @jcantrill

Refactor the custom application include/exclude to change from:  
```
    inputs:
    - application:
        namespaces:
        - app-ns    
        excludeNamespaces:
        - test*
        containers:
          include:
          - nginx     
          exclude:
          - logger          
```
to 
```
    inputs:
    - application:
      - includes:
        - namespace: app-ns
          container: "*"
        - container: nginx
          namespace: "*"
      - excludes:
        - namespace: "test*"
          container: "*"
        - container: logger
          namespace: "*"        
```
### Acceptance Criteria:
- Verify when an include/exclude entry only defines "namespace" that it applies to all containers (e.g. *)
- Verify when an include/exclude entry only defines "container" that it applies to all namespaces (e.g. *)
- Verify multiple includes combinations results in logs included from each of the matches
- Verify multiple excludes combinations results in logs excluded from each of the matches
- Verify only includes that are not in the excludes list are matched.  (i.e. excludes take precedence and are run first)
- Verify the CEE documentation is updated to reflect the changes
- Verify the enhancement documentation is updated to reflect the changes
- Verify "namespaces" field as marked as deprecated
- Verify the entries of the "namespaces" field are added to the list of "includes" in the generated configuration
- Verify related/linked bugs are resolved

#### TODO:
- create upstream docs
- update CEE slides

### Links
- https://issues.redhat.com/browse/LOG-5186
- Related: 
  * https://issues.redhat.com/browse/LOG-5167
  * https://issues.redhat.com/browse/LOG-5162


api/logging/v1/input_types.go
```diff
@@ -48,26 +44,33 @@ type Application struct {
 	// +optional
 	ContainerLimit *LimitSpec `json:"containerLimit,omitempty"`
 
-	// Containers is the spec of containers to include and exclude when collecting logs.
-	// This effectively is an 'AND' with the other fields
-	// of this input definition. The subfields of containers supports globs.  The includes
-	// defined here are combined with the included namespaces to form a set of namespace
-	// and container permutations.  The same logic is applied to the excluded containers and namespaces
+	// Includes is the set of namespaces and containers to include when collecting logs.
+	// Note: infrastructure namespaces are still excluded for "*" values and each must specifically included.
```
jcantrill, line 48: What are we trying to say here?

api/logging/v1/input_types.go
```diff
@@ -48,26 +44,33 @@ type Application struct {
...
+	//
+	// +optional
+	Includes []IncludeExcludeSpec `json:"includes,omitempty"`
+
+	// Excludes is the set of namespaces and containers to ignore when collecting logs.
+	// Takes precedence over Includes option. The exclude patterns are applied after the attempt to glob everything in include.
```
jcantrill, line 54: I believe we can strike the last sentence and just say "Takes precedence over includes"

api/logging/v1/input_types.go
```diff
@@ -48,26 +44,33 @@ type Application struct {
...
 	//
 	// +optional
-	Containers *InclusionSpec `json:"containers,omitempty"`
+	Excludes []IncludeExcludeSpec `json:"excludes,omitempty"`
 }
 
-// InclusionSpec defines a set of similar resources for inclusion or exclusion
-type InclusionSpec struct {
+type IncludeExcludeSpec struct {
```
jcantrill, line 60: IMO the better name for this struct is something like "NamespaceContainerSpec" given that is what it defines

internal/generator/vector/conf/complex.toml
```diff
@@ -306,6 +306,7 @@ max_read_bytes = 3145728
 glob_minimum_cooldown_ms = 15000
 auto_partial_merge = true
 include_paths_glob_patterns = ["/var/log/pods/test-ns_*/*/*.log"]
+exclude_paths_glob_patterns = ["/var/log/pods/*/*/*.gz", "/var/log/pods/*/*/*.tmp", "/var/log/pods/default_*/*/*.log", "/var/log/pods/kube*_*/*/*.log", "/var/log/pods/openshift*_*/*/*.log"]
```
jcantrill, line 309: @Clee2691 noted this is unnecessary given the include will never match the exclude which was a change I made for the original implementation. If it is more challenging in the implementation, I can be swayed to leave as-is

internal/generator/vector/input/viaq_test.go
```diff
@@ -90,10 +97,23 @@ var _ = Describe("inputs", func() {
 		Entry("with an application that specs specific namespaces and exclude namespaces", logging.InputSpec{
 			Name: "my-app",
 			Application: &logging.Application{
-				Namespaces:        []string{"test-ns-foo", "test-ns-bar"},
```
jcantrill, line 93: Do we cover this usecase where  using deprecated namespaces field?  The variant here is would now include all containers from these NSs

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
```diff
@@ -130,6 +130,13 @@ func gatherPipelineInputs(clf loggingv1.ClusterLogForwarder) (sets.String, bool)
 				if hasInfraNamespaces(input.Application.Namespaces) {
 					inputTypes.Insert(loggingv1.InputNameInfrastructure)
 				}
+				if len(input.Application.Includes) > 0 {
+					for _, in := range input.Application.Includes {
+						if infraNamespaces.MatchString(in.Namespace) {
```
jcantrill, line 135: Isn't there a need to check if this NS is an infra namespace?  This is the companion check to line 130 is it not?

jcantrill (commented):
/hold

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
cahartma: line135 checks for infra?    Not sure I understand the question.

internal/generator/vector/conf/complex.toml
cahartma: I'd prefer to keep it there for all.   I wrote it to handle "*" but think it makes more sense to have it present.

api/logging/v1/input_types.go
cahartma: Okay done.   Was taken from vector docs.

api/logging/v1/input_types.go
cahartma: If you select to include "*" namespace, you still won't get infra namespaces.    You must specify them each individually.

internal/validations/clusterlogforwarder/validate_clusterlogforwarder_service_account.go
jcantrill: :+1: I missed the fact "infraNamespaces" is a RE

PR #2406, opened 2024-03-14
LOG-5216: CLO raises many 'error removing stale http input services'

Clee2691:
### Description
This PR fixes the errors arising from using `client.Client` to list services because of the change in namespaces watched. `client.Reader` is needed to list objects from specific namespaces because it bypasses the cache.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5216


jcantrill (commented):
/approve
/lgtm

PR #2409, opened 2024-03-14
LOG-5053: ensure inputs has right setted type before clfSpec migration 

vparfonov:
### Description
This PR addressed to fix issue during migration CLF with not properly configured `Type` in `ReceiverSpec`

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5053
- Enhancement proposal:

### DIFF
--- a/api/logging/v1/input_receiver_types.go
+++ b/api/logging/v1/input_receiver_types.go
@@ -16,7 +16,6 @@ const (
 type ReceiverSpec struct {
 
 	// Type of Receiver plugin.
-	// +kubebuilder:validation:Enum:=http;syslog
 	// +optional
 	Type string `json:"type"`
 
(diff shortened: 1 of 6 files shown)


jcantrill:
I don't believe this fully resolves the issue.  The error says:

```
updated validation is too restrictive: [].spec.inputs[0].receiver.type: Required value",
```
which I thought we fixed by stating type was optional:
```
// The fields of this struct define the set of known Receiver types.
type ReceiverSpec struct {

	// Type of Receiver plugin.
	// +kubebuilder:validation:Enum:=http;syslog
	// +optional
	Type string `json:"type"`

	// The ReceiverTypeSpec that handles particular parameters
	// +optional
	*ReceiverTypeSpec `json:",inline,omitempty"`
}
```
but I'm wondering if because we add a validation to be an Enum where empty is not valid, it fails.  I would recommend removing the validation comment and then we can set the type to one or the other like in your code or fail if it is anything other then "http" or "syslog"

vparfonov:
we checked it here: https://github.com/openshift/cluster-logging-operator/blob/master/internal/validations/clusterlogforwarder/inputs/validate.go#L52

PR #2422, opened 2024-03-27
LOG-5298: Refactor Viaq model to be a filter

jcantrill:
### Description
This PR:
* refactors the viaq data model to be a filter so that it is easier to swap out
* Moves all RH/viaq specific changes to the viaq filter
* Adds log_type, log_source to the source so we can make decisions in transforms
* Allows the ability to split some viaq transforms into configurable filters (e.g. loglevel, journal debug level, etc)
* blocked on #2463 
### Links
https://issues.redhat.com/browse/LOG-5298

/hold


internal/generator/vector/input/source.go
```diff
@@ -42,20 +38,20 @@ var (
...
+// NewSource creates an input adapter to generate config for ViaQ sources to collect logs excluding the
 // collector container logs from the namespace where the collector is deployed
-func NewViaQ(input logging.InputSpec, collectorNS string, resNames *factory.ForwarderResourceNames, op framework.Options) ([]framework.Element, []string) {
+func NewSource(input logging.InputSpec, collectorNS string, resNames *factory.ForwarderResourceNames, op framework.Options) ([]framework.Element, []string) {
 	els := []framework.Element{}
 	ids := []string{}
 	switch {
 	case input.Name == logging.InputNameApplication:
-		els, ids = NewViaqContainerSource(input, collectorNS, "", infraExcludes)
+		els, ids = NewContainerSource(input, collectorNS, "", infraExcludes, logging.InputNameApplication, logging.InfrastructureSourceContainer)
```
vparfonov, line 48: Maybe make sense add new or rename const because it can confuse Input is `application` but source type is `InfrastructureSourceContainer`

jcantrill: Agree. I'll address in a separate PR

PR #2435, opened 2024-04-17
LOG-5372: Enable log level for vector

Clee2691:
### Description
This PR enables configuration of the log level for the Vector collector through an annotation in the `ClusterLogForwarder` resource.

This introduces a new annotation: `observability.openshift.io/log-level`.
The log level can be set to one of `trace`, `debug`, `info`, `warn`, `error`, `off`.

Vector's default log level is `info`.

```
apiVersion: "logging.openshift.io/v1"
kind: ClusterLogForwarder
metadata:
  name: instance
  namespace: openshift-logging
  annotations:
    observability.openshift.io/log-level: "debug"
spec:
...
```

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5372

### DIFF
--- a/internal/collector/vector/visitors.go
+++ b/internal/collector/vector/visitors.go
@@ -8,9 +8,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func CollectorVisitor(collectorContainer *corev1.Container, podSpec *corev1.PodSpec, resNames *factory.ForwarderResourceNames, namespace string) {
+func CollectorVisitor(collectorContainer *corev1.Container, podSpec *corev1.PodSpec, resNames *factory.ForwarderResourceNames, namespace, logLevel string) {
 	collectorContainer.Env = append(collectorContainer.Env,
-		corev1.EnvVar{Name: "VECTOR_LOG", Value: "WARN"},
+		corev1.EnvVar{Name: "VECTOR_LOG", Value: logLevel},
 		corev1.EnvVar{Name: "KUBERNETES_SERVICE_HOST", Value: "kubernetes.default.svc"},
 		corev1.EnvVar{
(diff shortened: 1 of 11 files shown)


internal/collector/vector/visitors.go
```diff
@@ -8,9 +8,9 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func CollectorVisitor(collectorContainer *corev1.Container, podSpec *corev1.PodSpec, resNames *factory.ForwarderResourceNames, namespace string) {
+func CollectorVisitor(collectorContainer *corev1.Container, podSpec *corev1.PodSpec, resNames *factory.ForwarderResourceNames, namespace, logLevel string) {
 	collectorContainer.Env = append(collectorContainer.Env,
-		corev1.EnvVar{Name: "VECTOR_LOG", Value: "WARN"},
```
jcantrill, line 13: the change from warn to default will bump the log output of the collector making it more verbose.  I wonder if that is an issue?

jcantrill (commented):
/hold
/approve

Other then the one question, lgtm.  Do we know for certain annotation edits will trigger a reconciliation?

internal/collector/vector/visitors.go
Clee2691: I did that because vector's docs state the default is info. I can make the default to warn instead so it will keep the same behavior as we previously had.

vparfonov:
@jcantrill and @Clee2691 We had issue for making Vector less talkative https://github.com/openshift/cluster-logging-operator/pull/1888

jcantrill (commented):
/lgtm

PR #2438, opened 2024-04-19
LOG-5381: add alerts metric and dashboard panels

vparfonov:
### Description
This PR introduces monitoring and alerting for Vector collector output buffer size.

#### Alert: 
Added alerts to notify when Vector output buffer disk usage reaches a critical threshold.
buffer-alert

####  Metrics Dashboards (2 panels):
Panel 1: Shows the absolute size of the Vector output buffer, summed by hostname, namespace, and CLF instance name 
buffer-metric-1
Panel 2: Shows the percentage of buffer size relative to the total disk space available on the node.
buffer-metric-2

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:


config/prometheus/collector_alerts.yaml
```diff
@@ -85,6 +85,17 @@ spec:
       labels:
         service: visualization
         severity: Warning
+    - alert: DiskBufferUsage
+      annotations:
+        message: "Collectors potentially consuming too much node disk, {{ $value }}% "
+        summary: "Detected consuming too much node disk on $labels.hostname host"
```
jcantrill, line 91: Does this need curly brackets?  The sample alert in the PR shows it as-is here

docs/administration/collector-metrics-and-alerts.adoc
```diff
@@ -0,0 +1,33 @@
+= Collector Metrics and Alerts
```
jcantrill, line 1: We may need to add a note here this only applies to "infra" namespaces and we have no details for collectors deployed to non-infra NS.  Unless you were able to learn if it is possible to collect metrics from non-infra namespaces by some enablement

jcantrill (approved):

docs/administration/collector-metrics-and-alerts.adoc
vparfonov: added in docs how to enable

PR #2445, opened 2024-04-24
[release-5.9] LOG-5156: Include rotated files in kube_log glob patterns

cahartma:
### Description
Simple change, but need to make sure it's complete.    
Vector config include/exclude glob patterns all end with `*.log`.  This results in rotated infra log files being sent to the app index.   This change replaces `*.log` with `*` when building the file patterns for each.    
This should definitely avoid any rotated files from being collected as the wrong index, but may also result in collecting unwanted files.   the `.gz` and `.tmp` extensions are still added to exclude in all cases I tested, but we should be sure the star is not picking up others.

/cc @Clee2691 @vparfonov
/assign @jcantrill

/cherry-pick master

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5156


internal/generator/vector/conf/complex.toml
```diff
@@ -128,8 +128,8 @@ type = "kubernetes_logs"
 max_read_bytes = 3145728
 glob_minimum_cooldown_ms = 15000
 auto_partial_merge = true
-include_paths_glob_patterns = ["/var/log/pods/default_*/*/*.log", "/var/log/pods/kube*_*/*/*.log", "/var/log/pods/openshift*_*/*/*.log"]
-exclude_paths_glob_patterns = ["/var/log/pods/*/*/*.gz", "/var/log/pods/*/*/*.tmp", "/var/log/pods/openshift-logging_*/gateway/*.log", "/var/log/pods/openshift-logging_*/loki*/*.log", "/var/log/pods/openshift-logging_*/opa/*.log", "/var/log/pods/openshift-logging_elasticsearch-*/*/*.log", "/var/log/pods/openshift-logging_kibana-*/*/*.log", "/var/log/pods/openshift-logging_logfilesmetricexporter-*/*/*.log"]
+include_paths_glob_patterns = ["/var/log/pods/default_*/*/*", "/var/log/pods/kube*_*/*/*", "/var/log/pods/openshift*_*/*/*"]
```
jcantrill, line 131: I have concerns this changes behavior.  This says..."read anything in these directories" but we only wish to explicitly read the "[0-9]*.log" files.  We also do not intend to read rolled over files since that implies we may read them twice.    "*/*/*" is appropriate IMO for excludes but not includes

cahartma: Agreed, if all we want is to exclude rotated files. I read the ticket differently, but see it now.  thanks.  Will make the change and let you know.

PR #2448, opened 2024-04-25
LOG-5384: Drop support for fluentd collector

jcantrill:
### Description
This PR:
* Removes support for fluent
* leaves TODO to fix up fluent test that should also apply to vector
* Should simply introduction of new CLF api

### Links
* https://issues.redhat.com/browse/LOG-5384


internal/collector/collector.go
```diff
@@ -153,7 +138,6 @@ func (f *Factory) NewPodSpec(trustedCABundle *v1.ConfigMap, forwarderSpec loggin
 
 	if f.isDaemonset {
 		podSpec.Volumes = append(podSpec.Volumes,
-			v1.Volume{Name: logContainers, VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: logContainersValue}}},
```
cahartma, line 156: Why remove for DS?  Isn't this meant to be for deployment only?

test/framework/functional/output_forward.go
```diff
@@ -95,7 +93,7 @@ func (f *CollectorFunctionalFramework) addForwardOutputWithConf(b *runtime.PodBu
 	}
 
 	log.V(2).Info("Adding container", "name", name)
-	b.AddContainer(name, utils.GetComponentImage(constants.FluentdName)).
+	b.AddContainer(name, "quay.io/openshift-logging/fluentd:5.9").
```
cahartma, line 96: nit, but still used in a few places, so why not add to constants package?   Also, did you remove the `FluentdName` and `FluentdImageEnvVar` from constants file?

internal/collector/collector.go
jcantrill: `/var/log/containers` path is deprecated in favor of `/var/log/pods` and has been for a long time.  We have no reason to mount it

test/framework/functional/output_forward.go
jcantrill: Updated where I could

alanconway (approved):
/lgtm

PR #2463, opened 2024-05-01
LOG-5381: fix bundle to include alert

jcantrill:
### Description
This PR:
* fixes the bundle for the changes introduced by #2438 

### Links
* https://issues.redhat.com//browse/LOG-5381

cc @vparfonov @Clee2691

### DIFF
--- a/bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
+++ b/bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
@@ -103,6 +103,18 @@ spec:
       labels:
         service: visualization
         severity: Warning
+    - alert: DiskBufferUsage
+      annotations:
+        message: 'Collectors potentially consuming too much node disk, {{ $value }}% '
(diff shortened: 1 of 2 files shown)


bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -103,6 +103,18 @@ spec:
       labels:
         service: visualization
         severity: Warning
+    - alert: DiskBufferUsage
+      annotations:
+        message: 'Collectors potentially consuming too much node disk, {{ $value }}% '
+        summary: Detected consuming too much node disk on $labels.hostname host
+      expr: "(label_replace(sum by(hostname) (vector_buffer_byte_size{component_kind='sink',
```
Clee2691, line 110: The expression is not in the format as the other alerts. Does that matter?

jcantrill: In what way?  I just ran "make bundle" so if the format is off then we need to understand why

PR #2478, opened 2024-05-08
LOG-5515: Refactors AWS Cloudwatch output to observability API

jcantrill:
### Description
This PR:
* Refactors AWS Cloudwatch output to observability API

### Links
* blocked by #2477

cc @Viaq/team-log-collection


internal/collector/cloudwatch.go
```diff
@@ -76,10 +75,11 @@ func AddWebIdentityTokenEnvVars(collector *v1.Container, output logging.OutputSp
 			Name:  constants.AWSRegionEnvVarKey,
 			Value: output.Cloudwatch.Region,
 		},
-		v1.EnvVar{
-			Name:  constants.AWSRoleArnEnvVarKey,
-			Value: cloudwatch.ParseRoleArn(secret),
-		},
+		// TODO: FIX ME
```
cahartma, line 78: What's the issue here?

internal/generator/vector/output/azuremonitor/azuremonitor.go
```diff
@@ -45,43 +43,33 @@ shared_key = "{{.SharedKey}}"
...
-			common.NewRequest(id, strategy),
-		},
-		TLSConf(id, o, secret, op),
-	)
-}
-
-func Output(id string, o logging.OutputSpec, inputs []string, secret *corev1.Secret, op framework.Options) framework.Element {
 	azm := o.AzureMonitor
-	return AzureMonitor{
+	e := AzureMonitor{
```
cahartma, line 54: nit:   more explicit name
```suggestion
	output := AzureMonitor{
```

internal/generator/vector/output/cloudwatch/cloudwatch_test.go
```diff
@@ -0,0 +1,221 @@
...
+				vectorTLSSecret: {
+					Data: map[string][]byte{
+						constants.ClientCertKey:      []byte("-- crt-- "),
+						constants.ClientPrivateKey:   []byte("-- key-- "),
+						constants.TrustedCABundleKey: []byte("-- ca-bundle -- "),
+					},
+				},
+				secretWithCredentials: {
+					Data: map[string][]byte{
+						constants.AWSCredentialsKey:  []byte("[default]\nrole_arn = " + roleArn + "\nweb_identity_token_file = /var/run/secrets/token"),
```
cahartma, line 98: I'm very confused by this (and see it in our 5.9 as well).   Why is there a role specified in a TLS secret?    Vector requires the env vars to be set in order to use the role.   I can't remember why it wouldn't work without the role and token path via ENV vars, but I'm fairly sure this TLS has anything to do with roles.  Am I confused?

internal/collector/cloudwatch.go
jcantrill: These tests fail because we are retrieving the info differently then we did in the past.  This PR only updates the output wiring where I changed the method signature.  I intentionally disabled things in a bunch of places to minimize the focus of the change so we'll need to revisit things like this once we come back around to the deployment aspect of our changes

internal/generator/vector/output/cloudwatch/cloudwatch_test.go
jcantrill: Conceptionally this is the same as before;  you spec the role and we read it from a secret and populate the env var.  This specific case is just convenience.  The new API allows for the admin to tell us which secret and key to us but from a config point of view, we don't care.  The tests were parsing a key out of these files are hard-coding them into the config

internal/generator/vector/output/azuremonitor/azuremonitor.go
jcantrill: This was part of the azure PR that already merged

cahartma (commented):
/lgtm

PR #2481, opened 2024-05-09
LOG-5515: Refactor http output to use OBS api

jcantrill:
### Description
This PR:
* Refactors Elasticsearch output to use observability API

### Links
* blocked by #2479 

/hold


internal/generator/vector/output/elasticsearch/elasticsearch.go
```diff
@@ -48,203 +43,55 @@ api_version = "v{{ .Version }}"
...
-			ComponentID: id,
-		})
-		if common.HasUsernamePassword(secret) {
-			hasBasicAuth = true
-			up := UserNamePass{
-				Username: common.GetFromSecret(secret, constants.ClientUsername),
-				Password: common.GetFromSecret(secret, constants.ClientPassword),
-			}
-			conf = append(conf, up)
+func BasicAuth(id string, o obs.OutputSpec, secrets helpers.Secrets) Element {
```
Clee2691, line 85: You refactored `HTTPAuth` into `common/auth/auth.go`, doesn't ` HTTPAuth()`  in that file do this? Shouldn't you use that instead of refactoring this method?

internal/generator/vector/output/elasticsearch/elasticsearch.go
```diff
@@ -48,203 +43,55 @@ api_version = "v{{ .Version }}"
...
-		BasicAuth(id, o, secret),
-	)
+	outputs := []Element{
+		sink,
+		common.NewAcknowledgments(id, strategy),
+		common.NewBatch(id, strategy),
+		common.NewBuffer(id, strategy),
+		common.NewRequest(id, strategy),
+		tls.New(id, o.TLS, secrets, op),
+		BasicAuth(id, o, secrets),
```
Clee2691, line 67: Should this be using `HTTPAuth` method from `common/auth/auth.go`?

jcantrill: pushed into separate PR

PR #2483, opened 2024-05-10
LOG-2811: Add lokistack output type to cluster forwarder

Clee2691:
### Description
- Refactored `Loki` to use observability API
- Added `Loki` output to API
- Migrates `LokiStack` output to `Loki` outputs crafting the URL with the tenant (current behavior of LokiStack)

/cc 
/assign 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-2811


api/observability/v1/clusterlogforwarder_types.go
```diff
@@ -171,16 +171,8 @@ type SecretKey struct {
...
-	//
-	// +kubebuilder:validation:Optional
-	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Key Name",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
-	Key string `json:"key,omitempty"`
-
-	// Use Secret if the value should be sourced from a Secret in the same namespace.
-	//
 	// +kubebuilder:validation:Optional
-	Secret *corev1.LocalObjectReference `json:"secret,omitempty"`
+	*SecretKey `json:",inline"`
```
alanconway: As per slack comments I think this should be
```
Secret *SecretKey `json:"secret, omitempty`
```
Slack comment https://redhat-internal.slack.com/archives/C03744G7KE1/p1715294246066419?thread_ts=1715267662.470479&cid=C03744G7KE1

alanconway (changes requested):
Apart from a comment on the auth debate, this looks good. 
Generating extra outputs and pipelines is somewhat complex but I think it is the only option right now - vector doesn't support any way to vary the base URL per message.
I'd like the code to be simpler, but I don't have any good idea how to do that right now....

api/observability/v1/clusterlogforwarder_types.go
Clee2691: I think we can leave it as is for now with @jcantrill 's auth changes.

internal/generator/vector/helpers/secrets.go
```diff
@@ -38,21 +33,28 @@ func (s Secrets) Value(key *v1.SecretKey) []byte {
...
-func (s Secrets) AsStringFromBearerToken(key *v1.BearerToken) string {
+func (s Secrets) AsStringFromBearerToken(key *obs.BearerToken) string {
 	if key.Secret != nil {
-		return s.AsString(&v1.SecretKey{
+		return s.AsString(&obs.SecretKey{
 			Secret: key.Secret,
 			Key:    key.Key,
 		})
+		// We reconcile SA token secret, so name and key will be known
+	} else if key.ServiceAccount != nil {
```
jcantrill, line 51: nit.. we can remove the "if" since the previous statement returns early

internal/generator/vector/output/loki/loki.go
```diff
@@ -132,43 +133,43 @@ func New(id string, o logging.OutputSpec, inputs []string, secret *corev1.Secret
...
-		},
-		TLSConf(id, o, secret, op),
-		BasicAuth(id, o, secret),
-		BearerTokenAuth(id, o, secret),
-	)
+
+	confTLS := tls.New(id, o.TLS, secrets, op)
+	return []Element{
+		CleanupFields(componentID, inputs),
+		viaq.DedotLabels(dedottedID, []string{componentID}),
```
jcantrill, line 140: drop this transform which is what I have done for the others.  I'm banking on us saying the "v2 viaq" is what we now support

internal/generator/vector/output/loki/loki_conf_test.go
```diff
@@ -252,34 +348,6 @@ var _ = Describe("Generate vector config", func() {
...
-						"token": []byte("token-for-internal-loki"),
-					},
-				},
-			},
-			ExpectedConf: withDefaultLogcollectorToken,
-		}),
-	)
-})
-
 func TestVectorConfGenerator(t *testing.T) {
```
jcantrill, line 351: Can we remove this and add a "suite_test.go" file like other packages

internal/logstore/observability/lokistack/lokistack.go
```diff
@@ -0,0 +1,298 @@
+package lokistack
```
jcantrill, line 1: Do these need to go in the test package? migration package?  I'm not following why we are reconciling anything for any logstore

internal/logstore/observability/lokistack/lokistack_test.go
```diff
@@ -0,0 +1,231 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+		}
+	})
+
+	It("single tenant - single lokiStack output", func() {
```
jcantrill, line 38: Can we write this as... `It("should .... ` with the expectation of the outcome

test/observability/helpers/generator.go
```diff
@@ -0,0 +1,42 @@
+package helpers
```
jcantrill, line 1: Consider reviewing the azure_test.go which is the pattern I applied to the other outputs which makes use of a matcher that takes `Element`.  You may find this file along with some setup code in the actual test is not required

api/observability/v1/output_types.go
```diff
@@ -612,6 +618,45 @@ type LokiStack struct {
...
+	// Loki queries can also query based on any log record field (not just labels) using query filters.
+	//
+	// +kubebuilder:validation:Optional
+	LabelKeys []string `json:"labelKeys,omitempty"`
+
+	// TenantKey is a meta-data key field to use as the TenantID,
+	// For example: 'TenantKey: kubernetes.namespace_name` will use the kubernetes
+	// namespace as the tenant ID.
+	//
+	// +kubebuilder:validation:Optional
```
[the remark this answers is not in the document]
Clee2691: This will be `.log_type` as the loki output formats the tenant with `{{}}`

internal/logstore/observability/lokistack/lokistack.go
Clee2691: I've removed this and moved those specific functions to the migration package for lokistack

internal/logstore/observability/lokistack/lokistack_test.go
Clee2691: I moved these tests to the loki migration tests and used table tests instead.

test/observability/helpers/generator.go
Clee2691: I've refactored the loki tests accordingly.

api/observability/v1/output_types.go
```diff
@@ -613,6 +619,46 @@ type LokiStack struct {
...
+	LabelKeys []string `json:"labelKeys,omitempty"`
+
+	// TenantKey is a meta-data key field to use as the TenantID,
+	// For example: 'TenantKey: kubernetes.namespace_name` will use the kubernetes
+	// namespace as the tenant ID.
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:default:=.log_type
```
jcantrill, line 658: I'm fairly certain we still require the `{{}}` since that is the indicator of templating for vector.  I anticipate this is the indicator as well to say its a dynamic versus static value.  We likely need to add more validation to this and other places that have a "tenant" value (e.g. index)

internal/generator/vector/output/loki/loki.go
```diff
@@ -223,105 +221,28 @@ func formatLokiLabelValue(value string) string {
...
-func Labels(id string, o logging.OutputSpec) Element {
+func Labels(id string, o obs.OutputSpec) Element {
 	return LokiLabels{
 		ComponentID: id,
 		Labels:      lokiLabels(o.Loki),
 	}
 }
 
-func Tenant(l *logging.Loki) Element {
+func Tenant(l *obs.Loki) Element {
```
jcantrill, line 239: I believe allowing the API to have curly's will alter the impl here

internal/migrations/observability/migrate_lokiStack.go
```diff
@@ -0,0 +1,160 @@
...
+	outputs, pipelines, migrationConditions = processForwarderPipelines(spec)
+
+	spec.Outputs = outputs
+	spec.Pipelines = pipelines
+
+	return spec, migrationConditions
+}
+
+// hasLokiStackOutput checks if a pipeline includes a lokistack output
+func hasLokiStackOutput(outputMap map[string]*obs.OutputSpec, outputRefs []string) (bool, string) {
```
jcantrill, line 27: Its possible we could have multiple lokistacks

Clee2691: I've changed migrations to account for this as well as added a bunch of tests for this.

internal/migrations/observability/migrate_lokiStack.go
```diff
@@ -0,0 +1,167 @@
...
+	outputs, pipelines, migrationConditions = processForwarderPipelines(spec)
+
+	spec.Outputs = outputs
+	spec.Pipelines = pipelines
+
+	return spec, migrationConditions
+}
+
+// hasLokiStackOutput checks if a pipeline includes a lokistack output
+func hasLokiStackOutput(outputMap map[string]*obs.OutputSpec, outputRefs []string) *sets.String {
```
jcantrill, line 28: nit.  since we need to change the other thing maybe make this plural?

jcantrill (commented):
2 small nits othewise lgtm

internal/migrations/observability/migrate_lokiStack.go
Clee2691: I changed the name of this function to better represent what it does: `findLokiStackOutputs()`

jcantrill (commented):
/lgtm

PR #2487, opened 2024-05-10
LOG-5515: Refactor kafka to use OBS API

jcantrill:
### Description
This PR:
* Refactors Kafka output to use observability API

### Links


internal/generator/vector/output/kafka/kafka.go
```diff
@@ -52,34 +48,45 @@ func (k *Kafka) SetCompression(algo string) {
...
+	}
+	elements := []Element{
+		sink,
+		Encoding(id, op),
+		common.NewAcknowledgments(id, strategy),
+		common.NewBatch(id, strategy),
+		common.NewBuffer(id, strategy),
+		SASLConf(id, o.Kafka.Authentication, secrets),
+	}
+	elements = append(elements,
```
Clee2691, line 82: Should we keep using the `MergeElements()` function?

jcantrill: :man_shrugging:   I'm indifferent to its use.  It's not any cleaner IMO but pleas use it when you feel it's necessary

PR #2489, opened 2024-05-10
LOG-5515: Refactor syslog to use OBS API

jcantrill:
### Description
This PR:
* Refactors syslog output to use observability API

### Links

### DIFF
--- a/internal/generator/vector/output/syslog/syslog.go
+++ b/internal/generator/vector/output/syslog/syslog.go
@@ -2,7 +2,8 @@ package syslog
 
 import (
 	"fmt"
-	"github.com/openshift/cluster-logging-operator/internal/generator/vector/filter/openshift/viaq"
+	obs "github.com/openshift/cluster-logging-operator/api/observability/v1"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/output/common/tls"
 	"net/url"
 	"regexp"
 	"strings"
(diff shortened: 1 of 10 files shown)


internal/generator/vector/output/syslog/syslog.go
```diff
@@ -128,31 +121,19 @@ func Output(id string, o logging.OutputSpec, inputs []string, secret *corev1.Sec
...
+func Encoding(id string, o obs.OutputSpec) Element {
 	return SyslogEncoding{
-		ComponentID:  id,
-		RFC:          RFC(o.Syslog),
-		Facility:     Facility(o.Syslog),
-		Severity:     Severity(o.Syslog),
-		AppName:      AppName(o.Syslog),
-		ProcID:       ProcID(o.Syslog),
-		MsgID:        MsgID(o.Syslog),
-		Tag:          Tag(o.Syslog),
```
Clee2691, line 140: What happened to the `Tag`, `LogSource`, and `PayloadKey`?

jcantrill: > What happened to the `Tag`, `LogSource`, and `PayloadKey`?

I removed them based upon [a conversation](https://redhat-internal.slack.com/archives/C03744G7KE1/p1715180838236909?thread_ts=1714754608.783289&cid=C03744G7KE1) where I wish to move away from any special behavior, but I'm questioning if this was the right call.  I actually am wondering if it makes sense to not offer them but if we should be doing some transformation to fill in these fields based record content, or continue to offer them as-is and allow the admin to choose. We should discuss.

PR #2497, opened 2024-05-16
Fix deploy Make target

xperimental:
### Description

Fixes `make deploy` by removing the `deploy-elasticsearch-operator` target which does not exist anymore.

/cc @cahartma


jcantrill (commented):
/lgtm

PR #2509, opened 2024-05-22
LOG-5462: Do not deploy logging ConsolePlugin if already managed by COO

xperimental:
### Description

Currently the logging-view-plugin is managed by CLO and deployed either when the ClusterLogging references a LokiStack as default storage or has a special annotation. In the future the logging UI will be managed by the Cluster Observability Operator (see rhobs/observability-operator#477). This PR makes CLO aware of the COO managing the logging UI, so that both operators do not interfere with each other.

### Links

- JIRA: LOG-5462


periklis (commented):
LGTM

PR #2515, opened 2024-05-23
LOG-5571: Initially wire collector reconcile into obs controller

jcantrill:
### Description
This PR:
* initially wires the collector deployment based upon obs into the new controller

### Links
* https://issues.redhat.com/browse/LOG-5571


internal/api/observability/output_types.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	return names.UnsortedList()
+}
+
+// SecretNames returns a unique set of unordered secret names
+func (outputs Outputs) SecretNames() []string {
+	secrets := set.New[string]()
+	for _, o := range outputs {
+		if o.TLS != nil {
+			secrets.Insert(SecretsForTLS(o.TLS.TLSSpec)...)
```
vparfonov, line 35: Should we `continue` here to next iteration?

cahartma: Not the way I read it.   We still need any other secrets that are non-TLS, correct?

internal/controller/observability/collector.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	isDaemonSet := ShouldDeployAsDaemonSet(clf.Annotations, clf.Spec.Inputs)
+	log.V(3).Info("Deploying as DaemonSet", "isDaemonSet", isDaemonSet)
+	factory := collector.New(collectorConfHash, clusterID, clf.Spec.Collector, secrets, clf.Spec, resourceNames, isDaemonSet, LogLevel(clf.Annotations))
+	if err = factory.ReconcileCollectorConfig(noOpEventRecorder, k8Client, k8Reader, clf.Namespace, collectorConfig, ownerRef); err != nil {
+		log.Error(err, "collector.ReconcileCollectorConfig")
+		return
+	}
+
+	reconcileDeployment := factory.ReconcileDaemonset
```
vparfonov, line 84: maybe other name will be better here: `reconciler` or `reconcileWorkloads`?

PR #2518, opened 2024-05-28
LOG-5572: Update obs conditions to meet enhancement

jcantrill:
### Description
This PR:
* modifies the obs status spec to meet the enhancement doc
* Makes status a list of conditions to allow use of console annotations

### Links
 https://issues.redhat.com/browse/LOG-5572

cc @vparfonov @Clee2691 @cahartma


api/observability/v1/conditions.go
```diff
@@ -14,10 +14,6 @@ limitations under the License.
 
 package v1
 
-import (
```
vparfonov, line 17: only the declaration of constants remained in this file

jcantrill: Correct.  For now I'm certain these are still required.  The re-definitions follows the pattern recommended for API conditions and allows us to use the existing kube types

PR #2521, opened 2024-05-29
Fix RBAC for new API resources in controller

xperimental:
### Description

The aim of this PR is to fix the automatic generation of the RBAC role from the kubebuilder annotations. The following steps were taken:

- Clean up duplicate RBAC definition and project name
- Fix name in revert script
- Fix operator name in default channel config
- Fix access to clusterversions on startup

One open question is whether the annotations should be grouped in a single place or if it's fine for them to live "near where they are used" like I did with the added annotation for `consoleplugins` in `version/cluster.go`.

/cc @jcantrill


config/rbac/role.yaml
```diff
@@ -126,6 +132,32 @@ rules:
   - oauthclients
   verbs:
   - '*'
+- apiGroups:
+  - observability.openshift.io
+  resources:
+  - clusterlogforwarders
+  verbs:
```
vparfonov, line 139: can we use "*"?

xperimental:
I think I'm tending towards "in one place", but I would not put it into the `api/...` packages, because I don't think the RBAC for the operator is technically part of the API.

Some of the existing `kubebuilder:rbac` annotations are already in `internal/controller`. For me, this does make more sense as a place for these annotations. I think I would vote for either putting all of them on the new "observability controller", split them by controller (there's probably some overlap though) or put it in a new file (for example `internal/controller/rbac.go`).

---

> I tested PR need to add next permissions:

I can see annotations for both these resources exist already, they are just missing the `list` verb. I've added the verb to the existing annotation.

jcantrill:
> ... "in one place" ... put it in a new file (for example internal/controller/rbac.go)

This is my vote :+1:

xperimental:
@jcantrill @vparfonov Rebased and updated based on last comments again. I ran it on a cluster and it looks like startup is working as intended again.

internal/controller/kubebuilder_rbac.go
```diff
@@ -0,0 +1,32 @@
...
+// +kubebuilder:rbac:groups=console.openshift.io,resources=consolelinks;consoleexternalloglinks;consoleplugins;consoleplugins/finalizers,verbs=get;create;update;delete
+// +kubebuilder:rbac:groups=logging.openshift.io,resources=*,verbs=*
+// +kubebuilder:rbac:groups=core,resources=pods;pods/exec;services;endpoints;persistentvolumeclaims;events;configmaps;secrets;serviceaccounts;services/finalizers,verbs=*
+// +kubebuilder:rbac:groups=route.openshift.io,resources=routes;routes/custom-host,verbs="*"
+// +kubebuilder:rbac:groups=apps,resources=deployments;daemonsets;replicasets;statefulsets,verbs=*
+// +kubebuilder:rbac:groups=authentication.k8s.io,resources=tokenreviews;subjectaccessreviews,verbs=create
+// +kubebuilder:rbac:groups=authorization.k8s.io,resources=subjectaccessreviews,verbs=create
+// +kubebuilder:rbac:groups=batch,resources=cronjobs,verbs=*
+// +kubebuilder:rbac:groups=config.openshift.io,resources=proxies;infrastructures,verbs=get;list;watch
+// +kubebuilder:rbac:groups=console.openshift.io,resources=consolelinks;consoleexternalloglinks,verbs=get;create;update;delete
```
vparfonov, line 15: some duplication with line: 6

internal/controller/kubebuilder_rbac.go
```diff
@@ -0,0 +1,32 @@
...
+// +kubebuilder:rbac:groups=core,resources=pods;pods/exec;services;endpoints;persistentvolumeclaims;events;configmaps;secrets;serviceaccounts;serviceaccounts/finalizers;services/finalizers;namespaces,verbs=*
+// +kubebuilder:rbac:groups=logging.openshift.io,resources=*,verbs=*
+// +kubebuilder:rbac:groups=monitoring.coreos.com,resources=prometheusrules;servicemonitors,verbs=*
+// +kubebuilder:rbac:groups=networking.k8s.io,resources=networkpolicies,verbs=create;delete
+// +kubebuilder:rbac:groups=oauth.openshift.io,resources=oauthclients,verbs=*
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=clusterroles;clusterrolebindings,verbs=*
+// +kubebuilder:rbac:groups=rbac.authorization.k8s.io,resources=roles;rolebindings,verbs=*
```
vparfonov, line 22: can be combined with line: 21

xperimental:
Review comments have been fixed, but CI seems to be broken today. Maybe it'll work later today/tomorrow :crossed_fingers:

PR #2523, opened 2024-05-30
Remove unused function

m1kola:
### Description

`OPERATOR_CONDITION_NAME` is being injected by operator lifecycle manager into operator's deployments so that operators can know the name of the relevant `OperatorCondition` CR for cases when operators want to report some conditions or block their upgrades.

It seems like logging operator has a function which reads `OPERATOR_CONDITION_NAME`, but the function is never used. I suspect that there was previously reliance on `OperatorCondition` which is gone, but this function was left behind.

/cc jcantrill
/assign alanconway

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:

### DIFF
--- a/version/cluster.go
+++ b/version/cluster.go
@@ -2,31 +2,17 @@ package version
 
 import (
 	"context"
-	"fmt"
+
 	log "github.com/ViaQ/logerr/v2/log/static"
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-logging-operator/internal/hostedcontrolplane"
-	"os"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-)
-
-const (
-	operatorConditionName = "OPERATOR_CONDITION_NAME"
 )
 
 var (
 	clusterVersion string
 	clusterID      string
 )
-
-// OperatorVersion get clo operator version from OPERATOR_CONDITION_NAME ENV variable
-func OperatorVersion() (string, error) {
-	operatorVersion, found := os.LookupEnv(operatorConditionName)
-	if !found {
-		return "", fmt.Errorf("%s must be set", operatorConditionName)
-	}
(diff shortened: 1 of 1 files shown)


jcantrill:
/hold

@xperimental The operator version was part of the published telemetry info.  Is that still the case? Seems it was used in [main](https://github.com/openshift/cluster-logging-operator/commit/085bdc1fd735d430a59a1ae92e24f18284c6e956#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R192)

m1kola:
@jcantrill as far as I see it is currently not used at all, but I assume you want to start using it again.

How important is it to determine version of the operator at runtime? Can you use `ldflags` to set version at build time? E.g. `-ldflags="-X 'main.Version=v1.0.0'"`?

/cc @xperimental

xperimental:
Yes, this was used in the telemetry code which I temporarily removed recently. I did not give this code a thought at the time.

The function already exists in previous versions, but I think it always was only meant as an "override". I can't find a mention of the environment variable anywhere in this repository or midstream though, so it might be an even older remnant.

So to me it looks as if it's probably fine to remove the unused code / env var and re-do this with the new telemetry implementation later.

m1kola:
> I can't find a mention of the environment variable anywhere in this repository or midstream though, so it might be an even older remnant.

OLM injects this env var into deployment. Note however that there is no guarantee that value of the env var will contain version. In current implementation it does

A bit of background: I submitted this PR because noticed that this function is unused while researching usage of `OperatorCondition` CR in operators. We are considering removing it in the next major version of OLM or replacing it with something else

jcantrill:
Based upon the discussion it seems there is supported reason to remove at this time.  Allowing it to merge

PR #2528, opened 2024-05-31
LOG-5604: Add input validation

jcantrill:
### Description
This PR:
* adds input validation
* Drops validations that should be addressed by API annotations

### Links
* https://issues.redhat.com/browse/LOG-5604

cc @vparfonov @Clee2691


internal/validations/observability/inputs/application.go
```diff
@@ -0,0 +1,57 @@
...
+			message = fmt.Sprintf(message, args...)
+		}
+		return internalobs.NewCondition(obs.ValidationCondition,
+			metav1.ConditionTrue,
+			reason,
+			message,
+		)
+	}
+
+	if spec.Application == nil {
```
vparfonov, line 32: this behavior not agreed with `//+kubebuilder:validation:Optional` in `InputSpec`

internal/validations/observability/inputs/receivers.go
```diff
@@ -0,0 +1,36 @@
...
+		return []metav1.Condition{
+			internalobs.NewCondition(obs.ValidationCondition,
+				metav1.ConditionTrue,
+				obs.ReasonValidationFailure,
+				message,
+			),
+		}
+	}
+
+	if spec.Receiver == nil {
```
vparfonov, line 25:
```suggestion
	if spec.Receiver.Type == obs.ReceiverTypeHTTP {
    if spec.Receiver.HTTP == nil {
        return newCond(fmt.Sprintf("%s has nil HTTP receiver spec", spec.Name))
    }

    if spec.Receiver.HTTP.Format == "" {
        return newCond(fmt.Sprintf("%s does not specify a format", spec.Name))
    }
 }
```

internal/validations/observability/inputs/application.go
jcantrill: @vparfonov i don't understand what you are expressing. AFAIK there is no way to require a type and enforce the associated spec

jcantrill (commented):
/hold cancel

PR #2529, opened 2024-05-31
LOG-5605: Add input secret validation

jcantrill:
### Description
This PR:
* adds input secret (and configmap) validation for receivers
* Adds utility functions to list keys from an inputTLS spec

### Links
* https://issues.redhat.com/browse/LOG-5605
* blocked by #2528 

cc @vparfonov @Clee2691

### DIFF
--- a/internal/validations/observability/common/validate.go
+++ b/internal/validations/observability/common/validate.go
@@ -1,0 +1,45 @@
+package common
+
+import (
+	"fmt"
+	obsv1 "github.com/openshift/cluster-logging-operator/api/observability/v1"
+	internalobs "github.com/openshift/cluster-logging-operator/internal/api/observability"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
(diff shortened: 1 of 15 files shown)


internal/validations/observability/common/validate.go
```diff
@@ -0,0 +1,52 @@
...
+	internalobs "github.com/openshift/cluster-logging-operator/internal/api/observability"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func ValidateConfigMapOrSecretKey(name string, keys []*obsv1.ConfigMapOrSecretKey, secrets map[string]*corev1.Secret, configMaps map[string]*corev1.ConfigMap) (conditions []metav1.Condition) {
+	if len(keys) == 0 {
+		return nil
+	}
+	for _, key := range keys {
```
vparfonov, line 15: nit: can we find other word instead of `key` to avoid `key.Key`, e.g: `tag`, `reference`

jcantrill: how about "configs" instead of keys   and "entry" instead of "key".  This is set of config with multiple entries in the set

PR #2532, opened 2024-06-03
fix region to be optional

jcantrill:
### Description
This PR makes CW region optional to match the sink spec

### DIFF
--- a/api/observability/v1/output_types.go
+++ b/api/observability/v1/output_types.go
@@ -305,7 +305,7 @@ type Cloudwatch struct {
 	// +nullable
 	Tuning *CloudwatchTuningSpec `json:"tuning,omitempty"`
 
-	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Optional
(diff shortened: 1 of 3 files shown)


api/observability/v1/output_types.go
```diff
@@ -305,7 +305,7 @@ type Cloudwatch struct {
 	// +nullable
 	Tuning *CloudwatchTuningSpec `json:"tuning,omitempty"`
 
-	// +kubebuilder:validation:Required
```
cahartma, line 308: Is required for our config.   Is only optional if they already have an env variable set (which they dont).

jcantrill: The client has no notion of a Default?

jcantrill: I guess maybe not based on their example? https://github.com/awslabs/aws-sdk-rust/blob/main/examples/examples/cloudwatchlogs/src/bin/list-log-groups.rs#L44

PR #2533, opened 2024-06-03
LOG:5606: fix splunk auth to be required

jcantrill:
### Description
This PR makes splunk token required given it is required by the splunk sink

ref: https://issues.redhat.com/browse/LOG-5606


jcantrill:
cc @vparfonov @Clee2691

vparfonov:
@jcantrill looks like it needs to run a regenerate bundle

PR #2535, opened 2024-06-04
LOG-5621: Add pipeline validations

jcantrill:
### Description
This PR:
* Adds pipeline validation to ensure references exist 

### Links
* https://issues.redhat.com/browse/LOG-5621

cc @Clee2691 @vparfonov

### DIFF
--- a/internal/validations/observability/pipelines/validate.go
+++ b/internal/validations/observability/pipelines/validate.go
@@ -1,0 +1,109 @@
+package pipelines
+
+import (
+	"fmt"
+	obs "github.com/openshift/cluster-logging-operator/api/observability/v1"
+	internalcontext "github.com/openshift/cluster-logging-operator/internal/api/context"
+	internalobs "github.com/openshift/cluster-logging-operator/internal/api/observability"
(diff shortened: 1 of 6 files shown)


internal/validations/observability/pipelines/validate.go
```diff
@@ -0,0 +1,109 @@
...
+		for _, field := range filter.PruneFilterSpec.NotIn {
+			if field == hostName {
+				found = true
+				break
+			}
+		}
+		if !found {
+			inListPrunes = true
+		}
+	}
```
vparfonov, line 97:
```suggestion
	if filter.PruneFilterSpec.NotIn != nil {
		for _, field := range filter.PruneFilterSpec.NotIn {
			 if field == hostName {
			   inListPrunes = true
			   break
		  }
	  }
	}
```

jcantrill: This is ported over from the previous tests and the recommended changes causes a test failure

vparfonov: oh, yes, my bad, i missed ! in`if !found`

PR #2536, opened 2024-06-04
refactor BearerToken api to not have duplicate key

jcantrill:
### Description
This PR:
* refactors BearerToken so that it allows only spec of serviceaccount or secret without duplicate keys


jcantrill:
cc @Clee2691 @cahartma

PR #2538, opened 2024-06-05
LOG-5599: Refactoring Vector configuration from Kubernetes Secrets to the ConfigMap

vparfonov:
### Description
This PR introduce the new feature from Vector called Secrets, more information can be fined here: 

- [Vector Secrets Management Highlights](https://vector.dev/highlights/2022-07-07-secrets-management/)
- [Vector Configuration Reference - Secrets](https://vector.dev/docs/reference/configuration/global-options/#secret)

This enhancement enables us to securely integrate sensitive data into Vector configurations.

#### Changes:
1. Integration of Vector Secrets
Replaced sensitive information in the Vector configuration with references to secrets managed by Vector. 
Example of Vector configuration:
 - secret loader configuration:
```
[secret.from_secret_mount]
type = "exec"
command = ["sh", "/usr/bin/read_secret_data.sh"]
```
- secrets can be read via `SECRET[<backend name>.<secret name>]`, e.g.:
```
[sinks.http_receiver.auth]
strategy = "basic"
user = "SECRET[from_secret_mount.http_receiver_username]"
password = "SECRET[from_secret_mount.http_receiver_password]"
``` 
2. Developed a script that reads sensitive data from mounted Kubernetes Secret files. This script is executed once at Vector startup, ensuring that the sensitive data is properly initialized and available for Vector to use.

3. Refactoring Vector configuration from Kubernetes Secrets to the ConfigMap without ability to compromise sensitive data. 

### Testing outputs:

- [x] Splunk
- [x] Azure Monitor Logs
- [x] Elasticsearch
- [x] Loki
- [x] HTTP

/cc @cahartma @Clee2691 @alanconway 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5599
- Enhancement proposal:


internal/api/observability/output_types_test.go
```diff
@@ -30,5 +34,55 @@ var _ = Describe("helpers for output types", func() {
 			}
 		})
 
+		It("should return an empty set of keys when authentication is not defined for an output", func() {
```
jcantrill, line 37: #2533 alters the API to make this required which means in theory these tests likely would never be run in prod

internal/collector/common/path.go
```diff
@@ -1,26 +1,25 @@
...
 
 import (
-	"fmt"
 	"path/filepath"
 )
 
 // SecretPath is the path for any secret visible to the collector
 func SecretPath(secretName string, file string) string {
-	return fmt.Sprintf("%q", filepath.Join("/var/run/ocp-collector/secrets", secretName, file))
+	return filepath.Join("/var/run/ocp-collector/secrets", secretName, file)
```
jcantrill, line 9: Are these used in config gen and for mounting volume paths?  I think one place needs quotes and the other does not which means we'll need changes to accommodate both

internal/collector/vector/utils.go
```diff
@@ -6,11 +6,13 @@ import (
...
-	DefaultDataPath  = "/var/lib/vector"
-	ConfigFile       = "vector.toml"
-	vectorConfigPath = "/etc/vector"
-	entrypointValue  = "/usr/bin/run-vector.sh"
+	RunVectorFile              = "run-vector.sh"
+	DefaultDataPath            = "/var/lib/vector"
+	ConfigFile                 = "vector.toml"
+	vectorConfigPath           = "/etc/vector"
+	entrypointValue            = "/usr/bin/run-vector.sh"
+	SecretDataReader           = "/usr/bin/read_secret_data.sh"
```
jcantrill, line 14: I think you can concate the file from below with the path to make this constant

internal/collector/vector/visitors.go
```diff
@@ -44,6 +45,11 @@ func CollectorVisitor(collectorContainer *corev1.Container, podSpec *corev1.PodS
 			Items:      []corev1.KeyToPath{{Key: RunVectorFile, Path: RunVectorFile}},
 			Optional:   utils.GetPtr(true),
 		}}},
+		corev1.Volume{Name: common.SecretDataReader, VolumeSource: corev1.VolumeSource{Secret: &corev1.SecretVolumeSource{
+			SecretName: resNames.ConfigMap,
+			Items:      []corev1.KeyToPath{{Key: SecretDataReaderScriptFile, Path: SecretDataReaderScriptFile}},
+			Optional:   utils.GetPtr(true),
```
jcantrill, line 51: This isn't optional any longer right? Same is really true of the runscript.

internal/collector/vector/visitors.go
```diff
@@ -44,6 +45,11 @@ func CollectorVisitor(collectorContainer *corev1.Container, podSpec *corev1.PodS
 			Items:      []corev1.KeyToPath{{Key: RunVectorFile, Path: RunVectorFile}},
 			Optional:   utils.GetPtr(true),
 		}}},
+		corev1.Volume{Name: common.SecretDataReader, VolumeSource: corev1.VolumeSource{Secret: &corev1.SecretVolumeSource{
```
jcantrill, line 48: This is no longer a "SecretVolumeSource now that you are creating a configmap

internal/controller/observability/collector.go
```diff
@@ -125,6 +132,20 @@ func GenerateConfig(k8Client client.Client, spec obs.ClusterLogForwarder, resour
 	return generatedConfig, err
 }
 
+func generateSecretReaderScript(k8Client client.Client, outputs internalobs.Outputs, namespace string) (config string, err error) {
+	var secrets []corev1.Secret
```
jcantrill, line 136: This is already done early in the controller https://github.com/openshift/cluster-logging-operator/blob/master/internal/controller/observability/clusterlogforwarder_controller.go#L60

internal/generator/vector/conf/complex.toml
```diff
@@ -4,6 +4,11 @@ data_dir = "/var/lib/vector/openshift-logging/my-forwarder"
 [api]
 enabled = true
 
+# Load sensitive data from secret mount with script: /usr/bin/read_secret_data.sh
+[secret.from_secret_mount]
+type = "exec"
+command = ["/usr/bin/read_secret_data.sh"]
```
jcantrill, line 10: do we need to set permissions on this when mounting the volume?

internal/generator/vector/helpers/helpers.go
```diff
@@ -63,3 +66,17 @@ func ConfigPath(name string, file string) string {
 func SecretPath(secretName string, file string) string {
 	return fmt.Sprintf("%q", filepath.Join("/var/run/ocp-collector/secrets", secretName, file))
 }
+
+func FromVectorSecretWithSecretPath(secretName, key string) string {
+	return fmt.Sprintf("SECRET[%s.%s_%s]", VectorSecretID, FormatComponentID(secretName), FormatComponentID(key))
```
jcantrill, line 71: There is a helper function called "MakeID" which takes ariable args which should facilitate this  and replace "FormatComponentID"

internal/generator/vector/helpers/helpers.go
```diff
@@ -63,3 +66,17 @@ func ConfigPath(name string, file string) string {
...
+}
+
+func FromVectorSecretWithSecretKey(secretKey *v1.SecretKey) string {
```
jcantrill, line 74: Why do we need this one and the other?  At a minim we should be able to call one from the other to reduce code duplication

internal/generator/vector/helpers/helpers.go
```diff
@@ -63,3 +66,17 @@ func ConfigPath(name string, file string) string {
 func SecretPath(secretName string, file string) string {
 	return fmt.Sprintf("%q", filepath.Join("/var/run/ocp-collector/secrets", secretName, file))
 }
+
+func FromVectorSecretWithSecretPath(secretName, key string) string {
```
jcantrill, line 70: This name is really long and not accurate when viewing the actual config. Consider:

```
helpers.SecretFrom(secretKey) or helpers.SecretConfigFrom(secretKey) 
```
or if we have a secrets package:
```
secrets.ConfigFrom(secretKey)
```

internal/generator/vector/output/common/vector_secret.go
```diff
@@ -0,0 +1,36 @@
+package common
+
+import (
+	"fmt"
+	"github.com/openshift/cluster-logging-operator/internal/generator/framework"
+)
+
+type VectorSecret struct {
```
jcantrill, line 8: Consider dropping "Vector" from this name since its already under a "vector" package

internal/collector/vector/visitors.go
jcantrill: Rethinking... to clarify:
* the "data" or our vault must be mounted placed in a secret and mounted as a secret volume
* the config can go in a configmap and is mounted as a configmpa volume

I can help with some of this path mounting since I have to work on how the CW token projections are integrated here

internal/collector/vector/visitors.go
vparfonov: will check

internal/generator/vector/conf/complex.toml
vparfonov: you mean run permissions?

internal/generator/vector/helpers/helpers.go
vparfonov: Yes, need to fix that, in previous code all `Authentication`  was `SecretKey` so was used `FromVectorSecretWithSecretKey`

internal/generator/vector/output/common/auth/bearer_token.go
```diff
@@ -23,11 +25,28 @@ token = "{{.Token}}"
...
+			},
+			Key: auth.Token.Secret.Key,
+		}
+		bt.Token = helpers.SecretFrom(key)
+		return bt
 	}
+	//TODO fixme
+
+	// We reconcile SA token secret, so name and key will be known
+	key := &obs.SecretKey{
```
jcantrill, line 44: This is where once #2541 merges we should fetch token from the SECRET vault, but it will need to retrieve it from the projected SA token volume.  All the other values are in the secret but this one is sourced from the file system

internal/generator/vector/output/common/secret_reader.go
```diff
@@ -0,0 +1,28 @@
...
+func GenerateSecretReaderScript(secrets []corev1.Secret) string {
+	var scriptBuilder strings.Builder
+	scriptBuilder.WriteString("#!/bin/bash\ncat <<EOF\n{\n")
+	var values []string
+	for _, secret := range secrets {
+		for i, _ := range secret.Data {
+			values = append(values, fmt.Sprintf(scriptTmpl,
+				helpers.Replacer.Replace(secret.Name),
+				helpers.Replacer.Replace(i),
+				common.SecretPath(secret.Name, i)))
```
jcantrill, line 22: This has to also include the key name to get the value of the actual file.  The mount root will be the secret name but there are keys underneath that are specific values.  Maybe changing the "i" to "key" will make it more obvious what we are doing

internal/validations/observability/outputs/validate_secrets_test.go
```diff
@@ -40,5 +40,15 @@ var _ = Describe("validating output secrets", func() {
 			conds := ValidateSecretsAndConfigMaps(spec, secrets, configMaps)
 			Expect(conds).To(Not(BeEmpty()))
 		})
+		It("should validate secrets if TLS nil", func() {
+			spec.TLS = nil
+			conds := ValidateSecretsAndConfigMaps(spec, secrets, configMaps)
+			Expect(conds).To(Not(BeEmpty()))
```
jcantrill, line 46: Why should this fail if TLS is empty?  It is a valid for an output to not have TLS settings

internal/validations/observability/outputs/validate_secrets_test.go
```diff
@@ -40,5 +40,15 @@ var _ = Describe("validating output secrets", func() {
...
+		})
+		It("should validate secrets if TLS is empty", func() {
+			spec.TLS = &obs.OutputTLSSpec{}
```
jcantrill, line 49: Same question here

internal/generator/vector/helpers/helpers.go
```diff
@@ -56,10 +60,27 @@ func ListenOnAllLocalInterfacesAddress() string {
...
+	if secretKey != nil && secretKey.Secret != nil && secretKey.Key != "" {
+		return fmt.Sprintf("SECRET[%s.%s_%s]",
+			VectorSecretID,
+			MakeSecretPath(secretKey.Secret.Name),
+			MakeSecretPath(secretKey.Key))
+	}
+	return ""
+}
+
+func MakeSecretPath(path string) string {
```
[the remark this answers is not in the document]
vparfonov: not, MakeID did lowercase, here we don't need that

internal/validations/observability/outputs/validate_secrets_test.go
vparfonov: code fail [here](https://github.com/openshift/cluster-logging-operator/blob/master/internal/validations/observability/outputs/validate.go#L23) NPE if TLS is nil

internal/controller/observability/collector.go
vparfonov: yes, but that Map contain also secret from Inputs, we don't need them in this case

internal/validations/observability/outputs/validate_secrets_test.go
jcantrill: > code fail here NPE if TLS is nil

Shouldnt the fix then be to fix the nil pointer in :arrow_up: ? It is acceptable to not provide TLS spec.  This just means you don't need these tls configuration

internal/collector/vector/visitors.go
```diff
@@ -25,6 +25,7 @@ func CollectorVisitor(collectorContainer *corev1.Container, podSpec *corev1.PodS
 	dataPath := GetDataPath(namespace, resNames.ForwarderName)
 	collectorContainer.VolumeMounts = append(collectorContainer.VolumeMounts,
 		corev1.VolumeMount{Name: common.ConfigVolumeName, ReadOnly: true, MountPath: vectorConfigPath},
+		corev1.VolumeMount{Name: common.SecretDataReader, ReadOnly: true, MountPath: SecretDataReaderPath, SubPath: SecretDataReaderFile},
```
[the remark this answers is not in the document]
vparfonov: Why not? Maybe i don't get that do you mean

internal/generator/vector/output/common/secret_reader_test.go
```diff
@@ -0,0 +1,45 @@
+package common
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Describe("#generateCollectorConfig", func() {
```
jcantrill, line 10: #GenerateSecretReaderScript

internal/generator/vector/output/common/secret_reader_test.go
```diff
@@ -0,0 +1,45 @@
...
+	Context("not dependent on ClusterLogging instance", func() {
```
jcantrill, line 11: remove context

internal/generator/vector/output/common/secret_reader_test.go
```diff
@@ -0,0 +1,45 @@
...
+		It("should generate appropriate vector config", func() {
```
jcantrill, line 12: It should generate a valid script to read secrets

internal/collector/vector/visitors.go
```diff
@@ -33,17 +34,10 @@ func CollectorVisitor(collectorContainer *corev1.Container, podSpec *corev1.PodS
...
-		}}},
+		corev1.Volume{Name: common.ConfigVolumeName, VolumeSource: corev1.VolumeSource{ConfigMap: &corev1.ConfigMapVolumeSource{LocalObjectReference: corev1.LocalObjectReference{Name: resNames.ConfigMap}}}},
 		corev1.Volume{Name: common.DataDir, VolumeSource: corev1.VolumeSource{HostPath: &corev1.HostPathVolumeSource{Path: dataPath}}},
-		corev1.Volume{Name: common.EntrypointVolumeName, VolumeSource: corev1.VolumeSource{Secret: &corev1.SecretVolumeSource{
-			SecretName: resNames.ConfigMap,
-			Items:      []corev1.KeyToPath{{Key: RunVectorFile, Path: RunVectorFile}},
-			Optional:   utils.GetPtr(true),
-		}}},
+		corev1.Volume{Name: common.EntrypointVolumeName, VolumeSource: corev1.VolumeSource{ConfigMap: &corev1.ConfigMapVolumeSource{LocalObjectReference: corev1.LocalObjectReference{Name: resNames.ConfigMap}}}},
+		corev1.Volume{Name: common.SecretDataReader, VolumeSource: corev1.VolumeSource{ConfigMap: &corev1.ConfigMapVolumeSource{LocalObjectReference: corev1.LocalObjectReference{Name: resNames.ConfigMap}}}},
```
jcantrill, line 40: This volumesource should be secret since the readerscript is full of secrets

vparfonov: oh, yes exactly

jcantrill: I see from your comment that we could leave this as a configmap since the secrets are obscured by files.  Given you have already completed the changes we should just leave it

internal/collector/vector/visitors.go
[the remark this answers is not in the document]
vparfonov: How i understand this:

- `SubPath` specifies key that we want to mount the mount source 
- `MountPath` specifies the target location inside the container.

If we mount the full map in one place, we don't need `SubPath`

jcantrill (commented):
/lgtm

PR #2547, opened 2024-06-11
LOG-5648: Refactor condition types

jcantrill:
### Description
This PR:
* Refactors ConditionTypes to consolidate validations into messages

### Links
* https://issues.redhat.com/browse/LOG-5648

cc @Clee2691 @vparfonov @cahartma


alanconway (changes requested):
Guidelines for conditions have been updated since last I looked, mostly they are more vague:

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

As I understand it, the convention is to have a single list of Conditions, which is actually a set (only one instance of each Type)
The only widely-used "standard" condition is "Ready".

I don't think it makes sense to use condition.type or condition.reason to include an output name, both are considered part of the API, supposed to be stable for scripts that want to check the status of the CLF. As such we should keep them to a small set.

After thinking about it I suggest we minimize our use of conditions, and use condition.message for all the dynamic info

- Single 'conditions' array
- Minimal list of condition types ("Ready", "Degraded") and reasons.
- Invalid configuration becomes a Reason for Ready=false or Degraded=true. All dynamic info about sub-components is in the human readable Message part for debugging, not useful for scripting.

Example:

conditions:
  - type: Ready
    status: false
    reason: Invalid
    message: |
      input fred: missing URLs
      output barney: unknown output type 'amqp'
      filter prune: cannot parse blahblah
      ...
  }
}

If we no longer support any "Degraded" conditions, then we can reduce it to just "Ready".

That way our API exposure for conditions is small, people who want to write scripts only have to check for Ready=true. The error information is all in one place, and is not part of the API so error messages are free to change.

I think I overdid things with the original design of separate conditions per component.
Users will want to test for the CLO has a whole being ready, not for individual inputs or outputs.
In any case we don't provide accurate "Ready" information for sub-components, we only detect configuration errors, not actual "readiness".

Sorry for being late with this.

internal/validations/observability/outputs/cloudwatch.go
```diff
@@ -1,51 +1,41 @@
...
-			return conds(obs.ReasonMissingSpec, fmt.Sprintf("%q requires auth configuration", spec.Name))
+			return []string{"IAMRole is nil"}
 		}
 		if authSpec.IAMRole.RoleARN == nil {
-			return conds(obs.ReasonMissingSpec, fmt.Sprintf("%q auth requires a RoleARN", spec.Name))
+			results = append(results, "RoleARN is missing")
 		}
 		if authSpec.IAMRole.Token == nil {
-			return conds(obs.ReasonMissingSpec, fmt.Sprintf("%q auth requires a token", spec.Name))
+			return append(results, "Token is missing")
```
Clee2691, line 34: I thought the token was optional if you define `IAMRole` along with the `RoleArn`?

jcantrill: @cahartma told me it was required...?

jcantrill: made required

PR #2558, opened 2024-06-14
Include missing verbs in RBAC configuration

xperimental:
### Description

While running the CLO on a test cluster I noticed that it still produced two errors regarding RBAC for the `apiservers` and `securitycontextcontraints` resources. It looks as if the verbs needed to watch these resources were missing.

This is a follow-up fix for #2521 .


jcantrill (commented):
/approve
/lgtm

PR #2559, opened 2024-06-17
LOG-5524: use get(object, path) function to get the value of a given path

vparfonov:
### Description
Use` get(object, path)` function to get the value of a given path: e.g.: for label `kubernetes.labels.my-typekey` will use `val = get!(value: ., path: ["kubernetes","labels","my-typekey"])`

/cc @cahartma @Clee2691 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5524
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/output/elasticsearch/elasticsearch.go
+++ b/internal/generator/vector/output/elasticsearch/elasticsearch.go
@@ -1,7 +1,9 @@
 package elasticsearch
 
 import (
+	"encoding/json"
 	"fmt"
+	log "github.com/ViaQ/logerr/v2/log/static"
 	"strings"
 
 	. "github.com/openshift/cluster-logging-operator/internal/generator/framework"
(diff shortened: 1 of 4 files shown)


internal/generator/vector/output/elasticsearch/elasticsearch.go
```diff
@@ -248,3 +249,15 @@ func BasicAuth(id string, o logging.OutputSpec, secret *corev1.Secret) []Element
 
 	return conf
 }
+
+func safePath(path string) string {
+	if !strings.Contains(path, "-") {
```
jcantrill, line 254: We can just make it always use "get" .  It should work for both cases

vparfonov: yes., it will work for sure

PR #2570, opened 2024-06-25
fix audit volume names

jcantrill:
### Description
This PR:
* fixes audit volume/volumeMount names that clashed
* updates constant names

cc @cahartma


cahartma (commented):
/lgtm

PR #2573, opened 2024-06-26
LOG-5475 OTLP Initial commit of API changes only

cahartma:
### Description
Initial commit of api changes for OTLP output.   Output generator changes are ready as well, but need to get this merged.    
Also removing schema field from logging api.

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- Parent-task:  https://issues.redhat.com/browse/LOG-5370
- Sub-tasks: https://issues.redhat.com/browse/LOG-5475 https://issues.redhat.com/browse/LOG-5624
- Enhancement proposal: https://github.com/openshift/enhancements/pull/1618


api/observability/v1/output_types.go
```diff
@@ -769,3 +774,50 @@ type Syslog struct {
...
+	//
+	// +kubebuilder:validation:Enum:=gzip;none
+	// +kubebuilder:default:=gzip
+	Compression string `json:"compression,omitempty"`
+}
+
+// Otlp provided configuration for sending OTLP/HTTP with JSON-encoded Protobuf payload requests.
+// https://opentelemetry.io/docs/specs/otlp/#json-protobuf-encoding
+type Otlp struct {
+	// URL to send log records to.
```
jcantrill, line 792: Based upon [this](https://github.com/openshift/enhancements/pull/1618#discussion_r1601613595) discussion this should state "must terminate".  Additionally, add [validation](https://book.kubebuilder.io/reference/markers/crd-validation) to enforce this change:

```
+kubebuilder:validation:Pattern
```
There are examples of usage elsewhere within the API, maybe for serviceAccount?

api/observability/v1/output_types.go
```diff
@@ -769,3 +774,50 @@ type Syslog struct {
...
+	//
+	// +kubebuilder:validation:Optional
+	// +nullable
+	Tuning *OtlpTuningSpec `json:"tuning,omitempty"`
+
+	// Headers specify optional headers to be sent with the request
+	// default value is: {"Content-Type"="application/json"}
+	//
+	// +kubebuilder:validation:Optional
+	Headers map[string]string `json:"headers,omitempty"`
```
jcantrill, line 817: This is not part of the enhancement API.  Do we have need to support this initially?

api/observability/v1/output_types.go
```diff
@@ -769,3 +774,50 @@ type Syslog struct {
...
+
+	// Timeout specifies the Http request timeout in seconds. If not set, 10secs is used.
+	//
+	// +kubebuilder:validation:Optional
+	Timeout int `json:"timeout,omitempty"`
```
jcantrill, line 822: This is not part of the enhancement API.  Do we have need to support this initially?

bundle/manifests/cluster-logging.clusterserviceversion.yaml
```diff
@@ -390,6 +390,16 @@ spec:
         path: outputs[0].lokiStack.authentication.username.key
         x-descriptors:
         - urn:alm:descriptor:com.tectonic.ui:text
+      - description: Name of the key used to get the value from the referenced Secret.
```
jcantrill, line 393: I would have expected more changes to this file.  Did you do `make generate bundle` ?

api/observability/v1/output_types.go
```diff
@@ -769,3 +774,50 @@ type Syslog struct {
...
+
+	// Compression causes data to be compressed before sending over the network.
+	// It is an error if the compression type is not supported by the output.
+	//
+	// +kubebuilder:validation:Enum:=gzip;none
+	// +kubebuilder:default:=gzip
+	Compression string `json:"compression,omitempty"`
+}
+
+// Otlp provided configuration for sending OTLP/HTTP with JSON-encoded Protobuf payload requests.
```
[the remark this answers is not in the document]
cahartma: The resourceLog format we send is protobuf mapped to JSON.   https://opentelemetry.io/docs/specs/otlp/#otlphttp
"OTLP/HTTP uses Protobuf payloads encoded either in binary formator in JSON. Regardless of the encoding the Protobuf schema of the messages is the same for OTLP/HTTP and OTLP/gRPC"

api/observability/v1/output_types.go
cahartma: Not initially, but there was already an existing example used in HTTP... and I used it initially to debug the OTLP requests and test with otel-collector.    I will remove it.

api/observability/v1/output_types.go
cahartma: Nope

api/observability/v1/output_types.go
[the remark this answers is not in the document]
cahartma: If you think we should remove the word 'protobuf' due to confusion, I could agree with that.

bundle/manifests/cluster-logging.clusterserviceversion.yaml
cahartma: I did it again, no other changes.    However, I did notice we still have ref to ES and fluent, and a lot of garbage text.

api/observability/v1/output_types.go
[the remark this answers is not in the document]
jcantrill: The spec defines the encoding.  I would prefer to remove any mention of it here because we are not using it.  The payload may be acceptable for it, we can integrate protobuf...but dont.  I think it makes more sense to remove the notion and add the comment in later if we actually do use it

api/observability/v1/output_types.go
```diff
@@ -769,3 +774,40 @@ type Syslog struct {
...
+
+	// Compression causes data to be compressed before sending over the network.
+	// It is an error if the compression type is not supported by the output.
+	//
+	// +kubebuilder:validation:Enum:=gzip;none
+	// +kubebuilder:default:=gzip
+	Compression string `json:"compression,omitempty"`
+}
+
+// OTLP provides configuration for sending logs via OTLP/HTTP requests with JSON-encoded payloads.
```
jcantrill, line 789: OTLP defines configuration for sending logs via OTLP using OTEL scemantic conventions
https://opentelemetry.io/docs/specs/otlp/#otlphttp

PR #2576, opened 2024-06-26
LOG-5710: Enable Cloudwatch output functional tests

Clee2691:
### Description
This PR refactors Cloudwatch output functional tests

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5710


test/functional/outputs/cloudwatch/forward_to_cloudwatch_test.go
```diff
@@ -148,13 +180,13 @@ var _ = Describe("[Functional][Outputs][CloudWatch] Forward Output to CloudWatch
 			if testfw.LogCollectionType == logging.LogCollectionTypeVector {
```
jcantrill, line 180: We should investigate removing since there is no need to have special handling for vector.  Either the test applies as-is (which is what I would expect) or the test can be removed entirely

PR #2579, opened 2024-06-26
LOG-5711: Enable elasticsearch output functional tests

Clee2691:
### Description
This PR refactors Elasticsearch output functional tests and removes the index tests.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5711


test/runtime/observability/cluster_log_forwarder.go
```diff
@@ -127,7 +127,7 @@ func (p *PipelineBuilder) Named(name string) *PipelineBuilder {
 	return p
 }
 
-func (p *PipelineBuilder) ToElasticSearchOutput(visitors ...func(output *obs.OutputSpec)) *ClusterLogForwarderBuilder {
+func (p *PipelineBuilder) ToElasticSearchOutput(auth obs.HTTPAuthentication, visitors ...func(output *obs.OutputSpec)) *ClusterLogForwarderBuilder {
```
jcantrill, line 130: If we require auth here should we also mark the API as required?  I chose this pattern for one of my recent submitssions because we "required" auth.  I'm open to leaving this here instead of deferring a user to use a visitor

Clee2691: We can keep with the visitor instead.

jcantrill (commented):
/lgtm

PR #2604, opened 2024-07-02
LOG-5742: Enable eventrouter e2e test

jcantrill:
### Description
This PR:
* Enables the e2e event router test
* Fixes ESv6 to require id_key
* Fixes ID to use object UID if it is a kubernetes event

### Links
https://issues.redhat.com/browse/LOG-5733


jcantrill:
cc @cahartma @Clee2691 

It's likely we could move this to a functional test

PR #2608, opened 2024-07-03
Remove omitempty on required fields

xperimental:
### Description

When using `+kubebuilder:validation:Required` to mark a field of the CRD as required it can not be combined with `omitempty`. If `omitempty` is present then the `Required` is silently ignored.

This PR removes `omitempty` from all attributes having the `+kubebuilder:validation:Required` annotation in the comment, which adds new required fields to the generated CRD.

### DIFF
--- a/api/observability/v1/output_types.go
+++ b/api/observability/v1/output_types.go
@@ -214,20 +214,20 @@ type AzureMonitorAuthentication struct {
 	//
 	// +kubebuilder:validation:Required
 	// +nullable
-	SharedKey *SecretKey `json:"sharedKey,omitempty"`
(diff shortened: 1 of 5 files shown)


api/observability/v1/output_types.go
```diff
@@ -444,13 +444,14 @@ type GoogleCloudLogging struct {
 type GoogleGloudLoggingID struct {
 	// Type is the ID type provided
 	//
+	// +kubebuilder:validation:Required
```
xperimental, line 447: This is the only time I added a `Required` because the type looked like it should require the use of `type`. Please confirm that this is actually the case.

PR #2609, opened 2024-07-03
Move enum annotations to type definitions

xperimental:
### Description

This PR moves the `kubebuilder:validation:Enum` annotations to the type of the field, if we have defined a custom type for it. That way they are close to where the definition of the constants and only need to be present once, in case a type is reused in different places.

This is just a "code style" change.


jcantrill (commented):
/lgtm

jcantrill (commented):
/lgtm

PR #2611, opened 2024-07-03
Move startup script to file and use go-embed

xperimental:
### Description

This PR moves the `run-vector.sh` startup script from a constant to a file that is injected into the executable using `go-embed`. This change makes the file more easily discoverable in the repository and syntax highlighting and linting tools also work better with a separate file instead of a string constant.

This is just a "code style" change.


jcantrill (commented):
/approve
/lgtm

PR #2613, opened 2024-07-03
LOG-5482 otlp output functional test 

cahartma:
### Description
Based on previous PR for transforms https://github.com/openshift/cluster-logging-operator/pull/2590

Added the first functional test to send application logs to otel-collector container in the testnamespace.  Can read the log file and parse the attribute keys and values if necessary.   

All is still changing and will be dependent on what we decide is our "resource/stream" for each log_type, and which attributes we send (both in the resource.attributes and logRecord.attributes)

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- Parent task:  https://issues.redhat.com/browse/LOG-5370
- Sub-task: https://issues.redhat.com/browse/LOG-5482 
- Other tasks:  https://issues.redhat.com/browse/LOG-5476 https://issues.redhat.com/browse/LOG-5477 https://issues.redhat.com/browse/LOG-5478 https://issues.redhat.com/browse/LOG-5479 https://issues.redhat.com/browse/LOG-5480
- Enhancement proposal: https://github.com/openshift/enhancements/pull/1618

### DIFF
--- a/internal/generator/vector/output/otlp/otlp.go
+++ b/internal/generator/vector/output/otlp/otlp.go
@@ -1,0 +1,168 @@
+package otlp
+
+import (
+	"fmt"
+	obs "github.com/openshift/cluster-logging-operator/api/observability/v1"
+	. "github.com/openshift/cluster-logging-operator/internal/generator/framework"
+	genhelper "github.com/openshift/cluster-logging-operator/internal/generator/helpers"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/elements"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
+	vectorhelpers "github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/output/common"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/output/common/auth"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/output/common/tls"
+	"sort"
+	"strings"
+)
+
+type Otlp struct {
+	ComponentID      string
+	Inputs           string
+	URI              string
+	common.RootMixin //TODO: remove??
+}
+
+func (p Otlp) Name() string {
+	return "vectorOtlpTemplate"
+}
+
+func (p Otlp) Template() string {
+	return `{{define "` + p.Name() + `" -}}
+[sinks.{{.ComponentID}}]
+type = "http"
+inputs = {{.Inputs}}
+uri = "{{.URI}}"
+method = "post"
+payload_prefix = "{\"resourceLogs\":"
+payload_suffix = "}"
+{{.Compression}}
+{{end}}
+`
+}
+
+// TODO: test this for otlp
+func (p *Otlp) SetCompression(algo string) {
+	p.Compression.Value = algo
+}
+
+const (
+	logSourceContainer    = string(obs.ApplicationSourceContainer)
+	logSourceNode         = string(obs.InfrastructureSourceNode)
+	logSourceAuditd       = string(obs.AuditSourceAuditd)
+	logSourceKubeAPI      = string(obs.AuditSourceKube)
+	logSourceOpenshiftAPI = string(obs.AuditSourceOpenShift)
+	logSourceOvn          = string(obs.AuditSourceOVN)
+
+	// For now we are grouping all container logs by their ns, pod and container names
+	// so there is no distinction here at the moment.
+	// TODO: revise or remove
+	logSourceContainerApp   = string(obs.ApplicationSourceContainer)
+	logSourceContainerInfra = string(obs.InfrastructureSourceContainer)
+)
+
+func New(id string, o obs.OutputSpec, inputs []string, secrets vectorhelpers.Secrets, strategy common.ConfigStrategy, op Options) []Element {
+	if genhelper.IsDebugOutput(op) {
+		return []Element{
+			elements.Debug(helpers.MakeID(id, "debug"), vectorhelpers.MakeInputs(inputs...)),
+		}
+	}
+	// TODO: create a pattern to filter by input so all this is not necessary
+	var els []Element
+	// Creates reroutes for 'container','node','auditd','kubeAPI','openshiftAPI','ovn'
+	rerouteID := vectorhelpers.MakeID(id, "reroute") // "output_my_id_reroute
+	els = append(els, RouteBySource(rerouteID, inputs))
+	// Container
+	transformContainerID := vectorhelpers.MakeID(id, logSourceContainer)                       // "output_my_id_container"
+	transformContainerInputID := vectorhelpers.MakeRouteInputID(rerouteID, logSourceContainer) // "output_my_id_reroute.container"
+	reduceContainerID := vectorhelpers.MakeID(id, "groupby", "container")
+	els = append(els, TransformContainer(transformContainerID, []string{transformContainerInputID}))
+	// Group by cluster_id, namespace_name, pod_name, container_name
+	els = append(els, GroupByContainer(reduceContainerID, []string{transformContainerID}))
+	// Journal
+	transformNodeID := vectorhelpers.MakeID(id, logSourceNode)
+	transformNodeRouteID := vectorhelpers.MakeRouteInputID(rerouteID, logSourceNode)
+	els = append(els, TransformJournal(transformNodeID, []string{transformNodeRouteID}))
+	// Audit
+	transformAuditHostID := vectorhelpers.MakeID(id, logSourceAuditd)
+	transformAuditHostRouteID := vectorhelpers.MakeRouteInputID(rerouteID, logSourceAuditd)
+	transformAuditKubeID := vectorhelpers.MakeID(id, logSourceKubeAPI)
+	transformAuditKubeRouteID := vectorhelpers.MakeRouteInputID(rerouteID, logSourceKubeAPI)
+	transformAuditOpenshiftID := vectorhelpers.MakeID(id, logSourceOpenshiftAPI)
+	transformAuditOpenshiftRouteID := vectorhelpers.MakeRouteInputID(rerouteID, logSourceOpenshiftAPI)
+	transformAuditOvnID := vectorhelpers.MakeID(id, logSourceOvn)
+	transformAuditOvnRouteID := vectorhelpers.MakeRouteInputID(rerouteID, logSourceOvn)
+	reduceSourceID := vectorhelpers.MakeID(id, "groupby", "source")
+	reduceHostID := vectorhelpers.MakeID(id, "groupby", "host")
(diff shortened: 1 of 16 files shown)


internal/generator/vector/output/otlp/transform.go
```diff
@@ -0,0 +1,254 @@
...
+import (
+	. "github.com/openshift/cluster-logging-operator/internal/generator/framework"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/elements"
+	"github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
+	"strings"
+)
+
+// VRL for OTLP transforms by route
+const (
+	CreateResourceAttributes = `
```
jcantrill, line 12: Is here any advantage to convert this to VRL files and use embed?

test/functional/outputs/otlp/app-log-sample.json
```diff
@@ -0,0 +1,149 @@
...
+          "scope": {},
+          "logRecords": [
+            {
+              "timeUnixNano": "1693227568573159188",
+              "observedTimeUnixNano": "1719950665726889376",
+              "severityNumber": 9,
+              "body": {
+                "stringValue": "Format me to a container message!"
+              },
+              "attributes": [
```
jcantrill, line 79: These should be resource scoped

test/functional/outputs/otlp/forward_to_otlp_test.go
```diff
@@ -0,0 +1,192 @@
...
+	obstestruntime "github.com/openshift/cluster-logging-operator/test/runtime/observability"
+	"time"
+)
+
+const (
+	timestamp     = "2023-08-28T12:59:28.573159188+00:00"
+	timestampNano = "1693227568573159188"
+)
+
+var _ = Describe("[Functional][Outputs][Otlp] Functional tests", func() {
```
jcantrill, line 20: OTLP

test/functional/outputs/otlp/forward_to_otlp_test.go
```diff
@@ -0,0 +1,192 @@
...
+			Expect(raw).ToNot(BeEmpty())
+
+			logs, err := otlp.ParseLogs(raw[0])
+			Expect(err).To(BeNil(), "Expected no errors parsing the logs")
+
+			// Access the first (and only) ResourceLog
+			// Inspect resource attributes
+			resourceLog := logs.ResourceLogs[0]
+			resource := resourceLog.Resource
+			namespaceName := resource.FindStringValue(resource.NamespaceNameAttribute())
```
jcantrill, line 66: Not to be solved here but we can add our "templates" here to automagically test these

test/helpers/otlp/attributes.go
```diff
@@ -0,0 +1,51 @@
...
+	NodeName           = "node.name"
+	ClusterID          = "cluster.id"
+	K8sNamespaceName   = "k8s.namespace.name"
+	K8sPodName         = "k8s.pod.name"
+	K8sContainerName   = "k8s.container.name"
+	OpenshiftLogSource = "openshift.log.source"
+	OpenshiftLogType   = "openshift.log.type"
+)
+
+func (r Resource) NodeNameAttribute() string {
```
jcantrill, line 13: I think this functions would be more valuable to call FindStringValue then needing to pair them together inthe test.  As-is its a function to return a constant that is already public

api/logging/v1/output_types.go
```diff
@@ -15,6 +15,7 @@ const (
 	OutputTypeSplunk             = "splunk"
 	OutputTypeHttp               = "http"
 	OutputTypeAzureMonitor       = "azureMonitor"
+	OutputTypeOtlp               = "otlp" // Value used only as a constant in 6.0 for wiring together validation logic
```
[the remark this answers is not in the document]
cahartma: Done, just verified is now working.  It used to be a workaround.

bundle/manifests/observability.openshift.io_clusterlogforwarders.yaml
```diff
@@ -14,7 +14,7 @@ spec:
     listKind: ClusterLogForwarderList
     plural: clusterlogforwarders
     shortNames:
-    - clf
+    - lf
```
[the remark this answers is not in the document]
cahartma: I'll revert to clf for now.   We can change it separately after deciding.

internal/generator/vector/filter/filter.go
```diff
@@ -26,8 +26,8 @@ type InternalFilterSpec struct {
 	// RemapFilter is a filter that uses a remap transformation
 	RemapFilter RemapFilter
 
-	//TranformFactory takes an id, inputs and returns an Element
-	TranformFactory func(id string, inputs ...string) framework.Element
+	//TransformFactory takes an id, inputs and returns an Element
+	TransformFactory func(id string, inputs ...string) framework.Element
```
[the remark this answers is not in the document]
cahartma: Done

internal/generator/vector/output/otlp/otlp.go
```diff
@@ -0,0 +1,129 @@
...
+	// Create sink and wrap in `resourceLogs`
+	sink := Output(id, o, []string{formatResourceLogsID}, secrets, op)
+	if strategy != nil {
+		strategy.VisitSink(sink)
+	}
+	return MergeElements(
+		els,
+		[]Element{
+			sink,
+			common.NewAcknowledgments(id, strategy),
```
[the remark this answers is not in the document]
cahartma: I'll add it for consistency of course, thanks.      Note: that field will never be included since we set specific fields when we create a new object at the end of our remaps.

internal/generator/vector/output/otlp/otlp_all.toml
```diff
@@ -0,0 +1,374 @@
...
+# Normalize container log records to OTLP semantic conventions
+[transforms.output_otel_collector_container]
+type = "remap"
+inputs = ["output_otel_collector_reroute.container"]
+source = '''
+  # Create base resource attributes
+  resource.attributes = []
+  resource.attributes = append( resource.attributes,
+      [{"key": "cluster.id", "value": {"stringValue": get!(.,["openshift","cluster_id"])}},
+  	{"key": "openshift.log.source", "value": {"stringValue": .log_source}}]
```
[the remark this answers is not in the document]
cahartma: Hah, it was there and we decided it needed we needed to group by source.    If we add type here, it will get squashed unless we also group and reduce by matching type.

internal/generator/vector/output/otlp/otlp_all.toml
```diff
@@ -0,0 +1,374 @@
...
+  r = {}
+  r.timeUnixNano = to_string(to_unix_timestamp(parse_timestamp!(.@timestamp, format:"%+"), unit:"nanoseconds"))
+  r.observedTimeUnixNano = to_string(to_unix_timestamp(now(), unit:"nanoseconds"))
+  # Convert syslog severity keyword to number, default to 9 (unknown)
+  r.severityNumber = to_syslog_severity(.level) ?? 9
+  r.body = {"stringValue": string!(.message)}
+  r.attributes = []
+  # Append logRecord attributes
+  r.attributes = append(r.attributes,
+      [{"key": "openshift.log.type", "value": {"stringValue": .log_type}}]
```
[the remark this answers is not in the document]
cahartma: Not unless we want to group/reduce by that as well.   All resource.attributes are squashed, but logRecord.attributes are unique.   Happy to discuss.

internal/generator/vector/output/otlp/otlp_all.toml
```diff
@@ -0,0 +1,374 @@
...
+  )
+  # Openshift and kubernetes objects for grouping containers (dropped before sending)
```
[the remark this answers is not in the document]
cahartma: Dropped after grouping/reducing, before send, in our final remap for all.   In the end, we only send resource.attributes and logRecords.  
```
  . = {
        "resource": {
           "attributes": .resource.attributes,
        },
        "scopeLogs": [
          {"logRecords": .logRecords}
        ]
      }
```

internal/generator/vector/output/otlp/route.go
```diff
@@ -0,0 +1,42 @@
...
+
+func (r Route) Template() string {
+	return `{{define "routeTemplate" -}}
+{{if .Desc -}}
+# {{.Desc}}
+{{end -}}
+[transforms.{{.ComponentID}}]
+type = "route"
+inputs = {{.Inputs}}
+route.container = '.log_source == "container"'
```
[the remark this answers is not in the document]
cahartma: Yes thats possible.  Can we please merge first so I can refactor while everything else I've done is still working!   I keep breaking things during refactoring trying to create everything dynamically.  I obviously need more tests now that we have a general idea of the data model for each type.

internal/generator/vector/output/otlp/transform.go
cahartma: I would strongly suggest "not now".    Lets merge a working version before refactoring.

test/functional/outputs/otlp/app-log-sample.json
cahartma: Please explain or elaborate a bit.

cahartma: You are correct that source should be resource scoped.   Will fix that.   Here is the latest example:  
https://github.com/openshift/cluster-logging-operator/blob/master/hack/logsamples/otlp_container.json

test/functional/outputs/otlp/forward_to_otlp_test.go
cahartma: Certainly thanks!   Will be adding more tests over the next few days and can begin that process after.

test/helpers/otlp/attributes.go
cahartma: This is meant to grow and end up as a documented list of all attributes we are using and whether they are logRecord or resource attributes.   Id prefer to keep as is for now.   Makes for easier testing and logical reference, imho.

internal/generator/vector/output/otlp/otlp_all.toml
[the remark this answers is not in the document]
jcantrill: In the context of container logs... this is unique to the resource and not the log records.  Same with all the other log types too:
* journal: group by: cluster, node
* audit: group by: cluster, source

log type should not factor into the grouping since its a constant within each of these groupings

test/helpers/otlp/attributes.go
jcantrill: Maybe a need to refactor after merge.... but

* They can still be constants that we use for documenting, they just would have to live in a different package
* the usage as-is IMO is a little verbose  for example:
```
r.FindStringValue(ClusterID)
```
instead of just:
```
r.ClusterID()
```
One has to lookup what the attribute they want instead of simply inspecting the available things.  We know there is a fixed set of items we may wish to retrieve and the design still allows to get those which are accessed less.

PR #2625, opened 2024-07-10
LOG-5745: Only container logs can be forwarded with groupName {{.kubernetesnamespace_name}} to CloudWatch

Clee2691:
### Description
This PR addresses missing fields when templating the `groupName` for `Cloudwatch`. This now allows a fallback value to be defined in case a field is missing.

The syntax allows for static values as well as path expressions enclosed in single curly brackets `{}`. There must be a fallback value when using path expressions and can be expressed with `||` followed by another path expression or a static value in quotes.

Examples:

- foo
- {.foo||"none"}
- foo_{.bar||"none"}
- foo_bar{.baz."foo/bar"||.foo||"missing"}

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5745

### DIFF
--- a/internal/generator/vector/output/cloudwatch/cloudwatch.go
+++ b/internal/generator/vector/output/cloudwatch/cloudwatch.go
@@ -1,6 +1,8 @@
 package cloudwatch
 
 import (
+	_ "embed"
+
 	"regexp"
 	"strings"
 
@@ -13,6 +15,7 @@ import (
 	genhelper "github.com/openshift/cluster-logging-operator/internal/generator/helpers"
 	. "github.com/openshift/cluster-logging-operator/internal/generator/vector/elements"
 	vectorhelpers "github.com/openshift/cluster-logging-operator/internal/generator/vector/helpers"
+	commontemplate "github.com/openshift/cluster-logging-operator/internal/generator/vector/output/common/template"
 )
 
 type Endpoint struct {
@@ -58,7 +61,7 @@ type = "aws_cloudwatch_logs"
 inputs = {{.Inputs}}
 region = "{{.Region}}"
 {{.Compression}}
-group_name = "{{.GroupName}}"
+group_name = "{{"{{ _internal.cw_groupname }}"}}"
 stream_name = "{{"{{ stream_name }}"}}"
 {{compose_one .SecurityConfig}}
 healthcheck.enabled = false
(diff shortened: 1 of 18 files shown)


internal/generator/vector/output/cloudwatch/cloudwatch_test.go
```diff
@@ -122,20 +122,20 @@ var _ = Describe("Generating vector config for cloudwatch output", func() {
...
-			Entry("when groupName is spec'd", "{{.log_type}}-foo", func(spec *obs.OutputSpec) {},
+			Entry("when groupName is spec'd", `{.log_type||"missing"}-foo`, func(spec *obs.OutputSpec) {},
 				framework.NoOptions, "cw_with_groupname.toml"),
-			Entry("when URL is spec'd", "{{.log_type}}", func(spec *obs.OutputSpec) {
+			Entry("when URL is spec'd", `{.log_type||"missing"}`, func(spec *obs.OutputSpec) {
 				spec.Cloudwatch.URL = "http://mylogreceiver"
 			}, framework.NoOptions, "cw_with_url.toml"),
-			Entry("when minTLS and ciphers is spec'd", "{{.log_type}}", nil, testhelpers.FrameworkOptionWithDefaultTLSCiphers, "cw_with_tls_and_default_mintls_ciphers.toml"),
-			Entry("when tls is spec'd", "{{.log_type}}", func(spec *obs.OutputSpec) {
+			Entry("when minTLS and ciphers is spec'd", `{.log_type||"missing"}`, nil, testhelpers.FrameworkOptionWithDefaultTLSCiphers, "cw_with_tls_and_default_mintls_ciphers.toml"),
```
jcantrill, line 130: How are TLS and ciphers relevent here?

internal/generator/vector/output/cloudwatch/cloudwatch.go
```diff
@@ -158,6 +176,64 @@ del(.source_type)
 	}
 }
 
+func GroupNameTemplateRemap(componentID string, inputs []string, groupName string) Element {
```
jcantrill, line 179: It should be possible to make this "generic" for any place we wish to template "internal/generators/vector/outputs/common" ... maybe common/template ?  groupName here is the templated field

internal/generator/vector/output/cloudwatch/cw_groupname_with_aws_credentials.toml
```diff
@@ -24,9 +24,18 @@ source = '''
...
-[transforms.cw_dedot]
+# Cloudwatch Groupname
+[transforms.cw_group_name]
 type = "remap"
 inputs = ["cw_normalize_streams"]
+source = '''
+._internal.cw_groupname = "app-" + to_string!(.log_type||"missing")
+'''
+
+
```
jcantrill, line 35: nit... extra space

internal/generator/vector/output/cloudwatch/cloudwatch_test.go
Clee2691: I didn't add these tests, I only converted them to obs api as well as now with the groupname changes. I can remove them if they aren't relevant

PR #2628, opened 2024-07-10
LOG-5771: Add otlp tech-preview annotation during validate

cahartma:
### Description
Not sure what final text we want, but the logic is here.

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5771


internal/validations/observability/outputs/otlp.go
```diff
@@ -0,0 +1,17 @@
+package outputs
+
+import (
+	internalcontext "github.com/openshift/cluster-logging-operator/internal/api/context"
+	"github.com/openshift/cluster-logging-operator/internal/constants"
+	"strings"
+)
+
+func ValidateOtlpAnnotation(context internalcontext.ForwarderContext) (results []string) {
```
jcantrill, line 9: my recollection is returning nothing means its valid, which means if found and if enabled... will fail

cahartma: It returns an empty array like the other one and appends it to the messages.      Validation only fails when len(messages) > 0, so this works afaik.

jcantrill: Doesn't this fall through when the annotation is missing which means it will return:
```
return []string{"OTLP output annotation not found"}
```

jcantrill (commented):
/hold

internal/validations/observability/outputs/otlp.go
jcantrill, line 9: This needs a unit test to support the changes

cahartma: In progress.

cahartma: That's correct.   If the annotation is missing, we want validation to fail, so we add a message.     The condition on line28 below is set only when there are messages populated... 
`internalobs.NewConditionFromPrefix(obsv1.ConditionTypeValidOutputPrefix, out.Name, false, obsv1.ReasonValidationFailure, strings.Join(messages, ",")))`
If messages is empty, we set "output is valid".    You wrote the same thing for ValidateCloudWatchAuth().

PR #2639, opened 2024-07-12
LOG-5801: fix cloudwatch with roleARN from projected SA

jcantrill:
### Description
This PR:
* fixes using CW output with token from projected SA
* Removes volumemount from container for projected SA since it should not require explicit mounting

### Links
https://issues.redhat.com/browse/LOG-5801


alanconway (approved):
Will let @cahartma LGTM as I'm not very familiar.

cahartma (commented):
Looks good.  I need to run this through once locally as well once merged.
/lgtm

PR #2640, opened 2024-07-12
LOG-5787: The migration process doesn't work.

Clee2691:
### Description
This PR addresses the migration of `logging.openshift.io` resources to the new `observability.openshift.io` API during the operator upgrade process.

When the operator starts, existing logging CLs and CLFs trigger a create event, prompting the operator to mark them with an annotation. This ensures that the controller will reconcile and migrate these resources to the new API.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5787

### DIFF
--- a/internal/controller/forwarding/forwarding_controller.go
+++ b/internal/controller/forwarding/forwarding_controller.go
@@ -2,11 +2,12 @@ package forwarding
 
 import (
 	"context"
-	"github.com/openshift/cluster-logging-operator/internal/api/initialize"
 	"strings"
 	"time"
 
+	"github.com/openshift/cluster-logging-operator/internal/api/initialize"
 	"github.com/openshift/cluster-logging-operator/internal/controller"
+
 	"github.com/openshift/cluster-logging-operator/internal/validations/clusterlogforwarder"
 	"github.com/openshift/cluster-logging-operator/internal/validations/clusterlogforwarder/conditions"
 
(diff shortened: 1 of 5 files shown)


internal/controller/forwarding/forwarding_controller.go
```diff
@@ -132,12 +133,30 @@ func (r *ReconcileForwarder) updateStatus(instance *logging.ClusterLogForwarder)
...
 }
 
+// AnnotateLoggingClusterLogForwarders gets a list of existing clusterlogforwarders.logging.openshift.io
+// and annotates them with "logging.openshift.io/needs-migration"
+func AnnotateLoggingClusterLogForwarders(k8sClient client.Client, apiReader client.Reader) error {
+	var clfList logging.ClusterLogForwarderList
+	if err := apiReader.List(context.TODO(), &clfList); err != nil {
+		return err
+	}
+	if len(clfList.Items) > 0 {
```
cahartma, line 143: Is this check redundant since the for loop will be skipped if clfList.Items is empty?

jcantrill:
Why do we need to mark CRs to be migrated instead of migrating all that DONT have an annotation?

/hold
/approve

Clee2691:
> Why do we need to mark CRs to be migrated instead of migrating all that DONT have an annotation?
> 

Because upgrading causes a `create` event of existing resources. There needs to be a way to gatekeep other `create` events meaning no NEW resources should trigger a reconcile. 

If we migrate resources that don't have an annotation, then NEW CLs and CLFs can be created and migrated which you said we do not want to do (the reason `create` events were completely disabled from the beginning)

jcantrill (commented):
/lgtm

PR #2641, opened 2024-07-13
LOG-5810: fix cloudwatch stream names

cahartma:
### Description
All app logs in cloudwatch had their stream_name set to default.    This was caused by the `del(.file)` we are putting in each log_type filter.     I've created `._internal.file` to store the filename and modified the cloudwatch stream_name assignment.   
Two small changes, then the remaining files are .toml files for testing.

note: No records are dropped if either of these fields (.file, .message) is not present.  I am curious though, what other sinks could have been sending .file previously?  other output types?

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
https://issues.redhat.com/browse/LOG-5807

### DIFF
--- a/internal/generator/vector/conf/complex.toml
+++ b/internal/generator/vector/conf/complex.toml
@@ -146,6 +146,7 @@ source = '''
 if .log_type == "audit" && .log_source == "auditd" {
   .openshift.cluster_id = "${OPENSHIFT_CLUSTER_ID:-}"
   ._internal.message = .message
+  ._internal.file = .file
   del(.file)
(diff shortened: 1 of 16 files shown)


jcantrill:
Why did we do previously? How can we make this more in line with where we are going?  We never sent file with fluent and if we introduced it with vector that was unintentional

cahartma:
There are a few lines above that, or you have different code than me:  
```
.stream_name = "default"

if ( .file != null) {
 .stream_name = "kubernetes" + replace!(.file, "/", ".")
}
```

PR #2644, opened 2024-07-15
LOG-5640: Add validation to reject when CW outputs spec different web identities

vparfonov:

### Description

/cc @cahartma @Clee2691 
/assign @jcantrill  

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5640
- Enhancement proposal:


internal/validations/observability/outputs/cloudwatch.go
```diff
@@ -2,37 +2,56 @@ package outputs
 
 import (
 	obs "github.com/openshift/cluster-logging-operator/api/observability/v1"
+	"reflect"
 )
 
-func ValidateCloudWatchAuth(spec obs.OutputSpec) (results []string) {
+func ValidateCloudWatchAuth(spec obs.OutputSpec, currentIamRoleAuthSpec *obs.CloudwatchAuthentication) ([]string, *obs.CloudwatchAuthentication) {
```
jcantrill, line 8: Maybe we should have discussed further.  I think its possible that maybe most of these individual validations can go away because we already have requirements declared in the [api](https://github.com/openshift/cluster-logging-operator/blob/master/api/observability/v1/output_types.go#L291) and further [here](https://github.com/openshift/cluster-logging-operator/blob/master/api/observability/v1/output_types.go#L331)


And in review we are missing the validation like [this](https://github.com/openshift/cluster-logging-operator/blob/master/api/observability/v1/input_types.go#L56) for the cloudwatch auth


I think the only thing we need to satisfy this issue is to:

* gather all CW auth that use IAMRole and add the roleARNs to a set
* Fail if len of set > 1

vparfonov: @jcantrill Good point

PR #2646, opened 2024-07-15
Fix remap for syslog output

cahartma:
quick VRL change to syslog remap

/cc @Clee2691 @vparfonov
/assign @jcantrill


cahartma:
Closing this after discussion with Jeff.   Syslog should have everything it needs by the time it gets to this remap.   As it is currently, it should behave the same as 5.9.

PR #2649, opened 2024-07-15
LOG-5745: Forwarding logs to cloudwatch with groupName {{.kubernetes.namespace_name}}, only container logs can be forwarded.

Clee2691:
### Description
This PR fixes the validation for `Cloudwatch's` `groupName`. 
This validation ensures that fallback values are included if `groupName` contains `{}` to indicate dynamic fields. 

A dynamic value MUST end with a static string not just another field.

#### Examples:
Valid:
1. {.foo||"none"}
2. {.foo||.bar||"missing"}
3. {.kubernetes.namespace_name||.log_type||"foo"}

Invalid:
1. {.foo}
2. {.bar||.baz}

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5745

### DIFF
--- a/api/observability/v1/output_types.go
+++ b/api/observability/v1/output_types.go
@@ -301,9 +301,15 @@ type Cloudwatch struct {
 	Region string `json:"region"`
 
 	// GroupName defines the strategy for grouping logstreams
+	// The GroupName can be a combination of static and dynamic values consisting of field paths followed by `||` followed by another field path or a static value.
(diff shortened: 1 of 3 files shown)


cahartma:
I assume we also will want to explain this in the CRD?

cahartma:
I've had to add even more to get this to validate.   dot notation is not allowed in the subsequent "fieldnames", although you've accounted for it in the first.    
`{.hostname || .kubernetes.namespace_name||"foo"}`

PR #2650, opened 2024-07-16
LOG-5787: The migration process doesn't work.

Clee2691:
### Description
This PR adds a nil check for annotations on `ClusterLogging` and `ClusterLogForwarders` and initializes the map if the annotations are empty before marking existing resources.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5787


jcantrill (commented):
/approve
/lgtm

PR #2652, opened 2024-07-16
LOG-5813: Move clf.name validation to be declarative

jcantrill:
### Description
This PR:
* fixes pipeline validation to relax the name length
* Moves the CLF.name validation to be declarative

### Links
https://issues.redhat.com/browse/LOG-5813

cc @Clee2691 @cahartma @vparfonov


cahartma (commented):
Is there still a need to restrict from our reserved namespace/name, or does the api take care of that?

jcantrill:
> Is there still a need to restrict from our reserved namespace/name, or does the api take care of that?

There are no special namespaces or names in the context of 6.0.  This assures we have a name for which we can create "routeable" resources and thus the reason it still exists here to ensure it dns1035 label

PR #2662, opened 2024-07-17
fix mismatched CW enum value

jcantrill:
### Description
this pr fixes the CW auth enum mismatch

cc @cahartma


cahartma (commented):
/lgtm

PR #2672, opened 2024-07-18
LOG-5839: Update cw stream_name to no longer depend on file and tag

cahartma:
### Description
Cleaner stream names by removing pod_id and setting our own specific values per log_type.

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-5839


jcantrill (approved):
/lgtm

PR #2674, opened 2024-07-18
LOG-5840: Enable skipped functional test for CW

jcantrill:
### Description
This PR:
* enables the forward to CW multiline exception test
* Removes a test that affects any output regarding logtype to wrong outputs which we test elsewhere.  Routing from source to output has changed since this test was introduced

### Links
https://issues.redhat.com/browse/LOG-5840

### DIFF
--- a/test/functional/outputs/cloudwatch/forward_to_cloudwatch_test.go
+++ b/test/functional/outputs/cloudwatch/forward_to_cloudwatch_test.go
@@ -18,7 +18,6 @@ import (
 	"github.com/openshift/cluster-logging-operator/internal/runtime"
 	"github.com/openshift/cluster-logging-operator/internal/utils"
 	"github.com/openshift/cluster-logging-operator/test/framework/functional"
-	testfw "github.com/openshift/cluster-logging-operator/test/functional"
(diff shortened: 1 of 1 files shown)


test/functional/outputs/cloudwatch/forward_to_cloudwatch_test.go
```diff
@@ -203,58 +198,6 @@ var _ = Describe("[Functional][Outputs][CloudWatch] Forward Output to CloudWatch
 		})
 	})
 
-	Context("When sending infrastructure log messages to CloudWatch", func() {
```
cahartma, line 206: Why remove infra completely?

jcantrill: I interpreted this as a test to ensure infra logs get to the infra tenant. The functional framework is not capable of supporting this test at this time because vector only collects logs from the node where it runs; we are unable to mock it.  Since this test was written we changed the wiring from source to sink.  The implication that I see is if logs are getting mixed in tenants, then it should be prevalent with any receiver and we have a multi-output tests

cahartma (commented):
/lgtm

PR #2675, opened 2024-07-18
LOG-5841: Add rotated files back to the container excludes list

cahartma:
### Description
Was fixed in 5.9 but somehow didnt make it when everything moved to 6.0.   I may do a diff of the toml to see if any other small items we may have missed.

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-5841
- orig:  https://issues.redhat.com/browse/LOG-5156


jcantrill (commented):
/lgtm

PR #2678, opened 2024-07-22
LOG-5850: skipping reconcile process if set Unmanaged state

vparfonov:
### Description

/cc @cahartma @Clee2691 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5850
- Enhancement proposal:


internal/controller/clusterlogging/clusterlogging_controller.go
```diff
@@ -67,6 +67,11 @@ func (r *ReconcileClusterLogging) Reconcile(ctx context.Context, request ctrl.Re
 		return ctrl.Result{}, err
 	}
 
+	// skip reconcile process if Unmanaged state
+	if instance.Spec.ManagementState == loggingv1.ManagementStateUnmanaged {
+		return ctrl.Result{}, nil
```
jcantrill, line 72: Can we add a reasoncode to the Ready condition and set it to false

vparfonov: @jcantrill Why set it to `false`?

vparfonov: Something like this:
```
cond := loggingv1.NewCondition(loggingv1.ConditionReady, corev1.ConditionFalse, 
			  loggingv1.ReasonUnmanagedState, "Skip reconcile if managementState is Unmanaged")
```

jcantrill: Because the condition is "ready" and the implication is we are not ready because we are unmanaged, so the condition does not apply

PR #2679, opened 2024-07-22
LOG-5790: Allow syslog templating in CLF API syntax

Clee2691:
### Description
This PR refactors `Syslog's` output encoding of `AppName`, `MsgID`, `ProcID`, and `PayloadKey` to allow templating using static and dynamic values.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5790


api/observability/v1/output_types.go
```diff
@@ -821,30 +821,69 @@ type Syslog struct {
 	// +kubebuilder:default:=user
 	Facility string `json:"facility,omitempty"`
 
-	// PayloadKey specifies record field to use as payload.
+	// PayloadKey specifies record field to use as payload. This supports template syntax
```
jcantrill, line 824: 1. Do we need to mention the "default" key is "message" and if you specify nothing then that is what is used?
2. Does this one need special restriction to only allow you to use the "default" or a path to a field?  This seems to be one field where a static value or combination doesn't make any sense

cahartma (commented):
/lgtm

api/observability/v1/output_types.go
Clee2691: I've refactored to allow field paths encased in brackets. It will default to `.message` when it is left blank

api/observability/v1/output_types.go
```diff
@@ -821,30 +821,70 @@ type Syslog struct {
...
+	// A dynamic value is encased in single curly brackets `{}` and MUST end with a static fallback value separated with `||`.
+	// Static values can only contain alphanumeric characters along with dashes, underscores, dots and forward slashes.
+	// Example:
+	// 1. foo-{.bar||"none"}
+	// 2. {.foo||.bar||"missing"}
+	// 3. foo.{.bar.baz||.qux.quux.corge||.grault||"nil"}-waldo.fred{.plugh||"none"}
 	//
+	// +kubebuilder:validation:Pattern:=`^(([a-zA-Z0-9-_.\/])*(\{(\.[a-zA-Z0-9_]+|\."[^"]+")+((\|\|)(\.[a-zA-Z0-9_]+|\.?"[^"]+")+)*\|\|"[^"]*"\})*)*$`
 	// +kubebuilder:validation:Optional
 	// TODO: DETERMIN HOW to default the app name that isnt based on fluentd assumptions of "tag" when this is empty
```
jcantrill, line 854: I don't know what we are defaulting now and not necessarily as part of this PR but we should default these to be the same as the logstream or resource info we use for OTLP.  I don't think we need to honor the 5.9 "defaults" but maybe we can do something in migration differently then what the new, OOTB default is

internal/generator/vector/output/syslog/syslog.go
```diff
@@ -122,17 +163,61 @@ func Output(id string, o obs.OutputSpec, inputs []string, secrets vectorhelpers.
...
+func Encoding(id string, o obs.OutputSpec, templatePairs []FieldVRLStringPair) []Element {
+	// Set the appropriate encoding fields that have been user templated
+	encodingFields := []Element{
+		SyslogEncoding{
+			ComponentID:  id,
+			RFC:          strings.ToLower(string(o.Syslog.RFC)),
+			Facility:     Facility(o.Syslog),
+			Severity:     Severity(o.Syslog),
+			AddLogSource: genhelper.NewOptionalPair("add_log_source", o.Syslog.Enrichment == obs.EnrichmentTypeKubernetesMinimal),
+			PayloadKey:   genhelper.NewOptionalPair("payload_key", "payload_key"),
```
jcantrill, line 204: is the initial value intended to be "payload_key"

PR #2680, opened 2024-07-22
LOG-5131: Backwards-compatible telemetry implementation for new API

xperimental:
### Description

This PR reintroduces the telemetry metrics in the cluster-logging-operator. The implementation produces a set of metrics compatible with the previous CLO version. The label-set on the info-metrics for inputs/outputs is based on the available types of 6.x, so some labels are missing from the previous version and some are added (for example, the "default" output).

It does not create any metrics for `loggingv1.ClusterLogging` or `loggingv1.ClusterLogForwarder` instances, but the `loggingv1alpha1.LogFileMetricExporter` metric is still present.

### Links

- JIRA: [LOG-5131](https://issues.redhat.com//browse/LOG-5131)

### DIFF
--- a/internal/metrics/telemetry/cluster_log_forwarder.go
+++ b/internal/metrics/telemetry/cluster_log_forwarder.go
@@ -1,0 +1,44 @@
+package telemetry
+
+import (
+	observabilityv1 "github.com/openshift/cluster-logging-operator/api/observability/v1"
+	internalobs "github.com/openshift/cluster-logging-operator/internal/api/observability"
+)
+
+func gatherForwarderInfo(forwarder *observabilityv1.ClusterLogForwarder) (pipelines uint, inputs, outputs []string) {
+	inputs = makeZeroStrings(len(forwarderInputTypes))
+	outputs = makeZeroStrings(len(forwarderOutputTypes))
+
+	outputMap := internalobs.Outputs(forwarder.Spec.Outputs).Map()
+
+	activeInputNames := map[string]bool{}
+	activeOutputTypes := map[string]bool{}
+	for _, pipeline := range forwarder.Spec.Pipelines {
+		for _, inputName := range pipeline.InputRefs {
+			activeInputNames[inputName] = true
+		}
+
+		for _, outputName := range pipeline.OutputRefs {
+			output, found := outputMap[outputName]
+			if found {
+				activeOutputTypes[string(output.Type)] = true
+			}
+		}
+	}
+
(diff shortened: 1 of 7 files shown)


internal/metrics/telemetry/cluster_log_forwarder.go
```diff
@@ -0,0 +1,51 @@
...
+	for i, v := range forwarderOutputTypes {
+		if activeOutputTypes[v] {
+			outputs[i] = "1"
+		}
+	}
+
+	return pipelines, inputs, outputs
+}
+
+func makeOutputMap(spec *observabilityv1.ClusterLogForwarderSpec) map[string]*observabilityv1.OutputSpec {
```
jcantrill, line 45: There is a similar function to this in internal/api/observability

internal/metrics/telemetry/cluster_log_forwarder.go
```diff
@@ -0,0 +1,51 @@
...
+				activeOutputTypes[string(output.Type)] = true
+			}
+		}
+	}
+
+	pipelines = uint(len(forwarder.Spec.Pipelines))
+
+	for i, v := range forwarderInputTypes {
+		if activeInputNames[v] {
+			inputs[i] = "1"
```
jcantrill, line 32: should we use "boolLabel" here?

jcantrill (commented):
/hold

internal/metrics/telemetry/cluster_log_forwarder.go
xperimental: I just created `boolLabel` for the "inline usage", I don't think it would add any value here, just another function call and an unnecessary conditional (not sure, if that's optimized away). But I am quite neutral on this decision...

Should I make the "0" / "1" strings constants?

xperimental:
@jcantrill After having spent a bit of time with the existing telemetry metrics in CLO 5.y, I think I came to the conclusion, that it would be better to replace this in 6.0 instead of producing a set of metrics compatible with the previous versions.

Please keep this PR on hold. I'll either update this one with a new implementation or provide a second PR so that we can compare.

jcantrill:
https://issues.redhat.com/browse/LOG-5865

xperimental:
fyi: The alternative implementation is [here](https://github.com/openshift/cluster-logging-operator/compare/master...xperimental:cluster-logging-operator:new-telemetry-6.0).

xperimental:
Closing, because the alternative has been merged.

PR #2683, opened 2024-07-23
LOG-5846: fix filter name after migration according to spec

vparfonov:
### Description
Fix filter name generated during migration process according to specification, must match to the: 
`^[a-z][a-z0-9-]*[a-z0-9]$`

/cc @cahartma @Clee2691 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5846
- Enhancement proposal:

### DIFF
--- a/internal/migrations/observability/api/pipelines.go
+++ b/internal/migrations/observability/api/pipelines.go
@@ -96,7 +96,7 @@ func generatePipelineFilters(loggingPipelineSpec logging.PipelineSpec, createdFi
 		}
 	}
 	if len(loggingPipelineSpec.Labels) != 0 {
-		openshiftLabelFilterName := fmt.Sprintf("filter-%s-%s", loggingPipelineSpec.Name, string(obs.FilterTypeOpenshiftLabels))
+		openshiftLabelFilterName := fmt.Sprintf("filter-%s-%s", loggingPipelineSpec.Name, openshiftLabelsFilterName)
 		addedFilterRefs = append(addedFilterRefs, openshiftLabelFilterName)
 		obsFilter := &obs.FilterSpec{
 			Type:            obs.FilterTypeOpenshiftLabels,
--- a/internal/migrations/observability/api/pipelines_test.go
+++ b/internal/migrations/observability/api/pipelines_test.go
@@ -27,7 +27,7 @@ var _ = Describe("#ConvertPipelines", func() {
 		Expect(actFilterRefs).To(Equal(expObsFilterRefs))
 	},
 		Entry("should generate detect multiline error as a filter and filterRef", logging.PipelineSpec{
-			Name:                  "detectPipeline",
+			Name:                  "detect-pipeline",
 			DetectMultilineErrors: true,
 		}, []obs.FilterSpec{
 			{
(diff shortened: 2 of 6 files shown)


internal/migrations/observability/api/pipelines_test.go
```diff
@@ -51,12 +51,12 @@ var _ = Describe("#ConvertPipelines", func() {
 			Labels: map[string]string{"foo": "bar"},
 		}, []obs.FilterSpec{
 			{
-				Name:            "filter-labelPipeline-openShiftLabels",
+				Name:            "filter-labelPipeline-" + openshiftLabelsFilterName,
```
Clee2691, line 54: The regex is saying no capital letters am I right? The `P` should be lowercase.
```suggestion
				Name:            "filter-labelpipeline-" + openshiftLabelsFilterName,
```

internal/migrations/observability/api/pipelines_test.go
```diff
@@ -51,12 +51,12 @@ var _ = Describe("#ConvertPipelines", func() {
...
 				Type:            obs.FilterTypeOpenshiftLabels,
 				OpenShiftLabels: map[string]string{"foo": "bar"},
 			},
 		},
-			[]string{"filter-labelPipeline-openShiftLabels"}),
+			[]string{"filter-labelPipeline-" + openshiftLabelsFilterName}),
```
Clee2691, line 59: Same comment as above...
```suggestion
			[]string{"filter-labelipeline-" + openshiftLabelsFilterName}),
```

internal/migrations/observability/api/pipelines_test.go
```diff
@@ -72,12 +72,12 @@ var _ = Describe("#ConvertPipelines", func() {
 				Type: obs.FilterTypeParse,
 			},
 			{
-				Name:            "filter-filterPipeline-openShiftLabels",
+				Name:            "filter-filterPipeline-" + openshiftLabelsFilterName,
```
Clee2691, line 75: same as above

internal/migrations/observability/api/pipelines_test.go
```diff
@@ -72,12 +72,12 @@ var _ = Describe("#ConvertPipelines", func() {
...
 				Type:            obs.FilterTypeOpenshiftLabels,
 				OpenShiftLabels: map[string]string{"foo": "bar"},
 			},
 		},
-			[]string{detectMultilineErrorFilterName, parseFilterName, "filter-filterPipeline-openShiftLabels"}),
+			[]string{detectMultilineErrorFilterName, parseFilterName, "filter-filterPipeline-" + openshiftLabelsFilterName}),
```
Clee2691, line 80: same as above

cahartma (commented):
/lgtm

internal/migrations/observability/api/pipelines.go
```diff
@@ -96,7 +96,7 @@ func generatePipelineFilters(loggingPipelineSpec logging.PipelineSpec, createdFi
 		}
 	}
 	if len(loggingPipelineSpec.Labels) != 0 {
-		openshiftLabelFilterName := fmt.Sprintf("filter-%s-%s", loggingPipelineSpec.Name, string(obs.FilterTypeOpenshiftLabels))
+		openshiftLabelFilterName := fmt.Sprintf("filter-%s-%s", loggingPipelineSpec.Name, openshiftLabelsFilterName)
```
jcantrill, line 99: why not downcase the name instead of introducing an unnecessary constant?

vparfonov: i think that have readable name make sense here

PR #2684, opened 2024-07-23
LOG-5842 Forwarder validation failing on old conditions

cahartma:
### Description
 Added `removeStaleStatuses` to prune conditions for any resource that may no longer be listed (name change for example).   
I tried to create a test for this, but am not just going to push the fix.   I may come back to this later.   

I also changed the input validation to use SetCondition() like the others, and added name to the other messages to be consistent between all Condition types.

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-5842

### DIFF
--- a/internal/api/observability/conditions.go
+++ b/internal/api/observability/conditions.go
@@ -5,6 +5,7 @@ import (
 	obs "github.com/openshift/cluster-logging-operator/api/observability/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeclock "k8s.io/utils/clock"
+	"strings"
 )
 
 // clock is used to set status condition timestamps.
@@ -50,4 +51,17 @@ func SetCondition(conditions *[]metav1.Condition, newCond metav1.Condition) bool
 	}
 	*conditions = append(*conditions, newCond)
 	return true
+}
+
+// PruneConditions keeps only those conditions that match names from the spec
+func PruneConditions(conditions *[]metav1.Condition, spec NameList) {
+	keepers := []metav1.Condition{}
+	for _, condition := range *conditions {
+		for _, name := range spec.Names() {
+			if strings.HasSuffix(condition.Type, name) {
+				keepers = append(keepers, condition)
+			}
+		}
+	}
+	*conditions = keepers
 }
--- a/internal/api/observability/forwarder.go
+++ b/internal/api/observability/forwarder.go
@@ -7,6 +7,11 @@ import (
 	"strings"
 )
 
+// NameList provides a list of names for any resource implementing Names
+type NameList interface {
+	Names() []string
+}
+
 // DeployAsDeployment evaluates the spec to determine if the collector will be deployed as a deployment.
 // Collector is not a daemonset if the only input source is an HTTP receiver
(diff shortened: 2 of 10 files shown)


internal/api/observability/conditions.go
```diff
@@ -51,3 +53,39 @@ func SetCondition(conditions *[]metav1.Condition, newCond metav1.Condition) bool
 	*conditions = append(*conditions, newCond)
 	return true
 }
+
+// PruneConditions keeps only those conditions that match an existing spec name
+func PruneConditions(conditions *[]metav1.Condition, context internalcontext.ForwarderContext) {
```
jcantrill, line 58: The context isnt being used and can be removed... or  make it the only arg and iterate through them here.  alternatively:

We are already sending the right set of conditions for a given spec.  We could introduce an interface that our internal/api/obs types implement and condense this and make it generic:

```
type NameList interface {
    Names() []string
}
func (outputs Outputs) Names []string{
}
func (inputs Inputs) Names []string{
}

func PruneConditions(conditions *[]metav1.Condition, names NameList) {
	keepers := []metav1.Condition{}
	for _, condition := range *conditions {
		for _, name := range names {
			if strings.HasSuffix(condition.Type, name) {
				keepers = append(keepers, condition)
			}
		}
        }
    *conditions = keepers
}

. . .

internalobs.PruneConditions(&forwarder.Status.Inputs, forwarder.Spec.Inputs)
internalobs.PruneConditions(&forwarder.Status.Outputs, forwarder.Spec.Outputs)
...
```

internal/controller/observability/clusterlogforwarder_controller.go
```diff
@@ -50,6 +50,8 @@ func (r *ClusterLogForwarderReconciler) Reconcile(ctx context.Context, req ctrl.
 		return defaultRequeue, err
 	}
 
+	removeStaleStatuses(r.Forwarder)
```
cahartma, line 53: @jcantrill Do we need some sort of error handling here?

internal/api/observability/forwarder.go
```diff
@@ -7,6 +7,11 @@ import (
 	"strings"
 )
 
+// NameList provides a list of names for any resource implementing Names
+type NameList interface {
```
cahartma, line 11: @jcantrill Any other suggestions for a good place to put this?  Should be somewhere in package `observability`

internal/api/observability/conditions.go
cahartma: Refactored

internal/api/observability/forwarder.go
jcantrill: I think this is fine unless we find other places where we might use. Maybe its a utility type? :man_shrugging: .  I envisioned this package as wrappers around the public API to provide useful functionality.

internal/controller/observability/clusterlogforwarder_controller.go
jcantrill: This is working on an existing struct where I don't see any opportunity for an error to be produced so there is nothing to handle.  At best it would be maybe a nil pointer :man_shrugging:

PR #2686, opened 2024-07-23
LOG-5849: Fix node log stream names set twice

cahartma:
### Description
quick fix

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-5849


jcantrill (approved):
/hold cancel

PR #2688, opened 2024-07-23
LOG-5790: Allow syslog templating in CLF API syntax

Clee2691:
### Description
This PR removes the `payload_key` config if it is not defined by the user.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5790

### DIFF
--- a/internal/generator/vector/output/syslog/syslog.go
+++ b/internal/generator/vector/output/syslog/syslog.go
@@ -193,18 +193,22 @@ func getEncodingTemplatesAndFields(s *obs.Syslog) EncodingTemplateField {
 }
 
 func Encoding(id string, o obs.OutputSpec, templatePairs []FieldVRLStringPair) []Element {
-	// Set the appropriate encoding fields that have been user templated
-	encodingFields := []Element{
-		SyslogEncoding{
-			ComponentID:  id,
-			RFC:          strings.ToLower(string(o.Syslog.RFC)),
-			Facility:     Facility(o.Syslog),
-			Severity:     Severity(o.Syslog),
-			AddLogSource: genhelper.NewOptionalPair("add_log_source", o.Syslog.Enrichment == obs.EnrichmentTypeKubernetesMinimal),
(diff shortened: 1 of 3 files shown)


internal/generator/vector/output/syslog/syslog.go
```diff
@@ -193,18 +193,22 @@ func getEncodingTemplatesAndFields(s *obs.Syslog) EncodingTemplateField {
...
+	sysLEncode := SyslogEncoding{
+		ComponentID:  id,
+		RFC:          strings.ToLower(string(o.Syslog.RFC)),
+		Facility:     Facility(o.Syslog),
+		Severity:     Severity(o.Syslog),
+		AddLogSource: genhelper.NewOptionalPair("add_log_source", o.Syslog.Enrichment == obs.EnrichmentTypeKubernetesMinimal),
+		PayloadKey:   genhelper.NewOptionalPair("payload_key", nil),
+	}
+	if o.Syslog.PayloadKey != "" {
+		sysLEncode.PayloadKey.Value = "payload_key"
```
jcantrill, line 205: I would look at optional pair and what it does when it results in a string but this logic can either be:
```
genhelper.NewOptionalPair("payload_key", o.Syslog.PayloadKey)
```
or  must be:
```
if o.Syslog.PayloadKey != "" {
		sysLEncode.PayloadKey.Value = o.Syslog.PayloadKey
```
you don't want it to be the literal 'payload_key' right?

internal/generator/vector/output/syslog/syslog.go
```diff
@@ -246,7 +250,7 @@ func Severity(s *obs.Syslog) string {
 func PayloadKey(plKey string) string {
```
jcantrill, line 250: Where is this used?  Is it still needed if we are using optionalpair ?

jcantrill (commented):
/hold

PR #2689, opened 2024-07-23
LOG-5862: Validation rule causing  error on install

cahartma:
### Description
The validation error can also be verified by trying to applying the CRD to the cluster.   I was not able to reproduce the `"CRD is not present"` errors, but I am assuming the first (validation) error is causing the requirementStatus to fail as well.

Also adding a shortname back into the CRD.  In case of upgrade scenario, user can now use `obsclf` (until we remove the old CRD)

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-5862

### DIFF
--- a/api/observability/v1/clusterlogforwarder_types.go
+++ b/api/observability/v1/clusterlogforwarder_types.go
@@ -300,8 +300,8 @@ type ClusterLogForwarderStatus struct {
 //
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:resource:categories=observability,shortName=clf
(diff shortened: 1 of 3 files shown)


api/observability/v1/clusterlogforwarder_types.go
```diff
@@ -301,7 +301,7 @@ type ClusterLogForwarderStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:resource:categories=observability,shortName=clf
```
Clee2691, line 303: Does the shortname of obsclf also have to be added here?

cahartma: Probably, good catch.

cahartma: It's what I get for "sneaking" in a few other changes.

PR #2693, opened 2024-07-25
OBSDOCS-1220 - Single source CR testing.

libander:
[OBSDOCS-1220](https://issues.redhat.com//browse/OBSDOCS-1220) - Single source CR testing.

Created an asciidoc file to test single sourcing CR config samples for common use cases. Appropriate use of tagged regions should enable CR samples from this file to be included in the product docs individually, creating lower maintenance overhead, and ensuring all CR samples are properly validated.


cahartma:
This is good for now as POC, but the file will need to be renamed, and probably move up one level, once merged.   
/lgtm

libander:
> This is good for now as POC, but the file will need to be renamed, and probably move up one level, once merged.   
> 
> /lgtm

Yep, agree. Also should hammer out the comment vs visible balance.

PR #2700, opened 2024-07-29
LOG-5890: Migrating syslog output failed due to syslog.rfc: Unsupported value

Clee2691:
### Description
This PR fixes migrating a syslog output that doesn't spec an `RFC` value. It will default to `RFC5424`. Additonally, this PR refactors migration to default values for `facility` & `severity` if none are specified as specified by the API.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5890


jcantrill (commented):
/approve
/lgtm

PR #2706, opened 2024-07-30
Fix per-tenant label keys not using global default keys

xperimental:
### Description

During a discussion I noticed that there is a case, when we need to explicitly pass the default label keys through to the Loki output: when a per-tenant provides custom keys, `ignoreGlobal=false` and there are no global overrides.

This PR adds this case to the implementation.


cahartma (commented):
/lgtm

PR #2718, opened 2024-08-01
LOG-5853: Move dedot vrl to post filter to fix drop test on labels

cahartma:
### Description
Moving the dedot vrl out of the `viaq` filter and appended to the list as a post filter, to always come after all filters.   Azure and cloudwatch output types still included dedot, so these were removed.
I also reworded the validation slightly on drop filter to avoid any confusion (my own).   
Updated the feature doc to emphasize that these are always "regex" expressions.   
TODO: we need to update our feature docs for v6 spec.  I have a folder started by Libby that we can begin to copy these specs to.   

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-5853


jcantrill (approved):

jcantrill (commented):
/lgtm

PR #2723, opened 2024-08-01
LOG-5922: fix the default ES secret

jcantrill:
### Description
This PR:
* fixes the secret for default ES after migration to use the same secrets used in 5.9

### Links
https://issues.redhat.com/browse/LOG-5922


jcantrill:
cc @Clee2691 @cahartma

PR #2726, opened 2024-08-02
LOG-5905: Migration document

jcantrill:
### Description
This PR:
* Provides a migration guide for administrators

### Links
https://issues.redhat.com/browse/LOG-5905
cc @Clee2691 @cahartma @vparfonov


cahartma:
This is very good.   Great work!      Customized "default" output in the logging.clf  may be useful here as well.  I can add that when I add CW token.   We can talk with Libby, but may need to copy (or link) the cr samples somehow in the referenced upstream directory. (rename needed)
/lgtm

PR #2736, opened 2024-08-07
LOG-6207: enable stream (stdout/stderr) information in logs for Vector

vparfonov:
### Description
This PR introduces handling of `kubernetes_log` `stream` field (e.g., stdout, stderr). https://vector.dev/docs/reference/configuration/sources/kubernetes_logs/#line-fields

- moved the original `stream` field to `kubernetes.stream` to better align with the log event structure
- added `kubernetes.stream` to the group_by parameter of the Detect Multiline Exceptions filter, ensures that exception detection is correctly handled 
- included `kubernetes.stream `in the group_by parameter for the OTLP output, to improves the granularity of log data sent through OTLP by differentiating between different streams.
- modified the functional tests to verify that log events are handled correctly for both stdout and stderr outputs. 

/cc @cahartma @Clee2691 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-6207
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/output/otlp/otlp_all.toml
+++ b/internal/generator/vector/output/otlp/otlp_all.toml
@@ -60,7 +60,8 @@ source = '''
   r.attributes = append(r.attributes,
       [{"key": "k8s.pod.uid", "value": {"stringValue": get!(.,["kubernetes","pod_id"])}},
       {"key": "k8s.container.id", "value": {"stringValue": get!(.,["kubernetes","container_id"])}},
-      {"key": "k8s.node.name", "value": {"stringValue": .hostname}}]
+      {"key": "k8s.node.name", "value": {"stringValue": .hostname}},
+      {"key": "log.iostream", "value": {"stringValue": get!(.,["kubernetes","container_iostream"])}}]
   )
   if exists(.kubernetes.labels) {for_each(object!(.kubernetes.labels)) -> |key,value| {
       r.attributes = append(r.attributes,
--- a/internal/generator/vector/output/otlp/transform.go
+++ b/internal/generator/vector/output/otlp/transform.go
@@ -67,7 +67,8 @@ if exists(.openshift.labels) {for_each(object!(.openshift.labels)) -> |key,value
 r.attributes = append(r.attributes,
     [{"key": "k8s.pod.uid", "value": {"stringValue": get!(.,["kubernetes","pod_id"])}},
     {"key": "k8s.container.id", "value": {"stringValue": get!(.,["kubernetes","container_id"])}},
-    {"key": "k8s.node.name", "value": {"stringValue": .hostname}}]
+    {"key": "k8s.node.name", "value": {"stringValue": .hostname}},
+    {"key": "log.iostream", "value": {"stringValue": get!(.,["kubernetes","container_iostream"])}}]
 )
 if exists(.kubernetes.labels) {for_each(object!(.kubernetes.labels)) -> |key,value| {
     r.attributes = append(r.attributes,
--- a/test/functional/filters/multilineexception/multiline_exception_test.go
+++ b/test/functional/filters/multilineexception/multiline_exception_test.go
@@ -1,6 +1,7 @@
 package multilineexception
 
 import (
+	"github.com/openshift/cluster-logging-operator/internal/constants"
 	testruntime "github.com/openshift/cluster-logging-operator/test/runtime/observability"
 	"strings"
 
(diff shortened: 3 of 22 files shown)


internal/generator/vector/output/otlp/transform.go
```diff
@@ -28,7 +28,8 @@ resource.attributes = append( resource.attributes,
 resource.attributes = append( resource.attributes,
     [{"key": "k8s.pod.name", "value": {"stringValue": get!(.,["kubernetes","pod_name"])}},
     {"key": "k8s.container.name", "value": {"stringValue": get!(.,["kubernetes","container_name"])}},
-    {"key": "k8s.namespace.name", "value": {"stringValue": get!(.,["kubernetes","namespace_name"])}}]
+    {"key": "k8s.namespace.name", "value": {"stringValue": get!(.,["kubernetes","namespace_name"])}},
+    {"key": "k8s.stream", "value": {"stringValue": get!(.,["kubernetes","stream"])}}]
```
cahartma, line 32: For OTLP output, if you are adding to the resource attributes, then it also needs to be included in the `group_by` for container logs.  https://github.com/openshift/cluster-logging-operator/blob/master/internal/generator/vector/output/otlp/group_by.go#L39

internal/generator/vector/output/otlp/otlp_all.toml
```diff
@@ -24,7 +24,8 @@ source = '''
   resource.attributes = append( resource.attributes,
       [{"key": "k8s.pod.name", "value": {"stringValue": get!(.,["kubernetes","pod_name"])}},
       {"key": "k8s.container.name", "value": {"stringValue": get!(.,["kubernetes","container_name"])}},
-      {"key": "k8s.namespace.name", "value": {"stringValue": get!(.,["kubernetes","namespace_name"])}}]
+      {"key": "k8s.namespace.name", "value": {"stringValue": get!(.,["kubernetes","namespace_name"])}},
+       {"key": "log.iostream", "value": {"stringValue": get!(.,["kubernetes","container_iostream"])}}]
```
jcantrill, line 28: We can revert this change since IOStream is part of the log entry where this grouping is a way to define the overall resource which is the container workload

test/functional/filters/multilineexception/multiline_exception_test.go
```diff
@@ -108,12 +106,34 @@ created by main.main
...
+				ToHttpOutput()
+			//LOG-2241
+			b.FromInput("multiline-log-ns").
+				Named("other").
+				WithMultilineErrorDetectionFilter().
+				ToElasticSearchOutput()
+		}),
+
+		// exceptions to stderr
+		Entry("of Java services to stderr stream", constants.STDERR, javaException, nil),
```
jcantrill, line 131: Consider reverting this and only update some of the existing tests to use STDERR.  The logic will suffice and it keeps from expanding the tests unnecessarily

internal/generator/vector/output/otlp/group_by.go
```diff
@@ -43,7 +43,7 @@ func GroupByContainer(id string, inputs []string) Element {
 		Inputs:      helpers.MakeInputs(inputs...),
 		MaxEvents:   "250",
 		GroupBy: MakeGroupBys(".openshift.cluster_id",
-			".kubernetes.namespace_name", ".kubernetes.pod_name", ".kubernetes.container_name"),
+			".kubernetes.namespace_name", ".kubernetes.pod_name", ".kubernetes.container_name", ".kubernetes.container_iostream"),
```
jcantrill, line 46: @vparfonov iostream here is not an attribute to group_by for OTLP since its simply an attribute of the log entry and not the resource which is the container

internal/generator/vector/output/otlp/transform.go
```diff
@@ -66,7 +66,8 @@ if exists(.openshift.labels) {for_each(object!(.openshift.labels)) -> |key,value
 r.attributes = append(r.attributes,
     [{"key": "k8s.pod.uid", "value": {"stringValue": get!(.,["kubernetes","pod_id"])}},
     {"key": "k8s.container.id", "value": {"stringValue": get!(.,["kubernetes","container_id"])}},
-    {"key": "k8s.node.name", "value": {"stringValue": .hostname}}]
+    {"key": "k8s.node.name", "value": {"stringValue": .hostname}},
+    {"key": "log.iostream", "value": {"stringValue": get!(.,["kubernetes","container_iostream"])}}]
```
jcantrill, line 70: This should be added to the attributes for "logdata" of containers and not the "resource" attribute.

cahartma: r.attributes is the logRecord attribute.   This line is correct.

cahartma: One more question:   
If you are setting the value to `.kubernetes.container_iostream`   then it may seem more accurate to map this to `k8s.container.iostream`  ??    or at least maybe not `openshift.log`, and instead `k8s.log`  ??

jcantrill: > r.attributes is the logRecord attribute.
Good catch

@vparfonov `log.iostream` is documented in the conventions.  Please leave as-is for now

PR #2740, opened 2024-08-08
LOG-4805: The log severity should be sorted from most priority to less

Clee2691:
### Description
This PR changes the way a log event's `level` is set. It will first match on well known, structured patterns and if none match, it will match on unstructured keywords.

The order of the level is as follows and is sorted in descending priority: `emergency`, `alert`, `critical`, `error`, `warn`, `notice`, `info`, `debug`.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links

- JIRA: https://issues.redhat.com/browse/LOG-4805


jcantrill (approved):
/approve

test/functional/normalization/loglevel/container_logs_test.go
```diff
@@ -57,7 +58,6 @@ var _ = Describe("[functional][normalization][loglevel] tests for message format
 		outputTestLog := logs[0]
 		Expect(outputTestLog).To(FitLogFormatTemplate(outputLogTemplate))
 	},
-		Entry("should recognize a WARN message", "warn", "Warning: failed to query journal: Bad message OS Error 74"),
```
vparfonov, line 60: Why do you remove this test?

test/functional/normalization/loglevel/container_logs_test.go
vparfonov, line 1: This is not related to your changes, but can you take a look at the tests from line 65 and below, look like description not fit to given parameters

test/functional/normalization/loglevel/container_logs_test.go
Clee2691: We determined that we would stick with the changes for the keyword order and just remove this specific test. This is our best effort for setting log severity.

PR #2741, opened 2024-08-08
LOG-5948: Fix projected token logic

cahartma:
### Description
Token is now required when CW auth type is roleARN.   Also, we are no longer projecting a SA token, if the token is specified from a secret.   

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-5948


internal/collector/collector_test.go
```diff
@@ -520,13 +556,14 @@ var _ = Describe("Factory#Deployment#NewPodSpec", func() {
...
 					MountPath: constants.TrustedCABundleMountDir,
 				}))
-			})
-		})
-
-		Context("and mounting volumes", func() {
-			It("should not mount host path volumes", func() {
-				Expect(podSpec.Volumes).NotTo(ContainElement(v1.Volume{Name: sourcePodsName, VolumeSource: v1.VolumeSource{HostPath: &v1.HostPathVolumeSource{Path: sourcePodsPath}}}))
-				Expect(podSpec.Volumes).To(HaveLen(6))
+				Expect(podSpec.Volumes).To(ContainElement(v1.Volume{
```
jcantrill, line 559: There is a custom matcher for this though in review I am unable to attest to its value now

jcantrill (approved):

PR #2748, opened 2024-08-09
LOG-5525: trigger collector reload on update

jcantrill:
### Description
This PR:
* triggers a collector reload on secret change

### Links
* https://issues.redhat.com/browse/LOG-5525


internal/api/observability/secrets.go
```diff
@@ -11,3 +17,83 @@ func NewSecretReference(keyName, secretName string) *obs.SecretReference {
 		SecretName: secretName,
 	}
 }
+
+// Secrets is a map of secrets
+type Secrets map[string]*corev1.Secret
+
+// HashMD5 returns an MD5 representation of the secrets
+func (s Secrets) HashMD5() string {
```
xperimental, line 25: I would recommend not using MD5 as a hashing algorithm. Either use a "proper cryptographic hash" like SHA256/512 or use a newer non-cryptographic hash like [fnv](https://pkg.go.dev/hash/fnv) to signal "this has no security implications".

internal/api/observability/secrets.go
```diff
@@ -11,3 +17,83 @@ func NewSecretReference(keyName, secretName string) *obs.SecretReference {
 		SecretName: secretName,
 	}
 }
+
+// Secrets is a map of secrets
+type Secrets map[string]*corev1.Secret
```
xperimental, line 22: We have many mentions of "Secret" in different forms already. Maybe `SecretBundle` or something like that to have a different name for it?

internal/collector/collector.go
```diff
@@ -177,6 +176,7 @@ func (f *Factory) NewCollectorContainer(inputs internalobs.Inputs, outputs inter
 	}
 	collector.Env = []v1.EnvVar{
 		{Name: "COLLECTOR_CONF_HASH", Value: f.ConfigHash},
+		{Name: "SECRETS_HASH", Value: f.Secrets.HashMD5()},
```
xperimental, line 179: Is this hash output somewhere? If not, then it might be an idea to store it in a Pod annotation instead (or do both and populate the environment variable from the Pod annotation).

internal/api/observability/secrets.go
jcantrill: Team consensus is leave as-is

internal/api/observability/secrets.go
```diff
@@ -11,3 +16,79 @@ func NewSecretReference(keyName, secretName string) *obs.SecretReference {
...
+
+// Hash64a returns an FNV-1a representation of the secrets
+func (s Secrets) Hash64a() string {
+	names := s.Names()
+	sort.Strings(names)
+	buffer := fnv.New64a()
+	//buffer := bytes.NewBufferString("")
```
xperimental, line 28:
```suggestion
```

internal/api/observability/secrets.go
```diff
@@ -11,3 +16,79 @@ func NewSecretReference(keyName, secretName string) *obs.SecretReference {
...
+		}
+		sort.Strings(keys)
+
+		for _, k := range keys {
+			v := secret.Data[k]
+			buffer.Write([]byte(k))
+			buffer.Write(v)
+		}
+	}
+	return fmt.Sprintf("%d", buffer.Sum64())
```
xperimental, line 45: So far I have seen general consensus to be "output hash as hex digits", should we do this here as well?

```suggestion
	return fmt.Sprintf("%x", buffer.Sum64())
```

internal/collector/deployment.go
```diff
@@ -20,7 +21,11 @@ func (f *Factory) ReconcileDeployment(k8sClient client.Client, namespace string,
 	tlsProfile, _ := tls.FetchAPIServerTlsProfile(k8sClient)
 	desired := f.NewDeployment(namespace, f.ResourceNames.DaemonSetName(), trustedCABundle, tls.GetClusterTLSProfileSpec(tlsProfile))
 	utils.AddOwnerRefToObject(desired, owner)
-	return reconcile.Deployment(k8sClient, desired)
+	return reconcile.Deployment(k8sClient, desired, func(current, desired apps.Deployment) (bool, string) {
+		currentHash, _ := current.Spec.Template.Annotations[annotationConfigHash]
+		desiredHash, _ := desired.Spec.Template.Annotations[annotationConfigHash]
+		return currentHash == desiredHash, ""
```
xperimental, line 27: Why not add this to `deployments.AreSame` instead of having an additional point where comparisons are made?

internal/utils/comparators/deployments/comparator.go
```diff
@@ -8,8 +8,10 @@ import (
 	apps "k8s.io/api/apps/v1"
 )
 
+type DeploymentComparator func(current apps.Deployment, desired apps.Deployment) (equal bool, reason string)
+
 // AreSame compares deployments for equality and return true equal otherwise false
-func AreSame(current *apps.Deployment, desired *apps.Deployment) (bool, string) {
+func AreSame(current *apps.Deployment, desired *apps.Deployment, additional ...DeploymentComparator) (bool, string) {
```
xperimental, line 14: As far as I can see this is currently called from exactly one place. I think it would be much simpler to just do the additional comparison here instead of adding an additional "comparison callback" that moves the comparison somewhere else.

Same applies to DaemonSet, as far as I can see.

xperimental (commented):
Does it make sense to replace the MD5-based implementation for `COLLECTOR_CONF_HASH` with fnv as well?

internal/utils/comparators/deployments/comparator.go
jcantrill: It is only separated because its a general use function.  Granted its only used in one place but in the interest of proper design it makes sense IMO to leave it separated.  Though...if its not populated or expected to be populated then it evaluates to success either way

jcantrill:
> Does it make sense to replace the MD5-based implementation for `COLLECTOR_CONF_HASH` with fnv as well?

I think it does but given where we are in the release cycle I think this should either be a z-stream change or a next release change.  It's more invasive then I would like to add as part of the changes here

internal/utils/comparators/daemonsets/comparator_test.go
```diff
@@ -58,6 +59,15 @@ var _ = Describe("daemonset#AreSame", func() {
...
 			Expect(ok).To(BeTrue())
 		})
+
+		It("should fail when the secret hash is different", func() {
+			current.Spec.Template.Annotations = map[string]string{
+				constants.AnnotationSecretHash: "foo",
+			}
+			ok, _ := daemonsets.AreSame(current, desired)
+
+			Expect(ok).To(BeFalse())
```
xperimental, line 69: Tests :+1: 

Should we also check the second return type?

```suggestion
			ok, reason := daemonsets.AreSame(current, desired)

			Expect(ok).To(BeFalse())
			Expect(reason).To(Equal("secretHash"))
```

internal/utils/comparators/deployments/comparator_test.go
```diff
@@ -58,6 +59,15 @@ var _ = Describe("deployments#AreSame", func() {
...
 			Expect(ok).To(BeTrue())
 		})
+
+		It("should fail when the secret hash is different", func() {
+			current.Spec.Template.Annotations = map[string]string{
+				constants.AnnotationSecretHash: "foo",
+			}
+			ok, _ := deployments.AreSame(current, desired)
+
+			Expect(ok).To(BeFalse())
```
xperimental, line 69: Similar to `deployments`:

```suggestion
			ok, reason := deployments.AreSame(current, desired)

			Expect(ok).To(BeFalse())
			Expect(reason).To(Equal("secretHash"))
```

internal/api/observability/secrets.go
xperimental: For me there is value in sticking to conventions, even when it's just an implementation detail. Avoids surprises when switching from one project to another.

But ... you know ... consider it "optional". :slightly_smiling_face:

PR #2749, opened 2024-08-12
LOG-5815:forcibly to_string before match regex to fix working with not string field (e.g. integer)

vparfonov:
### Description

/cc @cahartma @Clee2691 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5815
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/filter/drop/filter.go
+++ b/internal/generator/vector/filter/drop/filter.go
@@ -21,15 +21,15 @@ func (f *Filter) VRL() (string, error) {
 		condList := []string{}
 		for _, cond := range test.DropConditions {
 			if cond.Matches != "" {
-				condList = append(condList, fmt.Sprintf(`match(%s, r'%s')`, cond.Field, cond.Matches))
+				condList = append(condList, fmt.Sprintf(`match(to_string(%s) ?? "", r'%s')`, cond.Field, cond.Matches))
 			} else {
-				condList = append(condList, fmt.Sprintf(`!match(%s, r'%s')`, cond.Field, cond.NotMatches))
+				condList = append(condList, fmt.Sprintf(`!match(to_string(%s) ?? "", r'%s')`, cond.Field, cond.NotMatches))
 			}
 		}
 		// Concatenate the conditions with ANDs and add Vector's error coalescing.
 		// If any errors arise from the match such as, `cond.Field` not being a string or a field
 		// is not present in the record, then it will automatically evaluate to false for the condition and specific test.
-		vrlCondition := "(" + strings.Join(condList, " && ") + " ?? false)"
+		vrlCondition := "(" + strings.Join(condList, " && ") + ")"
 		vrlTests = append(vrlTests, vrlCondition)
 	}
 
--- a/internal/generator/vector/filter/drop/filter_test.go
+++ b/internal/generator/vector/filter/drop/filter_test.go
@@ -46,7 +46,7 @@ var _ = Describe("drop filter", func() {
 				},
 			}
 			Expect(NewFilter(spec).VRL()).To(matchers.EqualTrimLines(`
-!((match(.kubernetes.namespace_name, r'busybox') && !match(.level, r'd.+') ?? false) || (match(.log_type, r'application') ?? false) || (match(.kubernetes.container_name, r'error|warning') && !match(.kubernetes.labels.test, r'foo') ?? false))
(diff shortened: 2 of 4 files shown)


internal/generator/vector/filter/drop/filter.go
```diff
@@ -21,15 +21,15 @@ func (f *Filter) VRL() (string, error) {
...
+				condList = append(condList, fmt.Sprintf(`match(to_string!(%s), r'%s')`, cond.Field, cond.Matches))
 			} else {
-				condList = append(condList, fmt.Sprintf(`!match(%s, r'%s')`, cond.Field, cond.NotMatches))
+				condList = append(condList, fmt.Sprintf(`!match(to_string!(%s), r'%s')`, cond.Field, cond.NotMatches))
 			}
 		}
 		// Concatenate the conditions with ANDs and add Vector's error coalescing.
 		// If any errors arise from the match such as, `cond.Field` not being a string or a field
 		// is not present in the record, then it will automatically evaluate to false for the condition and specific test.
-		vrlCondition := "(" + strings.Join(condList, " && ") + " ?? false)"
```
Clee2691, line 32: As the comment above states, `?? false` was added so that if the field could not be matched with the regex, it would evaluate to `false` automatically. With your change vector will just abort outright. Is that the expected behavior?

internal/generator/vector/filter/drop/filter.go
```diff
@@ -21,15 +21,15 @@ func (f *Filter) VRL() (string, error) {
...
+		vrlCondition := "(" + strings.Join(condList, " && ") + ")"
```
jcantrill, line 32: I assume we still need this since it is possible fields do not exist.  cc @Clee2691

internal/generator/vector/filter/drop/filter.go
```diff
@@ -21,15 +21,15 @@ func (f *Filter) VRL() (string, error) {
 		condList := []string{}
 		for _, cond := range test.DropConditions {
 			if cond.Matches != "" {
-				condList = append(condList, fmt.Sprintf(`match(%s, r'%s')`, cond.Field, cond.Matches))
+				condList = append(condList, fmt.Sprintf(`match(to_string!(%s), r'%s')`, cond.Field, cond.Matches))
```
jcantrill, line 24: Do we have a test to support this? Maybe one for the 403 in the bug

internal/generator/vector/filter/drop/filter.go
vparfonov: @Clee2691 
If field not exist to_string will return empty string, so regex will be matched with empty string
cc @jcantrill

internal/generator/vector/filter/drop/filter.go
vparfonov: https://github.com/openshift/cluster-logging-operator/pull/2749/files#diff-6282421d5e27209320b53a3de7f8c3c2885b46e9ae27b608b2fb52786decfdecR85

internal/generator/vector/filter/drop/filter.go
Clee2691: It's not just a field doesn't exist, it is also if
> value is not an integer, float, boolean, string, timestamp, or null.
If the field is not one of those, the program will abort.

[Example](https://playground.vrl.dev/?state=eyJwcm9ncmFtIjoiLmJhciA9IHRvX3N0cmluZyEoLnRlc3QpIiwiZXZlbnQiOnsibWVzc2FnZSI6IkhlbGxvIFZSTCIsImZvbyI6ImRlbGV0ZSBtZSIsImh0dHBfc3RhdHVzIjoiMjAwIiwidGVzdCI6eyJiYXIiOiJmb28ifX0sImlzX2pzb25sIjpmYWxzZSwiZXJyb3IiOm51bGx9):
JSON message:
```
{
    "message": "Hello VRL",
    "foo": "delete me",
    "http_status": "200",
    "test":{
        "bar":"foo"
    }
}
```
VRL:
```
.bar = to_string!(.test)
```

Error:
```
1 │ .bar = to_string!(.test)
  │        ^^^^^^^^^^^^^^^^^ unable to coerce { bar: string } into string
```

vparfonov: I think `match` will have the same behavior

PR #2753, opened 2024-08-13
LOG-5856: Remove internal package reference from API

xperimental:
### Description

This PR removes the internal package reference from the `api/observability/v1` package. The affected code is moved to the "internal API package". It also removes the dependency to ginkgo as there are no existing tests yet.

/cc @cahartma
/assign @jcantrill 

### Links

- JIRA: [LOG-5856](https://issues.redhat.com//browse/LOG-5856)


jcantrill (commented):
/lgtm

PR #2757, opened 2024-08-14
LOG-5906: Fix metrics ServiceMonitor label selector

xperimental:
### Description

I noticed that I did not get any collector metrics anymore in the dashboard after a recent update. This PR should fix the selector used in the ServiceMonitor for picking up the collector metrics.

I have added a new label describing the "service type" of the created Service resources. It looked to me as if we have two types: metrics and input services.

/cc @cahartma
/assign @jcantrill

### Links

- JIRA: LOG-5906


jcantrill (commented):
/approve
/lgtm

PR #2758, opened 2024-08-14
LOG-5969: Update otlp attributes mapping and feature doc

cahartma:
### Description
Quick update to otel mappings (and doc) based on feedback and further investigation

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-5969

### DIFF
--- a/docs/features/logforwarding/outputs/otlp-forwarding.adoc
+++ b/docs/features/logforwarding/outputs/otlp-forwarding.adoc
@@ -75,7 +75,7 @@ The Semantic Conventions in OpenTelemetry define a *Resource* as an immutable re
 
 For example, a process producing telemetry that is running in a container has a container_name, a cluster_id, a pod_name, a namespace, and possibly a deployment or app_name.
 
-All of these *Attributes* are included in the *Resource* object.  This grouping and reducing of common attributes is a powerful tool when sending logs as telemetry data.
+All of these *Attributes* are included in the *Resource* object.  This grouping and reducing of common attributes is a key aspect of sending logs via OpenTelemetry.
 
 .OTEL References
 * https://opentelemetry.io/docs/specs/semconv/[Semantic Conventions]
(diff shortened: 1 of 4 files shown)


docs/features/logforwarding/outputs/otlp-forwarding.adoc
```diff
@@ -93,35 +93,53 @@ https://github.com/open-telemetry/opentelemetry-proto/blob/v1.3.1/examples/logs.
...
+kubernetes.namespace_name,        k8s.namespace.name
+kubernetes.pod_name,              k8s.pod.name
+kubernetes.container_id,          k8s.container.id
+kubernetes.pod_id,                k8s.pod.uid
+kubernetes.namespace_labels.*,    k8s.namespace.label.*
+kubernetes.labels.*,              k8s.pod.label.*
+openshift.labels.*,               openshift.label.*
+log_type,                         openshift.log.type
+log_source,                       openshift.log.source
+hostname,                         url.domain
```
jcantrill, line 113: hostname is listed twice.  Do these need to be separated by log_type?

docs/features/logforwarding/outputs/otlp-forwarding.adoc
```diff
@@ -93,35 +93,53 @@ https://github.com/open-telemetry/opentelemetry-proto/blob/v1.3.1/examples/logs.
...
 .ViaQ Data Model
 * https://github.com/openshift/cluster-logging-operator/blob/master/docs/reference/datamodels/viaq/v1.adoc
+
+=== References
+* https://opentelemetry.io/docs/specs/otlp/
+* https://opentelemetry.io/docs/specs/semconv/resource/container/
+* https://opentelemetry.io/docs/specs/semconv/attributes-registry/k8s/
+* https://opentelemetry.io/docs/specs/semconv/attributes-registry/process/
+* https://opentelemetry.io/docs/specs/semconv/attributes-registry/http/
+* https://opentelemetry.io/docs/specs/semconv/attributes-registry/url/
```
jcantrill, line 143: why URL?  What does it correspond to?

jcantrill (approved):
/hold

few comments otherwise lgtm

docs/features/logforwarding/outputs/otlp-forwarding.adoc
cahartma: Based on further understanding of the semconv, these are all "children" and siblings of the "registry" semconv.   The idea is to use the best match from any of the attributes.   url matched several of them exactly.  

> Attributes defined in the registry can be used in different semantic conventions. Attributes should be included in this registry before they are used in semantic conventions. Semantic conventions may override all the properties of an attribute except for the id and type in case it’s required for a particular context. In addition, semantic conventions specify the requirement level of an attribute in the corresponding context.

docs/features/logforwarding/outputs/otlp-forwarding.adoc
cahartma: They are indeed separated by log_type.    k8s.node.name is set for auditd logs as a resource attribute (grouped and reduced).    url.domain is set for API logs and is a logrecord attribute, so not grouped/reduced by.

cahartma: At some point, I can create a mapping for the specific fields within each log_type, and match it up with the ViaQ structs we have currently.

PR #2761, opened 2024-08-15
Updated LokiStack CR to use configmap openshift-service-ca.crt for CA certificate

dhgautam99:
As per current config, it is preposed to use secret "logcollector-token" secret to retrieve service-ca.crt but configmap named openshift-service-ca.crt which is part of RHOCP aldready holds the same service CA certificate. It is better to use this as it is provided by platform by default and it also holds updated service CA, if the service-ca operator rotates the certificate.

### Description

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: NA
- Enhancement proposal:

### DIFF
--- a/docs/administration/upgrade/to_v6.0.adoc
+++ b/docs/administration/upgrade/to_v6.0.adoc
@@ -303,7 +303,7 @@ spec:
     tls:
       ca:
         key: service-ca.crt
-        secretName: logcollector-token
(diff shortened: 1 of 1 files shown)


jcantrill (commented):
/ok-to-test
/approve
/lgtm

PR #2763, opened 2024-08-15
LOG-5978: Remove default values from API definition

xperimental:
### Description

This PR removes all `kubebuilder:default` markers except for the `managementState` which is present in all ClusterLogForwarder instances in an effort to fix creation of CLF resources using the OpenShift Console form view.

/cc @cahartma
/assign @jcantrill

### Links

- JIRA: [LOG-5978](https://issues.redhat.com//browse/LOG-5978)

### DIFF
--- a/api/observability/v1/output_types.go
+++ b/api/observability/v1/output_types.go
@@ -200,7 +200,6 @@ type BaseOutputTuningSpec struct {
 // DeliveryMode sets the delivery mode for log forwarding.
 //
 // +kubebuilder:validation:Enum:=atLeastOnce;atMostOnce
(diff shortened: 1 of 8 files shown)


api/observability/v1/output_types.go
```diff
@@ -1030,7 +1021,6 @@ type Syslog struct {
 	//     Emergency Alert Critical Error Warning Notice Informational Debug
 	//
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:default:=informational
```
jcantrill, line 1033: If we are removing this optional default then we need to pair it with an intializer (e.g. internal/api/initialize) or something else to provide a default in the generator

PR #2764, opened 2024-08-15
Update example resources distributed with CSV

xperimental:
### Description

This PR updates the example files distributed with the CSV. They are used as a starting point when creating a new resource in the OpenShift Console.

I have opted for a "application and infrastructure logs to local LokiStack" configuration for the ClusterLogForwarder. I think ideally we synchronize this with what we list as the "recommended configuration" in the documentation. I've already tried to use the naming conventions we had for the resources in the 5.y docs.

This PR also enables a new feature in the OpenShift Console: It now shows a "ClusterLogForwarder required" warning in the operator details when no resource has been created yet, prompting the user to create one. This button unfortunately does not use the same example file, but we could keep it synchronized manually. Currently I am using a "blank" CLF for that function.

/cc @cahartma
/assign @jcantrill

### DIFF
--- a/bundle/manifests/cluster-logging.clusterserviceversion.yaml
+++ b/bundle/manifests/cluster-logging.clusterserviceversion.yaml
@@ -34,28 +34,10 @@ metadata:
           "apiVersion": "observability.openshift.io/v1",
           "kind": "ClusterLogForwarder",
           "metadata": {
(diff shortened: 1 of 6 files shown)


jcantrill (commented):
/approve
/lgtm

PR #2769, opened 2024-08-16
LOG-4745: Default collector resources when undefined

jcantrill:
### Description
This PR:
* defaults the collector resources (e.g. memory,cpu) when the collector or collector resources are undefined
* defaults values as suggested by vector [upstream](https://vector.dev/docs/reference/configuration/sources/kubernetes_logs/#resource-limits)

### Links
https://issues.redhat.com/browse/LOG-4745

cc @cahartma @xperimental @Clee2691

### DIFF
--- a/internal/api/initialize/init_resources_test.go
+++ b/internal/api/initialize/init_resources_test.go
@@ -1,0 +1,60 @@
+package initialize
+
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	obs "github.com/openshift/cluster-logging-operator/api/observability/v1"
+	"github.com/openshift/cluster-logging-operator/internal/utils"
+	v1 "k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


internal/api/initialize/init_resources_test.go
```diff
@@ -0,0 +1,60 @@
...
+		forwarder = obs.ClusterLogForwarder{
+			Spec: obs.ClusterLogForwarderSpec{},
+		}
+	})
+
+	It("should apply the default resources when collector is not defined", func() {
+		forwarder := Resources(forwarder, utils.NoOptions)
+		Expect(forwarder.Spec.Collector.Resources).To(Equal(expResources))
+	})
+	It("should apply the default resources when resources are not defined defined", func() {
```
Clee2691, line 37: nit: You wrote defined twice
```suggestion
	It("should apply the default resources when resources are not defined", func() {
```

PR #2771, opened 2024-08-16
LOG-5909: Some configurations in otlp.tuning are not added into vector.toml

Clee2691:
### Description
This PR adds the appropriate tuning sections for the `OTLP` output. In addition, it adds unit tests for the tuning section for all output types.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-5909

### DIFF
--- a/internal/generator/vector/output/azuremonitor/azm_tuning.toml
+++ b/internal/generator/vector/output/azuremonitor/azm_tuning.toml
@@ -1,0 +1,22 @@
+
+[sinks.output_azure_monitor_logs]
+type = "azure_monitor_logs"
+inputs = ["pipelineName"]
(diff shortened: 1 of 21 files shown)


test/helpers/outputs/fakeadapter/fake_adapter.go
```diff
@@ -0,0 +1,25 @@
+package fakeadapter
```
jcantrill, line 1: Consider following the kubernetes pattern for client where they use a package like:

test/helpers/outputs/adapter/fake
internal/generator/vector/helpers/adapter/fake
internal/generator/vector/helpers/adapters/outputs/fake

Doing so would also mean you could drop "Fake" from method and type names

jcantrill (commented):
One comment otherwise looks good to me

PR #2762, opened 2024-08-15
LOG-5977: Update the API to use kube standards

jcantrill:
### Description
This PR:
* updates the API to use kube standards
* makes ES version required and not defaulted

### Links
https://issues.redhat.com/browse/LOG-5977

cc @cahartma @xperimental @Clee2691 @libander @anpingli

PR #2772, opened 2024-08-16
Upstream v6 doc updates

cahartma:


docs/administration/upgrade/to_v6.0.adoc
```diff
@@ -1,12 +1,33 @@
...
+Logging version 6.0 is a major change from earlier releases and is the realization of several longstanding goals.
 
-* Distinct operators to support logging components (e.g. collectors, storage, visualization)
-* Remove support of managed log storage and visualization based upon the Elastic products (i.e. Elasticsearch, Kibana)
-* Remove support of Fluentd log collector implementation
-* Remove support of the ClusterLogging.logging.openshift.io and ClusterLogForwarder.logging.openshift.io
+The main highlights of these changes are:
+
+* There is now distinct operators to support separate logging components (e.g. collectors, storage, visualization)
+* Cluster Logging Operator no longer implements managed log storage and visualization based upon the Elastic products (i.e. Elasticsearch, Kibana)
```
jcantrill, line 9: The original statement was about "Cluster Logging" and this change is more specific to the actual operator.  If we leave as "CLO" then I think it would better be described like:

The ClusterLoggingOperator no longer manages log storage or visualization of any kind including LokiStack resource or Elastic products (i.e. Elasticsearch, Kibana)

docs/administration/upgrade/to_v6.0.adoc
```diff
@@ -1,12 +1,33 @@
...
+* CLO has removed support of Fluentd log collector implementation
+* API changes remove support of CRD from API Group `logging.openshift.io/v1` of `kind: ClusterLogging` and `kind: ClusterLogForwarder`
```
jcantrill, line 11: I don't think most admins understand CRD.  This, IMO, would be better worded closer to the original statement:

Introduce the ClusterLogForwarder.observability.openshift.io API to replace both ClusterLogging.logging.openshift.io and ClusterLogForwarder.logging.openshift.io APIs

Though even :up: is slightly misleading since the CL resource represented something completely separate an the new API isn't really replacing it; it sort of replaces a very specific use-case

docs/administration/upgrade/to_v6.0.adoc
```diff
@@ -237,30 +263,39 @@ metadata:
...
 ----
-NOTE: In this example, application logs are written to the 'application-write' alias/index instead of 'app-write'
+. `index` can be a combination of dynamic and static values. Dynamic values are enclosed in curly brackets `{}`
+and MUST end with a "quoted" static fallback value separated with `||`.
+
+More details use: `oc explain clf.spec.outputs.elasticsearch.index`
+
+
+NOTE: In this example, application logs are written to the 'application-write' and 'infrastructure-write' index.
+Previous versions without index, would have instead wrote to 'app-write', 'infra-write'
```
jcantrill, line 297: ... written to..

docs/administration/upgrade/to_v6.0.adoc
```diff
@@ -297,17 +332,18 @@ spec:
...
-          key: token
-          secretName: logcollector-token
+          secret:
+            key: token
+            name: logcollector-token
     tls:
       ca:
         key: service-ca.crt
-        configMapName: openshift-service-ca.crt
+        secretName: logcollector-token
```
jcantrill, line 341: Replace with the change here https://github.com/openshift/cluster-logging-operator/pull/2761

docs/administration/upgrade/to_v6.0.adoc
```diff
@@ -323,11 +359,11 @@ apiVersion: "logging.openshift.io/v1"
 kind: ClusterLogForwarder
 spec:
   pipelines:
-   - name: application-logs
+   - name: app-logs
```
jcantrill, line 362: nit but there are inconsistencies in indention.  We should on of:

* align arrays with the container element
* indent all arrays same space from the container element

docs/administration/upgrade/to_v6.0.adoc
```diff
@@ -337,71 +373,212 @@ apiVersion: "observability.openshift.io/v1"
...
+          matches: debug
+    - name: prune-fields
+      type: prune
+      prune:
+        in:
+        - .kubernetes.labels.foobar
+        notIn:
+        - .message
+    - name: audit-logs
+      type: kubeAPIAudit
```
jcantrill, line 412: This may change with our standards PR change?

docs/administration/upgrade/to_v6.0.adoc
```diff
@@ -337,71 +373,212 @@ apiVersion: "observability.openshift.io/v1"
...
+        - 409
 ----
 ==== Validation & Status
 Most validations are enforced when a resource is created or updated which provides immediate feedback.  This is
 a departure from previous releases where all validation occurred post creation requiring inspection of the resource status location.  Some validation still occurs post resource creation for cases where is not possible to do so at creation or update time.
 
-Instances of the **ClusterLogForwarder.observability.openshift.io** must satisfy the following conditions before the operator will deploy the log collector: Authorized, Valid, Ready.  An example of these conditions is:
+Instances of the **ClusterLogForwarder.observability.openshift.io** must satisfy the following before the operator will deploy the log collector:
+
+*Conditions:* `Authorized, Valid, Ready` = "True"
```
jcantrill, line 424: This should probably state something about "status" field equals True

docs/administration/upgrade/to_v6.0.adoc
```diff
@@ -337,71 +373,212 @@ apiVersion: "observability.openshift.io/v1"
...
+      status: "True"
+      type: observability.openshift.io/Authorized
+    - message: ""
+      reason: ValidationSuccess
+      status: "True"
+      type: observability.openshift.io/Valid
+    - message: ""
+      status: "True"
+      type: observability.openshift.io/Ready
+  filtersStatus:
```
jcantrill, line 446: These become xxxConditions

docs/administration/upgrade/to_v6.0.adoc
```diff
@@ -337,71 +373,212 @@ apiVersion: "observability.openshift.io/v1"
...
-===== Red Hat Managed LokiStack
-===== Forwarding to CloudWatch using Short-Lived Token
-===== Forwarding to Elasticsearch using Custom Indices
-===== Forwarding with Input Receivers
+=== ClusterLogForwarder Only
+====
+.Forwarding to CloudWatch using Short-Lived Token (SA Token)
+[source, yaml]
+----
+apiVersion: observability.openshift.io/v1
```
jcantrill, line 493: This is a place for us to promote reuse and possible add tests to ensure they are valid.  Recommend moving this to separate files and using an include (i.e.) something like:
```
 .v6.0 Forwarding to Red Hat Managed LokiStack
[source,yaml]
 ----
include::../../../docs/reference/samples/forward-to-rh-managed-es.yaml[]
 ----
```

docs/administration/upgrade/to_v6.0.adoc
```diff
@@ -337,71 +373,212 @@ apiVersion: "observability.openshift.io/v1"
...
+          type: iamRole
+          iamRole:
+            roleARN:
+              secretName: role-for-sts
+              key: credentials
+            token:
+              from: serviceAccount
+...
+----
+.Alternative CW Auth spec using role and static token
```
jcantrill, line 516: This could be a complete sample we pull in or left as a snippet

docs/administration/upgrade/to_v6.0.adoc
```diff
@@ -337,71 +373,212 @@ apiVersion: "observability.openshift.io/v1"
...
+                name: cw-token
+...
+----
+.Alternative CW Auth spec using static key and secret
+[source, yaml]
+----
+        authentication:
+          type: awsAccessKey
+          awsAccessKey:
+            keyID:
```
jcantrill, line 538: keyId

jcantrill (changes requested):

docs/administration/upgrade/to_v6.0.adoc
cahartma: Yes,  but needs to merge

docs/administration/upgrade/to_v6.0.adoc
cahartma: Do you think we need additional details or a note otherwise, in this example?

jcantrill: I think we just need the correct fieldName which will be xxxxConditions once https://github.com/openshift/cluster-logging-operator/pull/2762 merges

docs/reference/samples/observability.filters-pipelines-unchanged.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: "observability.openshift.io/v1"
+kind: ClusterLogForwarder
+spec:
+  filters:
+    - name: my-multiline
+      type: detectMultilineException
+    - name: my-parse
+      type: parse
+    - name: my-labels
+      type: openShiftLabels
```
jcantrill, line 10: openshiftLabels

docs/reference/samples/observability.filters-pipelines-unchanged.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: "observability.openshift.io/v1"
+kind: ClusterLogForwarder
```
jcantrill, line 2: Are these intended to be working examples?  If so then they should at minimum have a name.  Namespace is implied with  the create command

cahartma: Understood.  I keep flipping back and forth between samples and snippets.   Will get this straightened out today.

docs/reference/samples/observability.output-loki-default.yaml
```diff
@@ -0,0 +1,30 @@
...
+  outputs:
+    - name: default-lokistack
+      type: lokiStack
+      lokiStack:
+        target:
+          name: lokistack-dev
+          namespace: openshift-logging
+        authentication:
+          token:
+            from: secret
```
anpingli, line 16: It is better to use serviceAccount. here?

cahartma: Happy to use yours if you have a suggestion for a good example to use here.

cahartma: @anpingli what do you think is the most popular/useful configuration here?

xperimental: I second @anpingli 's comment and would suggest changing it to the `serviceToken`:

```
      authentication:
        token:
          from: serviceAccount
```

docs/reference/samples/observability.output-loki-default.yaml
```diff
@@ -0,0 +1,32 @@
...
+  namespace: openshift-logging
+spec:
+  serviceAccount:
+    name: logcollector
+  outputs:
+    - name: default-lokistack
+      type: lokiStack
+      lokiStack:
+        target:
+          name: lokistack-dev
```
xperimental, line 14: If this is meant as "user documentation" instead of "developer example", we have, so far used `logging-loki` as an example name there:

```suggestion
          name: logging-loki
```

docs/administration/unmanaged.adoc
```diff
@@ -18,14 +18,14 @@ NOTE: Logging components in an unmanaged state are not supported unless a suppor
 NOTE: An unmanaged deployment will not receive updates until the `ClusterLogForwarder` custom resource is placed back into a managed state.
 ```
   spec:
-    managementState: "Managed"
+    managementState: "managed"
```
xperimental, line 21: I prefer the unquoted style, when it is not necessary to quote the values:

```suggestion
    managementState: managed
```

docs/reference/samples/observability.filters-pipelines-unchanged.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: "observability.openshift.io/v1"
```
xperimental, line 1: Suggest to change to unquoted:

```suggestion
apiVersion: observability.openshift.io/v1
```

docs/reference/samples/observability.output-es-managed.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: "observability.openshift.io/v1"
```
xperimental, line 1: Suggest to change to unquoted:

```suggestion
apiVersion: observability.openshift.io/v1
```

cahartma:
This file will change from its current revision due to discussion last week, and the resulting API changes from:   https://github.com/openshift/cluster-logging-operator/pull/2780

docs/reference/samples/observability.filters-pipelines-unchanged.yaml
cahartma: I think we are going to need both types of examples/samples here.   I've included two input spec examples, but not sure its necessary to show outputs for those.    I still think these are helpful to have somewhere, if we want to only provide full examples.   There are complete examples where I think it makes sense.  Also, I have many more examples.

PR #2773, opened 2024-08-19
LOG-5907: add trace log level detection

vparfonov:
### Description

/cc @Clee2691 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5907
- Enhancement proposal:


internal/generator/vector/filter/openshift/viaq/normalize.go
```diff
@@ -24,6 +24,8 @@ if !exists(.level) {
     .level = "info"
   } else if match!(.message, r'^D[0-9]+|level=debug|Value:debug|"level":"debug"') {
     .level = "debug"
+  } else if match!(.message, r'^D[0-9]+|level=trace|Value:trace|"level":"trace"') {
```
Clee2691, line 27: I think the first regex should start with `T` for trace,
```suggestion
  } else if match!(.message, r'^T[0-9]+|level=trace|Value:trace|"level":"trace"') {
```

vparfonov: nice catch

PR #2775, opened 2024-08-19
LOG-5980: fix setting token when 'otlp.authentication.token.from: serviceAccount'

vparfonov:
### Description

/cc @cahartma @Clee2691 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5980
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/output/otlp/otlp_test.go
+++ b/internal/generator/vector/output/otlp/otlp_test.go
@@ -2,6 +2,8 @@ package otlp
 
 import (
 	"fmt"
+	"github.com/openshift/cluster-logging-operator/internal/constants"
+	corev1 "k8s.io/api/core/v1"
 	"time"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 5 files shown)


internal/generator/vector/output/otlp/otlp_test.go
```diff
@@ -46,10 +65,13 @@ var _ = Describe("Generate vector config", func() {
 		if visit != nil {
 			visit(&outputSpec)
 		}
+		var conf []framework.Element
 		if tune {
 			adapter = *fake.NewOutput(outputSpec, secret, framework.NoOptions)
+			conf = New(helpers.MakeOutputID(outputSpec.Name), outputSpec, []string{"pipeline_my_pipeline_viaq_0"}, secret, adapter, op)
+		} else {
+			conf = New(helpers.MakeOutputID(outputSpec.Name), outputSpec, []string{"pipeline_my_pipeline_viaq_0"}, secret, nil, op)
```
Clee2691, line 73: This `else` isn't needed because `adapter` is `nil` unless it's set by `tune == true`.

vparfonov: if `tune` was set `true` one time for other tests after`adapter` will initialize

PR #2776, opened 2024-08-19
Add new Make targets for easier deployment using operator-sdk

alanconway:
/hold

PR created for code review, discard when done.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,13 +93,16 @@ ci-check: check
 # Note: Go has built-in build caching, so always run `go build`.
 # It will do a better job than using source dependencies to decide if we need to build.
 
-bin/functional-benchmarker: force
+.PHONY: bin/functional-benchmarker
+bin/functional-benchmarker:
 	go build $(BUILD_OPTS) -o $@ ./internal/cmd/functional-benchmarker
 
-bin/forwarder-generator: force
+.PHONY: bin/forwarder-generator
+bin/forwarder-generator:
 	go build $(BUILD_OPTS) -o $@ ./internal/cmd/forwarder-generator
 
-bin/cluster-logging-operator: force
+.PHONY: bin/cluster-logging-operator
+bin/cluster-logging-operator:
 	go build $(BUILD_OPTS) -o $@ ./cmd
 
 .PHONY: openshift-client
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -93,13 +93,13 @@ ci-check: check
 # Note: Go has built-in build caching, so always run `go build`.
 # It will do a better job than using source dependencies to decide if we need to build.
 
-bin/functional-benchmarker: force
```
alanconway, line 96: If you remove `force` the binary will never be rebuilt.
If file exists and has no dependencies - make will ignore it.
The purpse of `force` is to force make to run the `go` command and let go figure out the dependencies - if you want to avoid that at the make level, you need to add all the dependencies for the binary to the rule e.g. `$(shell find somewhere/*.go)`

alanconway: An alternative to `force` is `.PHONY`. I use `force` is by force of habit. .PHONY is a GNU extension, the `force` trick works for any make variant, e.g. Solaris or HP-UX . Yes, I am that old.

alanconway (commented):
I like it. Simpler and cleaner than the existing bundle build. I think the OVERLAY stuff was over-engineering on my part :) Please rip out the existing stuff when you merge this to keep it clean & simple :)

Makefile
```diff
@@ -402,6 +402,41 @@ cluster-logging-operator-install:
 cluster-logging-operator-uninstall:
 	olm_deploy/scripts/operator-uninstall.sh
 
-.PHONY: gen-dockerfiles
-gen-dockerfiles:
-	./hack/generate-dockerfile-from-midstream > Dockerfile
+REPOSITORY_BASE ?= $(error REPOSITORY_BASE needs to be set for the olm- targets to work)
```
alanconway, line 405: Can we use REGISTRY_BASE please? That's what loki does and I've followed that convention elsewhere. Would be nice to set just one env.var :)

Makefile
xperimental: I got the purpose of `force` and wanted to add `.PHONY` instead (but then forgot about that). I think we can rely on the GNU extension to be there, as we're using this for all the other targets.

Makefile
alanconway: I retract this comment :)

xperimental: Too late. I already made it possible to use both variants (with the preference being `REGISTRY_BASE`). :grin:

xperimental:
/retitle Add new Make targets for easier deployment using operator-sdk

PR #2779, opened 2024-08-22
LOG-5993: Sort names of secrets / configmaps so that order is stable

xperimental:
### Description

This sorts the slices returned by `Names()` for both secrets and configmaps so that the ordering is stable.

/cc @cahartma
/assign @jcantrill

### Links

- JIRA: [LOG-5993](https://issues.redhat.com//browse/LOG-5993)


jcantrill (commented):
/lgtm

PR #2780, opened 2024-08-22
WIP: Partially Revert "LOG-5977: Update the API to use kube standards"

alanconway:
This partially reverts commit 5a4dd8e0dc18ee96b52f3605ab8b5a98446bc900.

Some of the public API changes in that commit made us less compliant with the kube api conventions:
 https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#constants

Removed those changes that make use less compliant.

Some remaining enum values start with a lower case letter, but most are covered by this paragraph of
the api-conventions:

> When a proper name or the name of a command-line executable is used as a constant the proper name
> should be represented in consistent casing - examples: systemd, iptables, IPVS, cgroupfs, Docker
> (as a generic concept), docker (as the command-line executable). If a proper name is used which
> has mixed capitalization like eBPF that should be preserved in a longer constant such as
> eBPFDelegation.

In other words, enum values that are identifiers defined externally to the CLO should follow the
case conventions of the source of the identifier, for example:

- HTTP methods: GET;HEAD;POST;PUT;DELETE;OPTIONS;TRACE;PATCH
- Viaq field values: audit;application;infrastructure;receiver
- Vector compression identifiers: none;snappy;zstd;lz4

Please review and we can discuss what further changes are needed.


alanconway:
/close
New PR based on todays meeting will be available later today.

PR #2784, opened 2024-08-27
Upstream v6 changes doc and config samples

cahartma:
### Description
Updated docs snippets and made all examples into complete specs.   I still need to figure out what Libby wants to do with the samples.adoc 

/cc @Clee2691 @vparfonov
/assign @jcantrill


docs/administration/upgrade/v6.0_changes.adoc
```diff
@@ -566,78 +480,132 @@ spec:
...
+      inputRefs:
         - application
-      name: to-default
-      parse: json
+        - infrastructure
+      filterRefs:
+        - my-parse
       outputRefs:
-        - default
+        - es-managed
```
jcantrill: this does not match the output name

docs/administration/upgrade/v6.0_changes.adoc
```diff
@@ -566,97 +480,162 @@ spec:
...
-----
-.v6 ClusterLogForwarder
+  - name: my-pipeline
+    inputRefs:
+    - application
+    - infrastructure
+    filterRefs:
+    - my-parse
+    outputRefs:
+    - es-managed
```
jcantrill, line 562: name ref is incorrect

docs/administration/upgrade/v6.0_changes.adoc
```diff
@@ -566,97 +480,162 @@ spec:
...
-        ca:
-          key: ca-bundle.crt
-          secretName: collector
-        certificate:
-          key: tls.crt
-          secretName: collector
-        key:
-          key: tls.key
-          secretName: collector
+  - name: es-managed-app
```
jcantrill, line 587: I just thought about this, we could be clever and allow a single output def that relies upon an openshift.label to craft the index.  We create 2 pipelines to the same output and use the openshiftLabel filter on each pipeline to set "app" or "infra" and then use that for the index name

docs/reference/samples/observability.filters-pipelines-unchanged.yaml
```diff
@@ -28,4 +28,4 @@ spec:
     - name: my-output
```
jcantrill, line 28: We should consider naming any yaml which is not complete in such a way to indicate its not complete

cahartma: Include "Snippet" in the title, or what I prefer is the ellipses top and bottom?

docs/administration/upgrade/v6.0_changes.adoc
alanconway: That is clever :)

cahartma: Perfect, and I was also thinking, if they have created custom indexes using the 'structuredType' options in 5.9, we may want to show how to achieve that as well.

cahartma: Updated, please review.

docs/reference/samples/observability.filters-pipelines-unchanged.yaml
cahartma: Or do you mean the actual filename?   These should all be complete at this point.

docs/reference/samples/observability.inputs-receivers.yaml
```diff
@@ -44,3 +44,24 @@ spec:
...
+  outputs:
+    - name: my-http-output
+      type: http
+      http:
+        url: http://my-log-output:80
+    - name: my-sys-out
+      type: syslog
+      syslog:
+        rfc: RFC5424
+        url: tcp://127.0.0.1:24224
```
anpingli, line 67: how about  tcp://my-syslogservice.example.com:514

jcantrill (commented):
/lgtm

docs/reference/samples/observability.inputs-receivers.yaml
cahartma: Will update for future if you think it's helpful.  As long as the current example passes validation, we are good for now.

PR #2794, opened 2024-09-05
LOG-6044: validation failure when adding a namespace which contains 'kube'

Clee2691:
### Description
This PR modifies the regex that determines infrastructure namespaces so that namespaces containing, but not starting with, `kube` or `openshift` are treated as application namespaces. Only the namespace `default` is treated as infrastructure and any other namespace containing `default` is not.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-6044


jcantrill (commented):
/approve

PR #2798, opened 2024-09-10
LOG-6069: Enable http/https proxy support for HTTP output.

alanconway:
-------

### Description

Add a 'proxy' field to HTTP output configuration to enable use of a proxy.e! -->

/cc @jcantrill
/assign @jcantrill

### Links
- JIRA: [LOG-6069](https://issues.redhat.com//browse/LOG-6069)

### DIFF
--- a/api/observability/v1/output_types.go
+++ b/api/observability/v1/output_types.go
@@ -617,12 +617,18 @@ type HTTP struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Timeout",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:number"}
 	Timeout int `json:"timeout,omitempty"`
 
-	// Method specifies the Http method to be used for sending logs. If not set, 'POST' is used.
+	// Method specifies the HTTP method to be used for sending logs. If not set, 'POST' is used.
 	//
 	// +kubebuilder:validation:Enum:=GET;HEAD;POST;PUT;DELETE;OPTIONS;TRACE;PATCH
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="HTTP Method"
 	Method string `json:"method,omitempty"`
+
+	// ProxyURL URL of a HTTP or HTTPS proxy to be used instead of direct connection.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:XValidation:rule="self == '' ||  isURL(self)", message="invalid URL"
+	ProxyURL string `json:"proxyURL,omitempty"`
 }
 
 type KafkaTuningSpec struct {
--- a/internal/generator/vector/output/http/http.go
+++ b/internal/generator/vector/output/http/http.go
@@ -18,6 +18,7 @@ type Http struct {
 	Inputs      string
 	URI         string
 	Method      string
+	Proxy       string
 	common.RootMixin
 }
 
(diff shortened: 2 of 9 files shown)


api/observability/v1/output_types.go
```diff
@@ -593,12 +593,17 @@ type HTTP struct {
...
 	//
 	// +kubebuilder:validation:Enum:=GET;HEAD;POST;PUT;DELETE;OPTIONS;TRACE;PATCH
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="HTTP Method"
 	Method string `json:"method,omitempty"`
+
+	// ProxyURL if specified is used to proxy HTTP and HTTPS traffic.
+	//
+	// +kubebuilder:validation:Optional
+	ProxyURL string `json:"proxyURL,omitempty"`
```
jcantrill, line 606: @alanconway What is our motivation for not configuring all three options?  I have some thoughts why you designed it this way but would like to understand if they are the same as mine

- Needs URL validation: https://github.com/openshift/cluster-logging-operator/pull/2798/files#diff-4749b2553f93d0fa25f19c74b7fad9dc691f10c0bf9c8f4b46ab0681c4a9560dR663

alanconway: - enabled: redundant - if proxy config exists, it's enabled if not it's disabled.
- http vs. https: if user has different proxies for each they can set up different outputs.

You could argue that configuring http/https separately is more convenient, and for the general web proxy case it probably is. However in the log forwarding case, each output will either be secure or insecure based on the TLS configuration, there's no case where we will send both to the same output. So this PR enable the proxy proxy config to handle anything, and lets the TLS config determine whether we send secure or insecure.

internal/generator/vector/output/http/http.go
```diff
@@ -32,6 +33,11 @@ type = "http"
 inputs = {{.Inputs}}
 uri = "{{.URI}}"
 method = "{{.Method}}"
+{{with .Proxy -}}
+proxy.enabled = true
+proxy.http = {{.}}
```
samanthajayasinghe, line 38: @alanconway is this needs to change as ?
```
proxy.http =  "{{.Proxy}}"
```

api/observability/v1/output_types.go
alanconway: added url valiation

internal/generator/vector/output/http/http.go
alanconway: fixed

api/observability/v1/output_types.go
```diff
@@ -593,12 +593,18 @@ type HTTP struct {
...
 
-	// Method specifies the Http method to be used for sending logs. If not set, 'POST' is used.
+	// Method specifies the HTTP method to be used for sending logs. If not set, 'POST' is used.
 	//
 	// +kubebuilder:validation:Enum:=GET;HEAD;POST;PUT;DELETE;OPTIONS;TRACE;PATCH
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="HTTP Method"
 	Method string `json:"method,omitempty"`
+
+	// ProxyURL if specified is used to proxy HTTP and HTTPS traffic.
```
jcantrill, line 603: @alanconway do we need to explicitly state that SOCKS proxy is not supported?

alanconway: I don't think so, but I made the wording more explict to say we're talking about a HTTP proxy, not  a general socket proxy:

	// ProxyURL URL of a HTTP or HTTPS proxy to be used instead of direct connection.

PR #2805, opened 2024-09-23
LOG-6128: Fix console plugin api version

jgbernalp:
### Description
From OCP v4.17+, the ConsolePlugin v1alpha1 was deprecated in favor of ConsolePlugin v1. This causes the logging plugin to fail during reconciliation as the new version of the spec needs for 4.17+. This PR adds support for ConsolePlugin v1 when the operator is running on clusters with v4.17+

This is targeted directly to the 5.9 branch as 6.0 (master) has removed the console plugin

/cc @jcantrill  
/assign @alanconway 

### Links
- JIRA: [LOG-6128](https://issues.redhat.com/browse/LOG-6128)

### DIFF
--- a/internal/controller/clusterlogging/clusterlogging_controller.go
+++ b/internal/controller/clusterlogging/clusterlogging_controller.go
@@ -2,9 +2,10 @@ package clusterlogging
 
 import (
 	"context"
-	"github.com/openshift/cluster-logging-operator/internal/migrations/clusterlogging"
 	"strings"
 	"time"
+
+	"github.com/openshift/cluster-logging-operator/internal/migrations/clusterlogging"
 
 	"github.com/openshift/cluster-logging-operator/internal/hostedcontrolplane"
 	"github.com/openshift/cluster-logging-operator/internal/logstore/lokistack"
@@ -16,8 +17,10 @@ import (
 
 	"github.com/openshift/cluster-logging-operator/internal/factory"
 	"github.com/openshift/cluster-logging-operator/internal/k8s/loader"
+	"github.com/openshift/cluster-logging-operator/internal/utils"
 	validationerrors "github.com/openshift/cluster-logging-operator/internal/validations/errors"
 
+	consolev1 "github.com/openshift/api/console/v1"
 	consolev1alpha1 "github.com/openshift/api/console/v1alpha1"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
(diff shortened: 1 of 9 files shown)


jcantrill (commented):
/approve
/lgtm

anpingli:
Must we fix it?  
For 5.9:  CLO 5.9 don't support v4.17.  
For 5.8:  CLO 5.8 and EO 5.8 are extended to 4.17,  while Loki-operator 5.8 doesn't support OCP 4.17.

anpingli:
The consoleplugin works well on Openshift 4.17.    But the logging-view-plugin can not be replaced with LoggingUI Plugin when test on 4.14.

internal/controller/clusterlogging/clusterlogging_controller.go
```diff
@@ -197,22 +200,40 @@ func (r *ReconcileClusterLogging) SetupWithManager(mgr ctrl.Manager) error {
...
-		controllerBuilder.Watches(&consolev1alpha1.ConsolePlugin{},
-			handler.EnqueueRequestsFromMapFunc(func(_ context.Context, obj client.Object) []reconcile.Request {
-				// Schedule a reconciliation for ClusterLogging if "our" ConsolePlugin was changed
-				if obj.GetName() == console.Name {
-					return []reconcile.Request{
-						{
-							NamespacedName: types.NamespacedName{
-								Namespace: constants.OpenshiftNS,
-								Name:      constants.SingletonName,
+	if console.CapabilityEnabled(context.TODO(), mgr.GetClient(), r.ClusterVersion) {
```
alanconway, line 203: Repetetive, suggest something like:

``` 
  var consolePlugin metav1.Object
	if utils.IsVersionAheadOrEqual(r.ClusterVersion, "v4.17") {
    consolePlugin = &consolev1.ConsolePlugin{}
  } else {
    consolePlugin = &consolev1alpha1.ConsolePlugin{}
  }
	controllerBuilder.Watches(&consolev1.ConsolePlugin{},
		handler.EnqueueRequestsFromMapFunc(func(_ context.Context, obj client.Object) []reconcile.Request {
			// Schedule a reconciliation for ClusterLogging if "our" ConsolePlugin was changed
			if obj.GetName() == console.Name {
				return []reconcile.Request{
					{
						NamespacedName: types.NamespacedName{
							Namespace: constants.OpenshiftNS,
							Name:      constants.SingletonName,
						},
					},
				}
			}
			return nil
		}), builder.WithPredicates(predicate.GenerationChangedPredicate{}))

```

internal/utils/utils.go
```diff
@@ -393,3 +394,16 @@ func GetCollectorName(collectorType logging.LogCollectionType) string {
...
 	return "unknown"
 }
+
+func IsVersionAheadOrEqual(currentVersion, version string) bool {
+	if !strings.HasPrefix(currentVersion, "v") {
+		currentVersion = "v" + currentVersion
+	}
+	if version == "" {
+		return false
+	}
```
alanconway, line 404: Testing == "" after adding "v" will never pass.
```suggestion
	if version == "" {
		return false
	}
	if !strings.HasPrefix(currentVersion, "v") {
		currentVersion = "v" + currentVersion
	}
```

alanconway (changes requested):

PR #2812, opened 2024-10-01
Remove targets for building custom bundles

xperimental:
### Description

Removes the make targets and supporting files for creating custom bundles using overlays.

@alanconway please have a look, maybe there's more unused parts that I missed.


jcantrill:
@xperimental do we have alternate instructions to replace these?  @alanconway added these to satisfy a customer need to deploy release candidate or early releases.

xperimental:
@jcantrill Alan requested to remove these after I added the `olm-*` targets. The new target are no direct replacement of these targets though, the removal is more for "general simplification".

alanconway:
@jcantrill @xperimental these olm- targets are a perfectly good replacement for making "special" release bundle images, I already used them for the last HCP bundle. The are more self-explanatory than the old stuff since they don't require editing separate overlay config, just giving some parameters to make to control version & image destination.

alanconway (approved):

PR #2815, opened 2024-10-03
LOG-6085: Enable CLF spec'd to write to LokiStack using OTEL semantics

Clee2691:
### Description
This PR enables the `ClusterLogForwarder` to write `OTEL` data to Red Hat managed `Lokistack`.

A new field, `dataModel`, is added to the `Lokistack` output in order to designate the type of data sent to `Lokistack`. It supports 2 values: `viaq` or `otel`. The field can be omitted and data sent will default to `viaq`.

The `OTLP` output generation now includes additional attributes in order to preserve UI plugin functionality and will be included in **ALL** `OTLP` outputs.

These attributes are:
1. `log_type`
2. `log_source`
3. `kubernetes_pod_name`
4. `kubernetes_container_name`
5. `kubernetest_namespace_name`

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-6085
- Enhancement proposal: https://github.com/openshift/enhancements/pull/1684

### DIFF
--- a/api/observability/v1/output_types.go
+++ b/api/observability/v1/output_types.go
@@ -748,6 +748,8 @@ type LokiStackAuthentication struct {
 }
 
 // LokiStack provides optional extra properties for `type: lokistack`
+// +kubebuilder:validation:XValidation:rule="!has(self.labelKeys) || !has(self.dataModel) || self.dataModel == 'Viaq'", message="'labelKeys' cannot be set when data model is 'Otel'"
+// +kubebuilder:validation:XValidation:rule="!has(self.tuning) || self.tuning.compression != 'snappy' || !has(self.dataModel) || self.dataModel == 'Viaq'", message="'snappy' compression cannot be used when data model is 'Otel'"
 type LokiStack struct {
 	// Authentication sets credentials for authenticating the requests.
 	//
@@ -790,6 +792,19 @@ type LokiStack struct {
 	// +kubebuilder:validation:Optional
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Stream Label Configuration"
 	LabelKeys *LokiStackLabelKeys `json:"labelKeys,omitempty"`
+
+	// DataModel can be used to customize how log data is stored in LokiStack.
+	//
+	// There are two different models to choose from:
+	//
+	//  - Viaq
+	//  - Otel
+	//
+	// When the data model is not set, it currently defaults to the "Viaq" data model.
+	//
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Data Model"
+	DataModel LokiStackDataModel `json:"dataModel,omitempty"`
 }
 
 // LokiStackLabelKeys contains the configuration that maps log record's keys to Loki labels used to identify streams.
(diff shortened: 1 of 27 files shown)


internal/api/initialize/init_lokiStack.go
```diff
@@ -6,47 +6,86 @@ import (
...
+			Token: ls.Authentication.Token,
+		},
+		Tuning:    ls.Tuning,
+		LabelKeys: lokiStackLabelKeysForTenant(ls.LabelKeys, tenant, lokioutput.DefaultLabelKeys),
+	}
 }
 
-func lokiStackURL(lokiStackSpec *obs.LokiStack, tenant string) string {
+// GenerateOtlpSpec generates and returns an OTLP spec for the defined lokistack output
+// Note: OTLP does not support compression type `snappy` where loki does
```
jcantrill, line 65: Loki using the current ingestion point supports snappy which AFAIK is just a plain HTTP output (@xperimental?).  The restriction as it exists now is in the mark up validation which occurs before any of our validation code so it shouldn't be a hard blocker.  I presume the reason this may not work as-is because "snappy" is not a compression enum value for OTLP, otherwise I would not expect any other restrictions.  Maybe this is a place where an "internal" model should exist which would allow us to pass on this value as something other than an enum

internal/constants/constants.go
```diff
@@ -80,6 +80,8 @@ const (
 	SyslogReceiverPort = 10514
 
 	VolumeNameTrustedCA = "trusted-ca"
+
+	DataModelOtel = "otel"
```
jcantrill, line 84: Should we add this to the API as a "type" which would match other enums?

internal/generator/vector/output/otlp/otlp_all.toml
```diff
@@ -26,6 +26,16 @@ source = '''
       {"key": "k8s.container.name", "value": {"stringValue": get!(.,["kubernetes","container_name"])}},
       {"key": "k8s.namespace.name", "value": {"stringValue": get!(.,["kubernetes","namespace_name"])}}]
   )
+  # Append backward compatibility attributes
+  resource.attributes = append( resource.attributes,
+      [{"key": "log_type", "value": {"stringValue": .log_type}}]
```
jcantrill, line 31: I'm working in this area as well.  I think this may break the current tests because of grouping of some of the infra logs.  I believe we may need to couple this change with "log_source" and probably cluster node name....

xperimental (commented):
Tested it with my OTLP draft PR for Loki Operator and it seems to work as intended in both `dataModel` modes.

What did not work for me was switching from `Viaq` to `Otel` with an existing ClusterLogForwarder. It did create the Condition when the annotation was not present, but it also did not progress any further when adding the annotation. I'm assuming that there's an unintended feedback of the old CLF status into the new reconciliation loop again.

api/observability/v1/output_types.go
```diff
@@ -790,6 +790,18 @@ type LokiStack struct {
...
+	// DataModel is an enum specifying which data model will be sent to Lokistack.
+	//
+	// This field can have 1 of 2 values:
+	// 1. viaq
+	// 2. otel
+	//
+	// This field can also be omitted and the data model will default to `viaq`
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum:=viaq;otel
```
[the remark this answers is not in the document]
Clee2691: Looks like @xperimental commented on this and we can keep it as it is. https://github.com/openshift/cluster-logging-operator/pull/2815#discussion_r1792154011

internal/api/initialize/init_lokiStack.go
Clee2691: Maybe this is a question for @cahartma? What was the decision for only allowing `none` and `gzip` when the `HTTP` output defines these as the choices: `none;gzip;snappy;zlib`?

internal/validations/observability/outputs/lokistack.go
```diff
@@ -0,0 +1,30 @@
...
+// ValidateLokistackOTLPForAnnotation validates lokistacks designated to ingest OTLP data has the OTLP tech preview annotation
+func ValidateLokistackOTLPForAnnotation(context internalcontext.ForwarderContext) {
+	clf := context.Forwarder
+
+	cond := internalobs.NewCondition(obs.ConditionTypeValidLokistackOTLPOutputs, obs.ConditionTrue, obs.ReasonValidationSuccess, "")
+	for _, outSpec := range clf.Spec.Outputs {
+		if outSpec.Type == obs.OutputTypeLokiStack &&
+			(outSpec.LokiStack != nil && outSpec.LokiStack.DataModel == obs.LokiStackDataModelOpenTelemetry) {
+			// Check if OTLP tech preview annotation is defined
+			if _, ok := clf.Annotations[constants.AnnotationOtlpOutputTechPreview]; !ok {
```
jcantrill, line 19: Don't we only need to do this once since it is a "root" condition, meaning we can exit early.  Additionally, if we have success, can that not be set once while exiting?

jcantrill (approved):
/approve

PR #2820, opened 2024-10-10
LOG-6221: 'delivery: AtLeastOnce' cannot be set for 'syslog'

Clee2691:
### Description
This PR adds the `tuning` -> `deliveryMode` configuration option to the `syslog` receiver.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-6221


internal/generator/vector/output/syslog/syslog.go
```diff
@@ -145,6 +146,7 @@ func New(id string, o obs.OutputSpec, inputs []string, secrets observability.Sec
 
 	return append(syslogElements,
 		common.NewAcknowledgments(id, strategy),
+		common.NewBatch(id, strategy),
```
cahartma, line 149: This has been added, but not implemented.   NewBuffer is used for delivery mode and already exists here.   Should we additionally add maxWrite to tuning, since NewBatch is added here?

Clee2691: Actually it should be NewBuffer, not Batch. I shouldn't have actually added this in. The `socket` output doesn't support batch/request parameters that I can see

PR #2821, opened 2024-10-10
LOG-6205: Refactor OTLP attributes to match observability data model

jcantrill:
### Description
This PR:
* Refactors the OTLP attributes to match https://github.com/rhobs/observability-data-model/pull/6
* Blocked by https://github.com/openshift/cluster-logging-operator/pull/2736

### Links
https://issues.redhat.com/browse/LOG-6205

cc @cahartma @xperimental @Clee2691


internal/generator/vector/output/otlp/transform.go
```diff
@@ -13,32 +13,44 @@ const (
...
-resource.attributes = append( resource.attributes, 
-    [{"key": "k8s.cluster.uid", "value": {"stringValue": get!(.,["openshift","cluster_id"])}},
-    {"key": "openshift.log.source", "value": {"stringValue": .log_source}}]
-)
-`
-	HostResourceAttributes = `
-# Append auditd host attributes
-resource.attributes = append( resource.attributes,
-    [{"key": "k8s.node.name", "value": {"stringValue": .hostname}}]
+resource.attributes = append(resource.attributes,
```
xperimental, line 16:
```suggestion
resource.attributes = append(resource.attributes,
```

internal/generator/vector/output/otlp/transform.go
```diff
@@ -140,15 +164,19 @@ o = {
...
 `
 	BackwardCompatContainerResourceAttributes = `
 # Append backward compatibility attributes for container logs
 resource.attributes = append( resource.attributes,
-	[{"key": "kubernetes_pod_name", "value": {"stringValue": get!(.,["kubernetes","pod_name"])}},
-	{"key": "kubernetes_container_name", "value": {"stringValue": get!(.,["kubernetes","container_name"])}},
-	{"key": "kubernetes_namespace_name", "value": {"stringValue": get!(.,["kubernetes","namespace_name"])}}]
+	[{"key": "kubernetes.pod_name", "value": {"stringValue": get!(.,["kubernetes","pod_name"])}},
+	{"key": "kubernetes.container_name", "value": {"stringValue": get!(.,["kubernetes","container_name"])}},
+	{"key": "kubernetes.namespace_name", "value": {"stringValue": get!(.,["kubernetes","namespace_name"])}}]
```
xperimental, line 179: Do you know why this becomes `kubernetes_namespace_name` in the final configuration? (see for example `otlp_all.toml`)
This also applies to the other attributes and the result is that Loki currently does not pick them up as it watches for `kubernetes.namespace_name`.

test/helpers/types/otlp/types.go
```diff
@@ -37,21 +37,32 @@ type LogRecord struct {
...
 	Body                 StringValue `json:"body,omitempty"`
 	Attributes           []Attribute `json:"attributes,omitempty"`
 }
 
 type AttributeValue struct {
-	StringValue string      `json:"stringValue,omitempty"`
+	StringValue *string     `json:"stringValue,omitempty"`
 	Bool        bool        `json:"boolValue,omitempty"`
-	Int         int         `json:"intValue,omitempty"`
+	Int         *string     `json:"intValue,omitempty"` //proto example defines as a string
```
xperimental, line 50: The comment certainly raised an eyebrow, but the example is actually correct (from the spec):

> Note that according to [Protobuf specs](https://developers.google.com/protocol-buffers/docs/proto3#json) 64-bit integer numbers in JSON-encoded payloads are encoded as decimal strings, and either numbers or strings are accepted when decoding.

My guess is that this is to get around languages not having any native integer type (JavaScript)...

xperimental (commented):
Did a short test together with the current state of the Loki Operator draft PR. Some of the new attributes make it across, but the "compatibility attributes" sadly don't, so I did not test with the UI yet. Left some comments already.

internal/generator/vector/output/otlp/transform.go
cahartma: Sanitizing by de-dotting (de-slash, de-dash) occurs on the entire list of LabelKeys and/or DefaultLabelKeys and I was a compatibility issue fix from fluentd.   https://github.com/openshift/cluster-logging-operator/blob/master/internal/generator/vector/output/loki/loki.go#L187

It's been that way in the config since original vector fix.

internal/generator/vector/filter/openshift/viaq/normalize.go
```diff
@@ -80,7 +80,7 @@ if starts_with(pod_name, "eventrouter-") {
   }
 }
 `
-	RemoveStream       = `del(.stream)`
+	RemoveStream       = `.kubeernetes.container_iostream = del(.stream)`
```
Clee2691, line 83: Typo here:
`kubernetes`
```suggestion
	RemoveStream       = `.kubernetes.container_iostream = del(.stream)`
```

internal/generator/vector/output/otlp/transform.go
xperimental: Yes, that's true. But we're still on the OTLP side here, which can (and does) use dots. The label will be called `kubernetes_namespace_name` on the Loki side, but it's able to pull the information from an attribute called `kubernetes.namespace_name`.

xperimental: It's also possible that the file in the unit-test does not reflect the reality yet? While I did not see the correct information on Friday, it seems to work now connecting a CLO with this PR to a LokiStack. I wonder why it did not work on Friday.

internal/generator/vector/filter/openshift/viaq/normalize.go
jcantrill: This is replaced during rebase

internal/generator/vector/output/otlp/transform.go
jcantrill: I am just now updating unit tests for this PR so I will leave this comment open for now

PR #2823, opened 2024-10-11
LOG-5131: Add recording rules for telemetry collection

xperimental:
### Description

This PR introduces recording rules to reduce the possible number of metrics being sent to telemetry for a single installation of Cluster Logging Operator.

This also needs review from the monitoring team, I'll ping them separately.

/cc @jcantrill
/assign @cahartma

### Links

- JIRA: [LOG-5131](https://issues.redhat.com//browse/LOG-5131), [MON-4051](https://issues.redhat.com/browse/MON-4051)

### DIFF
--- a/bundle/manifests/cluster-logging-operator-telemetry_monitoring.coreos.com_v1_prometheusrule.yaml
+++ b/bundle/manifests/cluster-logging-operator-telemetry_monitoring.coreos.com_v1_prometheusrule.yaml
@@ -1,0 +1,19 @@
+apiVersion: monitoring.coreos.com/v1
+kind: PrometheusRule
+metadata:
+  name: cluster-logging-operator-telemetry
+spec:
+  groups:
+  - name: telemetry.rules
+    rules:
+    - expr: sum by (resource_namespace, version) (log_forwarder_pipelines)
+      record: openshift_logging:log_forwarder_pipelines:sum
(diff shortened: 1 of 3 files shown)


xperimental:
fwiw: The `NewFakeClient` function was only deprecated in a few versions of `controller-runtime`. We've updated to a newer version where it's not deprecated anymore, so just removing `//nolint` from the functions is enough, the call does not need to be replaced.

jan--f:
As mentioned in [Jira](https://issues.redhat.com/browse/MON-4051?focusedId=25767573&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25767573), I'd also recommend dropping the version label.
Afaiu this will only ever have one value per cluster and is additionally already exporter in `csv_succeeded`.  If my assumption is true (only ever one value per cluster) I won't block on this however, since it doesn't add additional series. There is however overhead in storing this duplicate information, that could be retrieved at query time.

xperimental:
@jan--f Thanks for the review. I think I'd like to have the `version` label in there for simplifying the queries. As you said, there should only be one version running at a time, except for the transition phase during the update of the operator. So it should not add any additional series.

My experience so far has been that the performance overhead when querying due to the additional "join" is higher than the storage overhead. I'm open to removing the label again in a subsequent patch if we see the overhead due to the label to be too large, though.

PR #2824, opened 2024-10-11
LOG-1814: Avoid use deprecated methods in Cluster Logging Operator

Clee2691:
### Description
This PR unifies fake client creation utilizing `NewFakeClient` in lieu of `NewClientBuilder`.

/cc @vparfonov @cahartma 
/assign @jcantrill

### Links
- JIRA: https://issues.redhat.com/browse/LOG-1814


xperimental:
fwiw: The `NewFakeClient` function was only deprecated in a few versions of `controller-runtime`. We've updated to a newer version where it's not deprecated anymore, so just removing `//nolint` from the functions is enough, the call does not need to be replaced.

alanconway (approved):
+1  The Builder does provide extra flexibility for more complicated tests (e.g. if you need to control the Schema being used etc.) but there's no need to use it when its not required, which is most of the time.

jcantrill (commented):
/lgtm

internal/reconcile/scc_test.go
```diff
@@ -36,11 +37,11 @@ var _ = Describe("reconciling ", func() {
 
 		globalScheme := scheme.Scheme
 		Expect(security.Install(globalScheme)).To(Succeed())
-		builder := fake.NewClientBuilder().WithScheme(globalScheme)
+
```
xperimental, line 40: If you want, you could also inline `globalScheme` above into the `Install()` call. It's just a fancy `AddToScheme` :slightly_smiling_face:

xperimental (commented):
lgtm

PR #2825, opened 2024-10-14
NO-JIRA: Update golang builder and dev tools for 1.22

periklis:
### Description
Minor updates on the code base to build with go1.22 as our next cross-product-component default builder.

/cc @xperimental @cahartma @jcantrill


jcantrill (commented):
/lgtm

PR #2827, opened 2024-10-14
LOG-6149: fix the CSV description regarding EO dependency

jcantrill:
### Description
This PR:
* Drops EO installation requirement that it must be installed prior to the CLO

ref: https://issues.redhat.com/browse/LOG-6149


jcantrill:
/override 
ci/prow/e2e-ocp-target-minus-one

PR #2829, opened 2024-10-15
[release-5.8] NO-JIRA: Update golang builder and dev tools for 1.22

periklis:
### Description
Minor updates on the code base to build with go1.22 as our next cross-product-component default builder.

/cc @xperimental @cahartma @jcantrill


jcantrill:
@periklis I hit retest but believe its safe to override this test if it fails again.  It passed already once

PR #2830, opened 2024-10-15
LOG-6236: Align syslog output implementation with spec RFC3164 and RFC5124

vparfonov:
### Description
This PR addressed to align Syslog Output with syslog specification RFC3164 and RFC5424 

### [RFC3164](https://datatracker.ietf.org/doc/html/rfc3164) 
Format: `<PRI>TIMESTAMP HOSTNAME TAG: MESSAGE`
Example:`<34>Oct 11 22:14:15 mymachine su[1234]: 'su root' failed for lonvick on /dev/pts/`

|   | journal |  application | infrastructure | note |
| ------------- | ------------- |------------- | ------------- |  ------------- |
|AppName  | SYSLOG_IDENTIFIER* | namespacePodContainer  | .log_source  | |
| ProcId | _PID*  | N/A | .auditID (if available) | |
| MsgId | N/A | N/A  | N/A  | will have no effect in settings |
| Payloadkey | .message  | .message  | .message  | |
| Severity | N/A  | .level  | information (6) | |
| Facility | N/A  | user (1)  | security(13) | |

*If `ProcId` available, it will aggregate with `AppName` in the `.tag` field: `appname[procid]`

### [RFC5424](https://datatracker.ietf.org/doc/html/rfc5424) 
Format: `<PRI> VERSION TIMESTAMP HOSTNAME APP-NAME PROCID MSGID [STRUCTURED-DATA] MESSAGE`
Example: `<PRI>1 2024-11-08T14:35:04.123Z ip-10-0-88-196.ec2.internal app-name 1234 ID47 [exampleSDID@32473 iut="3" eventSource="Application" eventID="1011"] This is a sample message`

|   | journal |  application | infrastructure |
| ------------- | ------------- |------------- | ------------- |
|AppName  | SYSLOG_IDENTIFIER | namespace_pod_container  | .log_source  |
| ProcId | _PID  | pod_id | .auditID (if available) |
| MsgId | .log_source | .log_source  | .log_sourcel  |
| Payloadkey | .message  | .message  | .message  |
| Severity | N/A  | .level  | information (6) |
| Facility | N/A  | user (1)  | security(13) |

## Changes: 
- add calculation for default value if field not configured according to the table above
- fix configuration for deployed syslog server in functional test to follow RFC3164 specification 

/cc @cahartma @Clee2691 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-6236
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/output/syslog/rfc3164_with_defaults.toml
+++ b/internal/generator/vector/output/syslog/rfc3164_with_defaults.toml
@@ -4,14 +4,14 @@ inputs = ["application"]
 source = '''
 . = merge(., parse_json!(string!(.message))) ?? .
 if .log_type == "infrastructure" && .log_source == "node" {
-    ._internal.syslog.tag = to_string!(.systemd.u.SYSLOG_IDENTIFIER || "")
-    ._internal.syslog.proc_id = to_string!(.systemd.t.PID || "")
+   ._internal.syslog.tag = to_string!(.systemd.u.SYSLOG_IDENTIFIER || "")
+   ._internal.syslog.proc_id = to_string!(.systemd.t.PID || "")
 }
 if .log_source == "container" {
     ._internal.syslog.tag = join!([.kubernetes.namespace_name, .kubernetes.pod_name, .kubernetes.container_name], "")
     ._internal.syslog.severity = .level
     ._internal.syslog.facility = "user"
-     #Remove non-alphanumeric characters
+    #Remove non-alphanumeric characters
    	._internal.syslog.tag = replace(._internal.syslog.tag, r'[^a-zA-Z0-9]', "")
 	#Truncate the sanitized tag to 32 characters
 	._internal.syslog.tag = truncate(._internal.syslog.tag, 32)
@@ -21,15 +21,12 @@ if .log_type == "audit" {
     ._internal.syslog.severity = "informational"
     ._internal.syslog.facility = "security"
 }
-.proc_id = "procID"
-.tag = "appName"
+.facility = to_string!(._internal.syslog.facility || "user")
+.severity = to_string!(._internal.syslog.severity || "informational")
+.proc_id = to_string!(._internal.syslog.proc_id || "-")
+.tag = to_string!(._internal.syslog.tag || "")
 if exists(.proc_id) && .proc_id != "-" && .proc_id != "" {
-   .tag = .tag + "[" + .proc_id  + "]"
-}
-if is_null(.plKey) {
-	.payload_key = .
-} else {
-	.payload_key = .plKey
+    .tag = .tag + "[" + .proc_id  + "]"
 }
 '''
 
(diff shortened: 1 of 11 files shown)


internal/generator/vector/output/syslog/rfc3164_with_app_name_setting.toml
```diff
@@ -0,0 +1,24 @@
+[transforms.example_parse_encoding]
+type = "remap"
+inputs = ["application"]
+source = '''
+. = merge(., parse_json!(string!(.message))) ?? .
+
+.tag = to_string!(.systemd.u.SYSLOG_IDENTIFIER[.systemd.t.PID] || .systemd.u.SYSLOG_IDENTIFIER || "-")
```
jcantrill, line 7: I think we want the fallback to be empty spaces for all of these.  Fluentd used "-" but I do not see where dash is anything in the spec

internal/generator/vector/output/syslog/syslog.go
```diff
@@ -170,23 +219,53 @@ func getEncodingTemplatesAndFields(s *obs.Syslog) EncodingTemplateField {
...
-			Field:     "msg_id",
-			VRLString: commontemplate.TransformUserTemplateToVRL(s.MsgId),
+			Field:     "severity",
+			VRLString: commontemplate.TransformUserTemplateToVRL(`{._internal.syslog.severity || "informational"}`),
 		})
 	}
 
-	if s.ProcId != "" {
+	if s.ProcId == "" {
+		s.ProcId = `{._internal.syslog.proc_id || "-"}`
```
jcantrill, line 238: My previous comment regarding "default" applies and the reason I really want to hash this out in the document of what the various scenarios look like.  IMO the fallback should not be `-`; it should be empty

vparfonov: yes for RFC3164 should be empty, but for RFC5164 `-`. I will update

internal/generator/vector/output/syslog/rfc3164_with_defaults.toml
```diff
@@ -0,0 +1,43 @@
+[transforms.example_parse_encoding]
+type = "remap"
+inputs = ["application"]
+source = '''
+. = merge(., parse_json!(string!(.message))) ?? .
+if .log_type == "infrastructure" && .log_source == "node" {
```
kattz-kawa, line 6: @vparfonov There might be logs with .log_type == "infrastructure" that do not have .log_source == "node". As far as I can see, I found that the logs from namespace_name=openshift-kube-apiserver and container_name=kube-apiserver do not have .log_source == "node". This line may need a conditional statement for container logs within the openshift-* namespace.

vparfonov: @kattz-kawa Good point, it can be .log_source == "container", but it will be not journal log something close to application logs.

alanconway: I think this logic is correct, could simplify slightly :
- if (node) - these are journald logs and we can map the syslog fields exactly from the logs.
  Note: 'if (node)' is equivalent to 'if (infra && node)' because all node logs are infra logs.
  The difference in format is really node vs. container, not application vs. infra.
- if (container) - container logs are formatted the same regardless of type (application or infra)
- if (audit) - audit logs formatted differently from either of the previous two cases.

alanconway (approved):
Looks good to me.

PR #2834, opened 2024-10-17
LOG-6267: fix provider so operator is evaluated as official

jcantrill:
### Description
back port of #2828 

cc @Clee2691


jcantrill:
Closing given last intended release of 5.9 already includes set of final issues

PR #2859, opened 2024-11-06
LOG-6324: Add level field to otlp logRecord attributes

cahartma:
### Description
This is a fix we are squeezing into 6.1, since the issue presents itself the UI console.

/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-6324

### DIFF
--- a/internal/generator/vector/output/otlp/transform.go
+++ b/internal/generator/vector/output/otlp/transform.go
@@ -99,7 +99,8 @@ if exists(.annotations) {for_each(object!(.annotations)) -> |key,value| {
 	ContainerLogAttributes = `
 r.attributes = append(r.attributes,
   [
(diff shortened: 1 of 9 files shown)


internal/generator/vector/output/otlp/transform.go
```diff
@@ -64,6 +64,7 @@ r.body = {"stringValue": to_string!(get!(.,["_internal","message"]))}
 	APILogAttributes = `
 r.attributes = append(r.attributes,
   [
+    {"key": "level", "value": {"stringValue": .level}},
```
xperimental, line 67: This seems to be for filtering audit logs. I don't think we need "level" in those log entries.

```suggestion
```

xperimental (commented):
Tested on a cluster, seems to work fine for container logs in application and infrastructure as well as journald logs in infrastructure.

Only one change: I don't think we need the level on `audit` logs.

PR #2870, opened 2024-11-14
LOG-6376: skip tls validation if it set in config for HTTP output

vparfonov:
### Description
This PR:
- fix Fluentd configuration generation in case `InsecureSkipVerify: true` by adding `tls_verify_mode none`
- Improve validation procedure for HTTP output, now will be fired error in case not secure URL and TLS settings 

/cc @cahartma @Clee2691 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-6376
- Enhancement proposal:

### DIFF
--- a/internal/generator/fluentd/output/http/http_test.go
+++ b/internal/generator/fluentd/output/http/http_test.go
@@ -226,8 +226,9 @@ var _ = Describe("Generate fluentd config", func() {
 	tls_private_key_path '/var/run/ocp-collector/secrets/http-receiver/tls.key'
 	tls_client_cert_path '/var/run/ocp-collector/secrets/http-receiver/tls.crt'
 	tls_ca_cert_path '/var/run/ocp-collector/secrets/http-receiver/ca-bundle.crt'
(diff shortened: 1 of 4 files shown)


internal/generator/fluentd/output/http/http_test.go
```diff
@@ -319,6 +320,90 @@ var _ = Describe("Generate fluentd config", func() {
...
 </label>
+`,
+		}),
+		Entry("with TLS config but not secure URL", helpers.ConfGenerateTest{
+			CLFSpec: logging.ClusterLogForwarderSpec{
+				Outputs: []logging.OutputSpec{
+					{
+						Type: logging.OutputTypeHttp,
+						Name: "http-receiver",
+						URL:  "http://my-logstore.com/logs/app-logs",
```
jcantrill, line 331: IMO this is a user error because non secure protocol + TLS makes no sense.  We should reject this as a validation error before even attempting to create a configuration

jcantrill (commented):
/approve
/lgtm

PR #2871, opened 2024-11-15
LOG-6204: Multi CLF validation error when adding a namespace which contains 'kube' or 'openshift'

Clee2691:
### Description
This PR adjusts `LOG-6204` to define the below infrastructure namespaces that are to be included when the `infrastructure` input is defined. It also adjusts the excluded list by adding the below infrastructure namespaces to exclude for `application` input sources.

The infrastructure namespaces are as follows:
1. `default`
2. `openshift`
3.  Any namespace prefixed with `openshift-`
4. `kube`
5.  Any namespace prefixed with `kube-`

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-6204


cahartma (commented):
/lgtm

PR #2890, opened 2024-12-02
[master] LOG-6346: fix lokistack validation logic

cahartma:
### Description
- removing the new condition type related to the annotation
- removing prefilter and instead validate annotation during output spec validation
   - updated reason, message, and status on the output condition
- if failed annotation for lokistack otel output, we discussed reverting to viaq, but is instead failing to deploy
   -  since we validate in the output we fail reconcile and thus do not deploy
   -  we need a new condition type (remove from outputs) if we still want to deploy lokistack as "viaq"
- removed "valid log-level" condition to only set when status is false 
- removed unused status package from old api

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-6346


internal/controller/observability/clusterlogforwarder_controller.go
```diff
@@ -95,6 +88,7 @@ func (r *ClusterLogForwarderReconciler) Reconcile(ctx context.Context, req ctrl.
 
 	if !validateForwarder(r.ForwarderContext) {
 		readyCond.Reason = obsv1.ReasonValidationFailure
+		readyCond.Message = "collector not ready"
```
jcantrill, line 91: Consider making this message more expressive.  Maybe we say something about checking the other conditions

internal/validations/observability/common/validate.go
```diff
@@ -3,42 +3,58 @@ package common
...
+		messages = append(messages, fmt.Sprintf("configmap[%s.%s] value is empty", configMapName, key))
+	}
+	return messages
+}
+
+// IsEnabledAnnotation checks if an annotation is set to either "true" or "enabled"
+func IsEnabledAnnotation(context internalcontext.ForwarderContext, annotation string) bool {
+	enabledValues := sets.NewString("true", "enabled")
+	if value, ok := context.Forwarder.Annotations[annotation]; ok {
+		if enabledValues.Has(value) {
```
jcantrill, line 55: Do we care about case?

internal/validations/observability/outputs/validate_tech_preview_test.go
```diff
@@ -0,0 +1,102 @@
...
+				}
+			})
+			It("should pass validation when type is not OTLP", func() {
+				out.Type = obs.OutputTypeHTTP
+				// Return value is empty when validation passes
+				Expect(ValidateTechPreviewAnnotation(context)).To(BeEmpty())
+			})
+		})
+
+		When("output type is LokiStack", func() {
```
jcantrill, line 98: We can remove this if their are no expectations.  Consider just adding a comment if that is something to provide additional clarification

jcantrill:
@cahartma I leave my review suggestions to your discretion.  If you choose to do nothing, just release the hold

internal/controller/observability/clusterlogforwarder_controller.go
cahartma: The `Ready` condition was confusing to me at first as to what it represented.   I thought this simple message explains the issue without any repetition.    
I had considered `reconcile did not finish`  or `unable to complete reconciliation of collector spec...`     But after testing, this brief "what is not ready" message, is only for validation failures, and since they always Fail together, the `Valid` condition already contains the details of why it failed.    
```
  {
    "lastTransitionTime": "2024-12-02T05:31:34Z",
    "message": "one or more of [conditions, inputs, outputs, pipelines, filters] have failed validation",
    "reason": "ValidationFailure",
    "status": "False",
    "type": "observability.openshift.io/Valid"
  },
  {
    "lastTransitionTime": "2024-12-02T05:31:34Z",
    "message": "collector not ready",
    "reason": "ValidationFailure",
    "status": "False",
    "type": "Ready"
  },
```

Other failures for `Ready` condition carry the error message through, since they do not have their own condition type.   Let me know what you want to change it to.

internal/validations/observability/common/validate.go
cahartma: We do.  Isn't everything else in the spec case sensitive?    Kubernetes standard is lowercase for annotations and labels, although I don't think they enforce it?    Also follows yaml standards afaik.

internal/validations/observability/outputs/validate_tech_preview_test.go
cahartma: I'd say we merge this for now and clean up later.    I was aiming to cherrypick the fix into 6.1 and 6.0 before z-stream freeze tomorrow.   :man_shrugging:

PR #2899, opened 2024-12-06
Expose `OPENSHIFT_CLUSTER_NAME` as an env variable

danilo-gemoli:
### Description
Expose the name of the cluster the operator is running on:
```sh
$ oc get infrastructure/cluster -o json | jq .status.infrastructureName
```
as an environment variable `OPENSHIFT_CLUSTER_NAME` on the collector container.
The previous major release `v5.x` of the operator used to set the cluster name as the default value of the `groupPrefix` field on the `ClusterLogForwarder`:
```yaml
apiVersion: logging.openshift.io
kind: ClusterLogForwarder
spec:
  outputs:
  - type: cloudwatch
    cloudwatch:
      groupPrefix: ...
```
check the validation function [here](https://github.com/openshift/cluster-logging-operator/blob/1c607cd51f0cf835d12b15faa251d68c95308b58/internal/validations/clusterlogforwarder/validate_clusterlogforwarderspec.go#L226-L235).
As far as I'm able to understand, that's not the case anymore in `v6.x`.
The old behavior was pretty convenient for a cluster admin because, with no configuration at all, it was possible to give meaningful names to log groups in AWS CloudWatch.
By exposing `OPENSHIFT_CLUSTER_NAME` it would allowed to mimic it once again:
```yaml
apiVersion: observability.openshift.io
kind: ClusterLogForwarder
spec:
  outputs:
  - type: cloudwatch
    cloudwatch:
      groupName: '{ get_env_var("OPENSHIFT_CLUSTER_NAME") + .log_type || "noname" }'
```

/cc @jcantrill 
/assign @jcantrill

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -135,6 +135,11 @@ func main() {
 		log.Error(err, "unable to retrieve the cluster version")
 		os.Exit(1)
 	}
+	clusterName, err := version.ClusterName(mgr.GetAPIReader())
+	if err != nil {
+		log.Error(err, "unable to retrieve the cluster name")
+		os.Exit(1)
+	}
 	migrateManifestResources(mgr.GetClient())
 
 	log.Info("Registering Components.")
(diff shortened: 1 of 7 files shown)


cahartma:
I suggest we discuss further in either a new Jira in the LOG project (@danilo-gemoli), or separate meeting where we can discuss and update this PR.

jcantrill:
Closing this issue as I don't see us accepting this as-is and further discussing as [recommended](url) is the right path forward.https://github.com/openshift/cluster-logging-operator/pull/2899#issuecomment-2528979911

PR #2907, opened 2024-12-12
LOG-6236: add docs about how to forwarding log to Syslog

vparfonov:
### Description
Add document with description how to configure Log Forwarding to the Syslog receiver 

/cc @Clee2691 @cahartma 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s): https://github.com/openshift/cluster-logging-operator/pull/2830
- Bugzilla:
- Github issue:
- JIRA:
- Enhancement proposal:

### DIFF
--- a/docs/features/logforwarding/outputs/syslog-forwarding.adoc
+++ b/docs/features/logforwarding/outputs/syslog-forwarding.adoc
@@ -1,0 +1,148 @@
+== Forwarding to Syslog
+
+Create a Cluster Log Forwarder instance by specifying corresponding parameters.
+
+.cluster-log-forwarder.yaml
+[source,yaml]
+----
+apiVersion: observability.openshift.io/v1
+kind: ClusterLogForwarder
+metadata:
+  name: instance
+  namespace: openshift-logging
+spec:
+  outputs:
+  - name: syslog
+    syslog:
+      rfc: RFC3164 # <1>
+      url: http://my-syslog.com # <2>
+      severity: informational # OPTIONAL parameter can be skipped will be calculated automatically <3>
+      facility: user # OPTIONAL parameter can be skipped will be calculated automatically <4>
+      payloadKey: message # OPTIONAL parameter can be skipped will be calculated automatically <5>
+      appName: myApp # OPTIONAL parameter can be skipped will be calculated automatically <6>
+      procId: 3144 # OPTIONAL parameter can be skipped will be calculated automatically <7>
+      msgId: MsgID # OPTIONAL parameter can be skipped will be calculated automatically <8>
+      enrichment: KubernetesMinimal # <9>
+    type: syslog
+  pipelines:
+    - name: demo-logs
+      inputRefs:
+        - application
+        - infrastructure
+        - audit
+      outputRefs:
+        - syslog
+  serviceAccount:
+    name: mine
+----
+Where `url` and `rfc` parameters are mandatory other are optional
+
+. `rfc`: Syslog RFC specification:  `RFC3164` or `RFC5424`
+. `url`: an absolute URL, with a scheme. Valid schemes are: `http`, `https`, `tcp`, `tls`, `udp` and `udps`
+. `severity`: values are defined in https://tools.ietf.org/html/rfc5424#section-6.2.1
+  The value can be a decimal integer or one of these case-insensitive keywords:
+
+     Emergency
+     Alert
+     Critical
+     Error
+     Warning
+     Notice
+     Informational
+     Debug
+
+. `facility`: values are defined in https://tools.ietf.org/html/rfc5424#section-6.2.1.
+  The value can be a decimal integer. Facility keywords are not standardized,
+  this API recognizes at least the following case-insensitive keywords
+  (defined by https://en.wikipedia.org/wiki/Syslog#Facility_Levels):
+
+    kernel user mail daemon auth syslog lpr news
+    uucp cron authpriv ftp ntp security console solaris-cron
+    local0 local1 local2 local3 local4 local5 local6 local7
+
+
+. `payloadKey`: specifies record field to use as payload. This supports template syntax to allow dynamic per-event values.
+  The `payloadKey` must be a single field path encased in single curly brackets `{}`.
+  Field paths must only contain alphanumeric and underscores. Any field with other characters must be quoted.
+  If left empty, Syslog will use  the whole message as the payload.
+  Examples:
+
+  {.bar}
+  {.foo.bar.baz}
+  {.foo.bar."baz/with/slashes"}
+
+. `appName`: is APP-NAME part of the syslog-msg header.
+  `appName` needs to be specified if using `rfc5424`. The maximum length of the final values is truncated to 48. This supports template syntax* to allow dynamic per-event values.
+
+.  `procId`: is `PROCID` part of the syslog-msg header.
+   `procId` needs to be specified if using `rfc5424`. The maximum length of the final values is truncated to 128. This supports template syntax* to allow dynamic per-event values.
+
+.  `msgId`: is `MSGID` part of the syslog-msg header. This supports template syntax to allow dynamic per-event values. MsgId needs to be specified if using `rfc5424`.  The maximum length of the final values is truncated to 32. This supports template syntax* to allow dynamic per-event values.
+
+. `enrichment`: `None` or `KubernetesMinimal`.
+Take in account that it applies only to container logs.
+`None` add no additional enrichment to the record.
+`KubernetesMinimal` adds `namespace_name`, `pod_name`, and `container_name` to the beginning of the message
+body (e.g. `+namespace_name=myproject, container_name=server, pod_name=pod-123, message={"foo":"bar"}+`).
+This may result in the message body being an invalid JSON structure.
+
+
+[NOTE]
+====
+*Template syntax
+
+The `appName`, `msgId` and `procId` can be a combination of static and dynamic values consisting of field paths followed by `||` followed by another field path or a static value.
+A dynamic value is encased in single curly brackets `{}` and MUST end with a static fallback value separated with `||`.
+Static values can only contain alphanumeric characters along with dashes, underscores, dots and forward slashes.
+
+Examples:
+
+  foo-{.bar||"none"}
+  {.foo||.bar||"missing"}
+  foo.{.bar.baz||.qux.quux.corge||.grault||"nil"}-waldo.fred{.plugh||"none"}
+====
+
+=== Defaults value of Syslog fields
+Will be calculating during processing log event if values not set:
+
+==== RFC3164
+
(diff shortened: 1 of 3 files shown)


jcantrill (commented):
/hold

docs/features/logforwarding/outputs/syslog-forwarding.adoc
```diff
@@ -0,0 +1,159 @@
...
+  Field paths must only contain alphanumeric and underscores. Any field with other characters must be quoted.
+  If left empty, Syslog will use the whole message as the payload key.
+  Examples:
+
+  {.bar}
+  {.foo.bar.baz}
+  {.foo.bar."baz/with/slashes"}
+
+. `appName`: is APP-NAME part of the syslog-msg header.
+  `appName` needs to be specified if using `rfc5424`. The maximum length of the final values is truncated to 48
```
jcantrill, line 75: We need a clarification here and similar in the remaining part of the doc.  This is a correct statement but it reads to me like we expect a user to set this in our API.  We do not need them to set it in the API because we are going to provide defaults.  I might even consider removing the statement all-to-gether.  cc @cahartma  thoughts?

docs/features/logforwarding/outputs/syslog-forwarding.adoc
```diff
@@ -0,0 +1,159 @@
...
+  This supports template syntax to allow dynamic per-event values.
```
jcantrill, line 76: I think we could pull all of these statements for all of these fields into  single statement.  Identify which ones support templating, what is the correct syntax, how it will be treated, etc.  This will remove the redundancy and repeated verbiage.

docs/features/logforwarding/outputs/syslog-forwarding.adoc
```diff
@@ -0,0 +1,159 @@
...
+Static values can only contain alphanumeric characters along with dashes, underscores, dots and forward slashes.
+Example:
+
+    foo-{.bar||"none"}
+    {.foo||.bar||"missing"}
+    foo.{.bar.baz||.qux.quux.corge||.grault||"nil"}-waldo.fred{.plugh||"none"}
+
+. `enrichment`: `None` or `KubernetesMinimal`.
+`None` add no additional enrichment to the record.
+`KubernetesMinimal` adds `namespace_name`, `pod_name`, and `collector_name` to the beginning of the message
```
jcantrill, line 110: This should apply only to container logs.  Did we "fix" vector to only apply these fields if the appropriate meta existed? I'm thinking of the case where we saw something like `pod_name= coontainer_name=`

jcantrill, line 110: should be container_name not collector_name

jcantrill (commented):
I was actually thinking the table would be part of the apis/observability

docs/features/logforwarding/outputs/syslog-forwarding.adoc
vparfonov: copy/past error from here https://github.com/openshift/cluster-logging-operator/blob/master/api/observability/v1/output_types.go#L1219

vparfonov: according to the Vector code, such situation possible if value no available, see: https://github.com/Clee2691/vector/blob/release-6.1/lib/codecs/src/encoding/format/syslog.rs#L335-L354

docs/features/logforwarding/outputs/syslog-forwarding.adoc
vparfonov: Maybe just add description that this is OPTIONAL and don't set if you are not sure, because i think in some cases `appName` can be useful

docs/features/logforwarding/outputs/syslog-forwarding.adoc
vparfonov: fixed

docs/features/logforwarding/outputs/syslog-forwarding.adoc
vparfonov: done

docs/features/logforwarding/outputs/syslog-forwarding.adoc
vparfonov: fixed

docs/features/logforwarding/outputs/syslog-forwarding.adoc
```diff
@@ -0,0 +1,149 @@
...
+        - syslog
+  serviceAccount:
+    name: mine
+----
+Where `url` and `rfc` parameters are mandatory other are optional
+
+. `rfc`: Syslog RFC specification:  `RFC3164` or `RFC5424`
+. `url`: an absolute URL, with a scheme. Valid schemes are: `http`, `https`, `tcp`, `tls`, `udp` and `udps`
+. `severity`: values are defined in https://tools.ietf.org/html/rfc5424#section-6.2.1
+  The value can be a decimal integer or one of these case-insensitive keywords:
```
jcantrill, line 43: Is this true?  Have we tried giving it a number?

docs/features/logforwarding/outputs/syslog-forwarding.adoc
```diff
@@ -0,0 +1,149 @@
...
+     Alert
+     Critical
+     Error
+     Warning
+     Notice
+     Informational
+     Debug
+
+. `facility`: values are defined in https://tools.ietf.org/html/rfc5424#section-6.2.1.
+  The value can be a decimal integer. Facility keywords are not standardized,
```
jcantrill, line 55: Same question here

docs/features/logforwarding/outputs/syslog-forwarding.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+    kernel user mail daemon auth syslog lpr news
+    uucp cron authpriv ftp ntp security console solaris-cron
+    local0 local1 local2 local3 local4 local5 local6 local7
+
+
+. `payloadKey`: specifies record field to use as payload. This supports template syntax to allow dynamic per-event values.
+  The `payloadKey` must be a single field path encased in single curly brackets `{}`.
+  Field paths must only contain alphanumeric and underscores. Any field with other characters must be quoted.
+  If left empty, Syslog will use the whole message as the payload key.
```
jcantrill, line 67: I think it should say:

... the whole message as the payload.

docs/features/logforwarding/outputs/syslog-forwarding.adoc
```diff
@@ -0,0 +1,149 @@
...
+
+. `appName`: is APP-NAME part of the syslog-msg header.
+  `appName` needs to be specified if using `rfc5424`. The maximum length of the final values is truncated to 48. This supports template syntax* to allow dynamic per-event values.
+
+.  `procId`: is `PROCID` part of the syslog-msg header.
+   `procId` needs to be specified if using `rfc5424`. The maximum length of the final values is truncated to 128. This supports template syntax* to allow dynamic per-event values.
+
+.  `msgId`: is `MSGID` part of the syslog-msg header. This supports template syntax to allow dynamic per-event values. MsgId needs to be specified if using `rfc5424`.  The maximum length of the final values is truncated to 32. This supports template syntax* to allow dynamic per-event values.
+
+. `enrichment`: `None` or `KubernetesMinimal`.
```
jcantrill, line 82: Add note this applies only to container logs.  Consider working it as "application logs" and/or "infrastructure container" logs

docs/features/logforwarding/outputs/syslog-forwarding.adoc
```diff
@@ -0,0 +1,149 @@
...
+|Severity|N/A|.level|information (6)|
+|Facility|N/A|user (1)|security (13)6|
+|===
+*If `ProcId` available, it will aggregate with `AppName` in the `.tag` field: `appname[procid]`
+
+== RFC5424
+
+Format: `+<PRI> VERSION TIMESTAMP HOSTNAME APP-NAME PROCID MSGID [STRUCTURED-DATA] MESSAGE+`
+
+Example: `+<PRI>1 2024-11-08T14:35:04.123Z ip-10-0-88-196.ec2.internal app-name 1234 ID47 [exampleSDID@32473 iut="3" eventSource="Application" eventID="1011"] This is a sample message+`
```
jcantrill, line 128: Consider removing structured data from the example.  AFAIK we do not explicitly do anything to populate this or allow a user to control what goes into this field

docs/features/logforwarding/outputs/syslog-forwarding.adoc
```diff
@@ -0,0 +1,149 @@
...
+Will be calculating during processing log event if values not set:
+
+==== RFC3164
+
+Format: `+<PRI>TIMESTAMP HOSTNAME TAG: MESSAGE+`
+
+Example: `+<34>Oct 11 22:14:15 mymachine su[1234]: 'su root' failed for lonvick on /dev/pts/+`
+
+|===
+| | journal |application|infrastructure|note
```
jcantrill, line 114: We have a little bit of conflicting statements here because of how we define log type.  Infra is container or journal where application is only container.  We should probably change app and infra to just be "container" and then add "audit" as another header since we process those.  Also the header could be something like:

Infrastructure Journal, Infrastructure/Application Container, Audit

docs/features/logforwarding/outputs/syslog-forwarding.adoc
vparfonov: yes, i added test to confirm

docs/features/logforwarding/outputs/syslog-forwarding.adoc
```diff
@@ -0,0 +1,148 @@
...
+  The `payloadKey` must be a single field path encased in single curly brackets `{}`.
+  Field paths must only contain alphanumeric and underscores. Any field with other characters must be quoted.
+  If left empty, Syslog will use  the whole message as the payload.
+  Examples:
+
+  {.bar}
+  {.foo.bar.baz}
+  {.foo.bar."baz/with/slashes"}
+
+. `appName`: is APP-NAME part of the syslog-msg header.
```
jcantrill, line 74: > We need a clarification here and similar in the remaining part of the doc. This is a correct statement but it reads to me like we expect a user to set this in our AP

Please update these fields to say Optional where appropriate as in the sample above.

jcantrill (changes requested):

jcantrill:
@vparfonov one comment otherwise this LGTM

docs/features/logforwarding/outputs/syslog-forwarding.adoc
vparfonov: added here:  https://github.com/openshift/cluster-logging-operator/pull/2907/files#diff-dcb126cbdb763df54cb71b630990f555a494cc9cb5406b0fb018b157a9d5e035R38

PR #2908, opened 2024-12-12
[master] LOG-6222: Include timestamp field to fix console

cahartma:
### Description
There is a lot more to this than initially written and we have had several meetings on the best way to resolve.   This fixes the console and aligns `timestamp` with the existing `@timestamp`   We are still awaiting a Loki update to time-based sharding that will  prevent any date concern as originally discussed in the ticket.   But can merge into master for the time being.

We should wait to mark the issue as resolved, until the Loki update is implemented.

/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-6222


test/helpers/types/types.go
```diff
@@ -195,7 +195,21 @@ type ViaQCommon struct {
...
+	// Value derived from legacy `@timestamp` for forward compatibility.
+	//
+	// format:
+	//
+	// * yyyy-MM-dd HH:mm:ss,SSSZ
+	// * yyyy-MM-dd'T'HH:mm:ss.SSSSSSZ
+	// * yyyy-MM-dd'T'HH:mm:ssZ
+	// * dateOptionalTime
+	//
+	// example: `2015-01-24 14:06:05.071000000 Z`
```
xperimental, line 211: Shouldn't this be:

```suggestion
	// example: `2015-01-24T14:06:05.071000000Z`
```

internal/generator/vector/output/splunk/splunk.go
```diff
@@ -95,11 +95,11 @@ func sink(id string, o obs.OutputSpec, inputs []string, index string, secrets ob
 
 func FixTimestampFormat(componentID string, inputs []string) Element {
 	var vrl = `
-ts, err = parse_timestamp(.@timestamp,"%+")
+.timestamp, err = parse_timestamp(.@timestamp,"%+")
```
xperimental, line 98: If I understand correctly, then this in the sink. Shouldn't all entries have both `.timestamp` and `.@timestamp` at this point already?

xperimental (commented):
Not tested, only looked at the diff so far.

internal/generator/vector/output/splunk/splunk.go
cahartma: `timestamp` comes from vector, and depending on source/sink is not always the value or format we want.   I assumed the parse was necessary here, so setting the new `timestamp` field to the same we already have VRL for (`@timestamp`) .    It's possible this formatting is not necessary for splunk.   I can try it again and see what splunk is expecting and/or accepts.

test/helpers/types/types.go
cahartma: Quite possibly, I just copied the current.   Let me check into it.

internal/generator/vector/output/splunk/splunk.go
xperimental: Seeing as VRL is typed, I would expect `.timestamp` to have a type of "timestamp" (think `time.Time` in Go) as opposed to being an opaque type ("string"), meaning that it should possible for the sink to convert it to whatever representation it needs, be that, for example, a unix-timestamp or something RFC3339 formatted.

test/helpers/types/types.go
cahartma: Updated

internal/generator/vector/output/splunk/splunk.go
cahartma: I will revert these splunk changes, and leave `timestamp` as is.

cahartma:
Seems this will need to wait for the Loki update.   Current functional tests are failing to capture 'old' messages.    Investigating a possible interim solution.

test/functional/normalization/message_format_test.go
```diff
@@ -44,7 +44,7 @@ var _ = Describe("[Functional][LogForwarding][Normalization] tests for message f
 
 		// Template expected as output Log
 		var outputLogTemplate = functional.NewApplicationLogTemplate()
-		outputLogTemplate.Timestamp = nanoTime
+		outputLogTemplate.TimestampLegacy = nanoTime
```
jcantrill, line 47: Why would we not want to additionally verify Timestamp = nanoTime?  Same applies in all these files

xperimental:
Can't we get rid of the `@timestamp` field in most outputs? (all except for ElasticSearch, maybe?)

cahartma:
> Can't we get rid of the `@timestamp` field in most outputs? (all except for ElasticSearch, maybe?)

I would say yes, but I'd suggest we do it in a future release, separate from this initial "console fix".

jcantrill:
> > Can't we get rid of the `@timestamp` field in most outputs? (all except for ElasticSearch, maybe?)
> 
> I would say yes, but I'd suggest we do it in a future release, separate from this initial "console fix".

@xperimental  The answer is "no" since its a breaking API change for ViaQ.  I agree the field is undesirable but removing it breaks anyone who actually depends upon it.  As much as we wish to push everyone to Otel, I believe we should still consider have a configurable version of ViaQ where we can clean up what exists now

PR #2912, opened 2024-12-17
LOG-6348: Collector startup script removes buffer lock files breaking locking code of vector

Clee2691:
### Description
This PR removes the deletion of the buffer lock files in `vector`'s startup script.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-6348

### DIFF
--- a/internal/collector/vector/run-vector.sh
+++ b/internal/collector/vector/run-vector.sh
@@ -7,24 +7,5 @@ VECTOR_DATA_DIR=%s
 echo "Creating the directory used for persisting Vector state $VECTOR_DATA_DIR"
 mkdir -p ${VECTOR_DATA_DIR}
 
-echo "Checking for buffer lock files"
-# Vector does not appear to release locks when the process terminates. Try to clean up for restart
-pushd ${VECTOR_DATA_DIR}
-  locks=$(ls -R . | awk '/:$/&&f{s=$0;f=0}
-/:$/&&!f{sub(/:$/,"");s=$0;f=1;next}
-NF&&f{ print s"/"$0 }' | grep \.lock)
-  rc=$?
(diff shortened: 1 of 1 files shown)


internal/collector/vector/run-vector.sh
```diff
@@ -7,24 +7,5 @@ VECTOR_DATA_DIR=%s
 echo "Creating the directory used for persisting Vector state $VECTOR_DATA_DIR"
 mkdir -p ${VECTOR_DATA_DIR}
```
xperimental, line 8: If I have seen this correctly, then the data directory is the base of a VolumeMount in the current configuration, meaning that it will always exist. If that is actually the case, I think it would be safe to completely remove the startup script and directly use vector as the entrypoint for the container.

jcantrill: The operator passes in the datadir and there actually is a conditional of the value based upon the "legacy" naming convention and other.  This behavior persists to support 'upgrade' behavior.

xperimental: As far as I can see the path can either be `/var/lib/vector` or `/var/lib/vector/<namespace>/<instance>`, but in both cases the path is directly used as the VolumeMount ([code](https://github.com/openshift/cluster-logging-operator/blob/9c07857c8b2c94e5b7d57b928816919b5cc30ebc/internal/collector/vector/visitors.go#L25)). I don't think the `mkdir` and thus the script is needed.

Clee2691: I'm inclined to leave this as is and we can revisit the start up script in another scope (JIRA)

PR #2921, opened 2025-01-10
LOG-6280: Only cleanup dashboard when owned by 'this' operator version

jcantrill:
### Description
This PR:
* Adds labels version, managedby to the collector dashboard configmap
* Modifies logic to only delete the dashboard on operator exit if the label set matches those expected by the operator
* Resolves the issue where the old version of the operator deleted the dashboard on_exit after the new version of the operator updated it

### Links
https://issues.redhat.com/browse/LOG-6280

### DIFF
--- a/internal/metrics/dashboard/dashboards.go
+++ b/internal/metrics/dashboard/dashboards.go
@@ -4,18 +4,18 @@ import (
 	"context"
 	_ "embed"
 	"fmt"
-
 	staticlog "github.com/ViaQ/logerr/v2/log/static"
+	"github.com/openshift/cluster-logging-operator/internal/constants"
 	"github.com/openshift/cluster-logging-operator/internal/reconcile"
 	"github.com/openshift/cluster-logging-operator/internal/runtime"
 	"github.com/openshift/cluster-logging-operator/internal/utils"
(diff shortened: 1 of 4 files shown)


internal/metrics/dashboard/dashboards.go
```diff
@@ -83,14 +87,25 @@ func ReconcileForDashboards(k8sClient client.Client, reader client.Reader) error
...
+		r, _ := labels.NewRequirement(k, selection.Equals, []string{v})
+		s = s.Add(*r)
+	}
+	options := &client.ListOptions{
+		LabelSelector: s,
+		Namespace:     DashboardNS,
 	}
-	return c.Delete(context.TODO(), cm)
+	configMaps := &corev1.ConfigMapList{}
+	if err := r.List(context.TODO(), configMaps, options); err != nil {
```
alanconway, line 102: How about Get(cm) and then compare the actual labels with the expected ones? 
The List() could find configmaps that have identical labels but a different name - its unlikely, but a Get is more efficient and direct.

jcantrill: @alanconway updated to check labels

PR #2940, opened 2025-01-27
LOG-6585: Fail validition if output is unreferenced

jcantrill:
### Description
This PR:
* validates each output is referenced by a pipeline or fails
* this will keep the operator from deploying a collector with configuration not accepted by vector

### Links
https://issues.redhat.com/browse/LOG-6585

cc @vparfonov @cahartma

### DIFF
--- a/internal/api/observability/pipelines.go
+++ b/internal/api/observability/pipelines.go
@@ -1,6 +1,9 @@
 package observability
 
-import obs "github.com/openshift/cluster-logging-operator/api/observability/v1"
+import (
+	obs "github.com/openshift/cluster-logging-operator/api/observability/v1"
+	"k8s.io/utils/set"
+)
 
 type Pipelines []obs.PipelineSpec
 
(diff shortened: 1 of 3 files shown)


internal/api/observability/pipelines.go
```diff
@@ -20,3 +23,13 @@ func (pipeline Pipelines) Map() map[string]obs.PipelineSpec {
...
+
+// ReferenceOutput iterates through the list of pipelines to see if any reference the given output
+func (pipeline Pipelines) ReferenceOutput(output obs.OutputSpec) bool {
+	for _, i := range pipeline {
+		if set.New(i.OutputRefs...).Has(output.Name) {
+			return true
+		}
+	}
+	return false
+}
```
vparfonov, line 35: small remark, maybe will be better to build the set once and then check:
```
func (pipeline Pipelines) ReferenceOutput(output obs.OutputSpec) bool {
    refs := set.New[string]() 
    for _, i := range pipeline {
        refs.Insert(i.OutputRefs...)
    }
    return refs.Has(output.Name) 
}
```

PR #2953, opened 2025-02-04
LOG-5706: Align syslog output implementation with spec RFC3164 and RFC5124 for Vector

vparfonov:
### Description
Backport for https://github.com/openshift/cluster-logging-operator/pull/2830

/cc @cahartma 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-5706
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/helpers/helpers.go
+++ b/internal/generator/vector/helpers/helpers.go
@@ -2,6 +2,7 @@ package helpers
 
 import (
 	"fmt"
+	"regexp"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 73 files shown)


jcantrill (commented):
/hold

internal/generator/vector/helpers/helpers.go
```diff
@@ -56,3 +59,49 @@ func ListenOnAllLocalInterfacesAddress() string {
 func EscapeDollarSigns(s string) string {
 	return strings.ReplaceAll(s, "$", "$$")
 }
+
+// TransformUserTemplateToVRL converts the user entered template to VRL compatible syntax
+// Example: foo-{.log_type||"none"} -> "foo-" + to_string!(.log_type||"none")
```
jcantrill, line 64: It may be that we need to reject these changes for 5.9 and earlier.  My brief review of the API is we don't support templating like 6.x and I'm hesitant to add that functionality here.

jcantrill (commented):
Let's talk about this further.  I wish I would have considered the larger implications of this change before you did the work as I think its adding in "features" we wish to not backport

PR #2956, opened 2025-02-06
LOG-6532: fix validation rule for Loki Tuning to allow it to be empty without failing validation

vparfonov:
### Description

/cc @cahartma 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-6532
- Enhancement proposal:

### DIFF
--- a/api/observability/v1/output_types.go
+++ b/api/observability/v1/output_types.go
@@ -772,7 +772,7 @@ type LokiStackAuthentication struct {
 
 // LokiStack provides optional extra properties for `type: lokistack`
 // +kubebuilder:validation:XValidation:rule="!has(self.labelKeys) || !has(self.dataModel) || self.dataModel == 'Viaq'", message="'labelKeys' cannot be set when data model is 'Otel'"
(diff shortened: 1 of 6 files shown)


api/observability/v1/output_types.go
```diff
@@ -739,6 +739,7 @@ type LokiTuningSpec struct {
 	// Compression causes data to be compressed before sending over the network.
 	//
 	// +kubebuilder:validation:Enum:=none;gzip;snappy
+	// +kubebuilder:default=none
```
jcantrill, line 742: Per our larger API discussions, @xperimental has recommended not setting any defaults because of the element of "surprise" when a CR is different from that spec'd by the user once fetched from storage.  Doing this means you will do nothing but the return will be set to none.

I believe the only way to resolve this is via an initializer that evaluates and modifies the CLF based upon the type and datamodel

vparfonov: Revise code again and think problem was in validation rule, added checking `!has(self.tuning.compression)`

PR #2958, opened 2025-02-07
LOG-6661: fix inconsistency for syslog facillity and severity templating

jcantrill:
### Description
This change:
* Restricts syslog.(severity|facility) to the template pattern supported elsewhere in the API
* Removes the capability to use the 5.x syslog template notation (e.g. `$$.message.foo`)
* Updates tests to validate changes

### Links
* https://issues.redhat.com/browse/LOG-6661

### DIFF
--- a/api/observability/v1/output_types.go
+++ b/api/observability/v1/output_types.go
@@ -1083,6 +1083,7 @@ type Syslog struct {
 	//     Emergency Alert Critical Error Warning Notice Informational Debug
 	//
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Pattern:=`^(([a-zA-Z0-9-_.\/])*(\{(\.[a-zA-Z0-9_]+|\."[^"]+")+((\|\|)(\.[a-zA-Z0-9_]+|\.?"[^"]+")+)*\|\|"[^"]*"\})*)*$`
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Severity",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
 	Severity string `json:"severity,omitempty"`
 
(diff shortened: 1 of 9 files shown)


internal/generator/vector/output/syslog/syslog_test.go
```diff
@@ -112,8 +112,8 @@ var _ = Describe("vector syslog clf output", func() {
 			spec.Syslog = &obs.Syslog{
 				URL:        "tls://logserver:6514",
 				RFC:        obs.SyslogRFC5424,
-				Facility:   "$$.message.facility",
-				Severity:   "$$.message.severity",
+				Facility:   `{.facility||"none"}`,
```
vparfonov, line 115: i think "none" is not correct fallback value

api/observability/v1/output_types.go
```diff
@@ -1099,6 +1100,7 @@ type Syslog struct {
 	//     local0 local1 local2 local3 local4 local5 local6 local7
 	//
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Pattern:=`^\{(\.[a-zA-Z0-9_]+|\."[^"]+")(\.[a-zA-Z0-9_]+|\."[^"]+")*\}$`
```
vparfonov, line 1103: pattern not match `kernel` and other words without `{}`

api/observability/v1/output_types.go
```diff
@@ -1083,6 +1083,7 @@ type Syslog struct {
 	//     Emergency Alert Critical Error Warning Notice Informational Debug
 	//
 	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Pattern:=`^\{(\.[a-zA-Z0-9_]+|\."[^"]+")(\.[a-zA-Z0-9_]+|\."[^"]+")*\}$`
```
vparfonov, line 1086: pattern not match word without `{}`

internal/generator/vector/output/syslog/syslog_test.go
jcantrill: I don't think it matters for the context of the test

PR #2961, opened 2025-02-13
WIP: Buffer tuning for https://issues.redhat.com/browse/RFE-6849

alanconway:
This PR has the proposed API changes needed for https://issues.redhat.com/browse/RFE-6849
The functionality is not implemented.


api/observability/v1/output_types.go
```diff
@@ -210,15 +210,48 @@ type BaseOutputTuningSpec struct {
...
+	// NOTE: There are trade-offs to changing buffer sizes.
+	// The defaults are calculated to provide good performance in most cases.
+	// It is not recommended to set this field unless you have done tests
+	// to determine a value that performs better than the defaults in your specific setting.
+	// See also []
+	//
+	// The following are some trade-offs to bear in mind:
+	//
+	// - MaxBufferSize is per-output, add up all the outputs for the total impact per-node.
+	// - With `deliveryMode: AtLeastOnce`, buffers are created on the node's `/var/log` partition.
```
jcantrill, line 243: buffers are created on `/var/lib` partition:  `data_dir = "/var/lib/vector/openshift-logging/my-forwarder"`

api/observability/v1/output_types.go
```diff
@@ -210,15 +210,48 @@ type BaseOutputTuningSpec struct {
...
+	//
+	// NOTE: There are trade-offs to changing buffer sizes.
+	// The defaults are calculated to provide good performance in most cases.
+	// It is not recommended to set this field unless you have done tests
+	// to determine a value that performs better than the defaults in your specific setting.
+	// See also []
+	//
+	// The following are some trade-offs to bear in mind:
+	//
+	// - MaxBufferSize is per-output, add up all the outputs for the total impact per-node.
```
jcantrill, line 242: it can be event more severe; it is sum per output of ALL deployed CLF instances.  We should consider how if the existing alert we have is sufficient and if there is anything in the operator we can do to guard against master nodes having issues

api/observability/v1/output_types.go
```diff
@@ -210,15 +210,48 @@ type BaseOutputTuningSpec struct {
...
+	// - With `deliveryMode: AtLeastOnce`, buffers are created on the node's `/var/log` partition.
+	//   Filling that partition can have severe consequences for the entire node.
+	// - With `deliveryMode: AtLeastOnce` the original log file itself is used as a replay buffer
```
jcantrill, line 245: This is not exactly correct since we were not able to impl acks given kubernetes_logs does not support acks at this time.  You could still lose logs collected but not sent

api/observability/v1/output_types.go
```diff
@@ -646,6 +679,14 @@ type KafkaTuningSpec struct {
 	// +kubebuilder:validation:Enum:=none;snappy;zstd;lz4
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Compression"
 	Compression string `json:"compression,omitempty"`
+
+	// MessageTimeout messages that are waiting to be sent are dropped after this timeout.
+	// Used to ensure that newer messages are not blocked indefinitely by older ones.
+	//
+	// A duration string is a possibly signed sequence of decimal numbers, each with optional fraction
+	// and a unit suffix, such as "300ms", "-1.5h" or "2h45m". Valid time units are "ns", "us" (or
```
jcantrill, line 687: Remove negative example.  Do we need to say it's "signed" since only positve numbers make sense.  We should additionally add a validation to restrict it to positive numbers

api/observability/v1/output_types.go
```diff
@@ -210,15 +210,48 @@ type BaseOutputTuningSpec struct {
 
 	// MinRetryDuration is the minimum time to wait between attempts to retry after delivery a failure.
 	//
+	// A duration string is a possibly signed sequence of decimal numbers, each with optional fraction
```
jcantrill, line 213: Same comment as below

api/observability/v1/output_types.go
alanconway: I pasted direct from the Go docs, thanks for catching.

api/observability/v1/output_types.go
alanconway: Argh, I did not know that. I assumed it was supported because it is for file logs, and k8s are file logs with decorations. That is unfortunate. I'll remove the comment.

alanconway:
@jcantrill all good catches, hopefully now all fixed. Let me know.

api/observability/v1/output_types.go
jcantrill: > Remove negative example. Do we need to say it's "signed" since only positve numbers make sense. We should additionally add a validation to restrict it to positive numbers

Even if this is straight from the go docs we should still correct the verbiage for our usecase.  IMO we should also restrict it to a precision that is supported by vector.  Do we think that a user can reasonably set it to use microsecond precision?   A negative duration is not valid in our context .  Additionally, I would like to see a validation here, if possilble,  to enforce a postive value

api/observability/v1/output_types.go
jcantrill: If we are to add buffer size then we also need to warn users that this is a cumulative issue where ALL forwarders deployed that declares disk buffer is an issue.  We run the risk of consuming too much node disk and destabilizing every workload on an individual node.  This is exacerbated on master nodes

api/observability/v1/output_types.go
alanconway: Yep, updated.

PR #2965, opened 2025-02-19
[release-6.0] LOG-6756: Add timestamp to fix console

cahartma:
### Description
Manual backport for 6.0.    This fixes the console and aligns `timestamp` with the existing `@timestamp`   We are still awaiting a Loki update to time-based sharding that will prevent any date concern as originally discussed in the ticket.   

We will want to wait until the Loki update is implemented before marking this as resolved.

/assign @jcantrill

### Links
- https://issues.redhat.com/browse/LOG-6755
- release-6.1: https://issues.redhat.com/browse/LOG-6755
- clone of 6.2: https://issues.redhat.com/browse/LOG-6222


hack/logsamples/viaq_journal.json
```diff
@@ -1,5 +1,5 @@
 {
-  "@timestamp": "2023-10-17T20:46:29.048949Z",
+  "timestamp": "2023-10-17T20:46:29.048949Z",
```
jcantrill, line 2: Why did @timestamp disappear for journal?

test/helpers/loki/receiver.go
```diff
@@ -61,6 +63,7 @@ func NewReceiver(ns, name string) *Receiver {
 			"-server.grpc-max-recv-msg-size-bytes", "20971520",
 			"-distributor.ingestion-rate-limit-mb", "200",
 			"-distributor.ingestion-burst-size-mb", "200",
+			"-validation.discover-log-levels=false",
```
jcantrill, line 66: Why does this fail now and didn't before this submission?

test/matchers/log_format_test.go
```diff
@@ -53,14 +53,14 @@ var _ = Describe("Log Format matcher tests", func() {
 	It("match same time value", func() {
 		timestamp := "2013-03-28T14:36:03.243000+00:00"
 		nanoTime, _ := time.Parse(time.RFC3339Nano, timestamp)
-		Expect(types.AllLog{ContainerLog: types.ContainerLog{ViaQCommon: types.ViaQCommon{Timestamp: nanoTime}}}).
-			To(FitLogFormatTemplate(types.AllLog{ContainerLog: types.ContainerLog{ViaQCommon: types.ViaQCommon{Timestamp: nanoTime}}}))
+		Expect(types.AllLog{ContainerLog: types.ContainerLog{ViaQCommon: types.ViaQCommon{TimestampLegacy: nanoTime}}}).
```
jcantrill, line 56: shouldnt we be testing for population of both values since we are sending both for backwards compatibility?

test/helpers/loki/receiver.go
cahartma: Updated version of loki is the cause.   Robert is aware and confirmed this is necessary.

test/matchers/log_format_test.go
cahartma: The original idea was we wanted to remove the original field because we determined it has no value.   As an alternative, we kept both to help ensure no regression issues.    Should we be concerned about testing an unused field?

hack/logsamples/viaq_journal.json
cahartma: Good catch, I can't recall off hand.    Will investigate, as I suspect I had a reason for doing so.

PR #2978, opened 2025-03-05
LOG-6583: Add Makefile logic to set build version

jcantrill:
### Description
This PR:
* Adds logic to generate build specific version information

### Links
https://issues.redhat.com/browse/LOG-6583


jcantrill:
@xperimental please comment why you don't like this approach or offer an alternative.  Implementation of konflux is going to require us to add some logic to set this at build time in a similar way that is provided by CPaaS now

xperimental:
@jcantrill Forgot about this one, but _I think_ we had discussed this a few months ago, but I am not sure anymore. Also don't know if our recent learnings about Konflux might have obsoleted this anyway.

PR #2979, opened 2025-03-11
LOG-6734: Splunk Event Metadata: Enhancements Doc

vparfonov:
### Description
This PR adds detailed documentation for forwarding logs to Splunk using Cluster Log Forwarder. 

It covers the use of Spunk metadata keys via Vector configuration parameters: `source`, `indexed_fields`, `sourceType`, and `host`. Added `payloadKey` parameter.

- How source can be dynamically derived from `.log_type` and `.log_source`
- Explanation of `indexed_fields`, their limitations, and storage implications
- Description of the extended remap function, allowing nested field support and automatic value conversion
- `payloadKey` specifies record field to use as payload
- `sourceType` determined automatically based on the type of the final event payload (`_json` or `generic_single_line`)
- `host` from `.hostname`, independent of user configuration

/cc 
/assign 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-6734
- Enhancement proposal:

### DIFF
--- a/docs/features/logforwarding/outputs/splunk-forwarding.adoc
+++ b/docs/features/logforwarding/outputs/splunk-forwarding.adoc
@@ -1,6 +1,7 @@
-=== Steps to forward to Splunk HTTP Event Collector (HEC)
+== Steps to forward to Splunk HTTP Event Collector (HEC)
 
-(https://dev.splunk.com/enterprise/docs/devtools/httpeventcollector/)
+More information about format of Splunk HTTP Event you can read in this document:
+https://docs.splunk.com/Documentation/Splunk/latest/Data/FormateventsforHTTPEventCollector[Format events for HTTP Event Collector]
 
 . Create a secret containing your `hecToken` using the following command:
 +
@@ -30,11 +31,16 @@ spec:
       type: splunk
       splunk:
         authentication:
-          token:
+          token: # <1>
             key: hecToken
             secretName: splunk-secret
-        url: 'http://example-splunk-hec-service:8088'
-        index: '{.log_type||"undefined"}'
+        url: 'http://example-splunk-hec-service:8088' # <2>
+        index: '{.log_type||"main"}' # <3>
+        source: '{.log_source||"undefined"}' # <4>
+        indexedFields: ['.log_type', '.log_source'] # <5>
+        payloadKey: '.kubernetes' # <6>
+        tuning:
+            compression: gzip # <7>
   pipelines:
     - name: my-logs
       inputRefs:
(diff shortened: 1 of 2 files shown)


docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -43,5 +48,113 @@ spec:
       outputRefs:
         - splunk-receiver
 ----
+1. Token points to the secret containing the Splunk HEC token used for authenticating requests.
+2. The base URL of the Splunk instance.
+3. Optional. The name of the index to send events to. If not specified, the default index defined within Splunk is used.
+4. Optional. Compression configuration, available to are: `none`, `gzip`. Default is `none`.
+5. Optional. The source of events sent to this sink.
```
jcantrill, line 55: Mention supports templating

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -43,5 +48,113 @@ spec:
...
+6. Optional. Fields to be added to Splunk index.
```
jcantrill, line 56: May need to mention it supports templating

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -30,11 +31,15 @@ spec:
...
       splunk:
         authentication:
-          token:
+          token: # <1>
             key: hecToken
             secretName: splunk-secret
-        url: 'http://example-splunk-hec-service:8088'
-        index: '{.log_type||"undefined"}'
+        url: 'http://example-splunk-hec-service:8088' # <2>
+        index: '{.log_type||"undefined"}' # <3>
```
jcantrill, line 38: We talked about adding `payloadKey` like syslog to allow setting the splunk `event` to something else.  This would imply the `sourceType` is dependent upon the payload.   Should we do this?

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -43,5 +48,113 @@ spec:
...
+In Splunk, the `source` field typically identifies the origin of a log event.
+To ensure consistency and meaningful categorization, the `source` value can be dynamically derived from the `log_type` and `log_source` fields,
+following Cluster Log Forwarder's conventions.
+
+Below the table with default value depends on log_type and log_source will be used if not set in configuration.
+
+|===
+| |pass:[Infrastructure Journal <br/><em>log_type:infrastructure<br/>log_source:node</em>]|pass:[Infrastructure/Application Container<br/><em>log_type:infrastructure\|application<br/>log_source:container</em>]|pass:[Audit<br/><em>log_type:audit<br/>log_source:auditd\|ovn\|openshiftAPI\|kubeAPI<em/>]|Note
+|`source`|SYSLOG_IDENTIFIER*|ns_name_podName_containerName|.log_source|
+|===
```
jcantrill, line 68: let's add sourceType and payloadkey here as well

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -30,11 +31,15 @@ spec:
...
             key: hecToken
             secretName: splunk-secret
-        url: 'http://example-splunk-hec-service:8088'
-        index: '{.log_type||"undefined"}'
+        url: 'http://example-splunk-hec-service:8088' # <2>
+        index: '{.log_type||"undefined"}' # <3>
+        tuning:
+            compression: gzip # <4>
+            source: '{.log_source}' # <5>
+            indexedFields: ['.log_type', '.log_source'] # <6>
```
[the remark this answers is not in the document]
vparfonov: I don't think we need to support templating here, because we will introduce the ability to use existing fields as index, after remapping by known rules. Not expecting a lot of variation with creating new fields because, it also required to add a field extractor on the Splunk side. At least this is my mind, we can discuss it.

docs/features/logforwarding/outputs/splunk-forwarding.adoc
vparfonov: Don't see the reason todo it, also Vector not support such configuration parameter

docs/features/logforwarding/outputs/splunk-forwarding.adoc
vparfonov: @jcantrill I forgot add the `host_key`. Will we expose ithttps://vector.dev/docs/reference/configuration/sinks/splunk_hec_logs/#host_key

vparfonov: As i understand, we decided to NOT expose `sourceType`, no?

docs/features/logforwarding/outputs/splunk-forwarding.adoc
jcantrill: Vector does support it by way of setting "message" to whatever value you want.  It will map that to the 'event'.  I'll defer in pushing on payload key mostly because I'm not certain we need to allow changing it.  The feature does not request it.  I'll defer to other opinions and input if there is a larger need

docs/features/logforwarding/outputs/splunk-forwarding.adoc
jcantrill: >As i understand, we decided to NOT expose sourceType, no?

If we are not exposing 'payloadkey' to populate the event then I see no reason to expose 'sourceType'.  We should always set it to '_json' like the example in the splunk documentation.  Additionally we should ensure the table we provide in this document identifies the value

> @jcantrill I forgot add the host_key. Will we expose it https://vector.dev/docs/reference/configuration/sinks/splunk_hec_logs/#host_key

IMO there is no need to expose this in our API.  We should either configure the key to be what we already use or set a field to the value of "hostname".  I would prefer to just configure it to be 'hostname'

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -43,5 +49,113 @@ spec:
...
+In Splunk, `indexed_fields` are fields that are extracted at *index time*, rather than at *search time*. This means their values are stored directly in the index alongside the raw event data, allowing for *faster search performance* on those fields.
+
+However, because `indexed_fields` increase storage usage, they should be used *sparingly* and only for *high-value fields* that provide significant search benefits.
+
+Since logs processed by Vector are typically *well-structured JSON*, Splunk can dynamically extract fields at search time. As a result, the use of `indexed_fields` is *not necessary by default*. That said, they may be valuable in *specific use cases*, such as *large datasets with frequent queries on specific fields*.
+In such cases, users can configure the required fields manually.
+
+==== Limitations
+
+* Only *top-level fields* are supported.
```
jcantrill, line 82: If we are adding templating which means we will copy a value to the root, there is no need to specify that here

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -43,5 +49,113 @@ spec:
...
+4. `source` - Optional. The source of events sent to this sink. This supports template syntax to allow dynamic per-event values.
+5. `indexedFields` - Optional. Fields to be added to Splunk index.
+6. `compression` - Optional. Compression configuration, available to are: `none`, `gzip`. Default is `none`.
+
+=== `source`
+In Splunk, the `source` field typically identifies the origin of a log event.
+To ensure consistency and meaningful categorization, the `source` value can be dynamically derived from the `log_type` and `log_source` fields,
+following Cluster Log Forwarder's conventions.
+
+Below the table with default value depends on log_type and log_source will be used if not set in configuration.
```
jcantrill, line 64: Let's expand this table to identify the 5?? splunk metadata fields and identify the values to which they will be set.

docs/features/logforwarding/outputs/splunk-forwarding.adoc
vparfonov: > Additionally we should ensure the table we provide in this document identifies the value

> IMO there is no need to expose this in our API. We should either configure the key to be what we already use or set a field to the value of "hostname". I would prefer to just configure it to be 'hostname'

https://github.com/openshift/cluster-logging-operator/pull/2979/files#diff-46e6d1852418f35b4db6dde8b446604e9d144d0cb4838d378ed8e2609156bca8R168

docs/features/logforwarding/outputs/splunk-forwarding.adoc
vparfonov: I think we need to add some information why we add such remapping

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -30,11 +31,16 @@ spec:
...
         authentication:
-          token:
+          token: # <1>
             key: hecToken
             secretName: splunk-secret
-        url: 'http://example-splunk-hec-service:8088'
-        index: '{.log_type||"undefined"}'
+        url: 'http://example-splunk-hec-service:8088' # <2>
+        index: '{.log_type||"undefined"}' # <3>
+        source: '{.log_source}' # <4>
```
jcantrill, line 39: Update with a fallback which is required by our templating format

docs/features/logforwarding/outputs/splunk-forwarding.adoc
jcantrill: From a user's perspective, our API is a "path" which does not necessarily have to be at the root.  We intend to flatten the path to be the key name with underscores.  ClusterLogForwarder this statement is inaccurate even though the implementation requires it at the root

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -43,5 +49,135 @@ spec:
...
+
+However, because `indexed_fields` increase storage usage, they should be used *sparingly* and only for *high-value fields* that provide significant search benefits.
+
+Since logs processed by Vector are typically *well-structured JSON*, Splunk can dynamically extract fields at search time. As a result, the use of `indexed_fields` is *not necessary by default*. That said, they may be valuable in *specific use cases*, such as *large datasets with frequent queries on specific fields*.
+In such cases, users can configure the required fields manually.
+
+==== Limitations
+
+* Only *top-level fields* are supported.
+* Field names may only include *letters, digits, dashes (`-`)*, and *underscores (`_`)*.
```
jcantrill, line 76: This is a restriction of the implementation but not the API.  We should state what is allowed in the API.  Additionally,  we should define how we transform the field name.  That does not go here since the header is "limitation".  Maybe we should reconsider the header since the limitations are at the implementation level but this is really defining specifics about "indexed fields"

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -43,5 +49,135 @@ spec:
...
+* Only *string values* are allowed.
```
jcantrill, line 77: We should state this as a transformation of the value since this is what we will do.

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -43,5 +49,135 @@ spec:
...
+
+==== Extended Support via Remap
+
+To provide more flexibility, we extend the functionality with a *custom remap function*.
```
jcantrill, line 81: I would remove the wording of Remap since that is an implementation detail of vector.  I think this section can be combined with the previous to identify the outcome (with the examples) of your choices in using the API:

* flattened key name
* string value and how objects or numbers will be converted
* replace of a top level key if it is indexed (i.e. kubernetes, openshift become json strings)

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -43,5 +49,135 @@ spec:
...
+|===
+| |pass:[Infrastructure Journal <br/><em>log_type:infrastructure<br/>log_source:node</em>]|pass:[Infrastructure/Application Container<br/><em>log_type:infrastructure\|application<br/>log_source:container</em>]|pass:[Audit<br/><em>log_type:audit<br/>log_source:auditd\|ovn\|openshiftAPI\|kubeAPI<em/>]|Note
+|`index`|||| not configured by default
+|`source`|SYSLOG_IDENTIFIER|ns_name_podName_containerName|.log_source|
+|`indexedFields`|||| not configured by default
+|`sourceType`|`_json`|`_json`|`_json`| *not configurable*
+|`host`|`.hostname`|`.hostname`|`.hostname`|*not configurable*
 
-NOTE: This will forward logs to the log type of the message.  The default index of the splunk server configuration is used when 'index' is not defined
+|===
```
jcantrill, line 183: Let's add 'message' to this table and state that it is the "ViaQ envelope which is a JSON string".  Please confirm this is currently what we send

vparfonov: @jcantrill you mean `event`?

jcantrill: My original statement I meant message but given this is for splunk and that is how users of splunk will associate the information in the splunk API, I agree 'event' is a better choice.  Since we talked about allowing this to be configurable then maybe we should also call it "EventKey" in our API instead of "PayloadKey" so the words are consistent with splunk's API.  @cahartma thoughts?  Does it make more sense to make our API consistent (i.e. Payload Key is what we use in syslog) or make it consistent with the sink (i.e. splunk users know it as 'event'0?

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -43,5 +49,146 @@ spec:
...
+
+
+NOTE: Use `payloadKey` carefully. Selecting a single field as the payload may cause other important information in the log to be dropped, potentially leading to inconsistent or incomplete log events.
+
+
+=== `sourceType` and `host`
+
+The `sourceType` and `host` fields are *not configurable* through the Cluster Log Forwarder API. They are set automatically as described below, independent of the `.log_type` or `.log_source` values.
+
+* `sourceType`: Set to `_json`
```
jcantrill, line 176: This statement will need to change depending upon the evaluation of what is sent to event:

* event == object:  _json
* event == string:  _text (??)
* event == array:   _json(??)
* event == numeric : _text(?? do we convert to a string)

PR #2990, opened 2025-03-18
LOG-4445: Forwarder must use hosted cluster_id on HCP cluster

Clee2691:
### Description
This PR updates https://github.com/openshift/cluster-logging-operator/pull/2280 to use the `hypershift API` instead of the unstructured object.

The `github.com/openshift/hypershift/api` package is added as a dependency.

New permissions are added to the role that CLO's `serviceAccount` is bound to: 
```
- apiGroups:
  - hypershift.openshift.io
  resources:
  - hostedcontrolplanes
  verbs:
  - list
  - get
```

Lastly, this PR adds the `VRL` transform that `kubeAPI` audit logs should go through as it was missing for http receiver inputs.
 
/cc @cahartma @vparfonov @alanconway 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-4445
- Related PRs
  - https://github.com/openshift/cluster-logging-operator/pull/2274


jcantrill (commented):
/approve

jcantrill:
/hold

pending feedback from QE team

alanconway (approved):
/lgtm

anpingli:
QE Test pass
-  The hcp cluster id is used when CLF is deployed in control panel namespace.
-  The management cluster ID is used when CLF is deployed in the other namespaces.

cmd/main.go
```diff
@@ -68,6 +68,7 @@ func init() {
 	utilruntime.Must(monitoringv1.AddToScheme(scheme))
 	utilruntime.Must(configv1.AddToScheme(scheme))
 	utilruntime.Must(securityv1.AddToScheme(scheme))
+	utilruntime.Must(hypershiftv1beta1.AddToScheme(scheme))
```
jcantrill, line 71: @Clee2691 @anpingli have we tested this in a non HCP cluster to ensure the operator does not fail to start because maybe it is missing the CRD for the hypershift apis?

jcantrill (commented):
/hold

cmd/main.go
Clee2691: Isn't that what the e2e tests basically confirm? CLO is able to be installed and send logs even with the hypershift scheme added. On each reconcile, it will try to fetch the `clusterID` and `version` so if it failed, none of those tests would have passed.

anpingli: Tested on regular cluster, the clusterID present correctly. No regression issue.

PR #2992, opened 2025-03-18
LOG-6084: Replace functional-benchmark plot library

jcantrill:
### Description
This PR:
* replaces the functional-benchmark plot library in support of testing [LOG-6084](https://issues.redhat.com//browse/LOG-6084)
* Modifies the http receiver to do some of the work previously done post processing
* Improves the report generation performance
* Removes console report
* Modifies the lost message graph to be a percentage

Example of results https://github.com/jcantrill/collector-experimental/blob/release-6.2/6.2.0/10m-01-5000lps-0512b-01/readme.md


internal/cmd/functional-benchmarker/main.go
```diff
@@ -12,18 +12,23 @@ import (
...
-	log "github.com/ViaQ/logerr/v2/log/static"
 	"github.com/openshift/cluster-logging-operator/internal/cmd/functional-benchmarker/config"
 	"github.com/openshift/cluster-logging-operator/internal/cmd/functional-benchmarker/reports"
 	"github.com/openshift/cluster-logging-operator/internal/cmd/functional-benchmarker/runners"
 	"github.com/openshift/cluster-logging-operator/internal/cmd/functional-benchmarker/stats"
 	"github.com/openshift/cluster-logging-operator/internal/utils"
 )
 
+var (
+	log = utils.InitLogger("functional-benchmarker")
```
Clee2691: Why aren't you using the `InitStaticLogger()` function here when you use it in previous files?

internal/cmd/functional-benchmarker/main.go
```diff
@@ -12,18 +12,23 @@ import (
...
+)
+
 // HACK - This command is for development use only
 func main() {
 	utils.InitLogger("functional-benchmarker")
```
Clee2691: Another instance of the logger?

internal/cmd/functional-benchmarker/reports/gonumplot/init.go
```diff
@@ -0,0 +1,9 @@
+package gonumplot
+
+import (
+	"github.com/openshift/cluster-logging-operator/internal/utils"
+)
+
+var (
+	log = utils.InitLogger("gonumplut")
```
Clee2691, line 8: `InitStaticLogger()`?

internal/cmd/functional-benchmarker/runners/cluster/init.go
```diff
@@ -0,0 +1,9 @@
+package cluster
+
+import (
+	"github.com/openshift/cluster-logging-operator/internal/utils"
+)
+
+var (
+	log = utils.InitLogger("cluster")
```
Clee2691, line 8: `InitStaticLogger()`?

internal/cmd/functional-benchmarker/stats/init.go
```diff
@@ -0,0 +1,9 @@
+package stats
+
+import (
+	"github.com/openshift/cluster-logging-operator/internal/utils"
+)
+
+var (
+	log = utils.InitLogger("stats")
```
Clee2691, line 8: `InitStaticLogger()`?

internal/cmd/functional-benchmarker/main.go
jcantrill: This one should be static because it would initialize the logger for the entire process that begin in main.

internal/cmd/functional-benchmarker/main.go
jcantrill: Removing this since we are initializing one with var

internal/cmd/functional-benchmarker/reports/gonumplot/init.go
jcantrill: I believe this should not be static as we do not want to replace the logger for the entire process.  This is similar to creating a logger for the package as part of the runtime controllers.  I'm not fully onboard with how loggers should be treated.

internal/cmd/functional-benchmarker/runners/cluster/init.go
jcantrill: same here as this becomes a package level logger..

internal/cmd/functional-benchmarker/stats/init.go
jcantrill: same

PR #2999, opened 2025-03-26
LOG-6860: Support Splunk Metadata keys in ClusterLogForwarder

vparfonov:
### Description
This PR addressed to extend the `ClusterLogForwarder` to support `Splunk` `metadata fields` when forwarding logs
so that we enable better integration with `Splunk`.

/cc @Clee2691 @cahartma 
/assign @jcantrill 

/cherry-pick 

### Links

- Depending on PR(s):
- Bugzilla:
- Github issue:
- JIRA: https://issues.redhat.com/browse/LOG-6860
- Enhancement proposal: https://github.com/openshift/cluster-logging-operator/pull/2979

### DIFF
--- a/api/observability/v1/output_types.go
+++ b/api/observability/v1/output_types.go
@@ -1045,6 +1045,53 @@ type Splunk struct {
 	// +kubebuilder:validation:Pattern:=`^(([a-zA-Z0-9-_.\/])*(\{(\.[a-zA-Z0-9_]+|\."[^"]+")+((\|\|)(\.[a-zA-Z0-9_]+|\.?"[^"]+")+)*\|\|"[^"]*"\})*)*$`
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Index",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
 	Index string `json:"index,omitempty"`
+
+	// IndexedFields are the list of fields to be indexed by Splunk, increase storage usage, they should be used sparingly and only for high-value fields that provide significant search benefits.
+	// Nested fields are flattened into top-level fields.
+	// Field paths are joined using dot notation, and unsupported characters are replaced with underscores (_).
+	// Non-string values are automatically converted to strings (e.g., 3 → "3", true → "true").
+	// Object values serialized as JSON strings (e.g., { status: 200 } → "{\"status\":200}").
+	//
+	// Examples: [`.kubernetes`, `.log_type`, '.kubernetes.labels.foobar', `.kubernetes.labels."foo-bar/baz"`]
+	// +nullable
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Indexed Fields"
+	IndexedFields []FieldPath `json:"indexedFields,omitempty"`
+
+	// Source identifies the origin of a log event.
+	// The Source can be a combination of static and dynamic values consisting of field paths followed by `||` followed by another field path or a static value.
+	// A dynamic value is encased in single curly brackets `{}` and MUST end with a static fallback value separated with `||`.
+	// Static values can only contain alphanumeric characters along with dashes, underscores, dots and forward slashes.
+	// If not specified will be detected according to .log_source and .log_type value.
+	// Details see in: docs/features/logforwarding/outputs/splunk-forwarding.adoc
+	//
+	// Example:
+	//
+	//  1. foo-{.bar||"none"}
+	//
+	//  2. {.foo||.bar||"missing"}
+	//
+	//  3. foo.{.bar.baz||.qux.quux.corge||.grault||"nil"}-waldo.fred{.plugh||"none"}
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Pattern:=`^(([a-zA-Z0-9-_.\/])*(\{(\.[a-zA-Z0-9_]+|\."[^"]+")+((\|\|)(\.[a-zA-Z0-9_]+|\.?"[^"]+")+)*\|\|"[^"]*"\})*)*$`
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Source",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	Source string `json:"source,omitempty"`
+
+	// PayloadKey specifies record field to use as payload.
+	// The PayloadKey must be a single field path.
+	//
+	// Field paths must only contain alphanumeric and underscores. Any field with other characters must be quoted.
+	//
+	// By default, payloadKey is not set, which means the complete log record is forwarded as the payload.
+	// Use payloadKey carefully. Selecting a single field as the payload may cause other important information in the
+	// log to be dropped, potentially leading to inconsistent or incomplete log events.
+	//
+	// Examples: `.kubernetes`, `.log_type`, '.kubernetes.labels.foobar', `.kubernetes.labels."foo-bar/baz"`
+	//
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Payload Key",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	PayloadKey FieldPath `json:"payloadKey,omitempty"`
 }
 
 // SyslogRFCType sets which RFC the generated messages conform to.
--- a/internal/generator/vector/helpers/helpers.go
+++ b/internal/generator/vector/helpers/helpers.go
@@ -5,7 +5,9 @@ import (
 	v1 "github.com/openshift/cluster-logging-operator/api/observability/v1"
 	"github.com/openshift/cluster-logging-operator/internal/constants"
 	"path/filepath"
+	"regexp"
 	"sort"
+	"strconv"
 	"strings"
 	"sync"
 
@@ -13,6 +15,9 @@ import (
 )
 
 const VectorSecretID = "kubernetes_secret"
+
+// Match quoted strings like "foo" or "foo/bar-baz"
+var quoteRegex = regexp.MustCompile(`^".+"$`)
 
 var (
 	Replacer         = strings.NewReplacer(" ", "_", "-", "_", ".", "_")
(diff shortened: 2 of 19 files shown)


api/observability/v1/output_types.go
```diff
@@ -1045,6 +1045,49 @@ type Splunk struct {
...
+	//Fields to be added to Splunk index
+	//
+	//Examples: [`.kubernetes`, `.log_type`, '.kubernetes.labels.foobar', `.kubernetes.labels."foo-bar/baz"`]
+	//
+	// +nullable
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Indexed Fields"
+	IndexedFields []FieldPath `json:"indexedFields,omitempty"`
+
+	// Source identifies the origin of a log event.
```
jcantrill, line 1058: Comment on how we "default" this value if not specified. It may be "good enough" to say to "see  clo document" for behavior when not defined

api/observability/v1/output_types.go
```diff
@@ -1045,6 +1045,49 @@ type Splunk struct {
...
+	//  2. {.foo||.bar||"missing"}
+	//
+	//  3. foo.{.bar.baz||.qux.quux.corge||.grault||"nil"}-waldo.fred{.plugh||"none"}
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Pattern:=`^(([a-zA-Z0-9-_.\/])*(\{(\.[a-zA-Z0-9_]+|\."[^"]+")+((\|\|)(\.[a-zA-Z0-9_]+|\.?"[^"]+")+)*\|\|"[^"]*"\})*)*$`
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Source",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	Source string `json:"source,omitempty"`
+
+	// PayloadKey specifies record field to use as payload.
```
jcantrill, line 1079: Same comment here about linking or stating the behavior when not specified

internal/generator/vector/helpers/helpers.go
```diff
@@ -80,3 +85,71 @@ func SecretFrom(secretKey *v1.SecretReference) string {
 	}
 	return ""
 }
+
+// generateQuotedPathSegmentArrayStr generates the final string of the array of array of path segments
```
jcantrill, line 89: Should be capitalized

internal/generator/vector/helpers/helpers.go
```diff
@@ -80,3 +85,71 @@ func SecretFrom(secretKey *v1.SecretReference) string {
...
+// and array of flattened path with replaced not allowed symbols to feed into VRL
+// E.g
+// [.kubernetes.namespace_labels."bar/baz0-9.test"] -> ([["kubernetes","namespace_labels","bar/baz0-9.test"]], ["_kubernetes_namespace_labels_bar_baz0-9_test"])
+func GenerateQuotedPathSegmentArrayStr(fieldPathArray []v1.FieldPath) (string, string) {
```
jcantrill, line 93: Where are the unit tests to support this complicated function?

internal/generator/vector/output/splunk/splunk_sink.toml
```diff
@@ -10,14 +10,33 @@ if err != null {
 	.@timestamp = ts
 }
 
+'''
```
jcantrill, line 13: Please modify the timestamp logic to use the "timestamp" field in lieu of '@timestamp' or maybe this is a different refactoring so its less changes to the toml files

internal/generator/vector/output/splunk/splunk_sink.toml
```diff
@@ -10,14 +10,33 @@ if err != null {
...
+if ._internal.log_type == "infrastructure" && ._internal.log_source == "node" {
+    ._internal.splunk.source = to_string!(._internal.systemd.u.SYSLOG_IDENTIFIER || "")
+}
+if ._internal.log_source == "container" {
+   	._internal.splunk.source = join!([._internal.kubernetes.namespace_name, ._internal.kubernetes.pod_name, ._internal.kubernetes.container_name], "_")
+}
+if ._internal.log_type == "audit" {
+   ._internal.splunk.source = ._internal.log_source
+}
+._internal.splunk.sourcetype = "_json"
```
jcantrill, line 28: This should map to the type based upon the payloadkey

internal/validations/observability/outputs/splunk.go
```diff
@@ -0,0 +1,19 @@
...
+	obs "github.com/openshift/cluster-logging-operator/api/observability/v1"
+	"strings"
+)
+
+func ValidateSplunk(spec obs.OutputSpec) (results []string) {
+	if spec.Splunk.PayloadKey != "" && spec.Splunk.IndexedFields != nil {
+		payload := string(spec.Splunk.PayloadKey)
+		for _, v := range spec.Splunk.IndexedFields {
+			if !strings.HasPrefix(string(v), payload) {
+				results = append(results, fmt.Sprintf("Indexed field: %s not part of payload: %s", v, payload))
```
jcantrill, line 14: are we trying to restrict they payload field from being part of indexed fields?  If so then the error should read that it can not or is not allowed

internal/validations/observability/outputs/splunk_test.go
```diff
@@ -0,0 +1,65 @@
...
+import (
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	obs "github.com/openshift/cluster-logging-operator/api/observability/v1"
+)
+
+var _ = Describe("validating Splunk", func() {
+	Context("#ValidateSplunkMetadata", func() {
+
+		It("should fail validation if meet indexed field not in payload", func() {
```
jcantrill, line 12: If i read this validation correctly, the "event" is always the entire vector log event (e.g. `.`)?  And you can only index fields that are part of what gets added to `.`?  The splunk docs would suggest otherwise but I guess vector's encoding does not allow defining the "event" content separate from "indexed field values"?  Maybe the "indexed" fields at the root level would need to be added to the "exclude" fields?

test/framework/functional/output_splunk.go
```diff
@@ -167,3 +184,47 @@ func (f *CollectorFunctionalFramework) ReadAppLogsByIndexFromSplunk(namespace, n
...
+		if results == "" || err != nil {
+			return false, err
+		}
+		return true, nil
+	})
+
+	if err != nil {
+		return "", err
+	}
+	println("\n\n\n\n\n\n\n")
```
jcantrill, line 204: should these be log statements?

test/functional/outputs/splunk/forward_to_splunk_test.go
```diff
@@ -194,4 +216,219 @@ var _ = Describe("Forwarding to Splunk", func() {
...
+			outputTestLog := appLogs[0]
+			Expect(outputTestLog.LogType).To(Equal(string(obs.InputTypeApplication)))
+			Expect(outputTestLog.Openshift.Sequence).To(BeEmpty()) //.openshift.sequence must be removed because it was set as root-level field openshift_sequence
+			Expect(outputTestLog.Openshift.ClusterID).To(Equal("functional"))
+			_, exist := outputTestLog.Kubernetes.Annotations["openshift.io/scc"]
+			Expect(exist).To(BeFalse()) // .kubernetes.annotations."openshift.io/scc" must be removed because was set as root-level field kubernetes_annotations_openshift_io_scc
+
+			//checking Splunk index fields with 'index=${index} | stats count by ${field_name}'
+			//not direct checking to be sure fields sent as indexed
+			//if filed is indexed will return counts 2*logs_count or logs_count otherwise
```
jcantrill, line 261: if field

test/functional/outputs/splunk/forward_to_splunk_test.go
```diff
@@ -194,4 +216,219 @@ var _ = Describe("Forwarding to Splunk", func() {
...
+		},
+			Entry("should send application logs with dynamic source field", `{.log_type||"missing"}`, "application", obs.InputTypeApplication),
+			Entry("should send application logs with static + dynamic source field", `foo-{.log_type||"missing"}`, "foo-application", obs.InputTypeApplication),
+			Entry("should send application logs with static + label with dot/slash source field", `foo-{.kubernetes.labels."slash/test.dot"||"missing"}`, "foo-application", obs.InputTypeApplication),
+			Entry("should send application logs with static + fallback value's source field", `foo-{.missing||"application"}`, "foo-application", obs.InputTypeApplication),
+			Entry("should send application logs with default source ", "", "", obs.InputTypeApplication),
+			Entry("should send audit logs with default source ", "", "", obs.InputTypeAudit),
+			Entry("should send journal logs with default source ", "", "", obs.InputTypeInfrastructure))
+
+		DescribeTable("with user defined payloadKey", func(payloadKey, expSourceType string, expression *regexp.Regexp) {
```
jcantrill, line 386: Please add a test for the audit log defined in the feature request and see how closely we can get to matching the request.  I don't believe we wish to test indexing every annotation but I'm interested to see how we populate "event" with the audit event message and maybe index a few other fields

internal/validations/observability/outputs/splunk_test.go
vparfonov: This checks not related to the Splunk requirements directly, more about our procedure how we remap/calculate fields. If indexed fields will be outside the given payload, we can't remap it correctly. So this warning to check given settings. I think adding "indexed" fields to exclude filter not make sens, we have `.log_source` and `.log_type` these should be enough.

internal/validations/observability/outputs/splunk.go
vparfonov: Indexed fields can't be outside of a given payload

internal/generator/vector/output/splunk/splunk.go
```diff
@@ -15,12 +17,78 @@ import (
...
+	value = get!(., payloadKey) 
+	if !is_null(value) {
+        internal = ._internal
+        . = {}
+        . = set!(., payloadKey, value)
+        ._internal = internal
+		if is_object(value) {
+			._internal.splunk.sourcetype = "_json"
+		} else {
+			._internal.splunk.sourcetype = "generic_single_line"
```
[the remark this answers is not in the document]
vparfonov: Maybe, if s_array() also should be "_json"? 
@jcantrill

internal/generator/vector/output/splunk/splunk.go
```diff
@@ -15,12 +17,78 @@ import (
...
+// If payload is object will be set 'sourcetype':"_json",
+// 'sourcetype': "generic_single_line" otherwise
+var payloadKeyTmpl = `
+payloadKey = %s
+if !is_null(payloadKey) {
+	value = get!(., payloadKey) 
+	if !is_null(value) {
+        internal = ._internal
+        . = {}
+        . = set!(., payloadKey, value)
```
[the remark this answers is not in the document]
vparfonov: yes, it set to 'event' field

internal/generator/vector/output/splunk/splunk_sink.toml
vparfonov: yes, lets do it in separate PR

internal/generator/vector/output/splunk/splunk_sink.toml
vparfonov: this used if no `payload settings`

internal/validations/observability/outputs/splunk.go
jcantrill: I don't believe this to be a reasonable restriction even though it is the behavior of the encoder.  Use the audit event example from the feature request where they are indexing annotations but the "event"/payload field comes from a sibling node in the original message.  We should be able to copy the fields in question to the "root" of the message so the encoder picks them up while setting the payload to something else.

internal/validations/observability/outputs/splunk_test.go
jcantrill: See my previous comment about the validation

api/observability/v1/output_types.go
```diff
@@ -1045,6 +1045,53 @@ type Splunk struct {
...
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Indexed Fields"
+	IndexedFields []FieldPath `json:"indexedFields,omitempty"`
+
+	// Source identifies the origin of a log event.
+	// The Source can be a combination of static and dynamic values consisting of field paths followed by `||` followed by another field path or a static value.
+	// A dynamic value is encased in single curly brackets `{}` and MUST end with a static fallback value separated with `||`.
+	// Static values can only contain alphanumeric characters along with dashes, underscores, dots and forward slashes.
+	// If not specified will be detected according to .log_source and .log_type value.
+	// Details see in: docs/features/logforwarding/outputs/splunk-forwarding.adoc
```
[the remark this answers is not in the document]
vparfonov: yes, i will change this after we will release

internal/validations/observability/outputs/splunk.go
vparfonov: Maybe you are right, this checks are overhead and not clear. I will remove this

internal/generator/vector/output/splunk/splunk_sink_with_indexed_fields.toml
```diff
@@ -0,0 +1,64 @@
...
+    if !is_null(value) {
+        new_key = replace(join!(field,"_"), r'[\./]', "_")
+        if !is_string(value) {
+          if is_object(value) {
+            value = encode_json(value)
+          } else {
+            value = to_string!(value)
+          }
+        }
+        . = remove!(., field, true)
```
jcantrill, line 43: looking at our document I don't see where we state we remove the indexed field from the payload

PR #3005, opened 2025-04-01
LOG-6879: fix operator permissions to update SCC

jcantrill:
### Description
forward port of #3003


jcantrill:
cc @Clee2691 @vparfonov

PR #3008, opened 2025-04-04
LOG-6790: Update cw forwarding to use credential file and profiles

Clee2691:
### Description
This PR adds the ability for the `ClusterLogForwarder` to utilize multiple distinct `role_arns` to authenticate to amazon's Cloudwatch on an STS enabled openshift cluster.

Changes:
1. Removes validation for multiple `role_arns`.
2. Adds validation to ensure valid `role_arns`.
3. Creates and reconciles a credentials file as a `configMap` with multiple profiles mounted in `/var/run/ocp-collector/config/<CONFIGMAP NAME>/credentials`.
4. The `cloudwatch` sink will specify the credentials file along with the profile name when `iamRole` is used as authentication.
5. No longer sets environment variables for `cloudwatch`.
6. Added documentation for an example workflow.

This PR depends on changes to the RH Vector build to include a fix for the `region` when authenticating to AWS.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links

- Depending on PR(s): https://github.com/ViaQ/vector/pull/197
- JIRA: https://issues.redhat.com/browse/LOG-6790


internal/collector/cloudwatch.credentials.tmpl
```diff
@@ -0,0 +1,7 @@
+{{- range . -}}
```
jcantrill, line 1: Please  refactor this into a "collector/cloudwatch" package

internal/collector/cloudwatch.go
```diff
@@ -1,57 +1,93 @@
...
 )
 
-// Add volumes and env vars if output type is cloudwatch and role is found in the secret
-func addWebIdentityForCloudwatch(collector *v1.Container, forwarderSpec obs.ClusterLogForwarderSpec, secrets observability.Secrets) {
-	if secrets == nil {
-		return
+type CloudwatchWebIdentity struct {
+	Name    string
+	RoleARN string
+	Token   string
```
jcantrill, line 31: Let's call this "TokenFile" to distiguish it from an actual token. Maybe consider "WebIdentityTokenFile"

internal/collector/cloudwatch.go
```diff
@@ -1,57 +1,93 @@
...
+	webIds := GatherAWSWebIdentities(outputs, secrets)
 
+	// No CW outputs
+	if webIds == nil {
+		return "", nil
+	}
+
+	// Execute Go template to generate credential profile(s)
+	w := &strings.Builder{}
+	err := CloudwatchCredentialsTemplate.Execute(w, webIds)
```
jcantrill, line 67: Will this return an array of  the template output?

internal/collector/cloudwatch.go
```diff
@@ -1,57 +1,93 @@
...
-	if secrets == nil {
-		return
+type CloudwatchWebIdentity struct {
+	Name    string
+	RoleARN string
+	Token   string
+}
+
+// ReconcileAWSCredentialsConfigMap reconciles a configmap with credential profile(s) for Cloudwatch output(s).
+func ReconcileAWSCredentialsConfigMap(k8sClient client.Client, reader client.Reader, namespace, name string, outputs []obs.OutputSpec, secrets observability.Secrets, configMaps map[string]*corev1.ConfigMap, owner metav1.OwnerReference) (*corev1.ConfigMap, error) {
```
jcantrill, line 35: Do we need the uncached reader other then k8sClient here?  Usually that is only an issue when trying to read a cluster resource that we don't own

internal/constants/constants.go
```diff
@@ -24,8 +24,9 @@ const (
 	AWSSecretAccessKey           = "aws_secret_access_key" //nolint:gosec
 	AWSAccessKeyID               = "aws_access_key_id"
 	AWSRoleSessionName           = "cluster-logging" // identifier for role logging session
-	AWSCredentialsKey            = "credentials"     // credrequest key to check for sts-formatted secret
-	AWSWebIdentityRoleKey        = "role_arn"        // manual key to check for sts-formatted secret
+	AWSCredentialsConfigMapName  = "aws-creds-configmap"
```
jcantrill, line 27: We are creating a "configmap".  We can drop "configmap" from the name.  It will help with my concerns about the name exceeding the name limits

internal/controller/observability/collector.go
```diff
@@ -72,6 +72,17 @@ func ReconcileCollector(context internalcontext.ForwarderContext, pollInterval,
...
 
+	credCm, err := collector.ReconcileAWSCredentialsConfigMap(context.Client, context.Reader, context.Forwarder.Namespace, context.Forwarder.Name, context.Forwarder.Spec.Outputs, context.Secrets, context.ConfigMaps, ownerRef)
+	if err != nil {
+		log.V(9).Error(err, "collector.ReconcileAWSProfileConfig")
+		return err
+	}
+
+	// Add generated credentials configmap to contexts to be mounted in pod
+	if credCm != nil {
+		context.ConfigMaps[credCm.Name] = credCm
```
jcantrill, line 83: Chance of collision wiht a spec'd secret name.   We may need an early validation to ensure there is none

internal/generator/vector/output/cloudwatch/cloudwatch.go
```diff
@@ -107,17 +109,22 @@ func sink(id string, o obs.OutputSpec, inputs []string, secrets observability.Se
...
 }
 
-func authConfig(auth *obs.CloudwatchAuthentication, secrets observability.Secrets) Element {
+func authConfig(outputName string, auth *obs.CloudwatchAuthentication, secrets observability.Secrets, options Options) Element {
 	authConfig := NewAuth()
 	if auth != nil && auth.Type == obs.CloudwatchAuthTypeAccessKey {
 		authConfig.KeyID.Value = vectorhelpers.SecretFrom(&auth.AWSAccessKey.KeyId)
 		authConfig.KeySecret.Value = vectorhelpers.SecretFrom(&auth.AWSAccessKey.KeySecret)
+	} else if auth != nil && auth.Type == obs.CloudwatchAuthTypeIAMRole {
+		if forwarderName, ok := options[OptionForwarderName]; ok {
```
jcantrill, line 124: Please look at utility methods on "options" .  I believe there is a function to make the retrieval and fetch/cast to a string cleaner

internal/generator/vector/conf/conf.go
```diff
@@ -70,6 +71,7 @@ func Conf(secrets map[string]*corev1.Secret, clfspec obs.ClusterLogForwarderSpec
 
 	outputMap := map[string]*output.Output{}
 	for _, spec := range clfspec.Outputs {
+		op[framework.OptionForwarderName] = forwarderName
```
jcantrill, line 74: Options are maps which are pass by ref;  I believe this is setting the same value over and over on the same map.  You should be able to move this out of the loop

internal/collector/cloudwatch.go
Clee2691: The template creates one final string that is a combination of all the web identities.
The string looks like the file `/internal/collector/cw_multiple_credentials`. This string is then written to the configmap

internal/collector/cloudwatch.go
Clee2691: Well the `reader` is used in [reconcile.configmap](https://github.com/openshift/cluster-logging-operator/blob/abf40334e35c27579a6ca98af5ec16749969cda2/internal/reconcile/configmaps.go#L20) probably because it is used for reconciling the `dashboard` configmap.

jcantrill: The dashboard, however, lives in an unmanaged namespace.  You should not need to pass the reader here and simply pass the client as the reader in the dependent function

internal/controller/observability/collector.go
jcantrill: If we have precedence and are certain there is limited chance of collision then leave it as-is

internal/collector/cloudwatch.go
Clee2691: I see what you mean. Will make the change

internal/controller/observability/collector.go
Clee2691: I think we will have precedence because it will overwrite what is spec'd when the data etc does not match

PR #3017, opened 2025-04-14
LOG-7014: fix parsing complex auditd logs for otlp mode

vparfonov:
### Description
This PR fixes the parsing of `auditd` log messages that contain multiple `msg` keys in a single line, such as:: 
`type=USER_AUTH msg=audit(1617908892.234:123): pid=1023 uid=0 auid=1000 ses=1 msg='PAM:authentication grantors=pam_unix acct="user1" exe="/usr/bin/sudo" hostname=? addr=? terminal=/dev/pts/0 res=success'`

When using the `parse_key_value()` VRL function on such messages, the `msg` key is returned as an array. However, for our parsing logic, only the first msg entry is relevant — in this case:
`msg=audit(1617908892.234:123)`

This PR ensures that only the first `msg` value is used during parsing.

/cc @cahartma @Clee2691 
/assign @cahartma @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: https://issues.redhat.com/browse/LOG-7014
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/output/otlp/transform.go
+++ b/internal/generator/vector/output/otlp/transform.go
@@ -110,7 +110,13 @@ if exists(kv.type) {
   r.attributes = push(r.attributes, {"key": "auditd.type", "value": {"stringValue": kv.type }})
 }
 if exists(kv.msg) {
-  trimmed = slice!(kv.msg, find!(kv.msg, "(") + 1, -2)
+  msg_str = ""
+  if is_array(kv.msg) {
+    msg_str = kv.msg[0] 
+  } else {
+    msg_str = kv.msg 
+  }
+  trimmed = slice!(msg_str, find!(msg_str, "(") + 1, -2)
   parts = split!(trimmed, ":")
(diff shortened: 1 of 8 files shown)


internal/generator/vector/output/otlp/transform.go
```diff
@@ -110,7 +110,13 @@ if exists(kv.type) {
   r.attributes = push(r.attributes, {"key": "auditd.type", "value": {"stringValue": kv.type }})
 }
 if exists(kv.msg) {
-  trimmed = slice!(kv.msg, find!(kv.msg, "(") + 1, -2)
+  msg_str = ""
+  if is_array(kv.msg) {
+    msg_str = kv.msg[0]
```
Clee2691, line 115: How do we know that only the first message is relevant?

vparfonov: This is standard `auditd` log format: `type=TYPE msg=audit():`

https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/7/html/security_guide/sec-understanding_audit_log_files#sec-Understanding_Audit_Log_Files

PR #3031, opened 2025-04-24
LOG-6863: safety parsing/read fields to avoid messing data in log event

vparfonov:
### Description
This PR addresses an issue where data in a Syslog log event could become corrupted after a `merge` operation. Such behavior can lead to unpredictable and hard-to-diagnose issues.

The fix involves `parsing` and `merging` data into a temporary value, from which the necessary information is extracted. This ensures that the original log event remains structured and valid.

/cc 
/assign 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA:
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/output/syslog/syslog.go
+++ b/internal/generator/vector/output/syslog/syslog.go
@@ -20,9 +20,42 @@ import (
 )
 
 const (
-	TCP       = `tcp`
-	TLS       = `tls`
-	ParsedMsg = "parsed_msg"
+	TCP         = `tcp`
+	TLS         = `tls`
+	ParsedMsg   = "parsed_msg"
+	defFacility = `{._internal.syslog.facility || "user"}`
+	defSeverity = `{._internal.syslog.severity || "informational"}`
+	defProcId   = `{._internal.syslog.proc_id || "-"}`
+	defTag      = `{._internal.syslog.tag || ""}`
+	defAppName  = `{._internal.syslog.app_name || "-"}`
+	defMsgId    = `{._internal.syslog.msg_id || "-"}`
+
+	nodeTag    = `._internal.syslog.tag = to_string!(.systemd.u.SYSLOG_IDENTIFIER || "")`
+	nodeProcId = `._internal.syslog.proc_id = to_string!(.systemd.t.PID || "")`
+
+	containerTag = `._internal.syslog.tag = join!([.kubernetes.namespace_name, .kubernetes.pod_name, .kubernetes.container_name], "")
+#Remove non-alphanumeric characters
+._internal.syslog.tag = replace(._internal.syslog.tag, r'[^a-zA-Z0-9]', "")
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -216,11 +216,11 @@ test-env: ## Echo test environment, useful for running tests outside of the Make
 	RELATED_IMAGE_LOG_FILE_METRIC_EXPORTER=$(IMAGE_LOGFILEMETRICEXPORTER) \
 
 .PHONY: test-functional
-test-functional: test-functional-benchmarker-vector
+test-functional:
 	RELATED_IMAGE_VECTOR=$(IMAGE_LOGGING_VECTOR) \
 	RELATED_IMAGE_LOG_FILE_METRIC_EXPORTER=$(IMAGE_LOGFILEMETRICEXPORTER) \
 	go test -race \
-		./test/functional/... \
+		./test/functional/outputs/syslog/... \
```
jcantrill, line 223: This means we will only test this package and none of the others

Makefile
```diff
@@ -216,11 +216,11 @@ test-env: ## Echo test environment, useful for running tests outside of the Make
 	RELATED_IMAGE_LOG_FILE_METRIC_EXPORTER=$(IMAGE_LOGFILEMETRICEXPORTER) \
 
 .PHONY: test-functional
-test-functional: test-functional-benchmarker-vector
+test-functional:
```
jcantrill, line 219: removing this dependency means we never compile the dev tool

internal/generator/vector/output/syslog/syslog.go
```diff
@@ -219,55 +228,30 @@ func getEncodingTemplatesAndFields(s obs.Syslog) EncodingTemplateField {
 		FieldVRLList: []FieldVRLStringPair{},
 	}
 
-	if s.Facility == "" {
-		s.Facility = `{._internal.syslog.facility || "user"}`
+	appendField := func(fieldName string, value *string, defaultVal string) {
+		if *value == "" {
```
jcantrill, line 232: Is it possible for "value" to be nil since its a pointer?  Do we need to ensure it's non-nil? why do we need to pass a pointer string?

internal/generator/vector/output/syslog/tcp_with_tuning.toml
```diff
@@ -22,6 +20,15 @@ if .log_type == "audit" {
    ._internal.syslog.severity = "informational"
    ._internal.syslog.facility = "security"
 }
+_tmp, err = parse_json(string!(.message))
```
jcantrill, line 23: I missing the intent of this section:
* parse message to temp
* error we set temp to the log event
* non error we set temp to the merge of the message and the log event
* set field parsed_msg

we do not appear to do anything with the parsed_msg so we do any of this?

vparfonov: By the way we can remove default calculation if value configured

PR #3032, opened 2025-04-29
[release-5.9]LOG-6963:Map cluster-wide proxy env variable for compatibility with Vector expectations

vparfonov:
### Description
Adding cluster-wide proxy automatically inject the environment variable, to support `Fluentd` collector  they was lowercased to `http_proxy`, `https_proxy` and `no_proxy`. In the same time, the Vector expects the variable to be named `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` (in uppercase) [1]. This mismatch causes the Vector not picked up proxy settings.

This PR adds a mechanism to support both collectors in the Vector value will be remapped according to current collector expectations.

[1] https://github.com/vectordotdev/vector/blob/f5b9265335c9df7bcbd4a57046738fb5e87a67c1/lib/vector-core/src/config/proxy.rs#L121

/cc @Clee2691 @cahartma 
/assign @jcantrill  

### Links

- Depending on PR(s): https://github.com/ViaQ/vector/pull/207
- GitHub issue:
- JIRA: https://issues.redhat.com/browse/LOG-6963
- Enhancement proposal:

### DIFF
--- a/internal/collector/collector.go
+++ b/internal/collector/collector.go
@@ -209,7 +209,8 @@ func (f *Factory) NewCollectorContainer(inputs logging.InputSpecs, secretNames [
 		{Name: "POD_IP", ValueFrom: &v1.EnvVarSource{FieldRef: &v1.ObjectFieldSelector{APIVersion: "v1", FieldPath: "status.podIP"}}},
 		{Name: "POD_IPS", ValueFrom: &v1.EnvVarSource{FieldRef: &v1.ObjectFieldSelector{APIVersion: "v1", FieldPath: "status.podIPs"}}},
 	}
-	collector.Env = append(collector.Env, utils.GetProxyEnvVars()...)
+
+	collector.Env = append(collector.Env, utils.GetProxyEnvVars(f.CollectorSpec.Type)...)
 
 	collector.VolumeMounts = []v1.VolumeMount{
 		{Name: f.ResourceNames.SecretMetrics, ReadOnly: true, MountPath: metricsVolumePath},
--- a/internal/collector/collector_test.go
+++ b/internal/collector/collector_test.go
@@ -173,39 +173,48 @@ var _ = Describe("Factory#Daemonset", func() {
 					Fail(fmt.Sprintf("Exp. collector to include env var: %s", name))
 				}
 
-				It("should add the proxy variables to the collector", func() {
-					_httpProxy := os.Getenv("http_proxy")
-					_httpsProxy := os.Getenv("https_proxy")
-					_noProxy := os.Getenv("no_proxy")
-					cleanup := func() {
-						_ = os.Setenv("http_proxy", _httpProxy)
-						_ = os.Setenv("https_proxy", _httpsProxy)
-						_ = os.Setenv("no_proxy", _noProxy)
-					}
-					defer cleanup()
+				DescribeTable("should add the proxy variables to the collector",
+					func(collectorType logging.LogCollectionType, setHttpProxy, setHttpsProxy, setNoProxy, expectedHttpProxy, expectedHttpsProxy, expectedNoProxy string) {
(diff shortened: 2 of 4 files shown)


cahartma:
This should most likely be fixed in 6.2 (and master), rather than 5.9.   Also, I'm not seeing how this is related to fixing the '@' issue in:  https://issues.redhat.com/browse/LOG-6963

cahartma:
Shouldn't we be merging this fix into master and release-6.2 first.   Is there a way we can resolve this in v6, then tell them to upgrade if they complain about not fixing in v5?

vparfonov:
> This should most likely be fixed in 6.2 (and master), rather than 5.9. Also, I'm not seeing how this is related to fixing the '@' issue in: [issues.redhat.com/browse/LOG-6963](https://issues.redhat.com/browse/LOG-6963)

Vector just don't picked up env value in lowercase, not related directly to the `@`

vparfonov:
> Shouldn't we be merging this fix into master and release-6.2 first. Is there a way we can resolve this in v6, then tell them to upgrade if they complain about not fixing in v5?

Maybe yes, original issue opened for 5.9 thats why i did it firstly

internal/collector/vector/run_script.go
```diff
@@ -11,6 +11,22 @@ VECTOR_DATA_DIR=%s
 echo "Creating the directory used for persisting Vector state $VECTOR_DATA_DIR"
 mkdir -p ${VECTOR_DATA_DIR}
 
+# Remap lowercase to uppercase if it's set
+if [ -z "${HTTP_PROXY:-}" ] && [ -n "${http_proxy:-}" ]; then
+  export HTTP_PROXY="$http_proxy"
+  echo HTTP_PROXY="$HTTP_PROXY"
```
jcantrill, line 17: If we are going to add a "log" statement then lets say what we are doing... "Setting HTTP_PROXY=...."

test/functional/misc/vector_api_cli_test.go
```diff
@@ -36,5 +35,19 @@ var _ = Describe("[Functional][Misc][API_CLI] Functional test", func() {
...
+			envVisitor := func(b *runtime.PodBuilder) error {
+				b.GetContainer(constants.CollectorName).AddEnvVar("http_proxy", "http://my_proxy:8080")
+				b.GetContainer(constants.CollectorName).AddEnvVar("https_proxy", "https://my_proxy:9090")
+				b.GetContainer(constants.CollectorName).AddEnvVar("no_proxy", "my-host,your-host")
+				return nil
+			}
+			Expect(framework.DeployWithVisitor(envVisitor)).To(BeNil())
+			out, err := framework.ReadCollectorLogs()
+			Expect(err).To(BeNil())
+			Expect(out).To(ContainSubstring(`HTTP_PROXY=http://my_proxy:8080`))
```
jcantrill, line 48: I believe a more accurate assertion would be to exec into the  pod, get the env vars, and check explicitly for the environment variable and value

jcantrill (commented):
/hold

test/functional/misc/vector_api_cli_test.go
vparfonov: yes, get the env vars will be better but with here var sets only in the process running that script, It does not persist globally in the container environment.

internal/collector/vector/run_script.go
cahartma: Is the proxy url actually something we want to put into a log?   Couldn't it contain user/pwd?

anpingli:
@vparfonov  According to my test result(Vector to Splunk).   https_proxy/http_proxy and HTTPS_PROXY/HTTP_RPOXY are same.   No matter which one is specified, the PROXY will be used. 
With HTTPS_RPOXY, the vector still fails to forward logs to splunk when  the proxy name incude encoded @.

PR #3033, opened 2025-04-30
LOG-7009: remove  parse and merge data from message to the log event

vparfonov:
### Description
This PR addresses several issues caused by problematic behavior introduced during the migration from `Fluentd` to `Vector`. These issues are related to parsing and merging data from the message into the log event, which can result in system information being overwritten, data being duplicated, or log events becoming corrupted. This fix also helps align the Syslog output with other output types because it was introduced only for Syslog output. 

/cc @Clee2691 @cahartma 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA:https://issues.redhat.com/browse/LOG-7009, https://issues.redhat.com/browse/LOG-7010, https://issues.redhat.com/browse/LOG-7008 
- Enhancement proposal

### DIFF
--- a/test/functional/outputs/syslog/forward_to_syslog_test.go
+++ b/test/functional/outputs/syslog/forward_to_syslog_test.go
@@ -131,30 +131,6 @@ var _ = Describe("[Functional][Outputs][Syslog] Functional tests", func() {
 			Expect(getProcID(fields)).To(Equal("myproc"))
 			Expect(getMsgID(fields)).To(Equal("mymsg"))
 		})
-		It("should take values of appname, procid, messageid from record", func() {
-			functional.NewClusterLogForwarderBuilder(framework.Forwarder).
-				FromInput(logging.InputNameApplication).
-				ToOutputWithVisitor(join(setSyslogSpecValues, func(spec *logging.OutputSpec) {
-					spec.Syslog.AppName = "$.message.appname_key"
-					spec.Syslog.ProcID = "$.message.procid_key"
-					spec.Syslog.MsgID = "$.message.msgid_key"
-				}), logging.OutputTypeSyslog)
-			Expect(framework.Deploy()).To(BeNil())
-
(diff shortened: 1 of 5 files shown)


test/functional/outputs/syslog/forward_to_syslog_test.go
```diff
@@ -131,30 +131,6 @@ var _ = Describe("[Functional][Outputs][Syslog] Functional tests", func() {
 			Expect(getProcID(fields)).To(Equal("myproc"))
 			Expect(getMsgID(fields)).To(Equal("mymsg"))
 		})
-		It("should take values of appname, procid, messageid from record", func() {
```
jcantrill, line 134: Let's try to enable these tests by introducing the JSON parse filter to demonstrate how this functionality should be work

test/functional/outputs/syslog/rfc3164_test.go
```diff
@@ -55,37 +55,9 @@ var _ = Describe("[Functional][Outputs][Syslog] RFC3164 tests", func() {
 		Expect(outputlogs[0]).To(MatchRegexp(expMatch), "Exp to find tag in received message")
 		Expect(outputlogs[0]).To(MatchRegexp(`{"index":.*1,.*"timestamp":.*1,.*"tag_key":.*"rec_tag"}`), "Exp to find the original message in received message")
 	},
-
-		Entry("should use the value from the record and include the message", "$.message.tag_key", "rec_tag", false),
```
jcantrill, line 59: Same comment here.  Also, let's add a test which demonstrates the bug no longer exists and proves we fixed it

test/functional/outputs/syslog/forward_to_syslog_test.go
vparfonov: 5.9 don't have JSON filter, i will do it in 6.x

PR #3034, opened 2025-04-30
Remove parse and merge data from message to the log event

vparfonov:
### Description
This PR addresses several issues introduced during the migration from Fluentd to Vector. These issues are related to parsing and merging data from the message field into the log event, which could lead to system information being overwritten, data duplication, or event corruption. The fix also brings Syslog output behavior in line with other output types, as the problem was specific to Syslog.

To retain the ability to extract values from JSON-formatted logs, a JSON Parser Filter must be added to the Log Forwarder spec.

To access specific fields from the parsed JSON, use the field names with the `structured` prefix.

Example:
Log message: `{"index":1,"timestamp":123456789,"facility_key":"local0","severity_key":"Informational"}`

Access paths:

- `{.structured.facility_key||"user"}`
- `{.structured.severity_key||"security"}` 

/cc @cahartma @Clee2691 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA:
- Enhancement proposal:


jcantrill (commented):
/approve

jcantrill:
This looks what I would expect.  I assume we will backport this against 6.2 which means lets find an issue to close it against

docs/features/logforwarding/outputs/syslog-forwarding.adoc
```diff
@@ -102,6 +102,58 @@ Examples:
...
+. `{.structured.severity_key||"security"}`
+
+==== How to add JSON Parser filter
+.cluster-log-forwarder.yaml
+[source,yaml]
+----
+apiVersion: observability.openshift.io/v1
+kind: ClusterLogForwarder
+metadata:
+  name: instance
```
Clee2691, line 128: Just a nit but I think we wanted to do away with the `instance` naming right? Can't remember but if that's not the case, ignore this.

internal/generator/vector/output/syslog/syslog.go
```diff
@@ -345,3 +307,92 @@ var keyre = regexp.MustCompile(`^\$(\.[[:word:]]*)+$`)
...
+			builder.WriteString(msgId + "\n")
+		}
+		if o.ProcId == "" || o.AppName == "" {
+			defaultRules = append(defaultRules, defaultRule{
+				cond:    `._internal.log_type == "infrastructure" && ._internal.log_source == "node"`,
+				appName: ifEmpty(o.AppName, nodeAppName),
+				procId:  ifEmpty(o.ProcId, nodeProcId2),
+			})
+		}
+		if o.AppName == "" || o.ProcId == "" || o.Severity == "" || o.Facility == "" {
```
Clee2691, line 356: You can combine the above if statement with this one because they both check if `o.ProcId` or `o.AppName` are empty.

internal/generator/vector/output/syslog/syslog.go
```diff
@@ -345,3 +307,92 @@ var keyre = regexp.MustCompile(`^\$(\.[[:word:]]*)+$`)
...
+
+	if o.RFC == obs.SyslogRFC3164 {
+		if o.ProcId == "" || o.AppName == "" {
+			defaultRules = append(defaultRules, defaultRule{
+				cond:    `._internal.log_type == "infrastructure" && ._internal.log_source == "node"`,
+				appName: ifEmpty(o.AppName, nodeTag),
+				procId:  ifEmpty(o.ProcId, nodeProcId),
+			})
+		}
+		if o.AppName == "" || o.Severity == "" || o.Facility == "" {
```
Clee2691, line 330: You check `o.AppName` is empty in the above if statement. You can get rid of `o.AppName` check in the above if.

vparfonov: Yep it looks a bit ugly, i will try to rework it somehow: but idea in step-by-step building config like this: 
```
if ._internal.log_type == "infrastructure" && ._internal.log_source == "node" {
    ._internal.syslog.app_name = to_string!(._internal.systemd.u.SYSLOG_IDENTIFIER||"-")
    ._internal.syslog.proc_id = to_string!(._internal.systemd.t.PID||"-")
}
if ._internal.log_source == "container" {
   ._internal.syslog.app_name = join!([._internal.kubernetes.namespace_name, ._internal.kubernetes.pod_name, ._internal.kubernetes.container_name], "_")
   ._internal.syslog.proc_id = to_string!(._internal.kubernetes.pod_id||"-")
   ._internal.syslog.severity = ._internal.level
   ._internal.syslog.facility = "user"
}
if ._internal.log_type == "audit" {
   ._internal.syslog.app_name = ._internal.log_source
   ._internal.syslog.proc_id = to_string!(._internal.auditID || "-")
   ._internal.syslog.severity = "informational"
   ._internal.syslog.facility = "security"
}
```
so it doesn't meeter what we checked above

PR #3043, opened 2025-05-16
LOG-7023: Add Prometheus alerting rule for collector 403 responses

Clee2691:
### Description
This PR introduces a new Prometheus alerting rule to detect an elevated rate of HTTP 403 Forbidden responses in the collector.

The associated [runbook](https://github.com/openshift/runbooks/pull/249) is pending PR review.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-7023
- runbook: https://github.com/openshift/runbooks/pull/249

### DIFF
--- a/bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
+++ b/bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
@@ -30,6 +30,31 @@ spec:
       labels:
         service: collector
         severity: Warning
+    - alert: CollectorHigh403ForbiddenResponseRate
+      annotations:
+        description: High rate of "HTTP 403 Forbidden" responses detected for collector
+          "{{ $labels.app_kubernetes_io_instance }}" in namespace {{ $labels.namespace
+          }} for output "{{ $labels.component_id }}". The rate of 403 responses is
+          {{ printf "%.2f" $value }}% over the last 2 minutes, persisting for more
+          than 5 minutes. This could indicate an authorization issue.
+        runbook_url: https://github.com/openshift/runbooks/blob/master/alerts/cluster-logging-operator/CollectorHigh403ForbiddenResponseRate.md
+        summary: At least 10% of sent requests responded with "HTTP 403 Forbidden"
+          for collector "{{ $labels.app_kubernetes_io_instance }}" in namespace {{
+          $labels.namespace }} for output "{{ $labels.component_id }}"
+      expr: |
+        sum(
+          irate(vector_http_client_responses_total{component_kind="sink", status="403"}[2m])
+        ) by (component_id, app_kubernetes_io_instance, namespace)
+        /
+        sum(
+          irate(vector_http_client_responses_total{component_kind="sink"}[2m])
+        ) by (component_id, app_kubernetes_io_instance, namespace)
+        * 100
+        > 10
+      for: 5m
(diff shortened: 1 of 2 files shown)


bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -30,6 +30,20 @@ spec:
...
+    - alert: CollectorHigh403ForbiddenResponseRate
+      annotations:
+        description: 'Collector: {{ $labels.app_kubernetes_io_instance }}, sink: {{
+          $labels.component_id }} is receiving HTTP 403 Forbidden response rate of
+          {{ $value | printf "%.2f" }} errors/sec over the last 2 minutes, persisting
+          for more than 5 minutes. This could indicate an authorization issue.'
+        summary: 'Collector: {{ $labels.app_kubernetes_io_instance }}, sink: {{ $labels.component_id
+          }} is receiving HTTP 403 Forbidden responses'
+      expr: |
+        sum by (component_id, app_kubernetes_io_instance) (irate(vector_http_client_responses_total{component_kind="sink", status="403"}[2m])) > 0
```
jcantrill, line 42: Do we need to include namespace here?  The instance = CLF.name right? Does this provide a unique enough check? If changing then the description and summary should also be updated

jcantrill:
/hold

pending other reviews

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
Clee2691: Yes that would be helpful. I will make the change

jcantrill:
@Clee2691 this  lgtm.  Can you please add a feature doc, if one doesn't exist, which names the alert and provides a brief summary of what it is.  Maybe this could be a jumping off point to link to the runbooks when they exist

Clee2691:
> @Clee2691 this lgtm. Can you please add a feature doc, if one doesn't exist, which names the alert and provides a brief summary of what it is. Maybe this could be a jumping off point to link to the runbooks when they exist

I've created the very first runbook here: https://github.com/openshift/runbooks/pull/249. This will be a good start  to other runbooks. At least it onboards it into the runbooks repo. I will add the link to the corresponding runbook to this alert.

config/prometheus/collector_alerts.yaml
```diff
@@ -28,6 +28,27 @@ spec:
       labels:
         service: collector
         severity: Warning
+    - alert: CollectorHigh403ForbiddenResponseRate
```
jcantrill, line 31: @xperimental @cahartma @Clee2691 We should consider the name here.  At minimum it should probably be prefixed as "LogCollector" and I'm wondering what we can do to remove "high" from the name.  Is there a clever way to include the actual threshold value in it? Could we maybe remove it all-together

Clee2691: I don't see the problem with this name for the alert. This alert is specific to 403 errors and that there is a high rate of them. 

Even in the [prometheus alerting rules](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/) page, the example on the page has the name of `HighRequestLatency`.

jcantrill (commented):
/lgtm

PR #3059, opened 2025-06-03
LOG-7267: add fallback option if syslog appname calculation fail

vparfonov:
### Description
This pull request addresses an issue where the default value calculation for the Syslog `appname` field would fail. This failure occurred specifically when a Prune filter removed one or more of the following Kubernetes metadata fields: `.kubernetes.namespace_name`, `.kubernetes.pod_name`, or `.kubernetes.container_name`.

The `join([.kubernetes.namespace_name, .kubernetes.pod_name, .kubernetes.container_name])` VRL function would encounter an error in such scenarios, preventing the `appname` field from being correctly initialized.

To resolve this, an error check has been added. If the `join()` operation encounters an error (due to missing fields), the `appname` field will now be gracefully initialized with an empty string.

Similar things were added for Syslog tag field.

/cc @cahartma @Clee2691 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: https://issues.redhat.com/browse/LOG-7267
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/output/syslog/syslog.go
+++ b/internal/generator/vector/output/syslog/syslog.go
@@ -40,15 +40,25 @@ const (
 	nodeProcIdRFC5424 = `._syslog.proc_id = to_string!(.systemd.t.PID || "-")`
 
 	// Default values for Syslog fields for application logType and infrastructure logType if source 'container'
-	containerAppName  = `._syslog.app_name = join!([.kubernetes.namespace_name, .kubernetes.pod_name, .kubernetes.container_name], "_")`
+	containerAppName = `._syslog.app_name, err = join([.kubernetes.namespace_name, .kubernetes.pod_name, .kubernetes.container_name], "_")
+if err != null {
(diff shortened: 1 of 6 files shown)


internal/generator/vector/output/syslog/syslog.go
```diff
@@ -40,15 +40,23 @@ const (
 	nodeProcIdRFC5424 = `._syslog.proc_id = to_string!(.systemd.t.PID || "-")`
 
 	// Default values for Syslog fields for application logType and infrastructure logType if source 'container'
-	containerAppName  = `._syslog.app_name = join!([.kubernetes.namespace_name, .kubernetes.pod_name, .kubernetes.container_name], "_")`
+	containerAppName = `._syslog.app_name, err = join([.kubernetes.namespace_name, .kubernetes.pod_name, .kubernetes.container_name], "_")
+if err != null {
```
jcantrill, line 44: Seems like we should log the error here and spam the collector logs.  This may be the only way for users to understand they need to set the value to something or not prune these fields

PR #3068, opened 2025-06-24
LOG-7222: Implement multi-stage parsing for log level determination

vparfonov:
#### Problem:
Determining the log level relied on `if/else` block to perform a keyword search across the entire raw message. This approach was had drawbacks: It could incorrectly assign a level if a log message contained multiple level keywords. For example, a structured INFO log with the word "ERROR" in its payload (... INFO ... message about a work ERROR) would be misclassified as error.

#### Changes:
This PR refactors the log level determination logic in our VRL script to use a robust, multi-stage parsing pipeline. Instead of immediately searching for keywords, we now attempt to parse the message using a series of specialized, embedded VRL functions. The fallback to a keyword search is now a last resort.
The new order of operations is as follows:
**Attempt 1** `parse_klog`:  we try parsing it as a Kubernetes klog message. This correctly handles logs originating from many components within our containerized environments.
**Attempt 2** `parse_logfmt`: next, we attempt to parse the message using the logfmt key-value format (e.g., level=error msg=...).
**Attempt 3** `parse_grok`: If all of the above fail, we use `parse_grok` with a predefined list of common text-based formats (e.g., Logback/Log4j). The first pattern to match determines the level.
**Fallback to Keyword Search**: Only if all structured parsing attempts fail do we fall back to the original if/else block to scan the raw message for keywords.

#### Benefits:
**Increased Accuracy**:  we now correctly classify messages and avoid false positives from the payload content.
**Improved Performance** (Expected, not confirmed): Using native, optimized VRL parse functions for their intended formats is generally more performant than executing a series of regex searches on every log. This moves our logic toward Vector's best practices.
**Enhanced Maintainability**: Adding support for a new text-based log format is now as simple as adding a new pattern to the parse_grok list, requiring no changes to the core script logic.

/cc @Clee2691 @cahartma 
/assign @cahartma 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: https://issues.redhat.com/browse/LOG-7222
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/filter/openshift/viaq/v1/normalize.go
+++ b/internal/generator/vector/filter/openshift/viaq/v1/normalize.go
@@ -26,54 +26,94 @@ if ._internal.log_source == "syslog" && exists(._internal.structured) {. = merge
 if ._internal.log_source == "container" && exists(._internal.structured) {.structured = ._internal.structured }
 `
 	SetLogLevel = `
+
 if !exists(._internal.level) {
-  level = "default"
+  level = null
   message = ._internal.message
 
-  # Match on well known structured patterns
-  # Order: emergency, alert, critical, error, warn, notice, info, debug, trace
-
-  if match!(message, r'^EM[0-9]+|level=emergency|Value:emergency|"level":"emergency"') {
-    level = "emergency"
-  } else if match!(message, r'^A[0-9]+|level=alert|Value:alert|"level":"alert"') {
-    level = "alert"
-  } else if match!(message, r'^C[0-9]+|level=critical|Value:critical|"level":"critical"') {
-    level = "critical"
-  } else if match!(message, r'^E[0-9]+|level=error|Value:error|"level":"error"') {
-    level = "error"
-  } else if match!(message, r'^W[0-9]+|level=warn|Value:warn|"level":"warn"') {
-    level = "warn"
-  } else if match!(message, r'^N[0-9]+|level=notice|Value:notice|"level":"notice"') {
-    level = "notice"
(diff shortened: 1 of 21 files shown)


internal/generator/vector/filter/openshift/viaq/v1/normalize.go
```diff
@@ -26,55 +26,95 @@ if ._internal.log_source == "syslog" && exists(._internal.structured) {. = merge
...
-  # Match on well known structured patterns
-  # Order: emergency, alert, critical, error, warn, notice, info, debug, trace
+  # attempt 1: parse as logfmt (e.g. level=error msg="Failed to connect")
+  
+  parsed_logfmt, err = parse_logfmt(message)  
+  if err == null && is_string(parsed_logfmt.level) {
+    level = downcase!(parsed_logfmt.level)
+  }
 
-  if match!(message, r'^EM[0-9]+|level=emergency|Value:emergency|"level":"emergency"') {
```
alanconway, line 36: I think this fallback would be better if we took the keyword that occurs _first_ in the log message, rather than the highest-priority one, e.g.
```
Apr 24 15:02:47 fedora gnome-session-binary[3056]: GnomeDesktop-WARNING: Could not create transient scope for PID 3303: GDBus.Error:org.freedesktop.DBus.Error.UnixProcessIdUnknown: Failed to set unit properties: No such process
```
This is a Warning with a message that contains the word Error. If you use the earliest keyword found "WARNING" you get this right, if you use the scariest keyword, "ERROR" you exaggerate the level.

vparfonov: done

PR #3070, opened 2025-06-26
LOG-7196: Enable annotations for kubeapi caching and configurable maxUnavailable

cahartma:
### Description
This feature adds the `use_apiserver_cache` config to the vector.toml and a configurable rollout strategy `maxUnavailable` to the DS. 
 Both of these are currently enabled via annotations in the CLF.      Future effort will move these to the forwarder spec.

### Verifying
Besides collector conditions and status, the following commands can be used to verify the two options:
`oc get cm/my-forwarder-config -o jsonpath='{.data.vector\.toml}' |grep  use_apiserver_cache`
`oc get ds/my-forwarder -oyaml -o jsonpath='{.spec.updateStrategy.rollingUpdate.maxUnavailable}'`

/cc @Clee2691 @vparfonov @jcantrill
/assign @xperimental 

### Links
- https://issues.redhat.com/browse/LOG-7196

### DIFF
--- a/internal/controller/observability/collector_features_test.go
+++ b/internal/controller/observability/collector_features_test.go
@@ -38,6 +38,15 @@ var _ = Describe("#EvaluateAnnotationsForEnabledCapabilities", func() {
 		Entry("enables debug for true", helpers.EnableDebugOutput, "true", AnnotationDebugOutput, "true"),
 		Entry("enables debug for True", helpers.EnableDebugOutput, "true", AnnotationDebugOutput, "True"),
 		Entry("disables debug for anything else", "", "", AnnotationDebugOutput, "abcdef"),
+
+		Entry("enables kube-cache for true", framework.UseKubeCacheOption, "true", AnnotationKubeCache, "true"),
+		Entry("enables kube-cache for True", framework.UseKubeCacheOption, "true", AnnotationKubeCache, "True"),
+		Entry("enables kube-cache for enabled", framework.UseKubeCacheOption, "true", AnnotationKubeCache, "enabled"),
+		Entry("disables kube-cache for anything else", "", "", AnnotationKubeCache, "bubbles"),
+
+		Entry("enables max-unavailable for value '10'", framework.MaxUnavailableOption, "10", AnnotationMaxUnavailable, "10"),
+		Entry("enables max-unavailable for value '99%'", framework.MaxUnavailableOption, "99%", AnnotationMaxUnavailable, "99%"),
+		Entry("disables max-unavailable option for anything not a number or percentage", "", "", AnnotationMaxUnavailable, "fluffy"),
 	)
 
 })
--- a/internal/generator/vector/input/container.go
+++ b/internal/generator/vector/input/container.go
@@ -37,7 +37,7 @@ var (
 )
 
 // NewContainerSource generates config elements and the id reference of this input and normalizes
-func NewContainerSource(spec obs.InputSpec, namespace, includes, excludes string, logType obs.InputType, logSource interface{}) ([]framework.Element, []string) {
+func NewContainerSource(spec obs.InputSpec, namespace, includes, excludes string, logType obs.InputType, logSource interface{}, useCache bool) ([]framework.Element, []string) {
 	base := helpers.MakeInputID(spec.Name, "container")
 	var selector *metav1.LabelSelector
 	if spec.Application != nil {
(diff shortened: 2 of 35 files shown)


internal/controller/observability/collector_features_test.go
```diff
@@ -38,6 +38,15 @@ var _ = Describe("#EvaluateAnnotationsForEnabledCapabilities", func() {
...
 		Entry("disables debug for anything else", "", "", AnnotationDebugOutput, "abcdef"),
+
+		Entry("enables kube-cache for true", framework.UseKubeCacheOption, "true", AnnotationKubeCache, "true"),
+		Entry("enables kube-cache for True", framework.UseKubeCacheOption, "true", AnnotationKubeCache, "True"),
+		Entry("enables kube-cache for enabled", framework.UseKubeCacheOption, "true", AnnotationKubeCache, "enabled"),
+		Entry("disables kube-cache for anything else", "", "", AnnotationKubeCache, "bubbles"),
+
+		Entry("enables max-unavailable for value '10'", framework.MaxUnavailableOption, "10", AnnotationMaxUnavailable, "10"),
+		Entry("enables max-unavailable for value '99%'", framework.MaxUnavailableOption, "99%", AnnotationMaxUnavailable, "99%"),
+		Entry("sets max-unavailable option to '100%' for anything else", framework.MaxUnavailableOption, "100%", AnnotationMaxUnavailable, "fluffy"),
```
vparfonov: "fluffy" ? :)

internal/generator/vector/source/kubernetes_logs.go
```diff
@@ -16,6 +16,7 @@ type KubernetesLogs struct {
 	IncludePaths       string
 	ExcludePaths       string
 	ExtraLabelSelector string
+	UseKubeCache       bool
```
vparfonov, line 19: as proposal, we can use here: 
`UseKubeCache typehelpers.OptionalPair`
in this case, we don't need to add 
`use_apiserver_cache = false` because this is default behavior

internal/generator/vector/input/container.go
```diff
@@ -51,6 +51,7 @@ func NewContainerSource(spec obs.InputSpec, namespace, includes, excludes string
 			IncludePaths:       includes,
 			ExcludePaths:       excludes,
 			ExtraLabelSelector: source.LabelSelectorFrom(selector),
+			UseKubeCache:       useCache,
```
vparfonov, line 54: we can use here: 
`conf.UseKubeCache = typehelpers.NewOptionalPair("use_apiserver_cache", true)`

anpingli:
Run some test with use_apiserver_cache enabled. The Vector works well.

cahartma:
> ... as what .spec.updateStrategy.rollingUpdate.maxUnavailable is specified.
@anpingli instead do you mean it matches the value set in the annotation?

internal/controller/observability/collector_features_test.go
anpingli: fluffy is rejected.
status:
  conditions:
  - lastTransitionTime: "2025-06-26T12:00:15Z"
    message: max-unavailable-rollout value "fluffy" must be an absolute number or
      a valid percentage

internal/generator/vector/source/kubernetes_logs.go
cahartma: If we provide the option to enable it, I think it helps to be explicit and set as false.   In the future this setting will most likely be enabled by default in a later release.

internal/generator/vector/input/container.go
cahartma: See comment below.   We plan to include this enabled by default in a future release if all goes well with initial customers.   Let me know if this makes sense related to your comment.

internal/controller/observability/collector_features_test.go
cahartma: Better than "foo" imho.  :)
This is an "annotations setting options" unit test so any invalid annotations value should set the default option value = 100%

cahartma: I see the unit test failed here.  Thats odd, I'll take a look.   
@anpingli this is correct and is validation of the forwarder I added to ensure it's very obvious for the user.    The above test is unit only.

internal/generator/vector/source/kubernetes_logs.go
vparfonov: make sense

internal/controller/observability/collector_features_test.go
cahartma: I've updated the unit test and the `EvaluateAnnotationsForEnabledCapabilities` method to check for valid values here as well.  Now it disables the option.   When we create the DS, we set the maxUnavailable value to 100% if the option is nil or invalid.

PR #3071, opened 2025-06-27
LOG-7340: add validation  that the Kafka URLs should be tcp or tls

vparfonov:
### Description

/cc @cahartma @Clee2691 
/assign @cahartma 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA:https://issues.redhat.com/browse/LOG-7340
- Enhancement proposal:

### DIFF
--- a/api/observability/v1/output_types.go
+++ b/api/observability/v1/output_types.go
@@ -692,7 +692,7 @@ type Kafka struct {
 	//
 	// The 'username@password' part of `url` is ignored.
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:XValidation:rule="self == '' ||  isURL(self)", message="invalid URL"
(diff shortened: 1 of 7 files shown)


api/observability/v1/output_types.go
```diff
@@ -692,7 +692,7 @@ type Kafka struct {
 	//
 	// The 'username@password' part of `url` is ignored.
 	// +kubebuilder:validation:Optional
-	// +kubebuilder:validation:XValidation:rule="self == '' ||  isURL(self)", message="invalid URL"
+	// +kubebuilder:validation:XValidation:rule="self == '' || (isURL(self) && (self.startsWith('tcp://') || self.startsWith('tls://')))",message="must be a valid URL with a tcp or tls scheme"
```
jcantrill, line 695: Are we certain 'http' is not an allowed protocol?  The bug description has reference to it.

vparfonov: According to our docs we support only `tcp` or `tls` scheme
https://docs.redhat.com/en/documentation/openshift_container_platform/4.15/html/logging/log-collection-and-forwarding-2#configuring-log-forwarding

PR #3073, opened 2025-07-02
support custom headers for Loki and Elasticsearch outputs

AndrewChubatiuk:
### Description
Add ability to set custom headers for Loki and Elasticsearch protocols, which are supported by Vector to allow to customize ingested logs from Openshift into VictoriaLogs

/cc @Clee2691 @cahartma
/assign @cahartma


internal/generator/vector/output/loki/with_headers.toml
```diff
@@ -0,0 +1,47 @@
...
+inputs = ["loki_receiver_remap_label"]
+endpoint = "https://logs-us-west1.grafana.net"
+out_of_order_action = "accept"
+healthcheck.enabled = false
+
+[sinks.loki_receiver.encoding]
+codec = "json"
+except_fields = ["_internal"]
+
+[sinks.loki_receiver.request]
```
vparfonov: @AndrewChubatiuk  Thanks for your contribution! Just a heads-up: according to the official docs, the Loki sink’s request block doesn’t support a headers field. https://vector.dev/docs/reference/configuration/sinks/loki/#request

AndrewChubatiuk: right, thanks, removed them for loki output

jcantrill:
@AndrewChubatiuk if the intent is to use this to write to VictoriaLogs then is there a functional test we can provide to confirm functionallity.  The larger question is what happens when RH logging to VictoriaLogs doesn't work and we are on the hook to support it with no backing tests.

AndrewChubatiuk:
@jcantrill added functional test with custom headers

AndrewChubatiuk:
@jcantrill done

PR #3074, opened 2025-07-02
LOG-7892: support format parameter for http sink

AndrewChubatiuk:
### Description
NDJSON (newline-delimited JSON) allows to send log events in a streaming fashion which results in a smaller memory footprint on a receiver side. Added `format` argument for HTTP sink, that accepts `json` and `ndjson` values to support NDJSON data ingestion

/cc @Clee2691 @cahartma @jcantrill
/assign @jcantrill

### DIFF
--- a/api/observability/v1/output_types.go
+++ b/api/observability/v1/output_types.go
@@ -664,6 +664,21 @@ type HTTPTuningSpec struct {
 	Compression string `json:"compression,omitempty"`
 }
 
+// HTTPFormat is used to define the data format of data to be sent.
+//
+// +kubebuilder:validation:Enum:=json;ndjson
+type HTTPFormat string
+
+func (s HTTPFormat) String() string {
+	return string(s)
+}
+
+// HTTPFormat type constants, must match JSON tags of HTTPFormat fields.
+const (
+	HTTPFormatJSON   HTTPFormat = "json"
+	HTTPFormatNDJSON HTTPFormat = "ndjson"
+)
+
 // HTTP provided configuration for sending json encoded logs to a generic HTTP endpoint.
 type HTTP struct {
 	URLSpec `json:",inline"`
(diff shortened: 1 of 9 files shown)


AndrewChubatiuk:
> are you still interested in submitting this contribution? If so, we'll need to create a tracking issue and review the API

yes, interested in this as well

vrutkovs:
/retitle [LOG-7892](https://issues.redhat.com//browse/LOG-7892): support NDJSON for http sink

AndrewChubatiuk:
hey @jcantrill
thanks for merging #3073
could you please review this PR as well?

api/observability/v1/output_types.go
```diff
@@ -698,6 +698,12 @@ type HTTP struct {
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:validation:XValidation:rule="self == '' ||  isURL(self)", message="invalid URL"
 	ProxyURL string `json:"proxyURL,omitempty"`
+
+	// LinePerEvent uses NDJSON instead of JSON to send data to remote destination.
+	//
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Line Per Event"
+	LinePerEvent bool `json:"line_per_event,omitempty"`
```
alanconway: Suggest a more general encoding enum:
```suggestion
 Encoding string `json:"encoding,omitempty"`
```
Initial values are "json" (default) and "ndjson" but it leaves room for other encodings if they are ever needed.

jcantrill: I'm not convinced we wish to call this "encoding".  Vector causes it "framing" which seems more accurate.  My largest concern is what if we wish to modify our output format to OTEL from ViaQ? OTEL is an "encoding" is it not?  We could "encode" the payload to be OTEL but write it to HTTP or Elasticsearch or whatever.

jcantrill: @AndrewChubatiuk My apology for the time it has been taking for this submission.  Please change this to an optinal field,  enum named "format" with values of 'json' (default) or 'ndjson'.

@alanconway I have chosen this field name according to how our AI friends answered: "is ndjson and encoding" which was "NDJSON, or Newline Delimited JSON, is not an encoding but a format for storing structured data where each line is a valid JSON object...."

AndrewChubatiuk: @jcantrill updated PR

PR #3093, opened 2025-08-11
[LOG-7505] Add support for cross-account AssumeRole

apahim:
## Summary
This PR adds support for cross-account CloudWatch log forwarding using AWS AssumeRole with automatically generated session names that include cluster and forwarder context.

## What's Changed
- **New AssumeRole Support**: Added `AssumeRole` field to `CloudwatchIAMRole` for cross-account access
- **Enhanced Security**: Added optional `ExternalID` field for additional security measures
- **Auto-Generated Session Names**: auto-generated with format: `<cluster_ID>-<CLF_Name>-<output_name>`. Example: `625f0e4d-cp-logs-out-to-cw`
- **Dual-Profile Authentication**: Implements proper authentication separation using AWS credentials profiles

### Cross-Account AssumeRole
```yaml
    outputs:
      - name: out-to-cw
        type: cloudwatch
        cloudwatch:
          region: us-east-1
          groupName: app-logs
          authentication:
            type: iamRole
            iamRole:
              roleARN:
                key: role_arn
                secretName: app-logs  # Initial role for authentication
              token:
                from: serviceAccount
              assumeRole:  # NEW: Cross-account assume role
                roleARN:
                  secretName: app-logs-cross  # Cross-account role ARN
                  key: role_arn
```

### Vector Configuration with AssumeRole

Here's the exact Vector configuration that's generated when using AssumeRole. The key
parts are:

1. CloudWatch Sink Configuration

```
# Cloudwatch Logs
[sinks.output_out_to_cw]
type = "aws_cloudwatch_logs"
inputs = ["output_out_to_cw_group_name"]
region = "us-east-1"
compression = "none"
group_name = "{{ _internal.output_out_to_cw_group_name }}"
stream_name = "{{ stream_name }}"
auth.credentials_file = "/var/run/ocp-collector/config/app-logs-aws-creds/credentials"
auth.profile = "output_out-to-cw"
healthcheck.enabled = false
```

2. AWS Credentials File (Referenced Above)

```
# Source profile for initial web identity authentication
[output_out-to-cw_source]
web_identity_token_file=/var/run/ocp-collector/serviceaccount/token
role_arn=arn:aws:iam::123456789012:role/app-logs
role_session_name=output-out-to-cw-source

# Assume role profile for cross-account access
[output_out-to-cw]
source_profile=output_out-to-cw_source
role_arn=arn:aws:iam::987654321098:role/app-logs-cross
role_session_name=625f0e4d-app-logs-out-to-cw
```

Key Points:

1. Authentication Method: Vector uses credentials_file + profile approach instead of
direct auth parameters
2. Profile Name: output_out-to-cw (matches the output name)
3. Dual-Profile Structure:
  - Source profile handles initial STS web identity authentication
  - Main profile handles the AssumeRole with custom session name
4. Session Name: 625f0e4d-app-logs-out-to-cw (cluster_ID-CLF_Name-output_name)
5. No Direct AssumeRole in Vector Config: The AssumeRole is handled entirely through the
 AWS credentials file

## AWS IAM Configuration for Cross-Account Access

STS on AWS

### Initial Role (Cluster Account, e.g. 123456789012:role/app-logs)

The initial role in your cluster account needs:

1. **Trust Policy to ServiceAccount (STS/OIDC)**:
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "Federated": "arn:aws:iam::123456789012:oidc-provider/example.com/289hcj7"
            },
            "Action": "sts:AssumeRoleWithWebIdentity",
            "Condition": {
                "StringEquals": {
                    "oidc-provider/example.com/289hcj7:sub": "system:serviceaccount:openshift-logging:app-logs"
                }
            }
        }
    ]
}
```

2. **AssumeRole Permissions**:

```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": "sts:AssumeRole",
            "Resource": "arn:aws:iam::123456789012:role/app-logs"
        }
    ]
}
```

### Target Role (Cross-account, e.g. 987654321098:role/app-logs-cross)

The target role in the external account needs:

1. **Trust Policy to source Role**:
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "AWS": "arn:aws:iam::123456789012:role/app-logs"
            },
            "Action": "sts:AssumeRole",
            "Condition": {
                "StringEquals": {
                    "sts:ExternalId": "your-unique-external-id"
                }
            }
        }
    ]
}
```

2. **CloudWatch Permissions**:
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Action": [
                "logs:CreateLogGroup",
                "logs:CreateLogStream", 
                "logs:PutLogEvents",
                "logs:DescribeLogGroups",
                "logs:DescribeLogStreams"
            ],
            "Resource": "*"
        }
    ]
}
```

## Backward Compatibility
✅ **No breaking changes for existing users**
- New assumeRole property is optional
- Existing IAM Role configurations without AssumeRole continue to use `output-{outputName}` session names
- New session name format only applies to new AssumeRole configurations
- All existing CloudWatch outputs remain fully functional

## Technical Details
- Session names include cluster metadata for better traceability
- Prevents mixed authentication modes that cause Vector parsing errors
- Comprehensive test coverage for all scenarios

## Testing Verified
- ✅ Cross-account log forwarding working in live cluster
- ✅ Session name format verified: `625f0e4d-cp-logs-out-to-cw`
- ✅ AWS credentials properly separated by profile
- ✅ Backward compatibility confirmed for existing configurations
- ✅ All unit tests passing

This resolves authentication issues in cross-account scenarios while maintaining full backward compatibility for existing CloudWatch outputs.


api/observability/v1/output_types.go
```diff
@@ -416,6 +416,15 @@ type CloudwatchAuthentication struct {
...
 	IAMRole *CloudwatchIAMRole `json:"iamRole,omitempty"`
+
+	// AssumeRole specifies an additional IAM role to assume after the initial authentication.
+	// This enables cross-account log forwarding where the initial role (from IAMRole or AWSAccessKey)
+	// is used to authenticate, and then this role is assumed to access CloudWatch in another account.
+	//
+	// +kubebuilder:validation:Optional
+	// +nullable
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Cross-Account Assume Role"
+	AssumeRole *CloudwatchAssumeRole `json:"assumeRole,omitempty"`
```
jcantrill: Based upon the API sample from the comments this structure would need to be  child of IAMRole; the type dictates the viability of the additional auth details.

api/observability/v1/output_types.go
```diff
@@ -447,6 +456,30 @@ type CloudwatchAWSAccessKey struct {
...
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="External ID Secret"
+	ExternalID *SecretReference `json:"externalID,omitempty"`
+
+	// SessionName is an optional identifier for the assumed role session.
+	// If not provided, a default session name will be generated.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Pattern:=`^[a-zA-Z0-9_+=,.@-]{2,64}$`
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Session Name",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	SessionName string `json:"sessionName,omitempty"`
```
jcantrill: - What is the benefit in exposing a sessionName if one is generated by the collector?  
- Is it consumed someplace external to the collector

If we need to configure the session name we likely can craft one using some combination of the CLF name and output name so the user does not need to provide it

api/observability/v1/output_types.go
```diff
@@ -447,6 +456,30 @@ type CloudwatchAWSAccessKey struct {
...
 }
 
+type CloudwatchAssumeRole struct {
+	// RoleARN points to the secret containing the ARN of the role to assume for cross-account access.
+	//
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Assume Role ARN Secret"
+	RoleARN SecretReference `json:"roleARN"`
+
+	// ExternalID points to the secret containing the external ID required for assuming the role.
```
jcantrill, line 468: I'm uncertain if this needs to be maintained in a secret or not.  The [documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_common-scenarios_third-party.html) is mildly conflicting: 
```AWS does not treat the external ID as a secret``` 
but it also says : 
```The external ID for a role can be seen by anyone with permission to view the role```

File: config/rbac/serviceaccount.yaml [hunk 4]
```diff
@@ -0,0 +1,6 @@
+---
```
jcantrill: We intentionally do not provide an SA to be used with the collector because log collection requires explicit permissions from the admin which allow specific log_types to be collected.  This was a 6.x design decision to make admins more responsible for permissions and decisions and to support deploying collectors to alternate namespaces. Details [here](https://docs.redhat.com/en/documentation/red_hat_openshift_logging/6.3/html/configuring_logging/configuring-log-forwarding)

docs/reference/cloudwatch-cross-account.adoc
```diff
@@ -0,0 +1,221 @@
+= CloudWatch Cross-Account Log Forwarding
```
jcantrill: This should go here https://github.com/openshift/cluster-logging-operator/tree/master/docs/features/logforwarding/outputs and may possibly be joined with https://github.com/openshift/cluster-logging-operator/blob/master/docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc

docs/reference/cloudwatch-cross-account.adoc
```diff
@@ -0,0 +1,221 @@
...
+        authentication:
+          type: iamRole
+          iamRole:
+            roleARN:
+              secretName: initial-role-secret
+              key: role_arn
+            token:
+              from: serviceAccount
+          # Cross-account assume role configuration
+          assumeRole:
```
jcantrill: As defined here this breaks the current design since the honoring the stanza is dictated about what 'type' is chosen.    Since we need to rely upon the role associated with the serviceaccount which has a 1:1 with the collector, we need an alternate way for an administrator to express  the assumed role and externalID but also provide the iamRole that is associated with the service account.

docs/reference/cloudwatch-cross-account.adoc
```diff
@@ -0,0 +1,221 @@
...
+          roleARN:
+            secretName: dev-account-role
+            key: role_arn
+----
+
+== Security Considerations
+
+1. **External ID**: Always use external IDs for cross-account role assumptions
+2. **Least Privilege**: Grant minimum required permissions to assumed roles
+3. **Session Names**: Use descriptive session names for auditing
```
jcantrill: we should be able to generate this based upon the CLF metadata: namespace, name, output name

docs/reference/samples/clusterlogforwarder-cloudwatch-cross-account.yaml
```diff
@@ -0,0 +1,111 @@
+---
```
jcantrill: I assume this should be moved to follow the STS doc

internal/collector/cloudwatch/cloudwatch.credentials.tmpl
```diff
@@ -1,7 +1,28 @@
-{{- range . -}}
+{{range . -}}
+{{if .AssumeRoleARN -}}
+# Base profile for initial authentication
+[output_{{ .Name }}_base]
```
jcantrill: Base looks to be the same as the "direct authentication profile" implying to me it is not required

internal/generator/vector/output/cloudwatch/cloudwatch.go
```diff
@@ -181,3 +186,19 @@ func ParseRoleArn(auth *obs.CloudwatchAuthentication, secrets observability.Secr
 	}
 	return ""
 }
+
+// ParseAssumeRoleArn search for matching valid assume role ARN
+func ParseAssumeRoleArn(auth *obs.CloudwatchAuthentication, secrets observability.Secrets) string {
```
jcantrill, line 204: Some how I would expect this to be included in the ParseRoleArn function

jcantrill (commented):
/hold

api/observability/v1/output_types.go
apahim: Good point, fixing.

api/observability/v1/output_types.go
apahim: This is an optional field, but the thinking is we will have Collectors in multiple clusters assuming a role in a central account, so this makes it easier to tell them apart in CloudTrail. Unless we template it with the cluster id or something unique to the cluster where the CLF is in, I'd like to keep this flexibility.

api/observability/v1/output_types.go
apahim: I had the same doubt. Let's keep it in the secret to be in the safe side.

config/rbac/serviceaccount.yaml
apahim: Right, removing.

docs/reference/cloudwatch-cross-account.adoc
apahim: +1

docs/reference/cloudwatch-cross-account.adoc
apahim: fixing

docs/reference/cloudwatch-cross-account.adoc
apahim: I would like a cluster identifier here. The simplest way is to allow for user-provided SessionName so I can template it ahead of CLF creation.

internal/collector/cloudwatch/cloudwatch.credentials.tmpl
apahim: Checking.

internal/generator/vector/output/cloudwatch/cloudwatch.go
apahim: Yeah, ok, merging.

docs/reference/samples/clusterlogforwarder-cloudwatch-cross-account.yaml
apahim: Ack

docs/reference/cloudwatch-cross-account.adoc
apahim: Well, I'm templating the cluster id here and removing it from the API. I won't include the namespace though, as we can always have more than one namespace in the input while piping them into a single output.

api/observability/v1/output_types.go
```diff
@@ -431,6 +431,15 @@ type CloudwatchIAMRole struct {
 	// +kubebuilder:validation:Required
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Token"
 	Token BearerToken `json:"token"`
+
+	// AssumeRole specifies an additional IAM role to assume after the initial authentication.
```
jcantrill: Let's be even more specific here:

```
AssumeRole specifies an IAM role to assume after the initial authentication .  This enables 
cross-account log forwarding where the initial role or access keys associated with the collector service account
is used to authenticate, and then this role is assumed in order to access CloudWatch in another 
account.  This requires a trust relationship has been established between the roles and the 
assumed roles has been granted the necessary CloudWatch permissions.
```

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+              from: serviceAccount
+            # Cross-account assume role configuration
+            assumeRole:
+              roleARN:
+                secretName: cross-account-secret
+                key: assume_role_arn
+              externalID:
+                secretName: cross-account-secret
+                key: external_id
+              # Note: sessionName is automatically generated
```
jcantrill: we can remove this since there is no obvious reference to a session name in the CLF

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+    - name: cross-account-logs
+      inputRefs:
+        - application
+      outputRefs:
+        - target-account-cloudwatch
+----
+
+==== AssumeRole Authentication Fields
+
+===== `assumeRole.roleARN` (required)
```
jcantrill: Remove this section and the bit regarding externalID since it is abstracted away in the config generation

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+Reference to a secret containing the ARN of the role to assume in the target account.
+
+===== `assumeRole.externalID` (optional)
+Reference to a secret containing the external ID required for assuming the role. This provides additional security by ensuring only entities with knowledge of the external ID can assume the role.
+
+===== Session Name (automatic generation)
```
jcantrill: This section is probably useful to users of AWS but need to rework how it fits into this document since we are removing the sections above

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+Session names for assumed role sessions are automatically generated by the operator to provide meaningful identification in AWS CloudTrail logs for auditing purposes.
+
+**Automatic Session Name Generation:**
+The operator generates session names using cluster metadata with the format:
+- Primary format: `{clusterId}-{clfName}-{outputName}`
```
jcantrill: CLF is multi-namespace so this must be the following to guarantee uniqueness: `{clusterId}-{namespace}-{clfName}-{outputName}`

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+===== `assumeRole.externalID` (optional)
+Reference to a secret containing the external ID required for assuming the role. This provides additional security by ensuring only entities with knowledge of the external ID can assume the role.
+
+===== Session Name (automatic generation)
+Session names for assumed role sessions are automatically generated by the operator to provide meaningful identification in AWS CloudTrail logs for auditing purposes.
+
+**Automatic Session Name Generation:**
+The operator generates session names using cluster metadata with the format:
+- Primary format: `{clusterId}-{clfName}-{outputName}` 
+- Fallback format: `output-{outputName}` (if cluster metadata unavailable)
```
jcantrill: what scenario is clusterID not available?  If that is possible then I would think the format will be `{namespace}-{clfName}-{outputName}`

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+- Always truncated to 64 characters maximum (AWS requirement)
```
jcantrill: This is a little concerning.  Maybe we should consider something like:
* construct value from: `{clusterId}-{namespace}-{clfName}-{outputName}`
* Hash it and limit it to.... 8 chars? 6?
* Replace the last 8 or 6 chars with the hash?... or place it on front, or remove the namespace ?

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+            ],
+            "Resource": "*"
+        }
+    ]
+}
+----
+
+=== Cross-Account Security Considerations
+
+1. **External ID**: Always use external IDs for cross-account role assumptions
```
jcantrill, line 292: It is recommended to always use ...

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+2. **Least Privilege**: Grant minimum required permissions to assumed roles
+3. **Session Names**: Use descriptive session names for auditing in CloudTrail
```
jcantrill: We can remove since we are not providing a way for this to be configured

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+kind: Secret
+metadata:
+  name: cross-account-external-id
+  namespace: openshift-logging
+stringData:
+  external_id: "unique-random-string-generated-per-relationship-2024"
+  # Do not use predictable values like account IDs or role names
+```
+
+**External ID Requirements:**
```
jcantrill, line 337: These are validations to add to the API and we can enforce with CEL, at least the length and acceptable characters

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+   - Verify region configuration matches
+
+3. **Invalid Role ARN**
+   - Ensure role ARN format is correct: `arn:aws:iam::ACCOUNT:role/ROLE-NAME`
+   - Verify the role exists in the target account
+
+Check Vector pod logs for assume role related errors:
+[source,bash]
+----
+kubectl logs -n openshift-logging <vector-pod-name>
```
jcantrill: collector-pod-name

internal/collector/cloudwatch/cloudwatch.credentials.tmpl
```diff
@@ -1,7 +1,29 @@
-{{- range . -}}
+{{range . -}}
+{{if .AssumeRoleARN -}}
+# Source profile for initial web identity authentication
+[output_{{ .Name }}_source]
+web_identity_token_file={{ .WebIdentityTokenFile }}
+role_arn={{ .RoleARN }}
+role_session_name=output-{{ .Name }}-source
```
jcantrill, line 7: This will have to change based upon the final format of the session name or maybe we should be more clever in specific generation since this is tied to the serviceaccount

internal/collector/cloudwatch/cloudwatch.credentials.tmpl
```diff
@@ -1,7 +1,29 @@
...
+{{- if .ExternalID }}
+external_id={{ .ExternalID }}
+{{- end }}
+{{- if .SessionName }}
+role_session_name={{ .SessionName }}
+{{- else }}
+role_session_name=output-{{ .Name }}
+{{- end }}
+{{else -}}
+# Direct authentication profile
```
jcantrill, line 22: Is it possible to always generate a source_profile and then simply always generate an "assume role" kind of section regardless is its assumed or not?  This would eliminate duplication.

internal/collector/cloudwatch/cloudwatch.credentials.tmpl
```diff
@@ -1,7 +1,29 @@
...
+[output_{{ .Name }}_source]
+web_identity_token_file={{ .WebIdentityTokenFile }}
+role_arn={{ .RoleARN }}
+role_session_name=output-{{ .Name }}-source
+
+# Assume role profile for cross-account access
+[output_{{ .Name }}]
+source_profile=output_{{ .Name }}_source
+role_arn={{ .AssumeRoleARN }}
+{{- if .ExternalID }}
```
jcantrill, line 13: See elsewhere in the code on how to use a helpers/optional_pair to eliminate the need for some of these if/else

internal/collector/cloudwatch/cloudwatch.go
```diff
@@ -29,12 +30,48 @@ type CloudwatchWebIdentity struct {
...
+	AssumeRoleARN string
+	ExternalID    string
+	SessionName   string
+}
+
+// generateSessionName creates an intelligent session name using cluster metadata
+// Format: <cluster_ID>-<CLF_Name>-<output_name>
+func generateSessionName(reader client.Reader, clfName, outputName string) string {
+	// Try to get cluster ID for meaningful session names
+	_, clusterID, err := version.ClusterVersion(reader)
```
jcantrill: This information is already available during collector reconciliation and can be passed in during the vector config generation.  Additionally, the value passed in attempts to get the hosted cluster ID or falls back to the management cluster ID

internal/collector/cloudwatch/cloudwatch.go
```diff
@@ -29,12 +30,48 @@ type CloudwatchWebIdentity struct {
...
+		sessionName = clusterID[:min(8, len(clusterID))] + "-" + outputName
+		if len(sessionName) > 64 {
+			sessionName = sessionName[:64]
+		}
+	}
+	return sessionName
+}
+
+// min returns the minimum of two integers
+func min(a, b int) int {
```
jcantrill: This can be replaced by min function https://tip.golang.org/ref/spec#Min_and_max

internal/collector/cloudwatch/cloudwatch.go
```diff
@@ -52,9 +89,9 @@ func ReconcileAWSCredentialsConfigMap(k8sClient client.Client, reader client.Rea
 }
 
 // GenerateCloudwatchCredentialProfiles generates AWS CLI profiles for a credentials file from spec'd cloudwatch role ARNs and returns the formatted content as a string.
-func GenerateCloudwatchCredentialProfiles(outputs []obs.OutputSpec, secrets observability.Secrets) (string, error) {
+func GenerateCloudwatchCredentialProfiles(reader client.Reader, clfName string, outputs []obs.OutputSpec, secrets observability.Secrets) (string, error) {
```
jcantrill: The clusterID is available from the forwarder context which is passed in during the reconciliation of the configmap for CW credentials

internal/generator/vector/output/cloudwatch/cloudwatch.go
jcantrill: This still looks like it is more or less the same as ParseRoleArn which makes me believe there is a way to refactor the function to incorporate doing this parsing for either the primary role or the assume role

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
jcantrill: Actually, revisiting this, the external ID is hidden in a secret so we can only evaluate post admission in a validator

cahartma:
All changes are included in https://github.com/openshift/cluster-logging-operator/pull/3114
Once merged, we can link and close this.

PR #3095, opened 2025-08-12
Add functional tests for log collection edge cases

vparfonov:
### Description
This PR introduces three new functional tests for log collection. These tests are addressed to verify the edge case behavior of Vector's file fingerprinting mechanism.

The tests confirm that CLO correctly handles both very short (but complete) log lines and very long (but incomplete) log lines.

#### How Vector's file fingerprinting works
To understand the purpose of these tests, it's important to know how Vector's fingerprinting strategy works. 

**The default strategy is line-based**:  While Vector's documentation states the default `fingerprint.strategy` is `checksum`, the source code shows that this legacy strategy is internally mapped to behave exactly like the modern `first_lines_checksum` strategy (`FirstLinesChecksum`). Therefore, the effective default behavior is to identify files based on their first complete line(s).

**The meaning of "file too small"**: A log file is considered "too small to fingerprint" if it does not yet contain a complete line, which means a line terminated by a newline character (`\n`). A file can be very large in terms of byte size, but if it lacks a final `\n`, Vector will consider it incomplete and wait before processing it. Similarly, **an empty (0-byte)** file is also considered "too small" because it too lacks a complete line.

This line-based approach is more robust for logs because a "line" represents a complete "event". It prevents Vector from ingesting partial log entries that are still being written to disk. 

#### Three new functional test cases have been added:
Test 1: short but complete line
Writes a single-character line followed by a newline (a\n) to a log file. This confirms that a file is considered valid as long as it contains a complete line, regardless of its small byte size.

Test 2: large but incomplete line
Writes a long string (1000+ characters) to a log file without a terminating newline character, confirms Vector emits a `Currently ignoring file too small to fingerprint` warning and does not ingest the log. On next step, writes normal log message, confirms system works as expected: log event send to the output.

Test 3: empty file creation 
Creates a new, empty (0-byte) log file, confirm that Vector emits a `Currently ignoring file too small to fingerprint` warning. This confirms that the default fingerprinting strategy requires content (at least one complete line) and will not process empty files immediately upon creation, which is the desired behavior.

Additionally, updates FAQs in `docs/administration/troubleshooting.md`

/cc @Clee2691 @cahartma 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: https://issues.redhat.com/browse/LOG-7075
- Enhancement proposal:

### DIFF
--- a/docs/administration/troubleshooting.md
+++ b/docs/administration/troubleshooting.md
@@ -4,8 +4,9 @@
 
 ## Frequently Asked Questions (FAQs)
 
-1. I've made changes to my `ClusterLogForwarder` (CLF) instance but the collectors are not redeployed/updated. Why is that?
-    - There could be an issue with one of the inputs, outputs, or pipelines of the CLF. Check the CLF status in one of 2 ways:
+### 1. I've made changes to my `ClusterLogForwarder` (CLF) instance but the collectors are not redeployed/updated. Why is that?
+    
+   - There could be an issue with one of the inputs, outputs, or pipelines of the CLF. Check the CLF status in one of 2 ways:
         1. Streamed events of the CLF.
             
             ```$ oc describe clf --show-events=true```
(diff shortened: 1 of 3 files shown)


jcantrill:
> The meaning of "file too small": A log file is considered "too small to fingerprint" if it does not yet contain a complete line, which means a line terminated by a newline character (\n). A file can be very large in terms of byte size, but if it lacks a final \n, Vector will consider it incomplete and wait before processing it.

The only situation where I can think this would exists is if the log producer was in the middle of writing while at the same time the collector tries to read the line.  This seems like a reasonable assumption given there is no coordination between read and write processing.  The container log driver will always terminate log events with a new line; this is the nature of how CRIO is designed with the format of each entry being:
```
<TIMESTAMP> F|P STDERR|STDOUT <MESSAGE>
```
Vector additionally only knows about containers on the node because of the information it has received from the kublet.  It would interesting to know "how" vector reads logs as in is it polling or doing a round robin of the files it knows about or is there some event based mechanism that triggers its attempt to read a file.  Either way there is a possible timing issue where the current line in the file does not meet the criteria to read it and subsequently produces the warning message.

Can you add an entry to our docs to capture the information about the message in the logs?  I was thinking maybe it should be in the [troubleshooting](https://github.com/openshift/cluster-logging-operator/blob/master/docs/administration/troubleshooting.md) doc where there is an FAQ

docs/administration/troubleshooting.md
```diff
@@ -4,12 +4,36 @@
...
             ```$ oc describe clf --show-events=true```
 
         2. Check the `status` section of the CLF instance `YAML` output.
         
-            ```$ oc get clf -oyaml```
\ No newline at end of file
+            ```$ oc get clf -oyaml```
+
+### 2. Warning Message: `Currently ignoring file too small to fingerprint`
+If you see this warning message in your logs, here is an explanation of what it means and what actions you should take.
```
jcantrill, line 19: We should also add in how the behavior of conmon impact collection and can cause the collector to generate the message.  Recall from the slack thread that conmon's behavior is to create an empty file.  This implies that if the container never produces a log message or the collector is checking the file before it has produced a log message then the warning message will appear.

vparfonov: done

jcantrill (commented):
/lgtm

PR #3097, opened 2025-08-13
Allow specification of a maximum line size to be applied after merging

vparfonov:
### Description
This PR adds the max_merged_line_bytes setting to the Vector collector configuration.
More information: [kubernetes_logs/#max_merged_line_bytes](https://vector.dev/docs/reference/configuration/sources/kubernetes_logs/#max_merged_line_bytes)

Currently, this parameter is not configurable via the Cluster Logging Forwarder spec and is hardcoded to 65536 bytes — which is twice the default value of [max_line_bytes](https://vector.dev/docs/reference/configuration/sources/kubernetes_logs/#max_line_bytes).

Using it helps prevent Vector from failing with the following error:
`vector::topology: An error occurred that Vector couldn't handle: failed to encode record: BufferTooSmall.
`

/cc 
/assign 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: https://issues.redhat.com/browse/LOG-7347
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/conf/complex.toml
+++ b/internal/generator/vector/conf/complex.toml
@@ -324,6 +324,7 @@ source = '''
 [sources.input_infrastructure_container]
 type = "kubernetes_logs"
 max_read_bytes = 3145728
+max_merged_line_bytes = 65536
 glob_minimum_cooldown_ms = 15000
 auto_partial_merge = true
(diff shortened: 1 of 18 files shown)


vparfonov:
Close for now until found better solution

PR #3101, opened 2025-08-29
LOG-7572: NetworkPolicy for ClusterLogForwarder

Clee2691:
### Description
This PR adds a default `NetworkPolicy` to allow all ingress and egress traffic for the collector pods.

This ensures logging continues to function in hardened environments that use `AdminNetworkPolicy` (ANP). 

To be effective against a cluster-wide Deny rule, this policy requires a corresponding high-priority `ANP` that delegates traffic decisions (action: `Pass`) for the collector pods.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-7572

### DIFF
--- a/api/observability/v1/output_types.go
+++ b/api/observability/v1/output_types.go
@@ -744,7 +744,7 @@ type Kafka struct {
 	Brokers []BrokerURL `json:"brokers,omitempty"`
 }
 
-// +kubebuilder:validation:XValidation:rule="self == '' || (isURL(self) && (self.startsWith('tcp://') || self.startsWith('tls://')))",message="each broker must be a valid URL with a tcp or tls scheme"
+// +kubebuilder:validation:XValidation:rule="isURL(self) && (self.startsWith('tcp://') || self.startsWith('tls://'))",message="each broker must be a valid URL with a tcp or tls scheme"
 type BrokerURL string
 
 type LokiTuningSpec struct {
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -78,6 +78,9 @@ rules:
   verbs:
   - create
   - delete
+  - list
+  - get
+  - watch
 - apiGroups:
   - observability.openshift.io
   resources:
--- a/docs/features/collection.adoc
+++ b/docs/features/collection.adoc
@@ -95,6 +95,7 @@ a|
 
 |https://issues.redhat.com/browse/LOG-3270[TLS Security Profile Compliance]
 |Comply with OCP cluster-wide cryptographic profiles for internal communication and allow configuration of outbound connection profiles. See link:./tls_security_profile.adoc[details]
+|https://issues.redhat.com/browse/LOG-7571[Network Policy]| Network policy in place for the collectors that allows all egress and ingress.
 |======
 
 === Tuning
--- a/docs/features/network_policy.adoc
+++ b/docs/features/network_policy.adoc
@@ -1,0 +1,104 @@
+= NetworkPolicy for Collector Pods
+
+The Cluster Logging Operator automatically creates and manages a `NetworkPolicy` for its collector pods to ensure they function in restrictive network environments, even if a cluster-wide `AdminNetworkPolicy` would otherwise block their traffic.
+
+== Overview
+
+When a `ClusterLogForwarder` is deployed, the operator creates a permissive `NetworkPolicy` that allows all ingress and egress traffic for the collector pods. This ensures that log collection can function properly even when:
+
+* Restrictive default `NetworkPolicies` are in place
+* `AdminNetworkPolicy` configurations limit pod communications
+* Namespace-level network restrictions are applied
+
+The `NetworkPolicy` is automatically created and removed along with the collector deployment lifecycle. 
+
+The `NetworkPolicy` can directly be edited by the cluster administrator and won't be reconciled by the operator upon updates.
+
+== NetworkPolicy Configuration
+
+The operator creates a `NetworkPolicy` for the collector with the following characteristics:
+
+```yaml
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: <COLLECTOR-INSTANCE-NAME>
+  namespace: <COLLECTOR-NAMESPACE>
+  labels:
+    app.kubernetes.io/name: vector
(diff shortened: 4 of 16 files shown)


api/observability/v1/output_types.go
```diff
@@ -744,7 +744,7 @@ type Kafka struct {
 	Brokers []BrokerURL `json:"brokers,omitempty"`
 }
 
-// +kubebuilder:validation:XValidation:rule="self == '' || (isURL(self) && (self.startsWith('tcp://') || self.startsWith('tls://')))",message="each broker must be a valid URL with a tcp or tls scheme"
+// +kubebuilder:validation:XValidation:rule="self == ” || (isURL(self) && (self.startsWith('tcp://') || self.startsWith('tls://')))",message="each broker must be a valid URL with a tcp or tls scheme"
```
jcantrill: You will need to revert this change as it will break deployment; I fought this as well and thought it was resolved.  Try setting it to `\"\'` to see if it will "stick".  I also do not recall why this is needed since: "brokers" is optional and I would expect an empty string to be an invalid URL

config/rbac/role.yaml
```diff
@@ -78,6 +78,11 @@ rules:
   verbs:
   - create
   - delete
+  - watch
```
jcantrill: can we just replace these with "*" ? following my comment below... we could maybe remove update and patch since we do not intend to control anything  but create and delete

docs/features/collection.adoc
```diff
@@ -95,6 +95,7 @@ a|
 
 |https://issues.redhat.com/browse/LOG-3270[TLS Security Profile Compliance]
 |Comply with OCP cluster-wide cryptographic profiles for internal communication and allow configuration of outbound connection profiles. See link:./tls_security_profile.adoc[details]
+|Network Policy| Network policy in place for the collectors that allows all egress and ingress.
```
jcantrill: Can we link to the epic?

docs/features/network_policy.adoc
```diff
@@ -0,0 +1,102 @@
+= NetworkPolicy for Collector Pods
+
+The Cluster Logging Operator automatically creates and manages a `NetworkPolicy` for its collector pods to ensure they function in restrictive network environments, even if a cluster-wide `AdminNetworkPolicy` would otherwise block their traffic.
```
jcantrill, line 3: This reminds me that we need an ingress NP for LFME since it is scraped by metrics.  Ideally this could simply be added to the manifest which means maybe we can wait?  I would be nice not to need to manage this policy

docs/features/network_policy.adoc
```diff
@@ -0,0 +1,102 @@
...
+
+== Overview
+
+When a `ClusterLogForwarder` is deployed, the operator creates a permissive `NetworkPolicy` that allows all ingress and egress traffic for the collector pods. This ensures that log collection can function properly even when:
+
+* Restrictive default `NetworkPolicies` are in place
+* `AdminNetworkPolicy` configurations limit pod communications
+* Namespace-level network restrictions are applied
+
+The `NetworkPolicy` is automatically created, updated, and removed along with the collector deployment lifecycle.
```
jcantrill: Let's mention this resource can be modified by administrators

docs/reference/operator/api_observability_v1.adoc
```diff
@@ -1238,15 +1238,15 @@ If segments contain characters outside of this range, the segment must be quoted
 
 Examples:
 
-- `.kubernetes.namespace_name`
+- `.kubernetes.namespace_id`
```
jcantrill, line 1241: Consider adding to a separate PR

internal/factory/network_policy.go
```diff
@@ -0,0 +1,39 @@
+package factory
+
+import (
+	"github.com/openshift/cluster-logging-operator/internal/constants"
+	"github.com/openshift/cluster-logging-operator/internal/runtime"
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NewNetworkPolicy creates a NetworkPolicy for clusterlogforwarder daemonsets.
```
jcantrill: Nit.  I would just say clfs .  We do support deployments too

internal/reconcile/networkpolicy.go
```diff
@@ -0,0 +1,30 @@
...
+	log "github.com/ViaQ/logerr/v2/log/static"
+	"github.com/openshift/cluster-logging-operator/internal/runtime"
+	networkingv1 "k8s.io/api/networking/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+// NetworkPolicy reconciles a NetworkPolicy to the desired spec returning an error
+// if there is an issue creating or updating to the desired state
+func NetworkPolicy(k8Client client.Client, desired *networkingv1.NetworkPolicy) error {
```
jcantrill, line 17: Allowing NP edit may affect this logic here.  I defer to your investigation

api/observability/v1/output_types.go
Clee2691: This was adding it by `make bundle` . When I use `make lint`, it reports that `make bundle` is needed.

docs/reference/operator/api_observability_v1.adoc
Clee2691: This was added when i used `make docs`. Unsure how I should proceed here.

docs/features/network_policy.adoc
Clee2691: Yes I think that would be possible since we can just target the LFME pods

jcantrill: Let's add a JIRA to deploy policy based upon the occurrence of an LFME deployment.  I think its reasonable to do the same and I don't want to have to wait for the OLM changes to land

docs/reference/operator/api_observability_v1.adoc
jcantrill: leave as-is

docs/features/network_policy.adoc
Clee2691: Added [JIRA](https://issues.redhat.com/browse/LOG-7675) for LFME NP

PR #3107, opened 2025-09-05
WIP - Enable opeator capabilities

jcantrill:
### Description
This PR
*  suggests a path by which we could enable operator capabilities using a configmap to selectively enable features
*  Tries to compartmentalize NetworkPolicy reconciliation


alanconway:
Is this for preview features? If not, why not just use the may YAML config to enable features? What's the benefit of a hidden second layer of configuration?

PR #3111, opened 2025-09-15
LOG-7622: The '.kubernetes.container_iostream' is not getting pruned in Logging 6.3 for Splunk output

Clee2691:
### Description
This PR resolves a bug where the prune transform failed to remove the `.kubernetes.container_iostream` field from log events. The removal logic has been corrected to ensure this field is successfully pruned.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-7622


internal/generator/vector/conf/complex.toml
```diff
@@ -29,6 +29,7 @@ type = "remap"
 inputs = ["input_audit_host"]
 source = '''
   . = {"_internal": .}
+  if exists(._internal.stream) {._internal.kubernetes.container_iostream = ._internal.stream}
```
jcantrill: This will never be a value from audit sources since they are file sources and kubernetes.   Can this be "removed" without  a large change?  If not, then I'm good with leaving it.

Clee2691: That is why I added the `exists()` since they won't have a value and will not set that value. I think I can add a condition however. I will play around with it.

cahartma (commented):
/lgtm

PR #3114, opened 2025-09-17
LOG-7688: Enable cloudwatch cross-account log forwarding

cahartma:
### Description
Building off the work contributed by @apahim.   Expanding the assumeRole functionality to any cloudwatch auth configuration.    Preparing auth modules for the s3 sink addition

/cc @Clee2691 @vparfonov 
/assign @jcantrill

### Links
- original issue: https://issues.redhat.com/browse/LOG-7505
- original PR:  https://github.com/openshift/cluster-logging-operator/pull/3093
- task: https://issues.redhat.com/browse/LOG-7688
- epic: https://issues.redhat.com/browse/LOG-7687


cahartma:
Forgot the documentation.

api/observability/v1/output_types.go
```diff
@@ -445,6 +454,32 @@ type CloudwatchAWSAccessKey struct {
...
+	RoleARN SecretReference `json:"roleARN"`
+
+	// ExternalID is the external ID to match when assuming the role.
+	// This is optional and is an additional security measure used to ensure that only the intended
+	// entity can assume the role.
+	//
+	// +kubebuilder:validation:Optional
+	// +nullable
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="External ID"
+	ExternalID string `json:"externalID,omitempty"`
```
jcantrill, line 480: - Doesn't this have a max length we should validate here?
- Any concerns this needs to be hidden in a secret?  Is it secret?

I think the document has some good points why this should be a secretRef

api/observability/v1/output_types.go
```diff
@@ -744,7 +779,7 @@ type Kafka struct {
 	Brokers []BrokerURL `json:"brokers,omitempty"`
 }
 
-// +kubebuilder:validation:XValidation:rule="isURL(self) && (self.startsWith('tcp://') || self.startsWith('tls://'))",message="each broker must be a valid URL with a tcp or tls scheme"
+// BrokerURL +kubebuilder:validation:XValidation:rule="self == ” || (isURL(self) && (self.startsWith('tcp://') || self.startsWith('tls://')))",message="each broker must be a valid URL with a tcp or tls scheme"
```
jcantrill: might need a rebase or revert.  I thought we resolved this issue. where we do not need to check "self"

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+2. **Role Assumption**: After initial authentication, the collector assumes a role in the target AWS account  
+3. **Log Forwarding**: Using the assumed role's credentials, the collector forwards logs to CloudWatch in the target account
+
+=== Cross-Account Configuration
+
+==== Prerequisites
+
+Before configuring cross-account log forwarding, ensure:
+
+1. **Trust Relationship**: The target account's IAM role must trust your cluster's initial role
```
jcantrill: Is this the cluster's initial role or the initial role spec'd in the output?

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+        - target-account-cloudwatch
+----
+
+==== AssumeRole Authentication Fields
+
+===== `assumeRole.roleARN` (required)
+Reference to a secret containing the ARN of the role to assume in the target account.
+
+===== `assumeRole.externalID` (optional)
+Reference to a secret containing the external ID required for assuming the role. This provides additional security by ensuring only entities with knowledge of the external ID can assume the role.
```
jcantrill: This description makes me think this should also be a secret.

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+        }
+    ]
+}
+----
+
+=== Cross-Account Security Considerations
+
+1. **External ID**: Always use external IDs for cross-account role assumptions
+2. **Least Privilege**: Grant minimum required permissions to assumed roles
+3. **Session Names**: Use descriptive session names for auditing in CloudTrail
```
jcantrill: we can remove this if we are generating it

docs/features/logforwarding/outputs/cloudwatch-sts-forwarding.adoc
```diff
@@ -153,6 +153,240 @@ spec:
...
+4. **Secret Management**: Protect secrets containing role ARNs and external IDs
+5. **Monitoring**: Monitor CloudTrail for AssumeRole activities
+
+==== External ID Storage in Secrets
+
+The cluster-logging-operator stores external IDs in Kubernetes secrets for the following security reasons:
```
jcantrill: This statement conflicts with the implementation

internal/collector/cloudwatch/credentials.go
```diff
@@ -0,0 +1,185 @@
...
+		sessionName = clusterID[:min(8, len(clusterID))] + "-" + outputName
+		if len(sessionName) > 64 {
+			sessionName = sessionName[:64]
+		}
+	}
+	return sessionName
+}
+
+// min returns the minimum of two integers
+func min(a, b int) int {
```
jcantrill: golang has a min function which can be used in place of this

internal/controller/observability/collector.go
```diff
@@ -29,6 +29,23 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+// ReconcileCollector adds scc, owner, options, context,
```
jcantrill: Can we make this more succinct regarding the description.  The verbosity lends itself to skew when the function is updated but the comment is not. All the information here can be understood by reading the calls and if the subsequent calls do not clearly identify intent then we should modify those function names.  I would say this comment should be something like:

```
ReconcileCollector generates and deploys the resources (i.e. ConfigMap, Secrets, Deployments, etc) to realize the spec of the ClusterLogForwarder
```

internal/controller/observability/collector.go
```diff
@@ -132,29 +151,29 @@ func ReconcileCollector(context internalcontext.ForwarderContext, pollInterval,
 	}
 
 	if err := reconcileWorkload(context.Client, context.Forwarder.Namespace, trustedCABundle, ownerRef); err != nil {
-		log.Error(err, "Error reconciling the deployment of the collector")
+		log.V(3).Error(err, "Error reconciling the deployment of the collector")
```
jcantrill: Do we really want to modify the verbocity to hide this issue if there is a problem with reconciliation?

internal/pkg/generator/forwarder/generator.go
```diff
@@ -50,7 +50,7 @@ func Generate(clfYaml string, debugOutput bool, client client.Client) (string, e
 	op := framework.Options{}
 	//k8shandler.EvaluateAnnotationsForEnabledCapabilities(forwarder.Annotations, op)
 	op[framework.ClusterTLSProfileSpec] = tls.GetClusterTLSProfileSpec(nil)
-	controller.SetKubeCacheOption(forwarder.Annotations, op)
+	controller.EvaluateTechPreviewAnnotations(forwarder.Annotations, op)
```
jcantrill: I would prefer these changes to go in separately from roleArn functionality

internal/validations/observability/outputs/cloudwatch.go
```diff
@@ -4,24 +4,72 @@ import (
...
+	if auth.Type == obs.CloudwatchAuthTypeIAMRole {
+		roleString = secrets.AsString(&auth.IAMRole.RoleARN)
+	}
+	return findSubstring(roleString)
+}
+
+// ParseAssumeRoleArn search for valid AWS assumeRole arn
+func ParseAssumeRoleArn(auth *obs.CloudwatchAuthentication, secrets observability.Secrets) string {
+	var roleString string
+	if auth.IAMRole != nil && auth.IAMRole.AssumeRole != nil {
```
jcantrill: elsewhere we do a switch on the type.  Any concerns of unintentionally overwriting the value

api/observability/v1/output_types.go
cahartma: I changed it to a string based on my research, as well as it is much simpler code and implementation.    AWS goes far enough to say its an antipattern and becomes a security risk to treat it as a secret.   The external ID for a role can be seen by anyone with permission to view the role.     We should chat further.

cahartma: max length is 1224 chars.   I didn't plan on a validation for this ID, but it could be added if you think its necessary.   It is validated when you created the trust-policy for the role at AWS.

internal/controller/observability/collector.go
cahartma: Yep, I didn't mean to leave this in.  I have a refactor that breaks up this monolith function.

internal/validations/observability/outputs/cloudwatch.go
cahartma: This is called from within a switch, so the type is already determined    The field path should prevent any confusion.   Maybe if both types were present?  a functional test should be able to determine if this works or not.

api/observability/v1/output_types.go
jcantrill: > I changed it to a string based on my research, as well as it is much simpler code and implementation. AWS goes far enough to say its an antipattern and becomes a security risk to treat it as a secret. The external ID for a role can be seen by anyone with permission to view the role. We should chat further.

Following up our conversation, I questioned the need to make this secret so this justification is satisfiesfactory

> I didn't plan on a validation for this ID, but it could be added if you think its necessary

Per our discussion, we know the required format and adding a declarative validation is "cheap" so if we can save a customer from a mismatch early it is to their benefit.  We should have examples of regex validations in output_types.go

api/observability/v1/output_types.go
cahartma: The issue is the validation needs to be moved to the field, not the type.     The type does not need any validation, and the e2e test fails.

internal/api/observability/output_types.go
```diff
@@ -146,7 +148,7 @@ func SecretReferences(o obsv1.OutputSpec) []*obsv1.SecretReference {
 		}
 	case obsv1.OutputTypeSyslog:
 	default:
-		log.V(0).Error(OutputTypeUnknown(o.Type), "Found unsupported output type while gathering secret names")
+		log.V(3).Error(OutputTypeUnknown(o.Type), "Found unsupported output type while gathering secret names")
```
jcantrill: Don't wish to hold this PR but this should be reverted to not hide the reason for terminating

PR #3115, opened 2025-09-17
LOG-7620: The '.openshift.sequence' not getting pruned in Logging 6.3

Clee2691:
### Description
This PR resolves a bug where the prune transform failed to remove the `.openshift.sequence` field from log events. The removal logic has been corrected to ensure this field is successfully pruned.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-7620

### DIFF
--- a/internal/generator/vector/filter/openshift/viaq/v1/normalize.go
+++ b/internal/generator/vector/filter/openshift/viaq/v1/normalize.go
@@ -149,5 +149,4 @@ if !exists(._internal.structured) {
 if exists(._internal.openshift) {.openshift = ._internal.openshift}
 if exists(._internal.dedot_openshift_labels) {.openshift.labels = del(._internal.dedot_openshift_labels) }
 `
-	SetOpenShiftSequence = `if ._internal.log_type != "receiver" { ._internal.openshift.sequence = to_unix_timestamp(now(), unit: "nanoseconds")}`
 )
--- a/internal/generator/vector/input/receiver_http_audit.toml
+++ b/internal/generator/vector/input/receiver_http_audit.toml
@@ -34,4 +34,5 @@ source = '''
   ._internal.log_type = "audit"
   ._internal.hostname = get_env_var("VECTOR_SELF_NODE_NAME") ?? ""
   ._internal.openshift = { "cluster_id": "${OPENSHIFT_CLUSTER_ID:-}"}
+  ._internal.openshift.sequence = to_unix_timestamp(now(), unit: "nanoseconds")
 '''
--- a/test/functional/filters/prune/prune_filter_test.go
+++ b/test/functional/filters/prune/prune_filter_test.go
@@ -15,6 +15,8 @@ import (
 	"github.com/openshift/cluster-logging-operator/test/helpers/kafka"
 	"github.com/openshift/cluster-logging-operator/test/helpers/loki"
 	"github.com/openshift/cluster-logging-operator/test/helpers/rand"
+	"github.com/openshift/cluster-logging-operator/test/helpers/splunk"
+	"github.com/openshift/cluster-logging-operator/test/helpers/types"
 	v1 "k8s.io/api/core/v1"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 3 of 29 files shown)


internal/generator/vector/filter/openshift/viaq/v1/normalize.go
```diff
@@ -149,5 +149,5 @@ if !exists(._internal.structured) {
 if exists(._internal.openshift) {.openshift = ._internal.openshift}
 if exists(._internal.dedot_openshift_labels) {.openshift.labels = del(._internal.dedot_openshift_labels) }
 `
-	SetOpenShiftSequence = `if ._internal.log_type != "receiver" { ._internal.openshift.sequence = to_unix_timestamp(now(), unit: "nanoseconds")}`
+	SetOpenShiftSequence = `if ._internal.log_type != "receiver" && exists(._internal.openshift.sequence) { ._internal.openshift.sequence = to_unix_timestamp(now(), unit: "nanoseconds")}`
```
jcantrill: Why are we replacing the value?

test/functional/filters/prune/prune_filter_test.go
```diff
@@ -77,6 +80,52 @@ var _ = Describe("[Functional][Filters][Prune] Prune filter", func() {
...
+
+			secret = runtime.NewSecret(f.Namespace, secretKey.SecretName,
+				map[string][]byte{
+					secretKey.Key: functional.HecToken,
+				},
+			)
+			f.Secrets = append(f.Secrets, secret)
+
+			Expect(f.Deploy()).To(BeNil())
+			time.Sleep(90 * time.Second)
```
jcantrill: Do we really need a 90s sleep to validate? I thought the deploy and read operations had polling to remove the need for sleeps

internal/generator/vector/input/receiver_http_audit.toml
```diff
@@ -34,4 +34,5 @@ source = '''
   ._internal.log_type = "audit"
   ._internal.hostname = get_env_var("VECTOR_SELF_NODE_NAME") ?? ""
   ._internal.openshift = { "cluster_id": "${OPENSHIFT_CLUSTER_ID:-}"}
+  ._internal.openshift.sequence = ""
```
jcantrill: why do we need to initialize this to empty spaces instead of the actual value?

test/functional/filters/prune/prune_filter_test.go
Clee2691: We still need to wait for the Splunk service to finish setting up before writing/reading from it. 
I just looked and see that our splunk functional tests have a new function `WaitOnSplunk()` that it uses to ensure the service is up. I will refactor to that.

internal/generator/vector/filter/openshift/viaq/v1/normalize.go
Clee2691: This value is set during the pipeline transform and is done **AFTER** the prune filter. My fix here is to initialize `.openshift.sequence` so that any subsequent pruning of the field will ensure this field is removed. 

I replace the value to keep the original behavior of setting the sequence during the pipeline transform. I can set the value during initialization as probably the delay between initialization -> pipeline transform is minimal?

internal/generator/vector/input/receiver_http_audit.toml
Clee2691: See [above](https://github.com/openshift/cluster-logging-operator/pull/3115#discussion_r2356856233) comment.

test/helpers/splunk/splunk.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"time"
+
+	"github.com/openshift/cluster-logging-operator/test/framework/functional"
+	. "github.com/onsi/gomega"
+)
+
+// WaitOnSplunk waits for Splunk to be ready by checking HEC health and service status
+func WaitOnSplunk(f *functional.CollectorFunctionalFramework) {
+	time.Sleep(20 * time.Second)
```
jcantrill, line 12: Do we still require the sleeps here and below?  Isn't the intent here to poll and not require a sleep?

Clee2691: Yes we do still need the sleeps for splunk to get setup. The polling of read/write returns an error if the splunk server isn't up which fails the test.

cahartma (commented):
/lgtm

PR #3122, opened 2025-09-26
LOG-7771: Enable LFME 'opt-in' NetworkPolicy

Clee2691:
### Description
This PR introduces an API to enable an optional `NetworkPolicy` for the `LogFileMetricExporter`.

Changes
  - Added NetworkPolicy field to LogFileMetricExporter CRD spec for opt-in network policy creation
  - Enhanced network policy factory to handle LFME-specific network policies
  - Updated LFME controller to respect the new NetworkPolicy configuration
  - Added comprehensive tests for the new network policy functionality
  - Currently only the `AllowIngressMetrics` and `AllowAllIngressEgress` rule set is supported for the LFME.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-7771


api/logging/v1alpha1/log_file_metrics_exporter_types.go
```diff
@@ -24,8 +24,31 @@ type LogFileMetricExporterSpec struct {
...
+	// Define the Network Policy for the LogFileMetricExporter
+	// +nullable
+	// +optional
+	//+operator-sdk:csv:customresourcedefinitions:type=spec,displayName="LogFileMetricExporter Network Policy",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	NetworkPolicy *NetworkPolicy `json:"networkPolicy,omitempty"`
+}
+
+type NetworkPolicy struct {
+	// NetworkPolicyRuleSetType is the type of network policy rule set to use
+	// +optional
```
jcantrill: I believe we need to make this required, declare it as an enum, and add "AllowAll" (name to match collector???)

internal/constants/constants.go
```diff
@@ -51,6 +51,7 @@ const (
 	VectorName                 = "vector"
 	KibanaName                 = "kibana"
 	LogfilesmetricexporterName = "logfilesmetricexporter"
+	ExporterPortName           = "exporter-port"
```
jcantrill: This is isolated to the LFME and lets be more explicit and imply call it "metrics".  It is the name of the port so we do not need to add 'port'.  I assume we could reuse the name for the collector as well

internal/factory/network_policy.go
```diff
@@ -1,41 +1,47 @@
...
 
-	// Configure the NetworkPolicy spec
-	np.Spec = networkingv1.NetworkPolicySpec{
-		PodSelector: metav1.LabelSelector{
-			MatchLabels: podSelector,
-		},
-		PolicyTypes: policyTypes,
-	}
+	npBuilder := runtime.NewNetworkPolicyBuilder(np)
+	npBuilder.WithPodSelector(podSelector)
```
jcantrill: Since this is a builder using a fluent style you should be able to chain this call into the previous statement

internal/metrics/logfilemetricexporter/metric_exporter_test.go
```diff
@@ -119,10 +123,93 @@ var _ = Describe("Reconcile LogFileMetricExporter", func() {
...
+			// Reconcile the LogFileMetricExporter
+			Expect(Reconcile(lfmeInstance, reqClient, reader, utils.AsOwner(lfmeInstance))).To(Succeed())
+
+			// Verify NetworkPolicy was not created
+			err := reqClient.Get(context.TODO(), networkPolicyKey, networkPolicyInstance)
+			Expect(errors.IsNotFound(err)).To(BeTrue(), "NetworkPolicy should not exist when not specified")
+		})
+
+		Context("when NetworkPolicy is specified", func() {
+			It("should successfully create and configure the NetworkPolicy for AllowIngressMetrics ruleset", func() {
```
jcantrill, line 162: Reviewing the AC in a compliment to me other comment, I see we did not specify "AllowAll" but I believe we should in order to be similar with intent of the collector. This also is should be an escape hatch should we not get the ruleset correct

internal/network/network_policy_test.go
```diff
@@ -148,15 +151,21 @@ var _ = Describe("Reconcile NetworkPolicy", func() {
...
+				networkingv1.PolicyTypeEgress,
 				networkingv1.PolicyTypeIngress,
 			}
 			Expect(policyInstance.Spec.PolicyTypes).To(Equal(expectedPolicyTypes))
 
-			// Verify ingress rules allow all traffic (empty rule)
+			// Verify ingress rules allow only the named metrics port
 			Expect(policyInstance.Spec.Ingress).To(HaveLen(1))
-			Expect(policyInstance.Spec.Ingress[0]).To(Equal(networkingv1.NetworkPolicyIngressRule{}))
+			Expect(policyInstance.Spec.Ingress[0].Ports).To(HaveLen(1))
```
jcantrill: I think we should be able to collapse these expectations to compare against an ingress spec

internal/runtime/network_policy.go
```diff
@@ -0,0 +1,307 @@
...
+// WithPodSelectorExpression sets the pod selector using match expressions
+func (builder *NetworkPolicyBuilder) WithPodSelectorExpression(matchExpressions []metav1.LabelSelectorRequirement) *NetworkPolicyBuilder {
+	builder.NetworkPolicy.Spec.PodSelector = metav1.LabelSelector{
+		MatchExpressions: matchExpressions,
+	}
+	return builder
+}
+
+// WithPolicyTypes sets the policy types (Ingress, Egress, or both)
+func (builder *NetworkPolicyBuilder) WithPolicyTypes(policyTypes []networkingv1.PolicyType) *NetworkPolicyBuilder {
```
jcantrill: Consider renaming to "WithPolicies"... you could even make it variable...

internal/runtime/network_policy.go
```diff
@@ -0,0 +1,307 @@
...
+	builder.NetworkPolicy.Spec.PolicyTypes = policyTypes
+	return builder
+}
+
+// WithIngressPolicyType adds Ingress to the policy types
+func (builder *NetworkPolicyBuilder) WithIngressPolicyType() *NetworkPolicyBuilder {
```
jcantrill, line 32: Isn't this the same as "AllowAllIngress" which is covered by another function?  I would say unless we are giving it an arg to add to ingress policy there is no need for this function.  Same comment applies to egress

internal/runtime/network_policy.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+// IngressRuleBuilder helps build ingress rules fluently
+type IngressRuleBuilder struct {
+	rule                 *networkingv1.NetworkPolicyIngressRule
+	networkPolicyBuilder *NetworkPolicyBuilder
+}
+
+// NewIngressRule starts building a new ingress rule
+func (builder *NetworkPolicyBuilder) NewIngressRule() *IngressRuleBuilder {
```
jcantrill, line 82: Consider 'AddIngressRule'

internal/runtime/network_policy.go
```diff
@@ -0,0 +1,307 @@
...
+	builder.WithIngressPolicyType()
+	rule := &networkingv1.NetworkPolicyIngressRule{}
+	return &IngressRuleBuilder{
+		rule:                 rule,
+		networkPolicyBuilder: builder,
+	}
+}
+
+// FromPodSelector adds a pod selector to the ingress rule
+func (irb *IngressRuleBuilder) FromPodSelector(matchLabels map[string]string) *IngressRuleBuilder {
```
jcantrill: Unless we are using this I would remove for now. Comment applies to any of the other functions as well

internal/runtime/network_policy.go
Clee2691: This is specific to the `NetworkPolicy` resource and they are called `PolicyTypes`. There are only 2 `PolicyTypes` and I have other builder methods to add just those specific types. So I don't think this needs to be changed.

internal/runtime/network_policy.go
Clee2691: These are not the same. If you just specify the `PolicyType` without an actual rule, even an empty one, it means deny all (ingress, egress). 

So here, adding `Ingress` as the policy type without `networkingv1.NetworkPolicyIngressRule{}` means DENY all ingress. Same applies for `Egress`

internal/runtime/network_policy.go
Clee2691: I think this should stay as `NewIngressRule` because it is a builder with `End()` being the actual "adding" of the ingress rule. I think this makes it more explicit that this returns an `IngressRuleBuilder` and not the final `Ingress` rule.

internal/constants/constants.go
Clee2691: Not to be confused with the [collecter metrics port name.](https://github.com/openshift/cluster-logging-operator/blob/9396fad2ef2cb6fb724fadd20b6db0430ad2f6dc/internal/collector/collector.go#L28)... But in the past I see it was named `logfile-metrics`. I can go with that.

internal/runtime/network_policy.go
Clee2691: I've decided to remove this and just use `WithIngressPolicyType` and `WithEgressPolicyType` which makes the build more explicit.

internal/runtime/network_policy.go
Clee2691: I am also trying to "pave" the way for the CLF NetworkPolicy building. I am not sure which ones I'll need. But I'll remove most of them and add them back in as needed.

api/logging/v1alpha1/log_file_metrics_exporter_types.go
Clee2691: Adding `AllowAllIngressEgress` as a rule set

internal/network/network_policy_test.go
Clee2691: ~~From above [comment](https://github.com/openshift/cluster-logging-operator/pull/3122#discussion_r2392181672)~~

internal/constants/constants.go
```diff
@@ -51,6 +51,7 @@ const (
 	VectorName                 = "vector"
 	KibanaName                 = "kibana"
 	LogfilesmetricexporterName = "logfilesmetricexporter"
+	LogFileMetricsPortName     = "logfile-metrics"
```
jcantrill: We do not need, AFAIK, a unique name to the deployment here because this is a separate deployment from the collector.  It's just a named endpoint to scrape metrics from where the name is only used to match up what is defined in the LFME unique service.   We should simply be able to call it "metrics"

internal/network/network_policy_test.go
Clee2691: I think this is the wrong comment here. I think it was supposed to go under another thread. So ignore.

internal/constants/constants.go
Clee2691: Ok I will make the change.

jcantrill (approved):
/approve
/lgtm

PR #3123, opened 2025-09-29
[release-6.3] Manual port of LOG-7620

Clee2691:
### Description
This is a manual port of [LOG-7620](https://issues.redhat.com/browse/LOG-7620) to `release-6.3`.

#### Original Description:
This PR resolves a bug where the prune transform failed to remove the `.openshift.sequence` field from log events. The removal logic has been corrected to ensure this field is successfully pruned.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-7805

### DIFF
--- a/test/helpers/splunk/splunk.go
+++ b/test/helpers/splunk/splunk.go
@@ -1,0 +1,33 @@
+package splunk
+
+import (
+	"time"
+
+	"github.com/openshift/cluster-logging-operator/test/framework/functional"
+	. "github.com/onsi/gomega"
+)
+
+// WaitOnSplunk waits for Splunk to be ready by checking HEC health and service status
+func WaitOnSplunk(f *functional.CollectorFunctionalFramework) {
+	time.Sleep(20 * time.Second)
+	Eventually(func() string {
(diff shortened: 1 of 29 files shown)


test/helpers/splunk/splunk.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"time"
+
+	"github.com/openshift/cluster-logging-operator/test/framework/functional"
+	. "github.com/onsi/gomega"
+)
+
+// WaitOnSplunk waits for Splunk to be ready by checking HEC health and service status
+func WaitOnSplunk(f *functional.CollectorFunctionalFramework) {
+	time.Sleep(20 * time.Second)
```
jcantrill, line 12: We shouldnt need this sleep since we have a poll which times out.  I can live with removing this in a separate change since your change moves the entire function

Clee2691: From my previous [comment](https://github.com/openshift/cluster-logging-operator/pull/3115#discussion_r2360696114) we still need this sleep for the initial start of Splunk.

PR #3129, opened 2025-10-08
LOG-7772: Enable 'opt-in' to Collector NetworkPolicy

Clee2691:
### Description
WIP Collector NetworkPolicy

### Links
- JIRA: https://issues.redhat.com/browse/LOG-7772


internal/controller/observability/collector.go
```diff
@@ -139,11 +139,19 @@ func ReconcileCollector(context internalcontext.ForwarderContext, pollInterval,
 		log.Error(err, "collector.ReconcileInputServices")
 		return err
 	}
-
+	networkPolicyName := fmt.Sprintf("%s-%s", constants.CollectorName, resourceNames.CommonName)
```
jcantrill, line 142: nit.  This can be moved inside the "if"

internal/network/ports.go
```diff
@@ -0,0 +1,197 @@
...
+			return 9093
+		}
+		return 9092
+	case obs.OutputTypeHTTP:
+		if scheme == "http" {
+			return 80
+		}
+		return 443 // https
+	}
+	return 0
```
jcantrill: I wonder if we should insert a "panic" here because we should only get here if there is an error in the logic or we missed a place to implement something for a new output

jcantrill (approved):
/approve

internal/controller/observability/collector.go
Clee2691: This cannot be moved because this policy name is needed for the `else` part of this statement for removal of the NetworkPolicy when the spec is removed on line 151.

internal/network/ports.go
Clee2691: Replaced with `panic`

PR #3148, opened 2025-11-04
NO-JIRA: replace deprecated --slow-spec-threshold with --poll-progress flags

vparfonov:
### Description
Replaced the deprecated Ginkgo flag `--slow-spec-threshold` (deprecated in Ginkgo v2) with the new progress reporting flags `--poll-progress-after` and `--poll-progress-interval`, because it generated excessive noise: 
```
You're using deprecated Ginkgo functionality:
=============================================
  --slow-spec-threshold is deprecated --slow-spec-threshold has been deprecated and will be removed in a future version of Ginkgo.  This feature has proved to be more noisy than useful.  You can use --poll-progress-after, instead, to get more actionable feedback about potentially slow specs and understand where they might be getting stuck.
```

Added:
`--poll-progress-after=300s` — starts showing progress if no spec completes for 300 seconds  (5 minutes)
`--poll-progress-interval=30s` — prints progress updates every 30 seconds afterward

Example:
`ginkgo -p -procs=1 -v --no-color --trace --poll-progress-after=300s --poll-progress-interval=30s --timeout=60m "$dir" | tee -a "$artifact_dir/test.log"`

/cc @Clee2691 @cahartma 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA:
- Enhancement proposal:


jcantrill (commented):
/approve
/lgtm

PR #3166, opened 2025-12-04
doc: Article on high volume log loss.

alanconway:
### Description

"This guide explains how to handle scenarios where high-volume logging can cause log loss in
OpenShift clusters, and how to configure your cluster to minimize this risk."

/cc @xperimental
/cc @cahartma
/assign @jcantrill

### Links
- JIRA: https://issues.redhat.com/browse/OBSDA-1211

doc: Article on high volume log loss.

"This guide explains how to handle scenarios where high-volume logging can cause log loss in
OpenShift clusters, and how to configure your cluster to minimize this risk.""

### DIFF
--- a/docs/administration/high-volume-log-loss.adoc
+++ b/docs/administration/high-volume-log-loss.adoc
@@ -1,0 +1,382 @@
+= High volume log loss
+:doctype: article
+:toc: left
+:stem:
+
+This guide explains how high log volumes in OpenShift clusters can cause log loss,
+and how to configure your cluster to minimize this risk.
+
+== Overview
+
+=== Log loss
+
+Container logs are written to `/var/log/pods`.
+The forwarder reads and forwards logs as quickly as possible with its available CPU/Memory.
+If the forwarder is too slow, in some cases adjusting its CPU/Memory may resolve the problem.
+
+There are always some _unread logs_, written but not yet read by the forwarder.
+
+_CRI-O_ (Container Runtime Interface - Open Container Initiative) captures logs from container
+stdout/stderr and writes them to files under `/var/log/pods`.
+It rotates log files, and deletes old files, to enforce per-container limits.
+CRI-O and the log forwarder act independently.
+There is no coordination or flow-control to ensure logs are forwarded before they are deleted.
+
+_Log Loss_ occurs when _unread logs_ are deleted by CRI-O _before_ being read by the forwarder.
+Lost logs are gone from the file-system, have not been forwarded anywhere, and cannot be recovered.
+
+NOTE: This guide focuses on _container logs_.
+The section <<Other types of logs>> briefly discusses other types of log.
+====
+Not all logs are container logs, the following types of logs are not discussed here but
+can be managed in similar ways:
+
+- Journald (node) logs: are
+====
+=== Log rotation
+
+CRI-O does the actual log rotation, but the rotation limits are specified via Kubelet.
+The parameters are:
+[horizontal]
+containerLogMaxSize:: Max size of a single log file (default 10MiB)
+containerLogMaxFiles:: Max number of log files per container (default 5)
+
+A container writes to one active log file.
+When the active file reaches `containerLogMaxSize` the log files are rotated:
+
+. the old active file becomes the most recent archive
+. a new active file is created
+. if there are more than `containerLogMaxFiles` files, the oldest is deleted.
+
+=== Best effort delivery
+
+OpenShift logging provides _best effort_ delivery of logs.
+There is limited capacity to store and forward logs reliably.
+If the load exceeds those limits, logs will be lost.
+
+This article discusses how you can tune these limits to minimize log loss under your expected loads.
+
+[WARNING]
+====
+**NEVER** abuse logs as a way to store or send application data - especially financial data.
+This is unreliable, insecure, and in all other ways inconceivable.
+Use appropriate tools that meet your reliability requirements for application data.
+For example: databases, object stores, or reliable messaging (Kafka, AMQP, MQTT).
+====
+
+=== Modes of operation
+
+[horizontal]
+writeRate:: long-term average logs per second per container written to `/var/log/pods`
+sendRate:: long-term average logs per second per container forwarded to the store
+
+During _normal operation_ `sendRate` keeps up with `writeRate` (on average).
+The number of unread logs is small, and does not grow over time.
+
+If `writeRate` exceeds `sendRate` (on average) for an extended period of time, unread logs accumulate.
+If this lasts long enough, log rotation will delete unread logs causing log loss.
+
+After a load surge ends, the system has to _recover_ by processing the accumulated unread logs.
+Until the backlog clears, the system is more vulnerable to log loss if there is another overload.
+
+== Metrics for logging
+
+Relevant metrics include:
+[horizontal]
+vector_*:: The `vector` process deployed by the log forwarder generates metrics for log collection, buffering and forwarding.
+log_logged_bytes_total:: The `LogFileMetricExporter` measures disk writes _before_ logs are read by the forwarder.
+  To measure end-to-end log loss it is important to measure data that is _not_ yet read by the forwarder.
+kube_*:: Metrics from the Kubernetes cluster.
+
+[CAUTION]
+====
+Metrics named `_bytes_` count bytes, metrics named `_events_` count log records.
+
+The forwarder adds metadata to the logs before sending so you cannot assume that a log
+record written to `/var/log` is the same size in bytes as the record sent to the store.
+
+Use event and byte metrics carefully in calculations to get the correct results.
+====
+
+=== Log File Metric Exporter
+
+The metric `log_logged_bytes_total` is the number of bytes written to each file in `/var/log/pods` by a container.
+This is independent of whether the forwarder reads or forwards the data.
+To generate this metric, create a `LogFileMetricExporter`:
+
+[,yaml]
+----
+apiVersion: logging.openshift.io/v1alpha1
+kind: LogFileMetricExporter
+metadata:
+  name: instance
+  namespace: openshift-logging
+----
+
+== Limitations
+
+Write rate metrics only cover container logs in `/var/log/pods`.
+The following are excluded from these metrics:
+
+* Node-level logs (journal, systemd, audit)
+* API audit logs
+
+This may cause discrepancies when comparing write vs send rates.
+The principles still apply, but account for this additional volume in capacity planning.
+
+=== Using metrics to measure log activity
+
+The PromQL queries below are averaged over an hour of cluster operation, you may want to take longer samples for more stable results.
+
+.*TotalWriteRateBytes* (bytes/sec, all containers)
+----
+sum(rate(log_logged_bytes_total[1h]))
+----
+
+.*TotalSendRateEvents* (events/sec, all containers)
+----
+sum(rate(vector_component_sent_events_total{component_kind="sink",component_type!="prometheus_exporter"}[1h]))
+----
+
+.*LogSizeBytes* (bytes): Average size of a log record on /var/log disk
+----
+sum(increase(vector_component_received_bytes_total{component_type="kubernetes_logs"}[1h])) /
+sum(increase(vector_component_received_events_total{component_type="kubernetes_logs"}[1h]))
+----
+
+.*MaxContainerWriteRateBytes* (bytes/sec per container): The max rate determines per-container log loss.
+----
+max(rate(log_logged_bytes_total[1h]))
+----
+
+NOTE: The queries above are for container logs only.
+Node and audit may also be forwarded (depending on your `ClusterLogForwarder` configuration)
+which may cause discrepancies when comparing write and send rates.
+
+== Other types of logs
+
+There are other types of logs besides container logs.
+All are stored under `/var/log`, but log rotation is configured differently.
+The same general principles of log loss apply, here are some tips for configuration.
+
+journald node logs:: The write-rate in is the total volume of logs from _local_ processes on the node.
+Rotation is controlled by local `journald.conf` configuration files.
+
+Linux audit node logs:: The write-rate is total of all auditable actions on the node.
+Rotation is controlled by `auditd`, which is configured by `/etc/auditd/auditd.conf`.
+
+Openshift and Kubernetes audit logs:: #TODO: link to existing docs and features for API audit.#
+
+ #TODO#: explain how to set node configuration in a cluster.
+
+== Recommendations
+
+=== Check forwarder CPU and Memory
+
+If the forwarder can't keep up with `writeRate`, there are two possible causes:
+- `sendRate` is to slow - the forwarder is often blocked waiting to send, which slows down reading once its internal buffers are full.
+- The _forwarder itself_ is too slow: the CPU and Memory limits  for the forwarder may be set too low slowing down the forwarder process itself.
+
+Adjusting CPU and memory for the forwarder is an easy solution for some logging problems
+and is always a good thing to check.
+
+However, if the real problem is `writeRate > sendRate`, then this won't solve all the problems.
+
+=== Estimate long-term load
+
+Estimate your expected steady-state load, spike patterns, and tolerable outage duration.
+The long-term average send rate *must* exceed the write rate (including spikes) to allow recovery after overloads.
+
+----
+TotalWriteRateBytes < TotalSendRateLogs × LogSizeBytes
+----
+
+=== Configure rotation
+
+Configure rotation parameters based on the _noisiest_ containers in your cluster,
+with the highest write rates (`MaxContainerWriteRateBytes`) that you want to protect.
+
+For an outage of length `MaxOutageTime`:
+
+.Maximum per-container log storage
+----
+MaxContainerSizeBytes = MaxOutageTime × MaxContainerWriteRateBytes
+----
+
+.Kubelet configuration
+----
+containerLogMaxFiles = N
+containerLogMaxSize = MaxContainerSizeBytes / N
+----
+
+NOTE: N should be a relatively small number of files, the default is 5.
+The files can be as large as needed so that `N*containerLogMaxSize > MaxContainerSizeBytes`
+
+=== Estimate total disk requirements
+
+Most containers write far less than `MaxContainerSizeBytes`.
+Total disk space is based on cluster-wide average write rates, not on the noisiest containers.
+
+.Minimum total disk space required
+----
+DiskTotalSize = MaxOutageTime × TotalWriteRateBytes × SafetyFactor
+----
+
+.Recovery time to clear the backlog from a max outage:
+----
+RecoveryTime = (MaxOutageTime × TotalWriteRateBytes) / (TotalSendRateLogs × LogSizeBytes)
+----
+
+[TIP]
+.To check the size of the /var/log partition on each node
+[source,console]
+----
+for NODE in $(oc get nodes -o name);
+  do echo "# $NODE"; oc debug -q $NODE -- df -h /var/log;
+done
+----
+
+==== Example
+
+The default Kubelet settings allow 50MB per container log:
+----
+containerLogMaxFiles: 5     # Max 5 files per container log
+containerLogMaxSize: 10MB   # Max 10 MB per file
+----
+
+Suppose we observe log loss during a 3-minute outage (forwarder is unable to forward any logs).
+This implies the noisiest containers are writing at least 50MB of logs _each_ during the 3 minute outage:
+
+----
+MaxContainerWriteRateBytes ≥ 50MB / 180s ≈ 278KB/s
+----
+
+Now suppose we want to handle an outage of up to 1 hour, without loss,
+rounding up to a maximum per-container write rate of 300KB/s.
+
+----
+MaxStoragePerContainerBytes = 300KB/s × 3600s ≈ 1GB
+
+containerLogMaxFiles: 10
+containerLogMaxSize: 100MB
+----
+
+For total disk space, suppose the cluster writes 2MB/s for all containers:
+
+----
+MaxOutageTime = 3600
+TotalWriteRateBytes = 2MB/s
+SafetyFactor = 1.5
+
+DiskTotalSize = 3600s × 2MB/s × 1.5 = 10GB
+----
+
+NOTE: `MaxStoragePerContainerBytes=1GB` applies only to the noisiest containers.
+The `DiskTotalSize=10GB` is based on the cluster-wide average write rates.
+
+=== Configure Kubelet log limits
+
+Here is an example `KubeletConfig` resource (OpenShift 4.6+). +
+It provides `50MB × 10 files = 500MB` per container.
+
+[,yaml]
+----
+apiVersion: machineconfiguration.openshift.io/v1
+kind: KubeletConfig
+metadata:
+  name: increase-log-limits
+spec:
+  machineConfigPoolSelector:
+    matchLabels:
+      machineconfiguration.openshift.io/role: worker
+  kubeletConfig:
+    containerLogMaxSize: 50Mi
+    containerLogMaxFiles: 10
+----
+
+You can modify `MachineConfig` resources on older versions of OpenShift that don't support `KubeletConfig`.
+
+=== Apply and verify configuration
+
+*To apply the KubeletConfig:*
+[,bash]
+----
+# Apply the configuration
+oc apply -f kubelet-log-limits.yaml
+
+# Monitor the roll-out (this will cause node reboots)
+oc get kubeletconfig
+oc get mcp -w
+----
+
+*To verify the configuration is active:*
+[,bash]
+----
+# Check that all nodes are updated
+oc get nodes
+
+# Verify the kubelet configuration on a node
+oc debug node/<node-name>
+chroot /host
+grep -E "(containerLogMaxSize|containerLogMaxFiles)" /etc/kubernetes/kubelet/kubelet.conf
+
+# Check effective log limits for running containers
+find /var/log -name "*.log" -exec ls -lah {} \; | head -20
+
+----
+
+== Alternative (non)-solutions
+
+This section presents what seem like alternative solutions at first glance, but have significant problems.
+
+=== Large forwarder buffers
+
+Instead of modifying rotation parameters, make the forwarder's internal buffers very large.
+
+==== Duplication of logs
+
+Forwarder buffers are stored on the same disk partition as `/var/log`.
+When the forwarder reads logs, they remain in `/var/log` until rotation deletes them.
+This means the forwarder buffer mostly duplicates data from `/var/log` files,
+which requires up to double the disk space for logs waiting to be forwarded.
+
+==== Buffer design mismatch
+
+Forwarder buffers are optimized for transmitting data efficiently, based on characteristics of the remote store.
+
+- *Intended purpose:* Hold records that are ready-to-send or in-flight awaiting acknowledgement.
+- *Typical time-frame:* Seconds to minutes of buffering for round-trip request/response times.
+- *Not designed for:* Hours/days of log accumulation during extended outages
+
+==== Supporting other logging tools
+
+Expanding `/var/log` benefits _any_ logging tool, including:
+
+- `oc logs` for local debugging or troubleshooting log collection
+- Standard Unix tools when debugging via `oc rsh`
+
+Expanding forwarder buffers only benefits the forwarder, and costs more in disk space.
+
+If you deploy multiple forwarders, each additional forwarder will need its own buffer space.
+If you expand `/var/log`, all forwarders share the same storage.
+
+=== Persistent volume buffers
+
+Since large forwarder buffers compete for disk space with `/var/log`,
+what about storing forwarder buffers on a separate persistent volume?
+
+This would still double the storage requirements (using a separate disk) but
+the real problem is that a PV is not a local disk, it is a network service.
+Using PVs for buffer storage introduces new network dependencies and reliability and performance issues.
+The underlying buffer management code is optimized for local disk response times.
+
+== Summary
+
+1. *Monitor log patterns:* Use Prometheus metrics to measure log generation and send rates
+2. *Calculate storage requirements:* Account for peak periods, recovery time, and spikes
+3. *Increase kubelet log rotation limits:* Allow greater storage for noisy containers
+4. *Plan for peak scenarios:* Size storage to handle expected patterns without loss
+
+TIP: The OpenShift console Observe>Dashboard section includes helpful log-related dashboards.
+
+
--- a/docs/administration/README.adoc
+++ b/docs/administration/README.adoc
@@ -5,3 +5,4 @@
 * Enabling event collection by link:deploy-event-router.md[Deploying the Event Router]
 * link:logfilemetricexporter.adoc[Collecting Container Log Metrics]
 * Example of a link:lokistack.adoc[complete Logging Solution] using LokiStack and UIPlugin
+* Configuring for link:large-volume.adoc[high volume log loss]


docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+  Rotation permanently deletes old logs to free disk space.
+- **Log Forwarder** collects logs by reading from `/var/log` files and sending to a remote store (often Loki).
+  - *Collected logs* have been read by the forwarder. They may be sent or still sending, but they are the forwarder's responsibility.
+  - *Uncollected logs* have not yet been read by the forwarder.
+
+.Log loss
+IMPORTANT: If uncollected logs are deleted by Kubelet before the forwarder can read them, they are **lost** permanently.
+
+.Filling `/var/log` will destabilize the node.
+WARNING: Kubelet rotation controls the size of log files in `/var/log`. The forwarder is responsible for controlling any other files (e.g. buffers) that it stores here.
```
jcantrill: Should we tweak this statement?  The forwarder is configured to store buffers and checkpoints under /var/lib/vector.  I suspect this is likely the same partition as /var/logs but that is probably dependent upon the node image being used

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+Send-rate exceeds write-rate. Logs are forwarded as quickly as they arrive.
+The set of uncollected logs is small and does not grow significantly.
+The forwarder has time to collect all logs before they are rotated.
+
+Overload (possible log loss)::
+Write-rate exceeds send-rate (on average) for an extended period of time.
+This could be caused by high write activity and/or slow sending.
+Uncollected logs accumulate as the forwarder falls behind.
+A brief overload (sometimes called a _burst_ or _spike_) can be handled without log loss if there is enough storage.
+An overload that continues long enough force rotation of uncollected logs **will** cause log loss.
```
jcantrill: forces...   and may? cause log loss?

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+
+Recovery (no log loss, but not yet back to normal)::
+After an overload, there are more uncollected logs accumulated in storage than during normal operation.
+We must consider the time it takes to reduce this backlog and get back to normal, before the next overload.
+
+IMPORTANT: It is theoretically impossible to prevent log loss under all conditions.
+The goal is to analyze your expected average and peak conditions, then provision storage accordingly.
```
jcantrill: The one concern I have of this statement is I suspect most customers have no control over this provisioning.  They use the OOTB node images provided by a RH cluster installation. I hope we have an SOP which describes how to create nodes which meet your requirements but I suspect most users do not know how to do this, can't or won't.

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+*Time to log loss during steady overload:*
+
+When write rate consistently exceeds send rate, logs accumulate until storage fills:
+
+```
+Time to log loss = Storage capacity ÷ (Write rate - Send rate)
+```
+
+*Example - Steady overload:*
+- Storage capacity: 500MB (kubelet default: 10MB × 5 files × 10 containers)
```
jcantrill: I think you wanted this to be a bullet list but it does not render that way

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+
+*Time to log loss during steady overload:*
+
+When write rate consistently exceeds send rate, logs accumulate until storage fills:
+
+```
+Time to log loss = Storage capacity ÷ (Write rate - Send rate)
+```
+
+*Example - Steady overload:*
```
jcantrill: sample from a worker node:
```
h-5.1# df -h /
Filesystem      Size  Used Avail Use% Mounted on
/dev/nvme0n1p4  120G   17G  104G  14% /
sh-5.1# lsblk
NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
nvme0n1     259:0    0   120G  0 disk 
|-nvme0n1p1 259:1    0     1M  0 part 
|-nvme0n1p2 259:2    0   127M  0 part 
|-nvme0n1p3 259:3    0   384M  0 part /boot
`-nvme0n1p4 259:4    0 119.5G  0 part /var/lib/kubelet/pods/3de9716f-099c-416a-b7cf-57148b9832d2/volume-subpaths/entrypoint/collector/6
                                      /var/lib/kubelet/pods/d1077521-9094-40b7-a6a4-fc66678a667d/volume-subpaths/nginx-conf/networking-console-plugin/1
                                      /var/lib/kubelet/pods/089962a6-ed0c-49d0-9644-69f701e9617e/volume-subpaths/nginx-conf/networking-console-plugin/1
                                      /var
                                      /sysroot/ostree/deploy/rhcos/var
                                      /sysroot
                                      /usr
                                      /etc
                                      /
```

which looks to me like most of the filesystem including lib, log, tmp are all on the same partition competing for the same disk.  This means there is less disk as the node density increases.  I'm not exactly certain how we reflect this information in the calculation.  Customer has more than 500mb so maybe we should frame this calculation in terms of the max container density, max log size, and max number of rollovers which would be the theoretical amount we could use if left unadjusted (minus the 256mb / buffered output)

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+After a peak logging period, the system needs time to drain accumulated logs. Calculate recovery time:
+```
+Recovery_time = Accumulated_logs / (Send_rate - Normal_log_rate)
+```
+
+*Key considerations:*
+
+- **Overlapping peaks:** If peaks occur closer together than recovery time, logs accumulate across multiple peaks
+- **Send rate degradation:** Network issues during recovery extend drain time
+- **Grace period:** Allow buffer between peaks for unexpected delays
```
jcantrill: Maybe we should not use the word 'buffer' since this is a configurable point on an output we do not expose and its something we do not want to encourage.  I think you are identifying a 'safety factor' here

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+- Peak accumulated: 1GB of logs
+- Normal log rate: 100 KB/s
+- Send rate: 400 KB/s
+- Recovery time: `1GB / (400-100) KB/s = 1GB / 300KB/s = 3,413 seconds (57 minutes)`
+
+If peaks occur more frequently than 57 minutes, you need larger storage or higher send capacity.
+
+== Recommendations
+
+The core recommendation is to _increase the log rotation limits of Kubelet_ based on
```
jcantrill: IMO we should emphasize this with one of those boxes or other to make it stand out

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+
+These are the key quantities to measure and/or estimate:
+
+- **Average log rate:** Normal sustained logging rate during typical operations
+- **Average send rate:** Normal sustained send rate during typical operations
+- **Peak log rate:** Maximum logging rate during high-activity periods (deployments, errors, etc.)
+- **Peak duration:** How long high-volume periods typically last
+- **Peak distribution:** How often do peak periods occur.
+
+==== OpenShift Logging metrics
```
jcantrill: Should we mention our collection dashboard?  Might be an opportunity for more people to use it and for additional improvement

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+
+OpenShift Logging provides Prometheus metrics to monitor log flows.
+Use these queries to measure your system:
+
+```promql
+# Log ingestion rate (bytes per second) - by namespace
```
jcantrill: These are not actually ingested logs but a metric produced by LFME.  Additionally it is by namespace, pod, and container

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+rate(log_logged_bytes_total[5m])
+
+# Log events ingestion rate (events per second)
+rate(log_logged_events_total[5m])
```
jcantrill: this metric does not exist

docs/administration/high-volume-log-loss.adoc
jcantrill: rate(vector_component_received_bytes_total{component_kind="source"}[5m]) is the actual metric which  maybe we should consider a recording rule? that makes it collector agnostic?

docs/administration/high-volume-log-loss.adoc
jcantrill: rate(vector_component_received_events_total{component_kind="source"}[5m])

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+
+# Vector component throughput (events per second)
+rate(vector_component_sent_events_total[5m])
```
jcantrill: consider adding the following to weed out prometheus traffic:

component_kind="sink", component_type!="prometheus_exporter"

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+rate(vector_component_sent_events_total{component_type="sink"}[5m])
+```
+
+*Understanding metric units:*
+
+Note that Vector metrics mix units - some report events/sec, others bytes/sec. For capacity planning, convert everything to bytes/sec:
+
+```promql
+# Calculate average log size (bytes per event)
+rate(log_logged_bytes_total[5m]) / rate(log_logged_events_total[5m])
```
jcantrill: We should consider comparing these to the existing dashboard

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+
+WARNING: This calculation only covers a single peak incident.
+For production systems, you must also consider multi-peak scenarios:
+
+```
+Total_storage = Single_peak_storage + (Peak_frequency / Recovery_rate) × Peak_storage
+```
+
+Where:
+- Peak_frequency = how often peaks occur (e.g., every 2 hours)
```
jcantrill: rendering issue

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+
+This approach uses large forwarder buffers rather than increasing kubelet limits.
+While forwarder configuration changes appear less intrusive than modifying kubelet parameters, this approach has significant drawbacks.
+
+==== Same disk, duplicate storage
+
+The naive expectation is that a forwarder buffer increases the total log buffer space to `/var/log + forwarder-buffer`. This is not the case.
+
+When the forwarder reads a log record, it _remains in `/var/log`_ until it is deleted by rotation.
+This means the forwarder buffer will usually be full of log data that is _still available in the log files_.
```
jcantrill: This is a true statement but we need to consider vector's checkpoint system.  Technically it is 'available' and it is correct that it would be both in the original location and the disk buffer, but it is effectively "gone" unless the checkout point is reset.  We are unable to utilize n/ack since kubernetes_log source does not support it.

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+
+Thus the _effective_ space for logs is more like `max(/var/log, forwarder-buffer)`.
```
jcantrill: Unless we modify our buffer settings, it is always smaller than /var/log because the default is 256Mb per output per collector.  The difference is we have "rescued" the log from being rotated off before it can be collected.  I think we need to be careful when saying "0" protection.

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+If the buffer is smaller than `/var/log` then it uses disk space, but provides **0** additional protection from log loss.
+If it is bigger you get some extra buffer, at the expense of doubling the disk storage required for `/var/log`
+
+It is _much_ more efficient to simply make `/var/log` bigger.
```
jcantrill: Following up on my previous comment I believe we need to consider available storage in context of the allotted disk, container density, max log, and max rotations.  This is the theoretical max and this is what will compete for space with all the other processes running on the node.  The main theme of the document is to 'resize disk' or '/var/log' but we are really advising changing the log retention and sizes. The difference between changing these settings and buffer is we are just changing the consumption point of the disk.  The volume limits on the front end are easier to monitor and calculate than on the back because you can have variable outputs and variable CLFs.  Additionally if you are mutliplexing the logs you are not disk buffering them on the back side multiple times.

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+Forwarder buffers are optimized for transmitting data efficiently, based on characteristics of the remote store.
+
+- **Intended purpose:** Hold records that are ready-to-send or in-flight awaiting acknowledgement.
+- **Typical timeframe:** Seconds to minutes of buffering for round-trip request/response times.
+- **Not designed for:** Hours/days of log accumulation during extended outages
+
+==== Supporting other logging tools
+
+Expanding `/var/log` benefits _any_ logging tool, including
+- `oc logs` for local debugging, or if there is a problem with the log collection.
```
jcantrill: rendering

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+- Standard Unix tools used when debugging by `rsh` to the node.
+
+Expanding forwarder buffers only benefits the forwarder, and costs more in disk space.
```
jcantrill, line 358: I would add something to the effect because it is buffered in a component dependent format (i.e. compression, encoding)

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+
+If you deploy multiple forwarders, each additional forwarder will need its own buffer space.
```
jcantrill, line 360: each output for each forwarder has its own buffer

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+If you expand `/var/log`, all forwarders share the same storage.
+
+==== Risk of self-sabotage
+
+Filling `/var/log` is a near-fatal problem.
+Storing forwarder buffers on `/var/log` means users need to be very aware of the _combined_ disk usage implied by separate Kubelet and Forwarder configurations.
```
jcantrill: buffers are written to /var/llb/..... some path that is specific to collector and output, but looks to be the same partition

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+This seems likely to lead to mistakes with serious consequence for less experienced users.
```
jcantrill: More emphatic:

This is likely...   or This can lead ...

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+
+Modifying kubelet parameters may seem like a bigger deal than adding buffer config to log forwarder.
+**It Is Not** - the calculations required, and the consequences of error are the same either way.
+
+So it seems better to make kubelet changes for 2 reasons:
```
jcantrill: It is better.... changes for two reasons:

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+- Log storage requirements are calculated and specified _once_ in kubelet config, not in separate kubelet and forwarder config.
```
jcantrill: rendering

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+- The "seriousness" of modifying Kubelet config encourages careful thought, exposing a harmless-looking forwarder "bufferSize" might encourage ill-considered use.
+
+==== Multiple forwarders
+
+It is possible to deploy multiple forwarders, which is useful in some cases.
+Deploying N forwarders multiplies all of the previously mentioned problems by N.
+
+=== Store buffers on a persistent volume
+
+The problems above stem from using `/var/log` for buffer storage.
```
jcantrill: Consider how we express buffer actually on /var/lib

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+   or other web service destination, and could cause log loss.
+
+3. The forwarder's buffering code is optimized for local file system response times.
+   It will not perform as well with remote storage.
+
+== Summary and recommendations
+
+1. **Monitor log patterns:** Use Prometheus metrics to measure log generation and send rates
+2. **Calculate storage requirements:** Account for peak periods, recovery time, and overlapping spikes
+3. **Increase kubelet log limits:** Use `KubeletConfig` to expand `/var/log` storage
```
jcantrill: This is not actually expanding /var/log but modifying the log retention and file size settings

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+4. **Plan for peak scenarios:** Size storage to handle expected peak patterns without loss
+
+Consider setting up alerts for:
+- High log generation rates: `rate(log_logged_bytes_total[5m]) > threshold`
+- Buffer fill levels: `vector_buffer_events > threshold`
+- Send rate degradation: `rate(vector_component_sent_events_total{component_type="sink"}[5m]) < threshold`
+
+Consider larger limits if you experience:
+- Frequent log loss during deployments or scaling events
```
jcantrill: rendering

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,362 @@
...
+
+== Summary and recommendations
+
+1. **Monitor log patterns:** Use Prometheus metrics to measure log generation and send rates
+2. **Calculate storage requirements:** Account for peak periods, recovery time, and overlapping spikes
+3. **Increase kubelet log limits:** Use `KubeletConfig` to expand `/var/log` storage
+4. **Plan for peak scenarios:** Size storage to handle expected peak patterns without loss
+
+Consider setting up alerts for:
+- High log generation rates: `rate(log_logged_bytes_total[5m]) > threshold`
```
jcantrill: rendering

jcantrill (changes requested):
Overall I believe this needs changes related to changing container log settings and not expanding capacity of /var/log

docs/administration/high-volume-log-loss.adoc
alanconway: Yes, good point.

docs/administration/high-volume-log-loss.adoc
alanconway: +1

docs/administration/high-volume-log-loss.adoc
alanconway: Agreed - as per your overall comment, I am rewriting to emphasize that the main point is adjusting rotation paramaters. Logs get lost because of per-container rotation, not because of the size of /var/log.

docs/administration/high-volume-log-loss.adoc
alanconway: +!, missing a newline

docs/administration/high-volume-log-loss.adoc
alanconway: Good info. As above - the main point is rotation parameters per-container. I think we can describe how to estimate the worst-case total log size, but I'm not sure I want to get into re-sizing partitions if you don't think you have enough storage.

docs/administration/high-volume-log-loss.adoc
alanconway: +1 I'm rewriting to make that clearer.

alanconway:
@jcantrill please re-review, I think I've addressed your comments.

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,327 @@
...
+Recovery time to clear the backlog from a max outage:
+
+----
+RecoveryTime = (MaxOutageTime × TotalWriteRateBytes) / (TotalSendRateLogs × LogSizeBytes)
+----
+
+=== Capacity planning example
+
+Consider a cluster with default kubelet settings:
+- 5 files × 10MiB = 50MB per container
```
jcantrill: Formatting issue maybe?

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,327 @@
...
+The total disk size requirement is based on the total rate over all containers.
+
+You need to include spikes in the long-term average, so you will have capacity to _recover_ after a spike.
+Recovery time to clear the backlog from a max outage:
+
+----
+RecoveryTime = (MaxOutageTime × TotalWriteRateBytes) / (TotalSendRateLogs × LogSizeBytes)
+----
+
+=== Capacity planning example
```
jcantrill: Do we need to mention something here about the maximum disk (partition?) available on a typical OCP node which is ~120G.  Also to consider is this example is for a single container but I believe OCP claims to support at least 200 containers for the density, maybe more.  These numbers are important and effectively cap the values you could use.

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,327 @@
...
+----
+
+The configuration rollout typically takes 10-20 minutes as nodes are updated in rolling fashion.
+
+== Why not use large buffers in the forwarder?
+
+An alternative approach is to use large buffers managed by the forwarder, rather than increasing kubelet limits.
+This seems attractive at first, but has significant drawbacks.
+
+=== Duplication of logs
```
jcantrill: Some place in this section it is worth mentioning that OCP standard node places all the storage requirements on the same partition.  All of these processess are keeping for the same storage. 

Alternately, I don't know if it is worth mentioning in the section about PVC as a buffer that this is likely to be network storage meaning you would be moving bytes across the network twice... write and read... which in theory is less efficient.  We currently have no way of claiming or attaching additional storage.

jcantrill (changes requested):

docs/administration/high-volume-log-loss.adoc
alanconway: Agreed, there's some implicit mention of that but I'll make it more prominent.

docs/administration/high-volume-log-loss.adoc
alanconway: Agreed. The PVC issue is discussed under "Why not use a large forwarder buffer" but I canadd a mention here also.

alanconway:
@jcantrill made changes to "Recommendations" to clarify your points - definite improvement.
Cosmetic changes to other sections.
Let me know how it reads to you

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+:stem:
+
+This guide explains how high log volumes in OpenShift clusters can cause log loss,
+and how to configure your cluster to minimize this risk.
+
+[WARNING]
+====
+#If your data requires guaranteed delivery *_do not send it as logs_*# +
+Logs were never intended to provide guaranteed delivery or long-term storage.
+Rotating disk files without any form of flow-control is inherently unreliable.
```
r2d2rnd: It's missing the context for understanding this phrase. We could rewrite it for better reflecting what's pretended. 

From the context, I could understand that it could be related to the Log Files generated where the rotation is done externally to Vector, then, not possible to hold the log until being read by Vector (this is explained later). But, from this isolated phrase, a person reading it, it can not get an understanding about what's pretended to say here.

We could indicate that "Vector has not control in how the logs are produced and rotated" not being the logs produced and the logs read by the collector a synchronous process.

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+Guaranteed delivery requires modifying your application to use a reliable, end-to-end messaging
```
r2d2rnd: We could probably mention the Red Hat Product: Red Hat AMQ that it's part of our portfolio https://www.redhat.com/en/technologies/jboss-middleware/amq

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+
+It is theoretically impossible to prevent log loss under all conditions.
+You can configure log storage to avoid loss under expected average and peak loads.
+====
+
+== Overview
+
+=== Log loss
+
+Container logs are written to `/var/log/pods`.
```
r2d2rnd, line 13: These are not the only logs that we collect and they are rotated: journald, audit OS and also other audit logs as KubeAPI audit logs and OVN Audit logs, etc.

Saying this here is relevant for the next explanations given in this document as the journald, audit OS, KubeAPI logs... they are not managed by the containerLogMaxSize and containerLogMaxFiles variables and they must be configured in other places

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+
+_Kubelet_ rotates log files and deletes old files periodically to enforce per-container limits.
+Kubelet and the forwarder act independently.
+There is no coordination or flow-control that can ensure logs get forwarded before they are deleted.
+
+_Log Loss_ occurs when _unread logs_ are deleted by Kubelet _before_ being read by the forwarder.
+footnote:[It is also possible to lose logs _after_ forwarding, we won't discuss that here.]
+Lost logs are gone from the file-system and have not been forwarded, so they likely cannot be recovered.
+
+=== Log rotation
```
r2d2rnd, line 36: As it's explained in the comment in Line 25, not all logs collected are managed by CRIO or the variables containerLogMaxSize and containerLogMaxFiles. Some not in this cathegory are: Audit OS, journald, KubeAPI Audit logs, etc.

Then, this sections needs to be more explicit for speaking only about the logs produced by containers or more generic and including the rest of the logs.

If we want to show the full picture, then, we could indicate how it can be modified to have slower retention for journald, for audit OS, KubeAPI audit....If it was needed help on this, let me know and I could provide the links

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+When the active file reaches `containerLogMaxSize` the log files are rotated:
+
+. the old active file becomes the most recent archive
+. a new active file is created
+. if there are more than `containerLogMaxFiles` files, the oldest is deleted.
+
+=== Modes of operation
+
+[horizontal]
+writeRate:: long-term average logs per second per container written to `/var/log`
```
r2d2rnd: We speak about writeRate, and we only mention per container, then, it leaves out other types of logs mentioned in the previous comments: journald, audit OS, KubeAPI...is this pretended?

/var/log is for all the logs, if we want to be explicit for logs coming from pods, we could indicate the path "/var/log/pods"

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+The forwarder reads and forwards logs as quickly as possible.
+There are always some _unread logs_, written but not yet read by the forwarder.
+
+_Kubelet_ rotates log files and deletes old files periodically to enforce per-container limits.
+Kubelet and the forwarder act independently.
+There is no coordination or flow-control that can ensure logs get forwarded before they are deleted.
+
+_Log Loss_ occurs when _unread logs_ are deleted by Kubelet _before_ being read by the forwarder.
+footnote:[It is also possible to lose logs _after_ forwarding, we won't discuss that here.]
+Lost logs are gone from the file-system and have not been forwarded, so they likely cannot be recovered.
```
r2d2rnd: The "probably" open a door that we could not probably want to leave open here. The collector will not read them if they were rotated, then, better to clarify it and being more precise on this as it's from time to time a question that arrives

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+
+== Overview
+
+=== Log loss
+
+Container logs are written to `/var/log/pods`.
+The forwarder reads and forwards logs as quickly as possible.
+There are always some _unread logs_, written but not yet read by the forwarder.
+
+_Kubelet_ rotates log files and deletes old files periodically to enforce per-container limits.
```
r2d2rnd: The Kubelet does not roate the log files. It's CRIO, the parameters are set in the Kubelet that it passes them to CRIO

We forget to mention here and not depending on CRIO:
  * journald logs rotated by the OS configuration for journald
  * audit OS logs managed by the audit configuration
  * KubeAPI audit logs that requires to pass some parameters to the Kubelet (that it's not CRIO which managed them as CRIO does for the logs coming from pods)

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+It is theoretically impossible to prevent log loss under all conditions.
+You can configure log storage to avoid loss under expected average and peak loads.
+====
+
+== Overview
+
+=== Log loss
+
+Container logs are written to `/var/log/pods`.
+The forwarder reads and forwards logs as quickly as possible.
```
r2d2rnd: This is not always correct. It's observed a lot of situations where it's limited the cpu/memory of the collector, then, "with the cpu and memory" resources given to the collector, it reads the most  that it can, that it does not mean as quickly as possible.

Then, one of the first steps when not reading fast enough, it's to verify the requests and limits set to the containers and the cpu and memory used by the collectors to observe if hitting cpu or memory limits.

Then, the collectors resources limitation can be (and it's a lot of times) one of the reasons of not able to process the most of the logs (reading and writing)

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+=== Modes of operation
+
+[horizontal]
+writeRate:: long-term average logs per second per container written to `/var/log`
+sendRate:: long-term average logs per second per container forwarded to the store
+
+During _normal operation_ sendRate keeps up with writeRate (on average).
+The number of unread logs is small, and does not grow over time.
+
+Logging is _overloaded_ when writeRate exceeds sendRate (on average) for some period of time.
```
r2d2rnd: This phrase is only true if logs are not dropped or filtered. We could perhaps to extend this phrase or being more explicit?

The Logging stack is also "overloaded" if the collector is using all the cpu or memory hitting the limits, then, it's (very) slow reading and sending the logs

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+=== Estimate long-term load
+
+Estimate your expected steady-state load, spike patterns, and tolerable outage duration.
+The long-term average send rate *must* exceed the write rate (including spikes) to allow recovery after overloads.
+
+----
+TotalWriteRateBytes < TotalSendRateLogs × LogSizeBytes
+----
+
+=== Configure Kubelet rotation
```
r2d2rnd: We are forgetting here about setting journald, KubeAPI logs rotation, audit OS rotation.

On the other said, it's mentioned continually about "Kubelet rotation", it's really CRIO, not the Kubelet which rotates the logs. The parameters are configured in the Kubelet, but these parameters are passed to CRIO that it's taking the standard output from the container and writing to /var/log and it's the responsible of rotating the container logs.

Then, we should probably to change in this document "Kubelet rotation" by CRIO rotation or "Configure in the Kubelet the parameters for the containers log rotation"

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+
+=== Log File Metric Exporter
+
+The metric `log_logged_bytes_total` is the number of bytes written to each file in `/var/log/pods` by a container.
+This is independent of whether the forwarder reads or forwards the data.
+To generate this metric, create a `LogFileMetricExporter`:
+
+[,yaml]
+----
+apiVersion: logging.openshift.io/v1alpha1
```
r2d2rnd, line 109: We could probably link to the Red Hat Documentation or sharing here an example with:
  * limits.{cpu,memory}, 
  * requests.{cpu,memory}
  * nodeSelector/tolerations

docs/administration/high-volume-log-loss.adoc
alanconway: You're right - this is a pet peeve of mine but it shouldn't be at the top of the article. I'll move it and give a bit more context.  This comment is about systems (esp. financial sector) that use logging as a way to deliver critical application data (e.g. billing or other financial transaction data) A log collector is absolutely not suited for this no matter how you tune it. Will clarify. (I used to work on AMPQ which does provide real delivery guarantees. I get triggered when people talk about "guaranteed" delivery of unstructured text written at random to stdout!)

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+
+== Recommendations
+
+=== Estimate long-term load
+
+Estimate your expected steady-state load, spike patterns, and tolerable outage duration.
+The long-term average send rate *must* exceed the write rate (including spikes) to allow recovery after overloads.
+
+----
+TotalWriteRateBytes < TotalSendRateLogs × LogSizeBytes
```
r2d2rnd, line 191: I feel that this formula is a simplification of a different reality where:

  1. it could not consider that the formula itself assumes that all the nodes are producing the same quantity/size of logs and this is not the reality.
  2. it assumes that not filtering (drop/pruning) is set
  3. The "TotalWriteRateBytes" comes from "log_logged_bytes_total" that it's a metric coming from measuring logs from pods. It's not easy to understand this formula when it's used  "TotalSendRateLogs" that can include logs collected by the audit logs (they can be so much verbose) and journald logs

### Assumption that all logs produce the same number of logs/size
The reality shows that usually only some of the collectors are assuming the majority of the load, then, you have some of the nodes where it can be produced the highest % of logs as they are containing the most verbose applications. Therefore, the nodes where running these applications and the collectors running on them are the most impacted. 

Later, it's shared the formula:
`DiskTotalSize = MaxOutageTime × TotalWriteRateBytes × SafetyFactor`

This `DiskTotalSize` metric could be "good", but, as the formula `TotalWriteRateBytes < TotalSendRateLogs × LogSizeBytes` is a simplification, it will be wrong and it will fail.

Let's put an example:
  * Node A: 300GB/day
  * Node B: 50GB/day
  * Node C: 30GB/day

The Disk needed in the Node A won't be the same that in the node B. Then, I could fail as not having enough disk in the Node A and in the nodes B and C, I have the most of the storage free.

Other example:
 * Node A: with a single pod generating 200G/day (this comes from the reality)
 * Node B: 250 pods generating not too many MB per day

Then, I feel this metric could not be good for calculating how much pressure will exist in the node and for being used later in other formulas.

A better metric, but also not good, it could be to make the measures by application/collector/node and using as baseline the highest value of them.

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+Total disk space is based on cluster-wide average write rates, not on the noisiest containers.
+
+.Minimum total disk space required
+----
+DiskTotalSize = MaxOutageTime × TotalWriteRateBytes × SafetyFactor
+----
+
+.Recovery time to clear the backlog from a max outage:
+----
+RecoveryTime = (MaxOutageTime × TotalWriteRateBytes) / (TotalSendRateLogs × LogSizeBytes)
```
r2d2rnd, line 227: As this metric takes  TotalWriteRateBytes that assumes that all the nodes are producing the same load, then, this time could not be correct. 

Let's take again the example:
  Node A: 300GB/day
  Node B: 50GB/day
  Node C: 30GB/day

Then, the Recovery time will be different from the collector in the Node A, Node B, and Node C.

Also, when recovering, in the Node A as the load is bigger and also probably at the node level, the time for recovering could be slower not only because of being needed to recover from the logs from the past, the node is producing more and more logs when the service is recovered that they need to be processed. 

In a production environment, I'd expect that the collector in the node A was suffering more than the other collectors as the receiver could be rejecting logs with "Too many" messages that the collector needs to retry.

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+----
+
+For total disk space, suppose the cluster writes 2MB/s for all containers:
+
+----
+MaxOutageTime = 3600
+TotalWriteRateBytes = 2MB/s
+SafetyFactor = 1.5
+
+DiskTotalSize = 3600s × 2MB/s × 1.5 = 10GB
```
r2d2rnd, line 271: This DiskTotalSize assums that when the logs are rotated, they are not compressed. Example of compression:

~~~
/host/var/log/pods/openshift-ovn-kubernetes_ovnkube-node-mrdgt_218324ca-0add-4885-9c30-80f7576d4f85/ovnkube-controller:
total 182032
-rw-------. 1 root root 37269952 Jan  7 14:21 0.log
-rw-------. 1 root root  4627032 Dec 26 00:22 0.log.20251219-234609.gz
-rw-------. 1 root root  4503073 Dec 30 07:30 0.log.20251226-002205.gz
-rw-------. 1 root root  4760160 Jan  4 04:30 0.log.20251230-073005.gz
-rw-------. 1 root root 52429256 Jan  4 04:30 0.log.20260104-043015
-rw-------. 1 root root 15676015 Jan 28 17:39 1.log
-rw-------. 1 root root  4794543 Jan 15 05:49 1.log.20260111-045408.gz
-rw-------. 1 root root  4766578 Jan 19 07:00 1.log.20260115-054909.gz
-rw-------. 1 root root  4789273 Jan 26 12:13 1.log.20260119-070010.gz
-rw-------. 1 root root 52431455 Jan 26 12:13 1.log.20260126-121325
~~~

In fact, if the logs are compressed, Vector will not read them, even when they remain in the nodes as per the "source" rule where the compressed files are excluded:
~~~
exclude_paths_glob_patterns = ["/var/log/pods/*/*/*.gz", "/var/log/pods/*/*/*.log.*", "/var/log/pods/*/*/
*.tmp", "/var/log/pods/openshift-logging_*/gateway/*.log", "/var/log/pods/openshift-logging_*/loki*/*.log
", "/var/log/pods/openshift-logging_*/opa/*.log", "/var/log/pods/openshift-logging_elasticsearch-*/*/*.lo
g", "/var/log/pods/openshift-logging_kibana-*/*/*.log", "/var/log/pods/openshift-logging_logfilesmetricex
porter-*/*/*.log"]
~~~

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+----
+
+NOTE: `MaxStoragePerContainerBytes=1GB` applies only to the noisiest containers.
+The `DiskTotalSize=10GB` is based on the cluster-wide average write rates.
+
+=== Configure Kubelet log limits
+
+Here is an example `KubeletConfig` resource (OpenShift 4.6+). +
+It provides `50MB × 10 files = 500MB` per container.
```
r2d2rnd, line 280: Caveats on this.

Making largers files will impact also to the node performance in different ways:

1.If the logs are compressed, then, the impact in the memory and cpu from the node will be bigger when compressing
2.When the collector is able to log forward again the logs, if it's pending to read 100G in the node, the impact on the disk will be big in terms of Disk I/O as the collector will try to read the quickest possible these logs. These logs are in "/var/log/pods", usually, the first disk from the node. If loads running in this node require low latency as it could be "etcd", they will suffer the impact

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,342 @@
...
+oc debug node/<node-name>
+chroot /host
+grep -E "(containerLogMaxSize|containerLogMaxFiles)" /etc/kubernetes/kubelet/kubelet.conf
+
+# Check effective log limits for running containers
+find /var/log -name "*.log" -exec ls -lah {} \; | head -20
+
+----
+
+The configuration rollout typically takes 10-20 minutes as nodes are updated in rolling fashion.
```
r2d2rnd: I'll remove the reference to the take 10-20 minutes. It depends a lot in how big is the cluster in the number of nodes, the API performance, the number of nodes allowed to be not available at the same  time...where the rollout of the nodes can take more than 1 hour.

Then, better remove the reference to the time.

r2d2rnd (commented):
Adding more feedback. Thank you for writing it

alanconway:
@r2d2rnd this is fantastic feedback, damn you. I have some work to do.
I've pushed a incomplete update covering some of your earlier comments (which I resolved)  but there's more to do:

- slow forwarder (collector CPU/mem) is equally important as slow sender, the system will be bottlenecked by whichever is worse. 
- You've got a lot of valuable extra detail and realism to work into examples
- other log types: not sure if the current section is enough - there's nothing on how to measure other log types using  metrics or otherwise

docs/administration/high-volume-log-loss.adoc
alanconway: I add an "other log types" section, but I think more work needed to treat them properly

docs/administration/high-volume-log-loss.adoc
alanconway: Good point, needs to be addressed.

docs/administration/high-volume-log-loss.adoc
alanconway: I've added an "Other log types" section but probably more is needed

docs/administration/high-volume-log-loss.adoc
alanconway: +1 - need more realistic examples or at least explanation of the issues.

docs/administration/high-volume-log-loss.adoc
alanconway: +1 need to clarify the calculation should be for the worst-case node. Can we use metrics to identify that?

docs/administration/high-volume-log-loss.adoc
alanconway: Never even thought about compression. Hmmm....

docs/administration/high-volume-log-loss.adoc
alanconway: Hah! Stupid Claude Code and its stupid assumptions, and the stupid author who didn't check Claud's work enough...

docs/administration/high-volume-log-loss.adoc
r2d2rnd: > +1 need to clarify the calculation should be for the worst-case node. Can we use metrics to identify that?

We can get metrics by collector/node. In fact, I usually needs to use by collector/node to know what's the one with worst behaviour and if this behaviour is in relation with the number/size of logs produced in the same node or it could be a different reason as it could be the node health, etc

docs/administration/high-volume-log-loss.adoc
```diff
@@ -0,0 +1,382 @@
...
+== Overview
+
+=== Log loss
+
+Container logs are written to `/var/log/pods`.
+The forwarder reads and forwards logs as quickly as possible with its available CPU/Memory.
+If the forwarder is too slow, in some cases adjusting its CPU/Memory may resolve the problem.
+
+There are always some _unread logs_, written but not yet read by the forwarder.
+
```
r2d2rnd, line 18: The log loss can be also produced by:
  - https://issues.redhat.com/browse/OBSDA-1206
  - quick pods as jobs that they are created/destroyed so quick that not able to read the logs

r2d2rnd (commented):
Othe causes for Log Loss

docs/administration/high-volume-log-loss.adoc
alanconway: More generally: when the pod log files are deleted (not just rotated) we could lose tail logs that we haven't read yet. If the pod was short-lived, that could be all the logs. Do you know the schedule for deleting log files after a pod is deleted? I know they can persist for a while but no idea how long.
Probably there's nothing we can do about this except note that it can happen in the document...

PR #3223, opened 2026-03-12
LOG-7261: Implement AzureLogIngestion output in CLO

Clee2691:
### Description
This PR adds a new `azureLogsIngestion` output type that sends logs to Azure Monitor via the Logs Ingestion API.
The `AzureMonitor` output is deprecated and will be removed in a future release. Microsoft will retire the Data Collecter API in September 2026.
                                                                                                            
What's new:                                               
  - Output type `AzureLogsIngestion` with required `DCR immutable ID`, `stream name`, and `URL`, plus optional `token scope`, `timestamp` field, and tuning
  - Two auth modes: `clientSecret` - long lived credentials and `workloadIdentity` - short lived tokens
  - Functional tests using a Mockoon mock server (mocks both the OAuth2 token endpoint and the DCR ingestion
   endpoint over TLS)
  - Refactored Azure test helpers into a shared test/helpers/azure/ package used by both azureMonitor and
  azureLogsIngestion
  - Add deprecation API comment to `AzureMonitor` output

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://issues.redhat.com/browse/LOG-7261
- Enhancement proposal: https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/workload-identity-auth-for-azure-monitor-logs.md

PR #3237, opened 2026-03-31
LOG-9190: Azure Monitor Logs Deprecation Alert

Clee2691:
### Description
Added alert for `AzureMonitorLogs` deprecation in favor of `AzureLogsIngestion` output.

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://redhat.atlassian.net/browse/LOG-9190

### DIFF
--- a/bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
+++ b/bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
@@ -52,6 +52,23 @@ spec:
         namespace: openshift-logging
         service: collector
         severity: info
+    - alert: ClusterLogForwarderAzureMonitorLogsDeprecation
+      annotations:
+        message: A ClusterLogForwarder, {{ $labels.namespace }}/{{ $labels.app_kubernetes_io_instance
+          }}, is configured to forward logs to Azure Monitor using the deprecated
+          HTTP Data Collector API.
+        summary: |-
+          The Azure Monitor HTTP Data Collector API used by the azureMonitor output type will be retired on September 14, 2026.
+          After this date, log forwarding to Azure Monitor using this output type will stop functioning.
+          Administrators must migrate to the Azure Monitor Logs Ingestion output.
+          See the Microsoft migration guide: https://learn.microsoft.com/en-us/azure/azure-monitor/logs/custom-logs-migrate
+          See the Microsoft Logs Ingestion API documentation: https://learn.microsoft.com/en-us/azure/azure-monitor/logs/logs-ingestion-api-overview
+      expr: |
+        log_forwarder_output_type{output="azureMonitor"} > 0
+      for: 1m
+      labels:
+        service: clusterlogforwarder
+        severity: warning
     - alert: CollectorNodeDown
       annotations:
         description: Prometheus could not scrape {{ $labels.namespace }}/{{ $labels.pod
--- a/config/prometheus/collector_alerts.yaml
+++ b/config/prometheus/collector_alerts.yaml
@@ -51,6 +51,21 @@ spec:
         namespace: openshift-logging
         service: collector
         severity: info
+    - alert: ClusterLogForwarderAzureMonitorLogsDeprecation
+      annotations:
+        message: "A ClusterLogForwarder, {{ $labels.namespace }}/{{ $labels.app_kubernetes_io_instance }}, is configured to forward logs to Azure Monitor using the deprecated HTTP Data Collector API."
+        summary: |-
+          The Azure Monitor HTTP Data Collector API used by the azureMonitor output type will be retired on September 14, 2026.
+          After this date, log forwarding to Azure Monitor using this output type will stop functioning.
+          Administrators must migrate to the Azure Monitor Logs Ingestion output.
+          See the Microsoft migration guide: https://learn.microsoft.com/en-us/azure/azure-monitor/logs/custom-logs-migrate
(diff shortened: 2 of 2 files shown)


bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -52,6 +52,24 @@ spec:
...
         service: collector
         severity: info
+    - alert: ClusterLogForwarderAzureMonitorLogsDeprecation
+      annotations:
+        message: A ClusterLogForwarder in namespace {{$labels.resource_namespace}}
+          is configured to forward logs to Azure Monitor using the deprecated HTTP
+          Data Collector API.
+        summary: |-
+          The Azure Monitor HTTP Data Collector API used by the azureMonitor output type will be retired on September 14, 2026.
+          After this date, log forwarding to Azure Monitor using this output type will stop functioning.
```
jcantrill, line 62: I wonder if we should be less firm and say... may stop functiontion.

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -52,6 +52,24 @@ spec:
...
+          Administrators must migrate to the Azure Monitor Logs Ingestion output.
```
jcantrill, line 63: We should add some sort of RH link or information here regarding our new API and feature.  @johnwilkins can you provide us with a link that will be used for this feature?  Can we establish it before hand

It may even be better to remove the MS links from the alert and put them in our documentation so users are forced through our docs

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -52,6 +52,24 @@ spec:
...
+          See the Microsoft migration guide: https://learn.microsoft.com/en-us/azure/azure-monitor/logs/custom-logs-migrate
+          See the Microsoft Logs Ingestion API documentation: https://learn.microsoft.com/en-us/azure/azure-monitor/logs/logs-ingestion-api-overview
+      expr: |
+        log_forwarder_output_type{output="azureMonitor"} > 0
+      for: 1m
+      labels:
+        namespace: openshift-logging
```
jcantrill: CLFs are not restricted to this namespace

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
```diff
@@ -52,6 +52,24 @@ spec:
...
+        service: clusterlogforwarder
```
jcantrill, line 70: how does this relate to the CLF name?

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
Clee2691: I think it is safe to say retired because Microsoft actually has verbiage saying it will stop functioning

> The Azure Monitor HTTP Data Collector API has been deprecated and will no longer be functional as of 9/14/2026. It's been replaced by the [Logs ingestion API](https://learn.microsoft.com/en-us/azure/azure-monitor/logs/logs-ingestion-api-overview).

ref: https://learn.microsoft.com/en-us/previous-versions/azure/azure-monitor/logs/data-collector-api?tabs=powershell

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
Clee2691: I'm not sure what you mean here. This alert is specific to the CLF's config so I believe labeling the service with `clusterlogforwarder` is the right move here. 

This is similiar to the `ClusterLogForwarderOutputErrorRate` alert which labels the service as `clusterlogforwarder`

jcantrill:
@Clee2691 I'm not certain we can cherry-pick this to older releases without determining what are strategy will be.  The verbiage may need change to identify the options we choose

Clee2691:
> @Clee2691 I'm not certain we can cherry-pick this to older releases without determining what are strategy will be. The verbiage may need change to identify the options we choose

Yes you are right. I was overly zealous with the cherry picks. I'll close them when they come up. Is there a way to cancel cherry-picks?

bundle/manifests/collector_monitoring.coreos.com_v1_prometheusrule.yaml
johnwilkins: Our current Azure-specifc module is here: https://docs.redhat.com/en/documentation/red_hat_openshift_logging/6.4/html-single/configuring_logging/index#logging-forwarding-azure_configuring-log-forwarding

Clee2691: @johnwilkins That specific doc will be obsolete after `Sept 14th 2026` since forwarding logs to Azure with the Data Collector API will be retired. We should not add that link to this alert.

What Jeff was saying was that we need an updated doc about the `AzureLogsIngestion` sink which will need to be written with the feature in this PR https://github.com/openshift/cluster-logging-operator/pull/3223

johnwilkins: Thanks. Jeff left comments in another PR, so that doc has been updated accordingly. I've created a doc Jira to create new content for this.

PR #3228, opened 2026-03-19
LOG-8972: Enhance cluster-logging-operator to react to cluster TLS Profile updates

jcantrill:
## Summary

This PR enhances the cluster-logging-operator to react to cluster TLS Profile updates, ensuring both the operator itself and deployed collectors use the cluster's TLS security configuration.

Fixes [LOG-8972](https://redhat.atlassian.net/browse/LOG-8972)
Blocked by #3238 

### Operator's Own TLS Configuration
- **TLS Conversion Helpers**: Added functions to convert OpenShift `TLSProfileSpec` to Go `crypto/tls.Config`
- **Metrics Server**: Operator's metrics endpoint now uses cluster TLS profile (cipher suites and min TLS version)
- **TLS Profile Watcher**: New controller that watches for APIServer TLS profile changes and gracefully restarts the operator pod to apply new configuration

### Part A: Operator TLS Configuration
1. Added `internal/tls/tls.go` conversion functions:
   - `CipherSuiteStringToID`: Convert cipher suite names to crypto/tls IDs
   - `TLSVersionToConstant`: Convert TLS version strings to crypto/tls constants
   - `TLSConfigFromProfile`: Create crypto/tls.Config from TLSProfileSpec
   - `GetTLSConfigOptions`: Get TLS options for controller-runtime manager

2. Created `internal/controller/tlsprofile/` watcher controller:
   - Monitors APIServer resource for TLS profile changes
   - Compares current profile to initial startup profile
   - Gracefully exits operator pod when changes detected
   - Kubernetes automatically restarts pod with new configuration

3. Updated `cmd/main.go`:
   - Fetch cluster TLS profile at startup
   - Configure metrics server with cluster TLS profile
   - Register TLS profile watcher controller

### When Cluster TLS Profile Changes

**Operator:**
1. TLS profile watcher detects the change
2. Operator pod exits gracefully (exit code 0)
3. Kubernetes restarts the pod (~10 seconds downtime)
4. New pod applies updated TLS configuration to metrics server

### TLS Profile Precedence
1. **Output-specific profile** (highest priority)
2. **Cluster TLS profile** (default when output doesn't specify)

### Unit Tests
- ✅ TLS conversion functions: 12/12 tests passing
- ✅ TLS profile watcher controller: 5/5 tests passing

### Verification
- ✅ `make build` - Success
- ✅ `make lint` - 0 issues
- ✅ All new unit tests passing

### Manual Testing Recommendations
1. Deploy operator in a test cluster
[...]

PR #3238, opened 2026-04-01
LOG-8977: Secure the metrics endpoint to restrict scraping to autorized client

jcantrill:
### Description
This PR:
* Enhances to the CLO to restrict scraping of metrics to users with a valid token and certs
* Adds an operator e2e test for validating operator scraping with metrics and certs
* Adds a test helper class to execute 'oc run' commands
* Fixes 'oc get' helper to provide stderr when the command errors
* blocks #3228 

### Links
* https://redhat.atlassian.net/browse/LOG-8977

cc @kabirbhartiRH @anpingli


config/prometheus/service.yaml
```diff
@@ -1,12 +1,14 @@
...
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: cluster-logging-operator-metrics-certs
   labels:
     name: cluster-logging-operator
   name: cluster-logging-operator-metrics
 spec:
   ports:
-    - name: http-metrics
+    - name: https-metrics
       port: 8686
```
vparfonov: `8443`?

config/default/manager_metrics_cert_patch.yaml
```diff
@@ -0,0 +1,31 @@
...
+metadata:
+  name: cluster-logging-operator
+spec:
+  template:
+    spec:
+      containers:
+      - name: cluster-logging-operator
+        args:
+        - --metrics-secure=true
+        - --metrics-bind-address=:8686
```
Clee2691: Should this be `8443` for the secure endpoint?

config/network-policy/clo-permissive-network-policy.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     - Ingress
     - Egress
   ingress:
-    # Allow metrics (Port 8686)
+    # Allow metrics (Port 8443)
     - ports:
       - port: http-metrics
```
Clee2691: Shouldn't this be `https-metrics`?

config/rbac/metrics_auth_role.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: metrics-auth-role
+  name: cluster-logging-operator-metrics-auth
```
Clee2691, line 4: I think this should actually continue to be a generic `metrics-auth`. The LFME PR #3247 will also need to use this `ClusterRole`.

jcantrill: This is likely true but the change was made here because the operator-sdk, when building the bundle has some logic where when it does not see 'cluster-logging-operator-metrics-auth' for this role, it creates separate bundle manifest files for the clusterrole and clusterrolebinding instead of adding them into the CSV and identifying the permissions the operator requires.

This is a side-affect of our config files which are a kustomize manifest and when you follow the normal artifact generation, you would normally set the 'namePrefix'  field which applies to all created resources.  We have variances and may be able to resolve them otherwise but this change is not that.

Clee2691: I can reconcile a `clusterrolebinding` with this `clusterrole` for the LFME

PR #3239, opened 2026-04-01
LOG-8973: Enhance the log-file-metrics-exporter to be configurable with TLS Profile that includes curves

Clee2691:
## Summary
- Added support for TLS curves when it gets implemented in github.com/openshift/api/pull/2583.

This PR also depends on https://github.com/ViaQ/log-file-metric-exporter/pull/43 so that the LFME actually honors the curves.

Fixes [LOG-8973](https://redhat.atlassian.net/browse/LOG-8973)

## Details

Curves are implicitly supported through ECDHE cipher suites — Go's `crypto/tls` handles curve negotiation automatically.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/tls/tls.go
+++ b/internal/tls/tls.go
@@ -18,6 +18,10 @@ var (
 	DefaultTLSCiphers = configv1.TLSProfiles[DefaultTLSProfileType].Ciphers
 	// DefaultMinTLSVersion is the default minimum TLS version for API servers
 	DefaultMinTLSVersion = configv1.TLSProfiles[DefaultTLSProfileType].MinTLSVersion
+	// DefaultTLSGroups are the default TLS groups (elliptic curves) for key exchange.
+	// TODO: When openshift/api is updated to include Groups in TLSProfileSpec,
+	// read from configv1.TLSProfiles[DefaultTLSProfileType].Groups directly.
+	DefaultTLSGroups = []string{"X25519", "secp256r1", "secp384r1", "X25519MLKEM768"}
 )
 
 // FetchAPIServerTlsProfile fetches tlsSecurityProfile configured in APIServer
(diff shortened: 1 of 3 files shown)


internal/controller/logfilemetricsexporter/logfilemetricsexporter_controller.go
```diff
@@ -120,9 +124,39 @@ func (r *ReconcileLogFileMetricExporter) SetupWithManager(mgr ctrl.Manager) erro
...
 }
 
+// mapAPIServerToLogFileMetricExporters maps APIServer changes to all LogFileMetricExporter resources
+func (r *ReconcileLogFileMetricExporter) mapAPIServerToLogFileMetricExporters(ctx context.Context, obj client.Object) []ctrl.Request {
+	if !tls.IsClusterAPIServer(obj) {
+		return nil
+	}
+
+	lfmeList := &loggingv1alpha1.LogFileMetricExporterList{}
+	if err := r.Client.List(ctx, lfmeList); err != nil {
```
jcantrill: Do we need a or maybe any of this logic given we only support a singleton of openshift-logging/instance?

internal/tls/tls.go
```diff
@@ -69,3 +72,37 @@ func GetClusterTLSProfileSpec(apiServerTLSProfile *configv1.TLSSecurityProfile)
 
 	return defaultProfile
 }
+
+// IsClusterAPIServer returns true if the object is the cluster APIServer resource
+func IsClusterAPIServer(obj client.Object) bool {
```
jcantrill: possibly duplicated in #3228

jcantrill (commented):
#3238 adds an e2e test for scraping metrics we may be able to build off for changes like this one

internal/tls/tls.go
Clee2691: Yes I am aware this is duplicated and will need to be removed when that other PR merges. This is so that the code can function for this PR.

internal/controller/logfilemetricsexporter/logfilemetricsexporter_controller.go
Clee2691: Probably not. I can remove this.

PR #3247, opened 2026-04-08
LOG-8978: Enhance log-file-metrics-exporter to only allow scraping of metrics from a secure endpoint

Clee2691:
### Description   
                                                                                
- Adds a `ClusterRoleBinding` for the LFME ServiceAccount to perform `TokenReview` and `SubjectAccessReview` API calls using the `system:auth-delegator`  `ClusterRole`, enabling server-side bearer token validation on the metrics endpoint.
- Passes `-secureMetrics` flag to the LFME binary so the auth middleware is active                                  
- Configures the `ServiceMonitor` to include the Prometheus bearer token when scraping                             
- Updates the e2e test to authenticate with a bearer token            

Should land #3238 first before this PR

This PR also requires changes in the `LFME` for metrics to be secured. 
See: https://github.com/ViaQ/log-file-metric-exporter/pull/42

/cc @cahartma @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://redhat.atlassian.net/browse/LOG-8978

### DIFF
--- a/internal/auth/rbac.go
+++ b/internal/auth/rbac.go
@@ -2,6 +2,7 @@ package auth
 
 import (
 	"fmt"
+
 	"github.com/openshift/cluster-logging-operator/internal/reconcile"
 	"github.com/openshift/cluster-logging-operator/internal/runtime"
 	"github.com/openshift/cluster-logging-operator/internal/utils"
@@ -9,6 +10,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+const systemAuthDelegatorClusterRoleName = "system:auth-delegator"
 
 // ReconcileRBAC reconciles the RBAC specifically for the service account and SCC
 func ReconcileRBAC(k8sClient client.Client, rbacName, saNamespace, saName string, owner metav1.OwnerReference) error {
@@ -23,6 +26,30 @@ func ReconcileRBAC(k8sClient client.Client, rbacName, saNamespace, saName string
 
 	desiredSCCRoleBinding := NewServiceAccountSCCRoleBinding(saNamespace, rbacName, desiredSCCRole.Name, saName, owner)
 	return reconcile.RoleBinding(k8sClient, desiredSCCRoleBinding)
+}
+
+// ReconcileMetricsAuthRBAC reconciles the ClusterRoleBinding that binds system:auth-delegator to the service account
+func ReconcileMetricsAuthRBAC(k8sClient client.Client, commonName, saNamespace, saName string) error {
+	name := fmt.Sprintf("%s-metrics-auth", commonName)
+	desiredMetricsAuthRoleBinding := NewMetricsAuthClusterRoleBinding(name, saNamespace, saName)
+	return reconcile.ClusterRoleBinding(k8sClient, desiredMetricsAuthRoleBinding.Name, func() *rbacv1.ClusterRoleBinding { return desiredMetricsAuthRoleBinding })
+}
+
+// NewMetricsAuthClusterRoleBinding binds the system:auth-delegator ClusterRole to the given service account.
+func NewMetricsAuthClusterRoleBinding(name, saNamespace, saName string) *rbacv1.ClusterRoleBinding {
+	return runtime.NewClusterRoleBinding(
+		name,
+		rbacv1.RoleRef{
+			APIGroup: rbacv1.GroupName,
+			Kind:     "ClusterRole",
+			Name:     systemAuthDelegatorClusterRoleName,
+		},
+		rbacv1.Subject{
+			Kind:      "ServiceAccount",
+			Name:      saName,
+			Namespace: saNamespace,
+		},
+	)
 }
 
 // NewMetaDataReaderClusterRoleBinding stubs a clusterrolebinding to allow reading of pod metadata (i.e. labels)
--- a/internal/metrics/logfilemetricexporter/metric_exporter.go
+++ b/internal/metrics/logfilemetricexporter/metric_exporter.go
@@ -51,6 +51,11 @@ func Reconcile(lfmeInstance *loggingv1alpha1.LogFileMetricExporter,
 		return err
 	}
 
+	if err := auth.ReconcileMetricsAuthRBAC(requestClient, resNames.CommonName, lfmeInstance.Namespace, resNames.ServiceAccount); err != nil {
+		log.Error(err, "logfilemetricexporter.ReconcileMetricsRBAC")
+		return err
+	}
+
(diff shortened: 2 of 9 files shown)


internal/auth/rbac.go
```diff
@@ -25,6 +25,56 @@ func ReconcileRBAC(k8sClient client.Client, rbacName, saNamespace, saName string
 	return reconcile.RoleBinding(k8sClient, desiredSCCRoleBinding)
 }
 
+// ReconcileMetricsRBAC reconciles RBAC for securing the metrics endpoint with token authentication.
+// It creates:
+// - A ClusterRole allowing tokenreviews and subjectaccessreviews (for server-side token validation)
+// - A ClusterRoleBinding binding the above role to the service account
+func ReconcileMetricsRBAC(k8sClient client.Client, name, saNamespace, saName string) error {
```
jcantrill, line 32: Do we need to create these cluster roles or can we defer to OLM to create them from the CSV since the operator will require the same permissions.  I thought this was part of the manifest deployment but I do not recall which SA we use for LFME

Clee2691: We do not need to create these `clusterroles` and I only have this here for tests to work. I've noted in the PR description that his can be removed when #3238 is merged and then we only need to reconcile the `clusterrolebinding`

internal/auth/rbac.go
```diff
@@ -25,6 +27,30 @@ func ReconcileRBAC(k8sClient client.Client, rbacName, saNamespace, saName string
 	return reconcile.RoleBinding(k8sClient, desiredSCCRoleBinding)
 }
 
+// ReconcileMetricsAuthRBAC reconciles the ClusterRoleBinding for the metrics auth role
+func ReconcileMetricsAuthRBAC(k8sClient client.Client, name, saNamespace, saName string) error {
```
jcantrill: I'm not certain we need to create this rolebinding.  Have we tested without it?  This binding is only required for clients that are not prometheus AFAIK and I do not expect us to need to cater to those.  

My testing of the CLO metrics endpoint where I'm inspecting in cluster metrics still shows operator metrics being received without any specfic bidings.

internal/metrics/logfilemetricexporter/metric_exporter.go
```diff
@@ -51,6 +51,11 @@ func Reconcile(lfmeInstance *loggingv1alpha1.LogFileMetricExporter,
 		return err
 	}
 
+	if err := auth.ReconcileMetricsAuthRBAC(requestClient, resNames.CommonName, lfmeInstance.Namespace, resNames.ServiceAccount); err != nil {
```
jcantrill, line 54: Same here.  I don't think this is necessary as prometheus discovers the endpoints to scrape through the servicemonitor.  Please verify

internal/reconcile/rbac.go
```diff
@@ -58,6 +58,19 @@ func ClusterRoleBinding(k8sClient client.Client, name string, generator func() *
 	return err
 }
 
+func ClusterRole(k8sClient client.Client, desired *rbacv1.ClusterRole) error {
```
jcantrill: We have to be careful about cluster objects as in the past there was an issue with finalizers and ownerrefs which would cause the API server to hang.  It was some issue with these objects not being namespaced but being owned by namespaced objects

internal/auth/rbac.go
Clee2691: @jcantrill We will need to reconcile the `clusterrole` because the one from the manifests are bundled into the CSV and the `cluster-logging-operator-metrics-auth` `clusterrole` is not generated as part of the manifests.

internal/auth/rbac.go
Clee2691: The `metrics-auth` binding is for the server side, not the client side. It grants the LFME's ServiceAccount permission to create `TokenReview` and `SubjectAccessReview` resources so it can validate incoming bearer tokens. Without it, the LFME  can't authenticate any scraper, including Prometheus. The operator metrics work because the operator's ServiceAccount already has this binding via `config/rbac/metrics_auth_role_binding.yaml`. It is in the CSV of the operator.

internal/metrics/logfilemetricexporter/metric_exporter.go
Clee2691: Again this is for the server not the client (prometheus)

internal/reconcile/rbac.go
Clee2691: There is a default ClusterRole: `system:auth-delegator` that actually provides both the `tokenreview` and `subjectaccessreview`. We could bind that in order to reduce cluster scoped resources. However, we would need to ensure proper garbage collection of the CRB... somehow.

I think the only other option here to avoid ownership issues is to add the metrics-auth CRB to the manifests. It would be available when no LFME exists but it would be managed by OLM.

Thoughts?

internal/reconcile/rbac.go
```diff
@@ -43,17 +44,40 @@ func RoleBinding(k8Client client.Client, desired *rbacv1.RoleBinding) error {
...
 func ClusterRoleBinding(k8sClient client.Client, name string, generator func() *rbacv1.ClusterRoleBinding) error {
-	wantRoleBinding := generator()
-	crb := runtime.NewClusterRoleBinding(name, rbacv1.RoleRef{})
-	op, err := controllerutil.CreateOrUpdate(context.TODO(), k8sClient, crb, func() error {
-		// Update the clusterrolebinding with our desired state
-		crb.RoleRef = wantRoleBinding.RoleRef
-		crb.Subjects = wantRoleBinding.Subjects
+	desired := generator()
+	existing := runtime.NewClusterRoleBinding(name, rbacv1.RoleRef{})
+	err := k8sClient.Get(context.TODO(), client.ObjectKeyFromObject(existing), existing)
```
jcantrill, line 49: What is different here that this is Get/Create/Update vs CreateOrUpdate?

internal/reconcile/rbac.go
```diff
@@ -43,17 +44,40 @@ func RoleBinding(k8Client client.Client, desired *rbacv1.RoleBinding) error {
...
+		log.V(3).Info("Recreating clusterRoleBinding", "name", name)
+		return k8sClient.Create(context.TODO(), desired)
+	}
+
+	existing.Subjects = desired.Subjects
+	log.V(3).Info("Updating clusterRoleBinding", "name", name)
+	return k8sClient.Update(context.TODO(), existing)
+}
+
+func ClusterRole(k8sClient client.Client, desired *rbacv1.ClusterRole) error {
```
jcantrill: Do we still need this now we are relying upon an existing role?

internal/reconcile/rbac.go
Clee2691: So with clusterrolebindings you cannot update the `roleref` on an existing resource. You actually have to delete the binding and recreate it.

internal/reconcile/rbac.go
Clee2691: I will remove this

PR #3251, opened 2026-04-13
OBSDA-1383: Make Splunk output sourcetype configurable in CLF

marpears:
### Description

This PR allows for configuration of the Splunk output source type in the CLF using a new `sourceType` field. This is so that we can support users who have defined [custom source types in Splunk](https://help.splunk.com/en/splunk-enterprise/get-started/get-data-in/10.2/configure-source-types/create-source-types).

The `sourceType` field can be used only when `payloadKey` is defined, and allows for a templated value so that it can be retrieved from a pod label. If `sourceType` is not defined, the current logic is preserved where it defaults to _json. If using `payloadKey` without `sourceType`, the source type used will be either _json or generic_single_line, depending on the structure of the final event payload.

CLF configuration example:

```
  outputs:
    - name: splunk-app-team-1
      splunk:
        authentication:
          token:
            key: hecToken
            secretName: splunk-app-team-1
        sourceType: '{.kubernetes.labels.splunk/sourcetype||"generic_single_line"}'
        url: 'https://splunk.customer.com:8088'
      type: splunk
```

/cc @Clee2691 @cahartma
/assign @jcantrill

### Links
- JIRA: [OBSDA-1383](https://redhat.atlassian.net/browse/OBSDA-1383)
- https://redhat.atlassian.net/browse/LOG-9366

## Summary by CodeRabbit

* **New Features**
  * Added sourceType option for Splunk outputs (static or templated); must be used with payloadKey. Default sourcetype behavior documented when omitted.

* **Validation**
  * CRD validation enforces sourceType only when payloadKey is set. Splunk source static values now allow ":" characters.

* **Documentation**
  * Expanded Splunk forwarding docs with examples, defaults, and guidance.

* **Tests**
  * New unit and end-to-end tests covering payloadKey + sourceType scenarios.


api/observability/v1/output_types.go
```diff
@@ -1178,6 +1178,22 @@ type Splunk struct {
 	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Source",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
 	Source string `json:"source,omitempty"`
 
+	// SourceType can be used to specify a pretrained or custom sourcetype in Splunk.
+	// If not specified, `sourceType` will be "_json" or be determined automatically when using `payloadKey` based on the type of the final event payload.
```
jcantrill: > ... be determined automatically ...

I generally do not like this change but this statement is inaccurate.  There is no 'automatic' determination as the value configured in the sink is the same as provided in the API. 

This field is a hint to the splunk receiver how to parse the incoming message field, which why they can't figure it out how to do this on the server-side is beyond me.  By allowing user's to configure this value, we are allowing them to misrepresent what the log format of the actual work load is.  Given logs come from various workloads from the entire cluster  to a single sink means the 'pre-trained' model may not apply to every log.  How do we guarantee that with this implementation?

The CLF is intentionally opinionated meaning the current implementation is correct: you either get a JSON blob of ViaQ format or generic text from some other part of the ViaQ payload.  IMO the proper way to add this feature is to evaluate the message field and dynamically set it based upon a pattern defined in the message.  The caveat is whether it can be set for individual log messages.

That being said, what does it mean to the user,  to set this value to say  "apache http" and the sink additionally handles some well-known java log4j formatted logs?  

- Does the receiver politely process everything?
- Are we going to be on the hook to "fix" the case where the logs don't match the sourceType

jcantrill (commented):
There is no functional test to support this addition.  Also note we do not backport features to earlier releases

api/observability/v1/output_types.go
marpears: Hi @jcantrill thanks for reviewing this PR. In reply to your comments:

The problem this feature is intended to resolve is a scenario where a user has configured custom source types in splunk to interpret particular app log message content and does not want to use the default source type of `_json`. 

As an example, custom source types can be configured to trim the event payload, extract key/value pairs from `.message` and transform them into fields for use with dashboards. 

The inability to define a source type with the CLO is particularly an issue for a user who is moving away from a self-managed log collector that allows source type to be defined, and has a mature suite of splunk custom source types and dashboards. When the source type is `_json` instead of the required custom source type, dashboards and search queries can be broken if they rely on fields which have not been extracted, and also splunk costs increase if the log event is not trimmed. Before the CLO supported native splunk forwarding, often users would implement a self-managed fluentd instance but the splunk hec plugin is now archived. This, and the native support for splunk forwarding is a reason why a user would be seeking to move to the CLO.

If a source type needs to be referenced for a particular application's log events, but is not appropriate for other apps, the ClusterLogForwarder would need multiple splunk outputs so that it does not misrepresent the format of log events for other applications. 

I have altered the text that referred to "automatic" source type determination to try and make this part clearer and added examples of CLF configurations that use the `sourceType` feature in docs/features/logforwarding/outputs/splunk-forwarding.adoc

jcantrill: > If a source type needs to be referenced for a particular application's log events, but is not appropriate for other apps, the ClusterLogForwarder would need multiple splunk outputs so that it does not misrepresent the format of log events for other applications.

This is exactly the concern I have since fundamentally it requires users to know the log format of the workloads on the cluster and how to wire them from source container to output.  This is precisely the behavior we explicitly wanted to avoid since before the introduction of CRIO; this was the problem to be solved with the initial fluend and regex of the source.  Frankly, IMO, it completely defeats the purpose of an operator and it being the thing to manage your deployments.

That being said, we discussed this change within our architecture meeting and landed on fundamentally what you describe based on a customer's usage and experience with Splunk.  Essentially, we would expect the code to be further changed to:

* defining no messagePayloadKey, the functionality remains as-is
* defining no sourceType with messagePayloadKey, the functionality remains as-is
* defining sourceType with messagePayloadKey, the functionality populates the spec'd sourceType
* add a BIG BOLD WARNING to the API and repo docs that basically say you get what you asked for and that care needs to be taken when crafting pipelines

I will re-review the comment and try to comment these items accordingly.  Additionally, I will review the vector config api to confirm that sourceType can use templates.  The best use of this feature, if vector supports it, would be for users to label their workloads with the sourceType and then the collector could use that value when forwarding logs.

api/observability/v1/output_types.go
```diff
@@ -1178,6 +1178,22 @@ type Splunk struct {
...
+	//
+	//  1. "access_combined"
+	//
+	//  2. "log4j"
+	//
+	//  3. "my:custom:sourcetype"
+	//
+	// +kubebuilder:validation:Optional
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="SourceType",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	SourceType string `json:"sourceType,omitempty"`
```
jcantrill, line 1195: - Add a validation rule similar to https://github.com/marpears/cluster-logging-operator/blob/b41e2067585f143ab6b4ffa656ceae5e9d779a16/api/observability/v1/output_types.go#L80 that allows setting this value only when the payloadKey is modified
- Add comment that identifies this field can be templated  along with a validation rule like https://github.com/marpears/cluster-logging-operator/blob/b41e2067585f143ab6b4ffa656ceae5e9d779a16/api/observability/v1/output_types.go#L1177
- Add a comment note about it is on the administrator to configure the pipeline so the sourceType matches the log entry.  The collector makes no effort or validation to ensure they match.

As an aside to the last point, I do not know how Splunk behaves with a mismatch which is a little concerning to me.

Also there are api validation tests here https://github.com/marpears/cluster-logging-operator/tree/b41e2067585f143ab6b4ffa656ceae5e9d779a16/test/e2e/collection/apivalidations

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -53,15 +54,102 @@ spec:
 2. `url`: The base URL of the Splunk instance.
 3. `index`: Optional. The name of the index to send events to. If not specified, the default index defined within Splunk is used. This supports template syntax to allow dynamic per-event values.
 4. `source`: Optional. The source of events sent to this sink. This supports template syntax to allow dynamic per-event values.
-5. `indexedFields`: Optional. Fields to be added to Splunk index.
-6. `payloadKey`: Optional. Specifies record field to use as payload.
-7. `compression`: Optional. Compression configuration, available to are: `none`, `gzip`. Default is `none`.
+5. `sourceType`: Optional. Can be used to specify a pretrained or custom sourcetype in Splunk. If not specified it will default to `_json` and be detected by .payloadkey.
```
jcantrill, line 57: There is nothing 'detected' by the collector or operator.  I recommend:
- linking to the list of pretrained sourcetypes

```
`sourceType`: Optional. Can be used to specify a pretrained or custom sourcetype in Splunk when the payloadKey is additionally defined.  If not specified, it will default to `_json`.
```

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -53,15 +54,102 @@ spec:
...
+6. `indexedFields`: Optional. Fields to be added to Splunk index.
+7. `payloadKey`: Optional. Specifies record field to use as payload.
+8. `compression`: Optional. Compression configuration, available to are: `none`, `gzip`. Default is `none`.
 
 === `source`
 In Splunk, the `source` field typically identifies the origin of a log event.
 To ensure consistency and meaningful categorization, the `source` value can be dynamically derived from the `log_type` and `log_source` fields,
 following Cluster Log Forwarder's conventions.
 
+=== `sourceType`
```
jcantrill, line 67: This would be a good place to add an example where the sourceType is defined as a pod label and the CLF spec's the sourceType as a templated field to use from the pod meta.  This highlights toe value of labels which is a core concept

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -166,18 +254,18 @@ Transformed field:
...
 By default, `payloadKey` is not set, which means the complete log record is forwarded as the payload.
 
+If `sourceType` is not defined, payloadKey can override the default `sourcetype` value of `_json` :
 
-NOTE: Use `payloadKey` carefully. Selecting a single field as the payload may cause other important information in the log to be dropped, potentially leading to inconsistent or incomplete log events.
+* `sourceType`:
+** `_json` — used when `payloadKey` points to an object.
+** `generic_single_line` — used when the payload is a primitive value (e.g., string, number, boolean).
 
+NOTE: Use `payloadKey` carefully. Selecting a single field as the payload may cause other important information in the log to be dropped, potentially leading to inconsistent or incomplete log events.
```
jcantrill, line 263: My expectation is the greatest value comes from when this points to the message field

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -166,18 +254,18 @@ Transformed field:
 
 By default, `payloadKey` is not set, which means the complete log record is forwarded as the payload.
 
+If `sourceType` is not defined, payloadKey can override the default `sourcetype` value of `_json` :
```
jcantrill, line 257: Please explain.  I don't understand what this means.

jcantrill (commented):
Please add e2e validation tests and at least one  functional test

api/observability/v1/output_types.go
marpears: Hi @jcantrill, I have updated this change so that sourceType requires payloadKey to be defined so that it works in alignment with the scenarios you described. I have also added warnings in the API and repo docs.

api/observability/v1/output_types.go
marpears: I have added :

1. A validation rule that only allows setting sourceType when the payloadKey is defined
2. A comment that identifies the sourceType field can be templated along with a validation rule
3. A warning to the API and docs to make clear it is the Administrator's responsibility to configure the pipeline so the sourceType matches the log entry and that the collector makes no effort or validation to ensure they match
4. E2E API validation tests for the payloadKey and sourceType fields
5. Functional tests to cover static and templated sourceType configurations

docs/features/logforwarding/outputs/splunk-forwarding.adoc
marpears: Updated text and added link to pretrained source types

docs/features/logforwarding/outputs/splunk-forwarding.adoc
marpears: Added an example where the sourceType is retrieved from a pod label using templating.

docs/features/logforwarding/outputs/splunk-forwarding.adoc
marpears: Agreed, updated this section to add a mention of using the message field as the payloadKey.

docs/features/logforwarding/outputs/splunk-forwarding.adoc
marpears: I've updated this section to hopefully make this clearer

docs/features/logforwarding/outputs/splunk-forwarding.adoc
```diff
@@ -37,10 +37,11 @@ spec:
         url: 'http://example-splunk-hec-service:8088' # <2>
         index: '{.log_type||"main"}' # <3>
         source: '{.log_source||"undefined"}' # <4>
-        indexedFields: ['.log_type', '.log_source'] # <5>
-        payloadKey: '.kubernetes' # <6>
+        sourceType: '{.log_source||"undefined"}' # <5>
```
jcantrill: consider changing this example to `{.kubernetes.labels."splunk/source_type"||"generic_single_line"}` or something similar as my recommendation would be to advise user's to label their workload. They should have an idea or be able to figure out which pre-defined type matches their workload.

internal/generator/vector/output/splunk/splunk.go
```diff
@@ -56,6 +56,25 @@ if !is_null(payloadKey) {
...
+// If payload is key is invalid, sourcetype will fall back to '_json'
+var payloadKeysourceTypeTmpl = `
+payloadKey = %s
+sourceType = %s
+if !is_null(payloadKey) {
+	value = get!(., payloadKey) 
+	if !is_null(value) {
+        internal = ._internal
+        . = {}
+        . = set!(., payloadKey, value)
```
jcantrill, line 69: This does not seem correct.  I might expect the 'message' field become the value from what was pulled from the payloadkey.  I'll have to review the functional tests to wrap my head around the correctness here.

Looking at the sample code maybe this hasn't change but just been extracted to a new var?

internal/generator/vector/output/splunk/splunk_sink_with_payloadkey_and_sourcetype.toml
```diff
@@ -0,0 +1,54 @@
...
+if ._internal.log_source == "container" {
+   	._internal.splunk.source = join!([._internal.kubernetes.namespace_name, ._internal.kubernetes.pod_name, ._internal.kubernetes.container_name], "_")
+}
+if ._internal.log_type == "audit" {
+   ._internal.splunk.source = ._internal.log_source
+}
+payloadKey = ["message"]
+sourceType = to_string!(._internal.foo||"missing")
+if !is_null(payloadKey) {
+	value = get!(., payloadKey)
```
jcantrill, line 30: @vparfonov is this right?  Really early in our processing pipeline we move everything to `_internal`?  Should the payloadKey be rooted there?  Is this functioning correctly now?

test/e2e/collection/apivalidations/api_validations_test.go
```diff
@@ -126,5 +126,15 @@ var _ = Describe("", func() {
...
+		Entry("should pass for Splunk with payloadKey", "splunk-payloadkey.yaml", func(out string, err error) {
+			Expect(err).ToNot(HaveOccurred())
+		}),
+		Entry("should pass for Splunk if sourceType is used with payloadKey", "splunk-payloadkey-and-sourcetype.yaml", func(out string, err error) {
+			Expect(err).ToNot(HaveOccurred())
+		}),
+		Entry("should fail for Splunk if sourceType is not used with payloadKey", "splunk-sourcetype.yaml", func(out string, err error) {
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(MatchRegexp("sourceType can only be set when payloadKey is defined"))
+		}),
```
jcantrill, line 158: Please add another failure scenario where the sourceType is templeted

jcantrill:
@marpears You will need to rebase and force push.

docs/features/logforwarding/outputs/splunk-forwarding.adoc
marpears: Hi @jcantrill, I've changed sourceType examples in the docs and api to use  `'{.kubernetes.labels."splunk/sourcetype"||"generic_single_line"}'` to encourage the use of workload labelling. Also I've updated tests that use templates to the same for consistency.

internal/generator/vector/output/splunk/splunk.go
marpears: I extracted the logic that sets the log event payload from var `payloadKeyTmpl` to new var `payloadKeysourceTypeTmpl` - so the sample code you have highlighted is the same. 

The new var is used when both `payloadKey` and `sourceType` have been defined, and doesn't need the logic to check if the payload is an object or not, and instead sets `._internal.splunk.sourcetype` to the user defined `sourceType` if the user-defined `payloadKey` exists in the log event.

I was weighing up whether to extend `payloadKeyTmpl` with extra logic to handle when sourceType is defined, or add a new var. I arrived at adding a new var in order to keep the resulting Vector config to a minimum.

test/e2e/collection/apivalidations/api_validations_test.go
marpears: Done. Also fixed a bug where the use of single quotes in execCMD was cutting short crYaml if it contained single quotes, causing the test to be compromised.

internal/generator/vector/output/splunk/splunk_sink_with_payloadkey_and_sourcetype.toml
vparfonov: yes, its ok, `payloadKey` comes from user config

bundle/manifests/cluster-logging.clusterserviceversion.yaml
```diff
@@ -82,7 +82,11 @@ metadata:
     categories: OpenShift Optional, Logging & Tracing, Observability
     certified: "false"
     containerImage: quay.io/openshift-logging/cluster-logging-operator:latest
+<<<<<<< HEAD
```
jcantrill: Fall out from rebase.  I would delete one or the other then run 'make bundle' again

api/observability/v1/output_types.go
```diff
@@ -1331,6 +1332,35 @@ type Splunk struct {
...
+	//  3. foo-{.bar||"none"}
+	//
+	//  4. {.foo||.bar||"missing"}
+	//
+	//  5. foo.{.bar.baz||.qux.quux.corge||.grault||"nil"}-waldo.fred{.plugh||"none"}
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Pattern:=`^(([a-zA-Z0-9-_.\/])*(\{(\.[a-zA-Z0-9_]+|\."[^"]+")+((\|\|)(\.[a-zA-Z0-9_]+|\.?"[^"]+")+)*\|\|"[^"]*"\})*)*$`
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="SourceType",xDescriptors={"urn:alm:descriptor:com.tectonic.ui:text"}
+	SourceType string `json:"sourceType,omitempty"`
```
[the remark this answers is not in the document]
jcantrill: @vparfonov @Clee2691 This pattern matcher is restricted by VRL.  Do we have any idea if VRL would allow us to add a colon for this type? Given our template utilities it may complicate the other fields and we may need to think about how we would accept this suggestion?

test/e2e/collection/apivalidations/splunk-sourcetype.yaml
```diff
@@ -0,0 +1,30 @@
...
+      type: application
+  outputs:
+    - name: splunk-myapp
+      splunk:
+        authentication:
+          token:
+            key: hecToken
+            secretName: splunk-myapp
+        sourceType: log4j
+        url: 'https://splunk.customer.com:8088'
```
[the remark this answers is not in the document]
marpears: This is intended to be rejected as a failure test case scenario.

marpears:
Hi @jcantrill, I think it would be worth extending this feature with a special option of `sourceType: none` which does not set sourcetype in the Splunk sink config in Vector. This is to handle a case where a user may wish for a source type to be enforced at the Splunk side.

According to the [Vector docs](https://vector.dev/docs/reference/configuration/sinks/splunk_hec_logs/#sourcetype), it does not mandate sourcetype. Splunk will use a sourcetype to httpevent if not set, so using `sourceType: none` will come with a warning that the user must configure this suitably at the Splunk side. 

Do you agree?

jcantrill:
> Hi @jcantrill, I think it would be worth extending this feature with a special option of `sourceType: none` which does not set sourcetype in the Splunk sink config in Vector. This is to handle a case where a user may wish for a source type to be enforced at the Splunk side.

What does it mean to enforce it on the the receiver side? What is the receiver behavior?  Does it still accept the logs forwarded by the collector?

marpears:
> > Hi @jcantrill, I think it would be worth extending this feature with a special option of `sourceType: none` which does not set sourcetype in the Splunk sink config in Vector. This is to handle a case where a user may wish for a source type to be enforced at the Splunk side.
> 
> What does it mean to enforce it on the the receiver side? What is the receiver behavior? Does it still accept the logs forwarded by the collector?

Hi @jcantrill, I've done a test which confirmed that when the collector doesn't define a source type, Splunk accepts the log event and uses a default source type of `httpevent`.

But on reflection, I think the complexities and use case of a special option to not set the source type in the collector needs some further thought. I do not want to detract from the core objective of this PR, which was to allow for a user-defined source type in the collector, so I'll park this idea.

jcantrill:
> > > Hi @jcantrill, I think it would be worth extending this feature with a special option of `sourceType: none` which does not set sourcetype in the Splunk sink config in Vector. This is to handle a case where a user may wish for a source type to be enforced at the Splunk side.
> 
> But on reflection, I think the complexities and use case of a special option to not set the source type in the collector needs some further thought. I do not want to detract from the core objective of this PR, which was to allow for a user-defined source type in the collector, so I'll park this idea.

We discussed this in a team meeting and I do also don't like the idea of a "magic" word.  My suggested alternative was to test an empty string.  What does that mean to splunk? I am assuming that vector will send an empty string if configured which is different than configurating nothing and getting 'httpevent'

marpears:
> > > > Hi @jcantrill, I think it would be worth extending this feature with a special option of `sourceType: none` which does not set sourcetype in the Splunk sink config in Vector. This is to handle a case where a user may wish for a source type to be enforced at the Splunk side.
> > 
> > 
> > But on reflection, I think the complexities and use case of a special option to not set the source type in the collector needs some further thought. I do not want to detract from the core objective of this PR, which was to allow for a user-defined source type in the collector, so I'll park this idea.
> 
> We discussed this in a team meeting and I do also don't like the idea of a "magic" word. My suggested alternative was to test an empty string. What does that mean to splunk? I am assuming that vector will send an empty string if configured which is different than configurating nothing and getting 'httpevent'

I've given that scenario a try, with the source type set to an empty string in the collector, it resulted in Splunk using a source type of an empty string.

I did another test to confirm the source type needs to be completely omitted from the sink output config in the collector for Splunk to assign a source type. My test was done with a simple unstructured log message, and Splunk assigned the source type of `httpevent`. I may repeat that using a log4j style log message to see if Splunk's [automatic source type matching](https://help.splunk.com/en/splunk-enterprise/get-started/get-data-in/10.2/configure-source-types/why-source-types-matter) assigns it to its pretrained `log4j` source type rather than using `httpevent`.

jcantrill:
> I did another test to confirm the source type needs to be completely omitted from the sink output config in the collector for Splunk to assign a source type. My test was done with a simple unstructured log message, and Splunk assigned the source type of httpevent. I may repeat that using a log4j style log message to see if Splunk's [automatic source type matching](https://help.splunk.com/en/splunk-enterprise/get-started/get-data-in/10.2/configure-source-types/why-source-types-matter) assigns it to its pretrained log4j source type rather than using httpevent.

I believe I interpret your test as:
* exclude sourceType config and vector sends 'httpEvent'
* config sourceType to "" and vector sends ""

I am having difficulty finding reference documentation that may explicitly state whether sourceType is required but did find some documentation that advises  to use the enterprise interface to modify the receiver config.  My interpretation is an admin can fix it after the fact.  Given my opinions regarding the need to set this value in context of various workloads on a kubernetes cluster, I would lean towards disregarding the collector configuration you proposed regarding 'none'.  Admins have several options with the existing ClusterLogForwarder solution that should meet the broader needs:

* Set it to an explicit value
* Set it a templated value that is populated based on pod label
* Set it to "" and figure out how to adjust it on the receiver side

The more urgent issue is determining if we can allow `:` in the sourceType pattern since this seems to be a pattern that is common for splunk.  I believe we can but it requires testing to ensure that is allowed using VRL.  It may also require adjustment for some of our helper functions.

marpears:
> > I did another test to confirm the source type needs to be completely omitted from the sink output config in the collector for Splunk to assign a source type. My test was done with a simple unstructured log message, and Splunk assigned the source type of httpevent. I may repeat that using a log4j style log message to see if Splunk's [automatic source type matching](https://help.splunk.com/en/splunk-enterprise/get-started/get-data-in/10.2/configure-source-types/why-source-types-matter) assigns it to its pretrained log4j source type rather than using httpevent.
> 
> I believe I interpret your test as:
> 
> * exclude sourceType config and vector sends 'httpEvent'
> * config sourceType to "" and vector sends ""
> 
> I am having difficulty finding reference documentation that may explicitly state whether sourceType is required but did find some documentation that advises to use the enterprise interface to modify the receiver config. My interpretation is an admin can fix it after the fact. Given my opinions regarding the need to set this value in context of various workloads on a kubernetes cluster, I would lean towards disregarding the collector configuration you proposed regarding 'none'. Admins have several options with the existing ClusterLogForwarder solution that should meet the broader needs:
> 
> * Set it to an explicit value
> * Set it a templated value that is populated based on pod label
> * Set it to "" and figure out how to adjust it on the receiver side
> 
> The more urgent issue is determining if we can allow `:` in the sourceType pattern since this seems to be a pattern that is common for splunk. I believe we can but it requires testing to ensure that is allowed using VRL. It may also require adjustment for some of our helper functions.

Hi @jcantrill, 

Just to clarify the previous test: 

* exclude sourcetype config and Splunk defaults to 'httpevent'. According to the [Vector docs](https://vector.dev/docs/reference/configuration/sinks/splunk_hec_logs/#sourcetype), Splunk defaults to 'httpevent' if sourcetype is not defined by Vector, rather than Vector sending 'httpevent'.
* config sourcetype to "" and Vector sends ""

That aside, I have made further changes to allow `:` in `sourceType` and also in `source` so a consistent separator pattern is permitted in these fields. Along with this I have created/updated api validation and functional tests. I didn't encounter any VRL issues with colon chars when exercising the change with the functional tests which cover a range of scenarios. The use of `:` is not permitted by Kubernetes in pod label values, so any instances of `:` have to be set directly in the CLF within the `sourceType` or `source` fields. I've added a note in the docs.

test/e2e/collection/apivalidations/splunk-templated-source-1.yaml
```diff
@@ -0,0 +1,30 @@
...
+      name: myapp
+      type: application
+  outputs:
+    - name: splunk-myapp
+      splunk:
+        authentication:
+          token:
+            key: hecToken
+            secretName: splunk-myapp
+        source: '{.kubernetes.labels."splunk/source"||"generic_single_line"}'
```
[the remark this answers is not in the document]
marpears: This is a new api validation test to fill a gap where the current "source" field didn't have any api validation tests.

api/observability/v1/output_types.go
jcantrill: nit. this should be updated as well

api/observability/v1/output_types.go
```diff
@@ -1327,10 +1328,39 @@ type Splunk struct {
...
+	//
+	// WARNING: The administrator is responsible for configuring the pipeline so the source type matches the log entry. The collector makes no effort or validation to ensure they match.
+	//
+	// If SourceType is not specified, the source type used is `_json`. If using PayloadKey without SourceType, the source type used will be either `_json` or `generic_single_line`, depending on the structure of the final event payload.
+	// Details in: docs/features/logforwarding/outputs/splunk-forwarding.adoc
+	//
+	// The SourceType can be a combination of static and dynamic values consisting of field paths followed by `||` followed by another field path or a static value.
+	// A dynamic value is encased in single curly brackets `{}` and MUST end with a static fallback value separated with `||`.
+	//
+	// Static values can only contain alphanumeric characters along with dashes, underscores, dots and forward slashes.
```
jcantrill: same here

jcantrill (commented):
few nits and otherwise looks ready to go.  Will also need another rebase because of the bundle conflict

marpears:
> few nits and otherwise looks ready to go. Will also need another rebase because of the bundle conflict

Thanks @jcantrill , can this be included in the 6.6 release?

PR #3243, opened 2026-04-02
LOG-7348: Add WIF support to CLO for googleCloud output

cahartma:
### Description
Simple CLO changes  to allow googeCloud WIF authentication.    WIF functionality will also require an update to the vector client.    The CLO changes are backwards compatible with existing vector auth functionality.

I've included a design document to highlight the API decision.    There are no changes to the googleCloud authentication API.   Service account tokens are always projected for GCP outputs to support WIF.   This does not present any performance or security issues and greatly simplifies the implementation.

Users provide an authentication secret reference pointing to a credentials file:
- type: "service_account" (current static authentication)
OR
- type: "external_account" (Workload Identity Federation)

Vector detects the type of file and handles the appropriate authentication flow.     

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- CLO (this PR):   https://redhat.atlassian.net/browse/LOG-7348
- Vector update:   https://redhat.atlassian.net/browse/LOG-9171
- Epic:   https://redhat.atlassian.net/browse/LOG-3577

PR #3259, opened 2026-04-24
LOG-7348: Add CLO logic for GCP Workload Identity Federation

Clee2691:
## Description
This PR supersedes https://github.com/openshift/cluster-logging-operator/pull/3243.

The GCP output API has been updated with a `Token` field. The token source for WIF can either be from the CLF's service account or from a secret.

This PR also introduces validation of the credentials file.

/cc @vparfonov 
/assign @jcantrill 

### Links

- JIRA: https://redhat.atlassian.net/browse/LOG-7348
- Vector update: https://redhat.atlassian.net/browse/LOG-9171
- Epic: https://redhat.atlassian.net/browse/LOG-3577


docs/features/logforwarding/outputs/googlecloud/gcp-authentication-design.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+== Design Decision
+
+The GCP authentication API uses a *simplified, credentials-only approach*. The authentication type (static token vs. Workload Identity Federation) is determined by *inspecting the credentials file content* at runtime, rather than requiring users to specify an explicit type field.
+
+== API Structure
+
+[source,go]
+----
+type GoogleCloudLoggingAuthentication struct {
```
jcantrill: If we are going to leave API here in user facing documentation then we should replace it with yaml.  User's will never interact with configuration via Go code.  I'm certain with the examples that follow we could like remove this section completely.

docs/features/logforwarding/outputs/googlecloud/gcp-authentication-design.adoc
```diff
@@ -0,0 +1,155 @@
...
+}
+----
+
+=== Workload Identity Federation (WIF) Authentication
+
+*Credentials File*: GCP external account credentials file (`external_account.json`)
+
+*Requirements*:
+
+* The external_account.json must reference the token file path: `/var/run/ocp-collector/serviceaccount/token`
```
jcantrill: This was the validation I believe we were discussing which would require us to inspect the secret content.  Don't we do something similar for AWS?

docs/features/logforwarding/outputs/googlecloud/gcp-authentication-design.adoc
```diff
@@ -0,0 +1,155 @@
...
+
+* Vector reads the credentials file and inspects the `"type"` field
+* If `"type": "external_account"` → Vector uses the projected token for WIF authentication
+* If `"type": "service_account"` → Vector ignores the token and uses the static key directly
+
+* Token is mounted at: `/var/run/ocp-collector/serviceaccount/token`
+
+This detection happens *at runtime in the Vector collector client*, which  supports both authentication methods through the same `credentials_path` configuration.
+
+=== Vector Configuration
```
jcantrill: not needed.

docs/features/logforwarding/outputs/googlecloud/gcp-authentication-design.adoc
```diff
@@ -0,0 +1,155 @@
...
+====
+The `audience` field MUST use your GCP *PROJECT NUMBER* (e.g., "12345678901"), NOT your project ID (e.g., "my-project"). Using the project ID will result in HTTP 403 Forbidden errors. Get your project number with:
+
+[source,bash]
+----
+gcloud projects describe PROJECT_ID --format="value(projectNumber)"
+----
+====
+
+== Implementation Details
```
jcantrill: I don't believe this section is required.  I would also consider retitling this to remove the word 'design'.  These documents are not intended to capture design decisions

docs/features/logforwarding/outputs/googlecloud/gcp-authentication-design.adoc
Clee2691: Yes we do parse the roleARN for AWS and validate it. If we go the secret inspection route, we could implement `type` for GCP auth. I am not opposed to it.

We could fail fast during the reconciliation instead of relying on failing during Vector runtime

Were these design decisions below talked about in any meetings? Trying to understand where they came from.
> *No file inspection in validation* - Validation code should not need to read secret contents
*Simplicity* - Eliminates complex file parsing, error handling, and caching logic

docs/features/logforwarding/outputs/googlecloud/gcp-authentication-design.adoc
Clee2691: Based on this and other comments in a similar vein, I think we can remove this doc altogether. There are the other 2 docs which are more instructional for forwarding to GCP

There is an [enhancement proposal](https://github.com/openshift/enhancements/blob/master/enhancements/cluster-logging/gcp-external-account-auth.md) which documents the more technical side so I think we can forgo this doc.

docs/features/logforwarding/outputs/googlecloud/gcp-authentication-design.adoc
Clee2691: See: https://github.com/openshift/cluster-logging-operator/pull/3259#discussion_r3150646186

docs/features/logforwarding/outputs/googlecloud/gcp-authentication-design.adoc
Clee2691: See: https://github.com/openshift/cluster-logging-operator/pull/3259#discussion_r3150646186

docs/features/logforwarding/outputs/googlecloud/gcp-authentication-design.adoc
jcantrill: If adding a 'type' is the only way we introduce the validation then I am satisfied with leaving it as-is.  @cahartma and I discussed the merit of adding type and always projecting the SA token.  The consensus was everything that is needed for auth is in the file so there should not be any reason to make the admin explicitly choose a type given we can already implicitly do that from the file content.

Clee2691: No, adding type isn't the only way to introduce validation, just that we CAN add type IF we do validation. This would align the cloud offerings like AWS and Azure with type, which was our initial plan.

api/observability/v1/output_types.go
```diff
@@ -718,14 +718,55 @@ type Elasticsearch struct {
...
+	// WorkloadIdentity contains the GCP Workload Identity Federation credentials.
+	//
+	// +kubebuilder:validation:Optional
+	// +nullable
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Workload Identity Credentials"
+	WorkloadIdentity *GoogleCloudLoggingWorkloadIdentity `json:"workloadIdentity,omitempty"`
+}
+
+// GoogleCloudLoggingServiceAccount contains GCP service account key credentials.
+type GoogleCloudLoggingServiceAccount struct {
```
vparfonov: Is this breaking changes for 6.5?

api/observability/v1/output_types.go
```diff
@@ -718,14 +718,55 @@ type Elasticsearch struct {
...
 // GoogleCloudLoggingAuthentication contains configuration for authenticating requests to a GoogleCloudLogging output.
+// +kubebuilder:validation:XValidation:rule="self.type != 'serviceAccount' || has(self.serviceAccount)", message="Additional type specific spec is required for authentication"
+// +kubebuilder:validation:XValidation:rule="self.type != 'workloadIdentity' || has(self.workloadIdentity)", message="Additional type specific spec is required for authentication"
 type GoogleCloudLoggingAuthentication struct {
-	// Credentials points to the secret containing the `google-application-credentials.json`.
+	// Type is the type of Google Cloud Logging authentication to configure.
+	//
+	// +kubebuilder:validation:Required
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Authentication Type"
+	Type GoogleCloudLoggingAuthType `json:"type"`
```
jcantrill: - Change to be optional
- Verbiage mentions default if undefined is serviceaccount
- Correct order so Credentials is as it was
-  May need Initializer
- Update docs accordingly

PR #3264, opened 2026-04-30
LOG-9365: fix regression in maxOpenShift version to allow OCP v4.20

jcantrill:
### Description
This PR:
* fixes the regression to keep the operator from blocking cluster upgrade to  OCP v4.20

### Links
https://redhat.atlassian.net/browse/LOG-9365

### DIFF
--- a/bundle/metadata/properties.yaml
+++ b/bundle/metadata/properties.yaml
@@ -1,3 +1,3 @@
 properties:
   - type: olm.maxOpenShiftVersion
-    value: 4.19
+    value: "4.20"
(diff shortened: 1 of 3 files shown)


bundle/metadata/properties.yaml
```diff
@@ -1,3 +1,3 @@
 properties:
   - type: olm.maxOpenShiftVersion
-    value: 4.19
+    value: 4.20
```
xperimental: Needs quotes (otherwise it will be treated as a number -> `4.2`):


```suggestion
    value: "4.20"
```

xperimental (changes requested):

xperimental (approved):

PR #3266, opened 2026-05-04
LOG-9379: Fix CVE-2026-33813 in golang.org/x/image

vparfonov:
### Description

Update golang.org/x/image from v0.25.0 to v0.39.0 to fix CVE-2026-33813, which addresses a Denial of Service vulnerability via malformed WEBP image parsing on 32-bit platforms.

Also upgrades Go version to 1.25.7 and updates the Dockerfile to use golang:1.25 as the builder base image.

/cc @Clee2691 @cahartma 
/assign @jcantrill 

### Links

- Depending on PR(s): https://github.com/openshift/release/pull/78748
- GitHub issue:
- JIRA: https://redhat.atlassian.net/browse/LOG-9379
- Enhancement proposal:


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM golang:1.24 AS builder
+FROM golang:1.25.7 AS builder
```
jcantrill, line 1: Please verify we do not require a similar change to the Dockerfile.art

jcantrill:
This may also require changes to the CI in /release and ART.  cc @xperimental

Dockerfile
vparfonov: Dockerfile.art alredy based on `go 1.25.7 `https://github.com/openshift/cluster-logging-operator/blob/28a725b1578d5c5971884b17a14be143b0032b9d/Dockerfile.art#L1

xperimental: The `FROM` line in `Dockerfile.art` is actually never used for the actual build. It's just "a similar image that we can actually pull". The situation has recently changed meaning that I can improve the situation (we might be able to reduce the number of Dockerfiles due to this again).

The actual build currently uses 1.25.8 (see [here](https://github.com/openshift-eng/ocp-build-data/blob/logging-6.5/streams.yml#L5)).

PR #2895, opened 2024-12-04
[release-6.1] LOG-6469: Fix lokistack tech-preview validation logic

cahartma:
### Description
This is the simplified 6.1 fix based on the fix for [LOG-6346](https://issues.redhat.com//browse/LOG-6346)  in master (6.2).

- removed lokistack specifics since the output type is migrated internally to otlp
- removing condition type and prefilter validation
   - instead validate annotation during output spec validation
   - updated reason and status on the output condition
- additional unit tests for validation logic   

/cc @Clee2691 @vparfonov
/assign @jcantrill

### Links
- original: https://issues.redhat.com/browse/LOG-6346  
  - https://github.com/openshift/cluster-logging-operator/pull/2890
- https://issues.redhat.com/browse/LOG-6469

PR #3270, opened 2026-05-05
LOG-9356: Implement Collector ServiceMonitor Changes

Clee2691:
### Description
This PR makes use of [metrics collection profiles](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/metrics-collection-profiles.md). 

The CMO supports metrics collection profiles that select `ServiceMonitors` by the `monitoring.openshift.io/collection-profile` label. This change adds that label to all `ServiceMonitors` created by CLO, enabling cluster admins to reduce metrics cardinality by switching to the minimal profile.                 
                                                                                                                                                               
Because the minimal `ServiceMonitor` has a different resource name (e.g., `minimal-collector`) but targets the same backing service, a s`erviceName` parameter was added to ensure TLS ServerName verification uses the correct service DNS name.

Minimal Metrics - Collector:
- `logcollector_component_event_unmatched_count`                                                            
- `vector_buffer_byte_size`                             
- `vector_buffer_discarded_events_total`                                                                    
- `vector_buffer_events`
- `vector_buffer_sent_events_total`                                                                         
- `vector_component_discarded_events_total`                                                                 
- `vector_component_errors_total`                                                                           
- `vector_component_received_bytes_total`                                                                   
- `vector_component_received_event_bytes_total`                                                             
- `vector_component_received_events_total`
- `vector_component_sent_bytes_total`                                                                       
- `vector_events_in_total`                              
- `vector_http_client_errors_total`
- `vector_http_client_requests_sent_total`                                                                  
- `vector_http_client_responses_total`
- `vector_open_files`
- 
Minimal Metrics - LFME:
- `log_logged_bytes_total`

/cc @vparfonov 
/assign @jcantrill

### Links
- JIRA: https://redhat.atlassian.net/browse/LOG-9356

## Summary by CodeRabbit

* **New Features**
  * Metrics scraping now creates separate "full" and "minimal" ServiceMonitors, tags them with a collection-profile label, and applies configurable relabeling and per-service TLS server names.
  * Added a Prometheus/Thanos test helper for runtime metric queries.

* **Tests**
  * Unit and e2e tests expanded to validate profile labeling, relabel rule ordering/content, TLS server name usage, and that full/minimal profiles expose expected metrics.

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -39,18 +39,21 @@ const (
 	InjectTrustedCABundleLabel = "config.openshift.io/inject-trusted-cabundle"
 
 	//ServiceAccountSecretPath is the path to find the projected serviceAccount token and other SA secrets
-	ServiceAccountSecretPath   = "/var/run/ocp-collector/serviceaccount"
-	TrustedCABundleMountFile   = "tls-ca-bundle.pem"
-	TrustedCABundleMountDir    = "/etc/pki/ca-trust/extracted/pem/"
-	ElasticsearchName          = "elasticsearch"
-	VectorName                 = "vector"
-	KibanaName                 = "kibana"
-	LogfilesmetricexporterName = "logfilesmetricexporter"
-	LogfilesmetricexporterPort = int32(2112)
-	MetricsPortName            = "metrics"
-	MetricsPort                = int32(24231)
-	PodSecurityLabelEnforce    = "pod-security.kubernetes.io/enforce"
-	PodSecurityLabelValue      = "privileged"
+	ServiceAccountSecretPath          = "/var/run/ocp-collector/serviceaccount"
+	TrustedCABundleMountFile          = "tls-ca-bundle.pem"
+	TrustedCABundleMountDir           = "/etc/pki/ca-trust/extracted/pem/"
+	ElasticsearchName                 = "elasticsearch"
+	VectorName                        = "vector"
+	KibanaName                        = "kibana"
+	LogfilesmetricexporterName        = "logfilesmetricexporter"
+	LogfilesmetricexporterPort        = int32(2112)
+	MetricsPortName                   = "metrics"
+	MetricsPort                       = int32(24231)
+	MetricsCollectionProfileFull      = "full"
+	MetricsCollectionProfileMinimal   = "minimal"
+	MetricsCollectionProfileTelemetry = "telemetry"
+	PodSecurityLabelEnforce           = "pod-security.kubernetes.io/enforce"
+	PodSecurityLabelValue             = "privileged"
 	// Disable gosec linter, complains "possible hard-coded secret"
 	CollectorSecretsDir         = "/var/run/ocp-collector/secrets" //nolint:gosec
 	ConfigMapBaseDir            = "/var/run/ocp-collector/config"
--- a/internal/metrics/relabel.go
+++ b/internal/metrics/relabel.go
@@ -1,0 +1,114 @@
+package metrics
+
+import (
+	"strings"
+
+	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+)
+
+type metricAllowlistConfig struct {
+	allowedMetrics []string
+}
+
+type metricDropConfig struct {
+	labelName      string
+	labelValue     string
+	excludeMetrics []string
+}
+
+var collectorMinimalAllowlist = &metricAllowlistConfig{
+	allowedMetrics: []string{
+		// Metrics used in alerts (collector_alerts.yaml)
+		"logcollector_component_event_unmatched_count",
+		"vector_http_client_errors_total",
+		"vector_http_client_requests_sent_total",
+		"vector_http_client_responses_total",
+		"vector_buffer_byte_size",
+		"vector_component_errors_total",
+		"vector_component_received_events_total",
+
+		// Metrics used in recording rules (collector_alerts.yaml, telemetry_rules.yaml)
+		"vector_component_received_bytes_total",
+
+		// Metrics used in dashboards (openshift-logging-dashboard.json)
+		"vector_component_sent_bytes_total",
+		"vector_component_received_event_bytes_total",
+		"vector_open_files",
+		"vector_component_discarded_events_total",
+
+		// Additional buffer and event metrics
+		"vector_buffer_discarded_events_total",
+		"vector_buffer_events",
+		"vector_buffer_sent_events_total",
+		"vector_events_in_total",
+	},
+}
+
+var collectorMinimalDropConfigs = []metricDropConfig{
+	{
+		labelName:  "component_kind",
+		labelValue: "transform",
+		excludeMetrics: []string{
+			"vector_component_received_bytes_total",
+			"vector_component_received_event_bytes_total",
+			"vector_component_received_events_total",
+			"vector_component_sent_bytes_total",
+		},
+	},
+}
+
+var collectorTelemetryAllowlist = &metricAllowlistConfig{
+	allowedMetrics: []string{
+		// Used in recording rule (telemetry_rules.yaml)
+		"vector_component_received_bytes_total",
+	},
+}
+
+var lfmeMinimalAllowlist = &metricAllowlistConfig{
+	allowedMetrics: []string{
+		// Used in recording rule (collector_alerts.yaml) and dashboard
+		"log_logged_bytes_total",
+	},
+}
+
+var lfmeTelemetryAllowlist = &metricAllowlistConfig{
+	allowedMetrics: []string{
+		// Used in recording rule (collector_alerts.yaml)
+		"log_logged_bytes_total",
+	},
+}
+
+var CollectorMinimalRelabelConfigs = buildRelabelConfigs(collectorMinimalAllowlist, collectorMinimalDropConfigs)
+var LFMEMinimalRelabelConfigs = buildRelabelConfigs(lfmeMinimalAllowlist, nil)
+var CollectorTelemetryRelabelConfigs = buildRelabelConfigs(collectorTelemetryAllowlist, nil)
+var LFMETelemetryRelabelConfigs = buildRelabelConfigs(lfmeTelemetryAllowlist, nil)
+var FullRelabelConfigs = buildRelabelConfigs(nil, nil)
+
+func buildRelabelConfigs(allowlist *metricAllowlistConfig, dropConfigs []metricDropConfig) []*monitoringv1.RelabelConfig {
+	configs := []*monitoringv1.RelabelConfig{
+		{
+			SourceLabels: []monitoringv1.LabelName{"__name__"},
+			TargetLabel:  "__name__",
+			Regex:        "(.*)-(.*)",
+			Replacement:  "${1}_${2}",
+		},
+	}
+
+	if allowlist != nil && len(allowlist.allowedMetrics) > 0 {
+		configs = append(configs, &monitoringv1.RelabelConfig{
+			Action:       "keep",
+			SourceLabels: []monitoringv1.LabelName{"__name__"},
+			Regex:        strings.Join(allowlist.allowedMetrics, "|"),
+		})
+	}
+
+	for _, drop := range dropConfigs {
+		configs = append(configs, &monitoringv1.RelabelConfig{
+			Action:       "drop",
+			SourceLabels: []monitoringv1.LabelName{monitoringv1.LabelName(drop.labelName), "__name__"},
+			Regex:        drop.labelValue + ";(" + strings.Join(drop.excludeMetrics, "|") + ")",
+		})
+	}
+
+	return configs
+}
--- a/internal/metrics/service_monitor.go
+++ b/internal/metrics/service_monitor.go
@@ -17,7 +17,7 @@ const (
 	prometheusBearerTokenFile = "/var/run/secrets/kubernetes.io/serviceaccount/token"
 )
 
-func newServiceMonitor(namespace, name string, owner metav1.OwnerReference, selector map[string]string, portName string) *monitoringv1.ServiceMonitor {
+func newServiceMonitor(namespace, name, serviceName string, owner metav1.OwnerReference, selector map[string]string, portName string, metricRelabelConfigs []*monitoringv1.RelabelConfig, profile string) *monitoringv1.ServiceMonitor {
 	var endpoint = []monitoringv1.Endpoint{
 		{
 			Port:            portName,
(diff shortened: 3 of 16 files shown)


internal/metrics/service_monitor.go
```diff
@@ -17,7 +17,26 @@ const (
...
+	{
+		SourceLabels: []monitoringv1.LabelName{"__name__"},
+		TargetLabel:  "__name__",
+		Regex:        "(.*)-(.*)",
+		Replacement:  "${1}_${2}",
+	},
+	{
+		Action:       "keep",
+		SourceLabels: []monitoringv1.LabelName{"__name__"},
+		Regex:        "logcollector_component_event_unmatched_count|vector_buffer_byte_size|vector_buffer_discarded_events_total|vector_buffer_events|vector_buffer_sent_events_total|vector_component_discarded_events_total|vector_component_errors_total|vector_component_received_bytes_total|vector_component_received_event_bytes_total|vector_component_received_events_total|vector_component_sent_bytes_total|vector_events_in_total|vector_http_client_errors_total|vector_http_client_requests_sent_total|vector_http_client_responses_total|vector_open_files",
```
jcantrill: Consider building this list from a declared array for easier inspection and maintenance by a developer like: https://github.com/jcantrill/cluster-logging-operator/commit/6e9be1ec79d856554cadb0931f582000459d021c#diff-29d727cb83c8ad2c51397ba153849bea7adb0c89fe31e5ebcdb30251958df381R4-R8

internal/metrics/service_monitor.go
```diff
@@ -17,7 +17,26 @@ const (
...
+	},
+	{
+		Action:       "drop",
```
jcantrill: Similar comment here

internal/metrics/service_monitor.go
jcantrill: this might facilitate externalization in future should we wish to do that outside the API (i.e. operator configMap)

simonpasquier:
Have you considered implementing [collection profiles](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/metrics-collection-profiles.md)?

The general idea is that the admin defines the level of details for metrics at the cluster level. We have 3 profiles:
* full (default) => collects all metrics available
* minimal => collects only metrics relevant for alerting rules, dashboards and telemetry
* telemetry => collects only metrics required for telemetry

We can help with the implementation (in summary, you need to define 1 additional service monitor per profile). Happy to discuss directly on your forum.

cc @jan--f and @r2d2rnd (who informed us about the work in progress, thanks!)

Clee2691:
> Have you considered implementing [collection profiles](https://github.com/openshift/enhancements/blob/master/enhancements/monitoring/metrics-collection-profiles.md)?
> 
> The general idea is that the admin defines the level of details for metrics at the cluster level. We have 3 profiles:
> 
>     * full (default) => collects all metrics available
> 
>     * minimal => collects only metrics relevant for alerting rules, dashboards and telemetry
> 
>     * telemetry => collects only metrics required for telemetry
> 
> 
> We can help with the implementation (in summary, you need to define 1 additional service monitor per profile). Happy to discuss directly on your forum.
> 
> cc @jan--f and @r2d2rnd (who informed us about the work in progress, thanks!)

@simonpasquier I've made changes to this to implement profiles. Can someone on the monitoring team also review?

internal/metrics/relabel.go
```diff
@@ -0,0 +1,98 @@
...
+			Regex:        "(.*)-(.*)",
+			Replacement:  "${1}_${2}",
+		},
+	}
+
+	if allowlist != nil && len(allowlist.allowedMetrics) > 0 {
+		configs = append(configs, &monitoringv1.RelabelConfig{
+			Action:       "keep",
+			SourceLabels: []monitoringv1.LabelName{"__name__"},
+			Regex:        strings.Join(allowlist.allowedMetrics, "|"),
```
jcantrill, line 101: do we need to be concerned with sorting so we don't thrash the servicemonitor on reconcile?

internal/metrics/relabel.go
```diff
@@ -0,0 +1,98 @@
...
+		})
+	}
+
+	for _, drop := range dropConfigs {
+		configs = append(configs, &monitoringv1.RelabelConfig{
+			Action:       "drop",
+			SourceLabels: []monitoringv1.LabelName{monitoringv1.LabelName(drop.labelName), "__name__"},
```
jcantrill, line 108: do we need to be concerned with sorting so we don't thrash the servicemonitor on reconcile?

jcantrill (commented):
/approve

internal/metrics/relabel.go
Clee2691: I don't think we need sorting because the metrics are hardcoded string slices so when strings.Join is called, they will be preserved in the same order.

internal/metrics/relabel.go
Clee2691: I don't think we need sorting because the metrics are hardcoded string slices so when strings.Join is called, they will be preserved in the same order.

test/e2e/collection/metrics/profile_test.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+	})
+
+	AfterAll(func() {
+		e2e.Cleanup()
+	})
+
+	Context("ServiceMonitor labels", func() {
+		It("should create a full-profile ServiceMonitor for the collector", func() {
+			output, err := oc.Literal().From(
```
jcantrill: nit. No need to change unless you really want... there are 'oc.Get' commands that will do this

test/e2e/collection/metrics/profile_test.go
```diff
@@ -0,0 +1,126 @@
...
+		if err := e2e.WaitForDaemonSet(clf.Namespace, clf.Name); err != nil {
+			Fail(fmt.Sprintf("Failed waiting for collector: %v", err))
+		}
+	})
+
+	AfterAll(func() {
+		e2e.Cleanup()
+	})
+
+	Context("ServiceMonitor labels", func() {
```
jcantrill, line 90: This could be a table test

test/helpers/prometheus/prometheus.go
```diff
@@ -0,0 +1,107 @@
...
+	log "github.com/ViaQ/logerr/v2/log/static"
+	. "github.com/onsi/ginkgo/v2"
+)
+
+const (
+	tokenCmd      = `oc create token prometheus-k8s -n openshift-monitoring` //nolint:gosec
+	thanosHostCmd = `oc get route thanos-querier -n openshift-monitoring -o jsonpath={.spec.host}`
+)
+
+func executeCmd(cmd string) (string, error) {
```
jcantrill, line 21: I'm wondering if we should also be using test/helper/cmd... May not be applicable since those commands were originally added for some async work and reading from a container

Clee2691: I don't think this is applicable here. We just want to run a command and get an output, the files in `test/helper/cmd` are for streaming I/O

internal/constants/constants.go
```diff
@@ -39,18 +39,20 @@ const (
...
+	TrustedCABundleMountDir         = "/etc/pki/ca-trust/extracted/pem/"
+	ElasticsearchName               = "elasticsearch"
+	VectorName                      = "vector"
+	KibanaName                      = "kibana"
+	LogfilesmetricexporterName      = "logfilesmetricexporter"
+	LogfilesmetricexporterPort      = int32(2112)
+	MetricsPortName                 = "metrics"
+	MetricsPort                     = int32(24231)
+	MetricsCollectionProfileFull    = "full"
+	MetricsCollectionProfileMinimal = "minimal"
```
simonpasquier: we're [adding](https://github.com/openshift/cluster-monitoring-operator/pull/2904) a new telemetry profile called `telemetry` which collects only metrics required for Red Hat Telemetry. It would be good to account for it already.

simonpasquier (commented):
There are quite a few telemetry metrics associated to the logging team: https://github.com/openshift/cluster-monitoring-operator/blob/c9f25136c1cc6f9f7facbf1d4f128849fbb8e0f1/manifests/0000_50_cluster-monitoring-operator_04-config.yaml#L1024-L1080

We'd need to make sure that these metrics are available in all profiles.

internal/constants/constants.go
Clee2691: The minimal profile already accounts for those identified metrics. Some of them are operator only metrics and we are not going to be implementing profiles for the operator. These service monitors are for the collector and the LFME itself not the operator.

I don't think we need a `telemetry` profile as the minimal is sufficient. I think we'd be down to 1 metric for the telemetry only profile. The LFME already has 1 metric in the minimal set anyways

jcantrill: > I don't think we need a telemetry profile as the minimal is sufficient. I think we'd be down to 1 metric for the telemetry only profile. The LFME already has 1 metric in the minimal set anyways

What if the profile chosen is 'telemetry'?  Would that mean we would provide nothing since we are dependent upon the client to use the desired servicemonitor?

Clee2691: @simonpasquier
The LFME SM would basically be duplicated if we have both `minimal` and `telemetry`. Is there a way around this?

@jcantrill I think it would be the "full" metrics if there isn't a matching label for `telemetry`
For the collector, I think we could provide a `telemetry` only SM which would include only 1 metric

simonpasquier: Sorry for the late reply. In short, the operator selects the union of
* service/pod monitors which have no `monitoring.openshift.io/collection-profile` label
* service/pod monitors which have the `monitoring.openshift.io/collection-profile: <configured profile>` label

If you deploy "full" and "minimal" service monitors but no "telemetry" service monitor then no metrics will be scraped if the user selects that profile in the CMO configuration.

It might not be what you want and in doubt, you can add a telemetry service monitor which is a duplicate of the minimal profile. Good news is that the telemetry profile isn't yet merged so you don't have to rush.

rexagod: Hello! The telemetry profile effort is up for reviews, but we'd preferably like for the logging team to introduce a telemetry profile before we go ahead with that just so any regressions may be detected in time.

The following metrics were contributed to the telemetry space by `@openshift/openshift-team-logging`, and would need to go into the telemetry profile manifest on the logging team's end. The list below was picked from [#2895](https://github.com/openshift/cluster-monitoring-operator/pull/2895/changes#diff-5f31160051d77fcbf858b33e062fc02d42c38a4b53ff34e36884c56f686fb007) and may not be up-to-date.

```

  ┌─────┬────────────────────────────────────────────────────────────────┬──────────────────────────────────────────────────┐
  │  #  │                          Metric Name                           │                   Description                    │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 1   │ eo_es_storage_info                                             │ Elasticsearch storage information                │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 2   │ eo_es_redundancy_policy_info                                   │ Elasticsearch redundancy policy information      │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 3   │ eo_es_defined_delete_namespaces_total                          │ Total defined delete namespaces in Elasticsearch │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 4   │ eo_es_misconfigured_memory_resources_info                      │ Elasticsearch misconfigured memory resources     │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 5   │ cluster:eo_es_data_nodes_total:max                             │ Maximum Elasticsearch data nodes                 │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 6   │ cluster:eo_es_documents_created_total:sum                      │ Total Elasticsearch documents created            │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 7   │ cluster:eo_es_documents_deleted_total:sum                      │ Total Elasticsearch documents deleted            │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 8   │ pod:eo_es_shards_total:max                                     │ Maximum Elasticsearch shards per pod             │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 9   │ eo_es_cluster_management_state_info                            │ Elasticsearch cluster management state           │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 10  │ log_logging_info                                               │ Logging configuration information                │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 11  │ log_collector_error_count_total                                │ Total log collector errors                       │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 12  │ log_forwarder_pipeline_info                                    │ Log forwarder pipeline information               │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 13  │ log_forwarder_input_info                                       │ Log forwarder input information                  │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 14  │ log_forwarder_output_info                                      │ Log forwarder output information                 │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 15  │ cluster:log_collected_bytes_total:sum                          │ Total bytes collected by logging                 │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 16  │ cluster:log_logged_bytes_total:sum                             │ Total bytes logged                               │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 17  │ openshift_logging:log_forwarder_pipelines:sum                  │ Total log forwarder pipelines                    │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 18  │ openshift_logging:log_forwarders:sum                           │ Total log forwarders                             │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 19  │ openshift_logging:log_forwarder_input_type:sum                 │ Log forwarder inputs aggregated by type          │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 20  │ openshift_logging:log_forwarder_output_type:sum                │ Log forwarder outputs aggregated by type         │
  ├─────┼────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────┤
  │ 21  │ openshift_logging:vector_component_received_bytes_total:rate5m │ Vector component received bytes rate (5m)        │
  └─────┴────────────────────────────────────────────────────────────────┴──────────────────────────────────────────────────┘
```

PR #3273, opened 2026-05-07
LOG-9383: Normalize NO_PROXY entries for Vector compatibility

vparfonov:
#### Summary
- Add `normalizeNoProxy()` to `GetProxyEnvVars()` that adds a dot-prefixed duplicate for each plain domain entry in `NO_PROXY`, so Vector's no-proxy crate matches subdomains correctly
- IPs, CIDRs, wildcards, and already dot-prefixed entries are left unchanged
- Example: `apps.example.com,.cluster.local,10.0.0.1` becomes `apps.example.com,.apps.example.com,.cluster.local,10.0.0.1`

#### Problem
When the OpenShift cluster-wide proxy sets noProxy: `apps.example.com`, Go-based components (using [httpproxy](https://pkg.go.dev/golang.org/x/net/http/httpproxy)) correctly bypass the proxy for `elasticsearch.apps.example.com`. However, Vector's `no-proxy` crate only does exact string matching for plain domain entries — subdomains are not matched. This means log forwarding traffic goes through the proxy when it should be bypassed. Users must use `.apps.example.com` (with leading dot) as a workaround, which differs from how everyother OpenShift component interprets noProxy.
  
 #### Root cause
The `no-proxy` crate's `NoProxyItem::Plain` variant does exact match only: `Self::Plain(source) => source == value`. Per Go's httpproxy, curl, Python, and Ruby conventions, `example.com` should match both the domain itself and all subdomains.

####  Workaround approach
This PR normalizes `NO_PROXY` values in CLO before passing them to the Vector container. For each plain domain entry, a `.domain` duplicate is added so Vector's `WithDot` matching covers subdomains.

  This is a workaround until the upstream fix lands:
  - Upstream issue/PR: https://github.com/jdrouet/no-proxy/pull/12
  - Vector discussion: https://github.com/vectordotdev/vector/discussions/25383

Once the no-proxy crate is fixed and Vector bumps the dependency, this normalization can be removed (marked with a comment in the code).

/cc @Clee2691 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: https://redhat.atlassian.net/browse/LOG-9383
- Enhancement proposal:

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proxy environment variable handling to ensure consistent normalization of exclusion lists across different entry formats (domains, IP addresses, CIDR notations).

* **Tests**
  * Expanded test coverage for proxy environment variable normalization with comprehensive test cases for various entry formats and edge cases.

### DIFF
--- a/internal/utils/utils.go
+++ b/internal/utils/utils.go
@@ -6,6 +6,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"net"
 	"os"
 	"os/exec"
 	"reflect"
@@ -328,6 +329,7 @@ func GetProxyEnvVars() []v1.EnvVar {
 				if len(constants.ExtraNoProxyList) > 0 {
 					value = strings.Join(constants.ExtraNoProxyList, ",") + "," + value
 				}
+				value = normalizeNoProxy(value)
 			}
 			envVars = append(envVars, v1.EnvVar{
 				Name:  strings.ToLower(envvar),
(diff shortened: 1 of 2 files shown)


internal/utils/utils.go
```diff
@@ -338,6 +340,43 @@ func GetProxyEnvVars() []v1.EnvVar {
...
+		}
+		if net.ParseIP(host) != nil {
+			continue
+		}
+		if _, _, err := net.ParseCIDR(entry); err == nil {
+			continue
+		}
+		result = append(result, "."+entry)
+	}
+	return strings.Join(result, ",")
```
jcantrill, line 377: Consider sorting this before returning so we do reduce the risk of thrashing the collector

jcantrill:
After our discussion, I do not believe we should alter the behavior but advise to fix the definition or documentation to explain 'no_proxy'.  My concern is what if the intention is to only no_proxy a specific host instead of all the subdomains.  This would alter that behavior.  Given there is no spec, IMO the correct way to resolve the matter is either we document how to make it behave regardless of implementation or we land the rust PR you have proposed.

PR #3277, opened 2026-05-13
LOG-9389: Payload Too Large errors when sending logs to Azure monitor using Log Ingestion API

Clee2691:
### Description

This PR sets the default `maxWrite` value to `1,000,000` bytes for the `AzureLogsIngestion` output tuning parameter. This aligns with the maximum size of an API call enforced by Azure’s [Logs Ingestion API service limits](https://learn.microsoft.com/en-us/azure/azure-monitor/fundamentals/service-limits#logs-ingestion-api).

/cc @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://redhat.atlassian.net/browse/LOG-9389

## Summary by CodeRabbit

* **Documentation**
  * Deprecation date for the legacy Azure HTTP Data Collector API set to September 14, 2026; Azure Monitor output flagged for retirement with migration guidance to Azure Logs Ingestion.
  * Examples updated and guidance added about Azure 1MB request body limit.

* **Configuration & Behavior**
  * Default batching/max request size for Azure Logs Ingestion capped at 1MB.

* **Tests & Validation**
  * Added validation and tests to enforce and verify the 1MB maxWrite limit.

### DIFF
--- a/docs/features/logforwarding/outputs/azure/azure-logs-ingestion-forwarding.adoc
+++ b/docs/features/logforwarding/outputs/azure/azure-logs-ingestion-forwarding.adoc
@@ -4,7 +4,7 @@ https://learn.microsoft.com/en-us/azure/azure-monitor/logs/logs-ingestion-api-ov
 
 The `azureLogsIngestion` output type sends log events to Azure Monitor using the Logs Ingestion API and a Data Collection Rule (DCR). This is the recommended replacement for the legacy HTTP Data Collector API used by `azureMonitor`.
 
-Since the HTTP Data Collector API is deprecated, and will be removed in **September** 2026, it is recommended to use the Logs Ingestion API instead. The `AzureMonitor` output type will be removed in the future.
+Since the HTTP Data Collector API is deprecated, and will be removed in **September 14th 2026**, it is recommended to use the Logs Ingestion API instead. The `AzureMonitor` output type will be removed in the future.
 
 ==== Prerequisites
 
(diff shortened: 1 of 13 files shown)


Clee2691:
We should consider enforcing a hard limit of 1,000,000 bytes for `maxWrite`. Since Azure's API rejects anything larger, allowing a higher configuration value offers no functional benefit.

Thoughts? @jcantrill @vparfonov

jcantrill:
> We should consider enforcing a hard limit of 1,000,000 bytes for `maxWrite`. Since Azure's API rejects anything larger, allowing a higher configuration value offers no functional benefit.

Agree.  It should be possible to restrict in the API declaratively so any changes are rejected immediately. The concerning behavior, especially for audit logs, is I believe larger messages will get dropped entirely; we should confirm this.  Additionally we should understand if we at a minimum get 'discarded' metrics.  This would make it consistent with the other outputs and hopefully we can resolve with @vparfonov truncation work.

Clee2691:
> > We should consider enforcing a hard limit of 1,000,000 bytes for `maxWrite`. Since Azure's API rejects anything larger, allowing a higher configuration value offers no functional benefit.
> 
> Agree. It should be possible to restrict in the API declaratively so any changes are rejected immediately. The concerning behavior, especially for audit logs, is I believe larger messages will get dropped entirely; we should confirm this. Additionally we should understand if we at a minimum get 'discarded' metrics. This would make it consistent with the other outputs and hopefully we can resolve with @vparfonov truncation work.

We do get `vector_component_discarded_events_total` metrics for these discarded events. Large singe log messages ( >1MB) will be dropped. Confirmation is the hard limit that is stated in Azure. We already see `413`s when the collector starts.

I've tested setting the limit to 1MB and I do not see any 413s with the current logs produced.

This is something the customer has to be aware of when forwarding to Azure using the Logs Ingestion API

PR #3278, opened 2026-05-14
LOG-9359: Make ignore_older_secs configurable for audit inputs

vparfonov:
### Description

The default `ignore_older_secs` for audit file sources is hardcoded to `3600` seconds (1 hour). This is too short for host audit logs (`/var/log/audit/audit.log`) and other audit sources that commonly have gaps exceeding one hour in production environments. When the collector restarts (e.g., during a ClusterLogForwarder update) and a file hasn't been modified within that window, audit logs are either silently lost or duplicated (due to upstream vector#17208).

 This PR adds an `AuditInputTuningSpec` with an ignoreOlder field (`*time.Duration`, consistent with output tuning specs) to the ClusterLogForwarder API, allowing users to configure this threshold per audit input. CEL validation enforces a minimum of 1 second. The default remains 3600s for backward compatibility.

#### Example usage:
```
inputs:
  - name: my-audit
    type: audit
    audit:
      sources:
        - auditd
      tuning:
        ignoreOlder: 604800  # 7 days
```

####  Changes:
  - Added AuditInputTuningSpec struct and Tuning field to Audit in the API types
  - Updated all four audit source generators (auditd, kubeAPI, openshiftAPI, ovn) to read from the tuning spec
  - Regenerated deepcopy and CRD manifests
  - Added unit tests with fixture files

/cc 
/assign 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA:
- Enhancement proposal:

## Summary by CodeRabbit

* **New Features**
  * Audit input tuning: add an optional per-audit-input duration setting (ignoreOlder) to control how old an audit log file may be before the collector ignores it on restart (default 3600s, minimum 1s).
  * UI/CRD schema updated so this tuning appears in the operator console.

* **Documentation**
  * API docs updated for audit tuning and expanded Google Cloud Logging authentication/token configuration.


api/observability/v1/input_types.go
```diff
@@ -264,6 +264,21 @@ var (
...
+	// IgnoreOlderSeconds specifies the maximum number of seconds since the last modification
+	// of an audit log file before the collector ignores it. When the collector restarts, files
+	// that have not been modified within this time window may not be collected.
+	// Increase this value for audit sources with infrequent writes to prevent data loss.
+	// The default value is 3600 (1 hour).
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Minimum=1
+	// +operator-sdk:csv:customresourcedefinitions:type=spec,displayName="Ignore Older Seconds"
+	IgnoreOlderSeconds *int64 `json:"ignoreOlderSeconds,omitempty"`
```
jcantrill: replace with a duration for Consistency with output tuning.  Looks like there are library functions to address minimum https://kubernetes.io/docs/reference/using-api/cel/#kubernetes-quantity-library

alanconway:
Suggest making it a generic input parameter, could apply to all inputs.

Make sense to increase the limit. Beware the tradeoff: this will increase the amount of old log data read on start-up, vector will read more old rotation files than it otherwise would have. Set the limit high enough for the problems you know about, but not too much higher. E.g. if the worse case is 1d maybe 2d is high enough - just an example, you have the data.

jcantrill:
> Suggest making it a generic input parameter, could apply to all inputs.
> 

The proposed change is consistent with how we defined tuning for Outputs where they are type specific and application input tuning.  It may make sense to add this parameter in future to other inputs but I recommend waiting until we see an actual need given this is the first report for audit since the 6.x API has been in existence

PR #3306, opened 2026-06-08
fix(cloudwatch): cap maxWrite at 1MB and validate configuration

vparfonov:
### Description

`CloudWatch Logs PutLogEvents API` has a 1MB (`1,048,576` bytes) limit per request. Previously, `ClusterLogForwarder` allowed setting tuning.maxWrite to values exceeding this limit (e.g., 10M), causing batch upload failures with "Upload too large" errors and event drops.

Implement two layers of protection:

1. Validation: Reject `ClusterLogForwarder` specs at admission time if `CloudWatch` output maxWrite exceeds 1MB, providing users with clear error messages. 

2. Runtime capping: The generator caps `maxWrite` at 1MB in the Vector configuration (only when tuning is configured), ensuring safe operation even if validation is bypassed.

/cc @kabirbhartiRH 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: https://redhat.atlassian.net/browse/LOG-9448
- Enhancement proposal:

## Summary by CodeRabbit

* **Bug Fixes**
  * CloudWatch sink now enforces the 1MB-per-request API limit by clamping oversized batches to prevent send errors.
  * Validation now flags CloudWatch outputs configured with max-write sizes above the 1MB API limit.

* **New Features**
  * A default batch size is applied when none is configured to ensure compliance with the CloudWatch limit.
  * Generated configs/templates now set the CloudWatch batch max to 1MB.

* **Tests**
  * Added tests covering the CloudWatch max-write validation.

### DIFF
--- a/internal/generator/vector/output/aws/cloudwatch/cloudwatch.go
+++ b/internal/generator/vector/output/aws/cloudwatch/cloudwatch.go
@@ -22,6 +22,9 @@ import (
 const (
 	groupNameField                  = "cw_group_name"
 	templatedInternalGroupNameField = `{{ _internal.` + groupNameField + ` }}`
+
+	// CloudwatchDefaultMaxBytes CloudWatch Logs PutLogEvents API has a 1MB per request limit
+	CloudwatchDefaultMaxBytes = 1_048_576
 )
 
 func New(id string, o *adapters.Output, inputs []string, secrets observability.Secrets, op utils.Options) (_ string, sink types.Sink, tfs api.Transforms) {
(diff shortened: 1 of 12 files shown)


internal/generator/vector/output/aws/cloudwatch/cloudwatch.go
```diff
@@ -42,7 +45,12 @@ func New(id string, o *adapters.Output, inputs []string, secrets observability.S
 		} else {
 			s.Compression = sinks.CompressionType(o.GetTuning().Compression)
 		}
-		s.Batch = common.NewApiBatch(o)
+		if batch := common.NewApiBatch(o); batch != nil {
```
jcantrill, line 48: Do we always get a batch?  What happens when the user does not define tuning?  This condition exists regardless of us having a tuning or not so we should consider ALWAYS setting the max value

PR #3327, opened 2026-07-02
fix(syslog): quote payloadKey path to prevent VRL evaluation as expression

vparfonov:
#### Root Cause:
The payloadKey path (e.g., .message) was injected unquoted into the VRL template:
`payload_key = .message    # evaluates as VRL expression, reads the VALUE of .message`

This resolved `.message` to its runtime value (e.g., "Hello World"), which was then treated as a field path looking for a field literally named "Hello World", which doesn't exist.

For payloadKey: `"{.message}"` this only produced noisy warnings (since `.message` was left unchanged). For any other field (e.g., `{.kubernetes.namespace_name}`), it silently failed to set `.message` to the intended value.

#### Fix:
Quote the path as a string literal and strip the leading dot before splitting:
```
payload_key = ".message"   # string literal, stays as a path
  if starts_with(payload_key, ".") {
    payload_key = slice!(payload_key, 1)
  }
  path = split!(payload_key, ".")
  value, err = get(., path)
```

/cc @Clee2691 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: https://redhat.atlassian.net/browse/LOG-9562
- Enhancement proposal:

## Summary by CodeRabbit

* **New Features**
  * Improved syslog message extraction when using a configured payload key.
  * Enhanced handling for payload keys coming from structured JSON content (RFC5424).
* **Bug Fixes**
  * Removed inconsistent skip-and-warning behavior when the configured payload field is missing or empty.
  * When lookup fails, syslog now reliably falls back to a cleaned-up message derived from the event.
* **Tests**
  * Added functional coverage for RFC5424 payload key behavior with structured input.

### DIFF
--- a/internal/generator/vector/output/syslog/syslog.go
+++ b/internal/generator/vector/output/syslog/syslog.go
@@ -125,16 +125,14 @@ for_each(excluded_fields) -> |_index, field| {
 
 	payloadKeyConfiguredVRL = `
 # Payload key configured, going to use the payload key for .message field
-payload_key = %s
-if payload_key != null && payload_key != "" {
-  payload_key = string!(payload_key)
-  path = split!(payload_key, ".")
-  value, err = get(., path)
-  if err == null && value != null {
-    .message = value
-  } else {
-    log("payload_key not found in event, skipping", level: "warn")
-  }
+payload_key = "%s"
+if starts_with(payload_key, ".") {
+  payload_key = slice!(payload_key, 1)
+}
+path = split!(payload_key, ".")
+value, err = get(., path)
+if err == null && value != null {
+  .message = value
 } else {
   excluded_fields = ["_internal", "_syslog"]
   temp = .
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -234,11 +234,11 @@ test-env: ## Echo test environment, useful for running tests outside of the Make
 	IMAGE_TLS_SCANNER=$(IMAGE_TLS_SCANNER) \
 
 .PHONY: test-functional
-test-functional: test-functional-benchmarker-vector
```
jcantrill, line 237: The compilation and basic testing of this tool was dropped

Makefile
```diff
@@ -234,11 +234,11 @@ test-env: ## Echo test environment, useful for running tests outside of the Make
...
+test-functional:
 	RELATED_IMAGE_VECTOR=$(IMAGE_LOGGING_VECTOR) \
 	RELATED_IMAGE_LOG_FILE_METRIC_EXPORTER=$(IMAGE_LOGFILEMETRICEXPORTER) \
 	go test -race \
-		./test/functional/... \
+		./test/functional/outputs/syslog/... \
```
jcantrill: the other functional tests were removed

jcantrill (approved):
/hold

PR #3330, opened 2026-07-02
fix(syslog): guard facility/severity conversion against null values

vparfonov:
### Description

Add `!is_null()` guard to both facility and severity conversion checks so null values skip conversion entirely, letting the syslog encoder use its default (`Informational`).

/cc 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: NO-JIRA
- Enhancement proposal:


jcantrill (commented):
/approve
/lgtm
/label release/v6.6

PR #3331, opened 2026-07-03
[release-6.5]fix(kubernetes_log source): fix VRL crashing on null (not exist) namespace_name when pod metadata annotation fails

vparfonov:
### Description

When Vector's kubernetes_logs source fails to annotate an event with pod metadata (e.g. pod already deleted), the kubernetes.namespace_name field is null, causing string!() to abort the remap transform.

Add fallback VRL logic that extracts kubernetes metadata (namespace, pod name, pod UID, container name) from the log file path, which is always available even when the Kubernetes API lookup fails. The file path format `/var/log/pods/<ns>_<pod>_<uid>/<container>/<n>.log` is parsed via regex.

Also use fallible `string()` with null coalescing in the `match_any` call as defense-in-depth in case the file path parsing also fails.

/cc @Clee2691 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: https://redhat.atlassian.net/browse/LOG-9584
- Enhancement proposal:

### DIFF
--- a/internal/generator/vector/conf/complex.toml
+++ b/internal/generator/vector/conf/complex.toml
@@ -342,10 +342,20 @@ inputs = ["input_infrastructure_container"]
 source = '''
   . = {"_internal": .}
   if exists(._internal.stream) {._internal.kubernetes.container_iostream = ._internal.stream}
+
+if exists(._internal.file) && !exists(._internal.kubernetes.namespace_name) {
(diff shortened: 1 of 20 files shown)


jcantrill:
@vparfonov lets find a time to discuss this.  We should NEVER not have this base meta which is available from the file path.  This failure is because it was purged.  Maybe we just need fallback logic to extract it from the file or maybe this is a prime candidate for a vector enhancement.

PR #3332, opened 2026-07-03
fix(kubernetes_log source): fix VRL crashing on null (not exist) namespace_name when pod metadata annotation fails

vparfonov:
### Description

When Vector's kubernetes_logs source fails to annotate an event with pod metadata (e.g. pod already deleted), the kubernetes.namespace_name field is null, causing string!() to abort the remap transform.

Add fallback VRL logic that extracts kubernetes metadata (namespace, pod name, pod UID, container name) from the log file path, which is always available even when the Kubernetes API lookup fails. The file path format `/var/log/pods/<ns>_<pod>_<uid>/<container>/<n>.log` is parsed via regex.

Also use fallible `string()` with null coalescing in the `match_any` call as defense-in-depth in case the file path parsing also fails.

/cc @Clee2691 
/assign @jcantrill 

### Links

- Depending on PR(s):
- Port-Forward: https://github.com/openshift/cluster-logging-operator/pull/3331
- GitHub issue:
- JIRA:
- Enhancement proposal:

## Summary by CodeRabbit

* **Bug Fixes**
  * Made namespace-based infrastructure vs application log classification tolerant of missing or null Kubernetes namespace values by using a null-safe namespace check.
  * Prevented routing failures caused by strict namespace string conversions.
  * When Kubernetes metadata is absent, Kubernetes metadata is now recovered from the log file path to keep log classification working consistently.
  * Preserved existing matching behavior for `default`, `openshift*`, and `kube*` namespaces.
* **Tests**
  * Added a functional test covering metadata recovery for deleted pods with missing annotations.

### DIFF
--- a/internal/generator/vector/conf/complex.toml
+++ b/internal/generator/vector/conf/complex.toml
@@ -344,10 +344,20 @@ inputs = ["input_infrastructure_container"]
 source = '''
 . = {"_internal": .}
 if exists(._internal.stream) {._internal.kubernetes.container_iostream = ._internal.stream}
+
+if exists(._internal.file) && !exists(._internal.kubernetes.namespace_name) {
+  parsed_path, err = parse_regex(string!(._internal.file), r'^/var/log/pods/(?P<namespace>[^_]+)_(?P<pod>.+)_(?P<uid>[a-f0-9-]{36})/(?P<container>[^/]+)/\d+\.log$')
+  if err == null {
+    ._internal.kubernetes.namespace_name = parsed_path.namespace
+    ._internal.kubernetes.pod_name = parsed_path.pod
+    ._internal.kubernetes.pod_id = parsed_path.uid
+    ._internal.kubernetes.container_name = parsed_path.container
+  }
+}
 ._internal.log_source = "container"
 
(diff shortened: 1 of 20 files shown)


internal/generator/vector/conf/complex.toml
```diff
@@ -344,10 +344,20 @@ inputs = ["input_infrastructure_container"]
 source = '''
 . = {"_internal": .}
 if exists(._internal.stream) {._internal.kubernetes.container_iostream = ._internal.stream}
+
+if exists(._internal.file) && !exists(._internal.kubernetes.namespace_name) {
```
jcantrill, line 348: Can we add a functional test to verify this?  You would need to mock the container source to be a file so we could craft a path that simulates a delete pod: ref: https://github.com/openshift/cluster-logging-operator/blob/main/test/framework/functional/vector/deploy.go#L91-L91  This one is to mock journal sources but you can provide the framework with a config visitor to do a similar thing for a singular test.

Otherwise lgtm

vparfonov: add the test: `container_metadata_fallback_test.go`

PR #3338, opened 2026-07-07
feat: make terminationGracePeriodSeconds configurable via field of CollectorSpec

vparfonov:
#### Problem
The collector pods have `terminationGracePeriodSeconds` hardcoded to `10 seconds`. When `deliveryMode: AtLeastOnce` is configured, Vector uses disk buffers to persist log data. During shutdown, Vector needs time to flush in-memory events to disk and close connections to remote sinks. If the grace period is too short, Kubernetes sends `SIGKILL` while a write is in progress, corrupting the disk buffer and causing `CrashLoopBackOff` with `InvalidProtobufPayload` errors on restart.

#### Solution
Expose `terminationGracePeriodSeconds` as a configurable field in `spec.collector`, allowing users to set the appropriate grace period for their environment.
- New optional field: `spec.collector.terminationGracePeriodSeconds`
- Default remains 10 seconds (backward compatible — no change for existing users)

Example

 ```
 apiVersion: observability.openshift.io/v1
  kind: ClusterLogForwarder
  metadata:
    name: my-forwarder
    namespace: openshift-logging
  spec:
    collector:
      terminationGracePeriodSeconds: 60
    ...
```
 
/cc @Clee2691 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: https://redhat.atlassian.net/browse/LOG-9636
- Enhancement proposal:

## Summary by CodeRabbit

* **New Features**
  * Added a configurable termination grace period for collector pods, with a default of 10 seconds when not set.
  * Updated product documentation in the UI/manifest to describe the new collector setting and clarify several logging/template configuration examples.

* **Bug Fixes**
  * Collector pod settings now respect user-provided termination grace period values instead of always using the default.

### CODE REVIEW

**Comment by vparfonov** [comment]
@coderabbitai review

File: internal/collector/collector_test.go [hunk 1]
```diff
@@ -134,6 +134,28 @@ var _ = Describe("Factory#Daemonset", func() {
 
 		Describe("when creating the podSpec", func() {
 
+			Context("and evaluating terminationGracePeriodSeconds", func() {
+				It("should default to 10 when not specified", func() {
+					Expect(*podSpec.TerminationGracePeriodSeconds).To(Equal(int64(10)))
+				})
+
+				It("should use spec value when specified", func() {
```
jcantrill, line 142: nit.  How is this test different from the one which follows?

internal/collector/collector_test.go
```diff
@@ -411,6 +433,28 @@ var _ = Describe("Factory#Deployment", func() {
...
+					factory.CollectorSpec = obs.CollectorSpec{
+						TerminationGracePeriodSeconds: utils.GetPtr[int64](60),
+					}
+					podSpec = *factory.NewPodSpec(nil, obs.ClusterLogForwarderSpec{}, "1234", tls.GetClusterTLSProfileSpec(nil), constants.OpenshiftNS)
+					Expect(*podSpec.TerminationGracePeriodSeconds).To(Equal(int64(60)))
+				})
+
+				It("should allow custom values via spec", func() {
```
jcantrill: Same question here

vparfonov: remove duplicate

PR #3352, opened 2026-07-20
Log 9591 prefix collector ConfigMap with clf-

stilwank:
## Summary
- Renames the collector ConfigMap from `{name}-config` to `clf-{name}-config` so it no longer collides with LokiStack's `{name}-config`.
- On upgrade/reconcile, deletes the legacy `{name}-config` ConfigMap only when it is owned by the ClusterLogForwarder.
- Replaces the earlier LokiStack name-conflict validation approach (per review feedback).
### Problem
Nothing prevents a `LokiStack` and a `ClusterLogForwarder` from sharing the same `metadata.name` in the same namespace.
Both operators previously created ConfigMap `{name}-config`, so CLO could overwrite Loki's config and Loki pods CrashLoopBackOff.

This is always reproducible and was reported against Logging 6.5.1.

Ref: [LOG-9591](https://redhat.atlassian.net/browse/LOG-9591)

#### New files:

- `internal/factory/resource_names.go` — ConfigMap name is now `clf-{name}-config`.
- `internal/collector/config.go` — after reconciling the new ConfigMap, remove legacy `{name}-config` if owned by this CLF.
 - Removed LokiStack lookup / `NameConflict` validation (anti-pattern for the collector operator).
 

#### Test plan:

- [x] Unit tests pass
- [x] Same-named LokiStack + ClusterLogForwarder: `lokistack-config` (Loki) + `clf-lokistack-config` (Vector)
- [x] Loki pods healthy; collector pods Running
- [x] Legacy CLF-owned `{name}-config` removed after reconcile

#### Links: 
JIRA: https://redhat.atlassian.net/browse/LOG-9591

## Summary by CodeRabbit

* **New Features**
  * Added validation to detect naming conflicts between a `ClusterLogForwarder` and a `LokiStack` in the same namespace.
  * Reports a clear name-conflict status condition when both resources would target the same configuration.
* **Bug Fixes**
  * Prevents deployment when a naming conflict is detected.
  * Ensures conflicting collectors are undeployed safely.
* **Tests**
  * Added coverage for conflicts and valid scenarios across namespaces and resource names.


internal/validations/observability/validate_name.go
```diff
@@ -0,0 +1,48 @@
...
+
+	lokiStack := &unstructured.Unstructured{}
+	lokiStack.SetGroupVersionKind(lokiStackGVK)
+
+	if err := forwarderContext.Client.Get(context.TODO(), key, lokiStack); err != nil {
+		if errors.IsNotFound(err) || meta.IsNoMatchError(err) {
+			internalobs.RemoveConditionByType(&clf.Status.Conditions, obs.ConditionTypeName)
+			return
+		}
+		return
```
vparfonov:
```suggestion
      log.Error(err, "failed to validate LokiStack name", "name", clf.Name, "namespace", clf.Namespace)
		return
```

internal/validations/observability/validate_name.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	configMapName := clf.Name + "-config"
+	condition := internalobs.NewCondition(obs.ConditionTypeName, obs.ConditionFalse, obs.ReasonNameConflict, "")
+	condition.Message = fmt.Sprintf(
+		`ClusterLogForwarder name %q conflicts with LokiStack %q in namespace %q: both use ConfigMap %q`,
+		clf.Name, clf.Name, clf.Namespace, configMapName,
+	)
```
vparfonov:
```suggestion
condition := internalobs.NewCondition(obs.ConditionTypeName, obs.ConditionFalse, obs.ReasonNameConflict,
      fmt.Sprintf(`ClusterLogForwarder name %q conflicts with LokiStack %q in namespace %q: both use ConfigMap %q`,
          clf.Name, clf.Name, clf.Namespace, configMapName))
```

jcantrill:
@stilwank I would suggest an alternate solution that does not impose this kind of dependency on LokiStack which seems like an anti-pattern given this operator is for the log collector.  Consider, modifying the logic to prefix the name with `clf`.  For a CLF named "foo" this would translate to a configmap named `clf-foo-config`.  This would require logic to ensure upgrade addresses the name change.

cc @xperimental

stilwank:
@jcantrill Updated per your suggestion: 

Dropped the LokiStack dependency and renamed the collector ConfigMap to `clf-{name}-config`, with upgrade cleanup that deletes the legacy `{name}-config` only when owned by the ClusterLogForwarder. Verified on cluster with same-named LokiStack + CLF -->  both ConfigMaps coexist and Loki stays healthy.

jcantrill:
@xperimental per our slack discussion which was originally commented in issue, the solution should be to ensure the operator is unable to take ownership of an object it did not create and for customer's to name the original resources in such a way that derived objects do not have name conflicts.  I recommend this PR be closed given we prefer not to solve it this way.

PR #3348, opened 2026-07-16
[release-6.2] test(metrics): add functional tests verifying component_sent_bytes_total carries component_id labels

Clee2691:
### Description
This PR adds functional tests to the `cloudwatch` and `http` outputs to verify `component_sent_bytes_total` carries `component_id` labels

Relies on vector pr: https://github.com/ViaQ/vector/pull/284

/cc @vparfonov 
/assign @jcantrill 

### Links
- JIRA:  https://redhat.atlassian.net/browse/LOG-7893

PR #3379, opened 2026-07-29
test(metrics): add functional tests verifying component_sent_bytes_total carries component_id labels

Clee2691:
### Description
This PR is a manual cherry-pick of https://github.com/openshift/cluster-logging-operator/pull/3348.

This PR also adds a test for AWS S3 along with setup for metrics auth. 

/cc @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://redhat.atlassian.net/browse/LOG-9671

## Summary by CodeRabbit

- **Tests**
  - Added functional-test helpers to set up access to the collector in-cluster `/metrics` endpoint and poll until expected metrics appear.
  - Added/extended regression coverage for CloudWatch, S3, and HTTP outputs to verify `component_id` labeling and ensure `region` is present where expected.
- **Bug Fixes**
  - Strengthened metric validations to catch duplicate or incomplete label sets in output sent-byte metrics.

### DIFF
--- a/test/framework/functional/metrics.go
+++ b/test/framework/functional/metrics.go
@@ -1,0 +1,87 @@
+package functional
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+
+	"github.com/openshift/cluster-logging-operator/internal/constants"
+	"github.com/openshift/cluster-logging-operator/internal/runtime"
+)
+
+// SetupMetricsRBAC creates the cluster-scoped RBAC resources needed to scrape
+// the collector's /metrics endpoint from within a test pod. Names are prefixed
+// with the framework's namespace so parallel test packages cannot collide.
+// The returned function deletes all created resources and should be registered
+// with DeferCleanup or called in AfterEach.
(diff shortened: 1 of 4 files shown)


test/framework/functional/metrics.go
```diff
@@ -0,0 +1,87 @@
...
+	if err = f.Test.Create(metricsReaderRole); err != nil {
+		return nil, nil, nil, err
+	}
+
+	metricsReaderBinding = runtime.NewClusterRoleBinding(
+		roleName,
+		runtime.NewClusterRoleRef(roleName),
+		runtime.NewServiceAccountSubject("default", f.Namespace),
+	)
+	if err = f.Test.Create(metricsReaderBinding); err != nil {
```
jcantrill, line 37: I don't recall if we have a cleanup function for functional tests but do we need to be concerned about cleaning up these clusterrolbinding in a test cluster because of potentially "false positives" due to essentially "name" collision?

Clee2691: This is not a concern because the name of the binding and other rbac resources is derived like this:

`roleName := fmt.Sprintf("%s-%s-metrics-reader", f.Test.NS.Name, f.Name)`

The test's namespace name is unique for every tests so there shouldn't be any collision especially since the NS name is generated from `UniqueNameForTest()`

PR #3381, opened 2026-07-30
Log 9591 Do not overwrite ConfigMaps owned by other resources

stilwank:
## Summary
- Prevents Cluster Logging Operator from updating resources that already exist but are not owned by the desired owner (e.g. ClusterLogForwarder).
- Surfaces the conflict as a deployment error on the CLF (`Ready=False` / `DeploymentError`) so users can rename or clean up conflicting resources.
- Replaces the earlier ConfigMap rename approach (closed PR #3352) per review consensus with @jcantrill / @xperimental.

### Problem
When a `LokiStack` and a `ClusterLogForwarder` share the same `metadata.name`, CLO previously overwrote Loki’s ConfigMap `{name}-config` (and could also create/collide on other same-named objects such as DaemonSet/Service/ServiceMonitor). That could leave Loki CrashLoopBackOff and is a general name-collision hazard, not Loki-specific.
Ref: [LOG-9591](https://redhat.atlassian.net/browse/LOG-9591)
### Solution
- Add shared helper `internal/utils/ownership.go` (`EnsureCanUpdateOwnedResource`) that refuses update when existing `ownerReferences` do not match the desired owner (UID-based).
- Wire the check into reconcile paths for:
  ConfigMap, DaemonSet, Deployment, Service, ServiceAccount, ServiceMonitor, NetworkPolicy, Role/RoleBinding, ClusterRoleBinding, and trusted-CA ConfigMap (`trust_bundle.go`).
- Unit tests cover create/update of owned objects and refusal for foreign-owned / unowned objects.

### Notes
- Customers should still use different names for LokiStack and ClusterLogForwarder.
- Shared SCC handling is unchanged (different ownership model).
- Symmetric fix in the Loki Operator is tracked separately.
### Test plan
- [x] Unit tests (`go test ./internal/utils/ ./internal/reconcile/`)
- [x] LokiStack first → same-named CLF → Loki ConfigMap unchanged; CLF Ready=False with refuse-overwrite message
- [x] Different CLF name → collector resources created; CLF Ready=True
- [x] Owned collector resources still update normally on reconcile
### Links
JIRA: https://redhat.atlassian.net/browse/LOG-9591
Supersedes: https://github.com/openshift/cluster-logging-operator/pull/3352

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented reconciliation from overwriting resources owned by a different controller, including ConfigMaps, Deployments, DaemonSets, NetworkPolicies, Services, ServiceAccounts, ServiceMonitors, and RBAC objects.
  * Resources with conflicting ownership now remain unchanged and return a clear error.
  * Continued updates when ownership matches, including valid shared-owner scenarios.
  * Ensured owner references stay synchronized when permitted.

* **Tests**
  * Expanded coverage for ownership conflicts, matching owners, shared ownership, and protected resource state across supported resource types.

### DIFF
--- a/internal/reconcile/rbac.go
+++ b/internal/reconcile/rbac.go
@@ -16,7 +16,7 @@ import (
 func Role(k8Client client.Client, desired *rbacv1.Role) error {
 	role := runtime.NewRole(desired.Namespace, desired.Name)
 	op, err := controllerutil.CreateOrUpdate(context.TODO(), k8Client, role, func() error {
-		if err := utils.EnsureCanUpdateOwnedResource(role, desired.OwnerReferences); err != nil {
+		if err := utils.EnsureCanUpdateOwnedResource(role, desired.OwnerReferences...); err != nil {
 			return err
 		}
 		role.Rules = desired.Rules
@@ -41,7 +41,7 @@ func RoleBinding(k8Client client.Client, desired *rbacv1.RoleBinding) error {
 		return err
 	}
 
-	if err := utils.EnsureCanUpdateOwnedResource(existing, desired.OwnerReferences); err != nil {
+	if err := utils.EnsureCanUpdateOwnedResource(existing, desired.OwnerReferences...); err != nil {
 		return err
 	}
 
@@ -72,7 +72,7 @@ func ClusterRoleBinding(k8sClient client.Client, name string, generator func() *
 		return err
 	}
 
-	if err := utils.EnsureCanUpdateOwnedResource(existing, desired.OwnerReferences); err != nil {
+	if err := utils.EnsureCanUpdateOwnedResource(existing, desired.OwnerReferences...); err != nil {
 		return err
 	}
 
--- a/internal/reconcile/rbac.go
+++ b/internal/reconcile/rbac.go
@@ -6,6 +6,7 @@ import (
 
 	log "github.com/ViaQ/logerr/v2/log/static"
 	"github.com/openshift/cluster-logging-operator/internal/runtime"
+	"github.com/openshift/cluster-logging-operator/internal/utils"
 	rbacv1 "k8s.io/api/rbac/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -15,7 +16,9 @@ import (
 func Role(k8Client client.Client, desired *rbacv1.Role) error {
 	role := runtime.NewRole(desired.Namespace, desired.Name)
 	op, err := controllerutil.CreateOrUpdate(context.TODO(), k8Client, role, func() error {
-		// Update the role with our desired state
+		if err := utils.EnsureCanUpdateOwnedResource(role, desired.OwnerReferences); err != nil {
+			return err
+		}
 		role.Rules = desired.Rules
 		role.OwnerReferences = desired.OwnerReferences
 		return nil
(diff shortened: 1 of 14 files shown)


internal/utils/ownership.go
```diff
@@ -0,0 +1,56 @@
...
+)
+
+// EnsureCanUpdateOwnedResource returns nil when the object has not been persisted yet
+// (create path) or when it is already owned by all of the desired owners.
+// If the object exists and is not owned by the desired owners (foreign-owned or
+// unowned), it returns an error and callers must not mutate the object.
+//
+// Ownership is matched by OwnerReference UID so additional ownerRefs on the
+// object do not cause false conflicts (LOG-9591).
+func EnsureCanUpdateOwnedResource(obj metav1.Object, desiredOwners []metav1.OwnerReference) error {
```
jcantrill: Consider modifying the signature to be `(obj metav1.Object, desiredOwners ...metav1.OwnerReference)` so you can pass a single owner without needing to wrap it an array everytime you make a call.

jcantrill: How does this differ from `utils.HasSameOwner` and why can't they functionally be combined?

internal/utils/ownership.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	if IsOwnedByDesired(obj.GetOwnerReferences(), desiredOwners) {
+		return nil
+	}
+	return ResourceOwnershipConflictError(obj)
+}
+
+// IsOwnedByDesired reports whether every desired owner UID is present among current owners.
+// An empty desired owner list only matches when current owners are also empty.
+func IsOwnedByDesired(current, desired []metav1.OwnerReference) bool {
```
jcantrill: The utils also has a set that simplies this logic

internal/reconcile/rbac.go
```diff
@@ -65,6 +72,10 @@ func ClusterRoleBinding(k8sClient client.Client, name string, generator func() *
 		return err
 	}
 
+	if err := utils.EnsureCanUpdateOwnedResource(existing, desired.OwnerReferences...); err != nil {
```
jcantrill, line 75: Need to verify but I'm not certain Cluster scoped objects have ownerrefs.  There were issues in the past where you could wedge deletion of a namespaced object because it claimed ownership over cluster scoped object.  This would mean there needs to be an alternate check about "ownership" (i.e. the managed by? label?)

internal/reconcile/rbac.go
```diff
@@ -75,6 +86,7 @@ func ClusterRoleBinding(k8sClient client.Client, name string, generator func() *
 	}
 
 	existing.Subjects = desired.Subjects
+	existing.OwnerReferences = desired.OwnerReferences
```
jcantrill, line 89: applies here too

internal/utils/ownership.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// hasSameOwnerUIDs reports whether current and desired have the same owner UIDs.
+// An empty desired owner list only matches when current owners are also empty.
+func hasSameOwnerUIDs(current, desired []metav1.OwnerReference) bool {
+	return ownerUIDSet(current).Equal(ownerUIDSet(desired))
+}
+
+func ownerUIDSet(refs []metav1.OwnerReference) sets.Set[string] {
+	uids := sets.New[string]()
```
jcantrill, line 34: utils package also has a set that could be used here facilitates set creation... and may negate the need for this function

PR #3383, opened 2026-08-03
fix(security): CVE-2026-10609 verify CLF creator authorization for SA token usage (LOG-9441)

vparfonov:
### Description

This PR addresses CVE-2026-10609, a privilege escalation vulnerability where users could reference any `ServiceAccount` in a `ClusterLogForwarder (CLF)` and forward its token to external log outputs without having explicit permission to use that `ServiceAccount`.

#### Root Cause
The operator did not validate whether a user had permission to use the `ServiceAccount` specified in a CLF. This allowed privilege escalation if a cluster admin had granted a user's pod SCC permissions but not SA usage rights the user could then create a CLF to steal that SA's token.

#### Solution

- **Added a ValidatingAdmissionWebhook**: Intercepts all `CREATE` and `UPDATE` operations for `ClusterLogForwarder` resources.
- **SubjectAccessReview (SAR) Enforcement**: Performs an unconditional SAR check using the requesting user's full identity (Username, UID, and Groups) to verify they possess the use verb on the referenced `ServiceAccount`.
- **Zero-Downtime Upgrades**: Because the check is performed purely at the admission webhook level, existing CLFs already in `etcd` will continue to run without interruption upon operator upgrade.

#### Additionally
- fix a few formatting issue
- update `hack/run-linter` script shebangs to `#!/usr/bin/env bash` for better portability

/cc @Clee2691 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: https://redhat.atlassian.net/browse/LOG-9441
- Enhancement proposal:

## Summary by CodeRabbit

* **New Features**
  * Added admission validation for ClusterLogForwarder resources that use ServiceAccount tokens.
  * Requests now verify authorization to use referenced ServiceAccounts.
  * Added automatic tracking of the requesting identity for auditing and validation.
  * Enabled mutating and validating webhooks for ClusterLogForwarder creation and updates.

* **Bug Fixes**
  * Preserved compatibility for resources without modifier metadata.
  * Improved permission handling for infrastructure and audit log inputs.

* **Tests**
  * Added unit and end-to-end coverage for authorized, denied, and legacy ServiceAccount usage.

### DIFF
--- a/internal/webhook/clusterlogforwarder_webhook.go
+++ b/internal/webhook/clusterlogforwarder_webhook.go
@@ -1,0 +1,86 @@
+package webhook
+
+import (
+	"context"
+	"fmt"
+
+	obs "github.com/openshift/cluster-logging-operator/api/observability/v1"
+	authorizationapi "k8s.io/api/authorization/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// +kubebuilder:webhook:path=/validate-observability-openshift-io-v1-clusterlogforwarder,mutating=false,failurePolicy=fail,sideEffects=None,groups=observability.openshift.io,resources=clusterlogforwarders,verbs=create;update,versions=v1,name=vclusterlogforwarder.observability.openshift.io,admissionReviewVersions=v1
+
+type ClusterLogForwarderValidator struct {
+	Client client.Client
+}
+
+var _ webhook.CustomValidator = &ClusterLogForwarderValidator{}
+
+func SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(&obs.ClusterLogForwarder{}).
+		WithValidator(&ClusterLogForwarderValidator{Client: mgr.GetClient()}).
+		Complete()
+}
+
+func (v *ClusterLogForwarderValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	return v.validateSAUsage(ctx, obj)
+}
+
+func (v *ClusterLogForwarderValidator) ValidateUpdate(ctx context.Context, _ runtime.Object, newObj runtime.Object) (admission.Warnings, error) {
+	return v.validateSAUsage(ctx, newObj)
+}
+
+func (v *ClusterLogForwarderValidator) ValidateDelete(_ context.Context, _ runtime.Object) (admission.Warnings, error) {
+	return nil, nil
+}
+
+// validateSAUsage verifies the requesting user has permission to use the referenced ServiceAccount.
+// This check is unconditional — any SA can be granted SCC permissions and mounted to workloads,
+// so we validate usage regardless of whether outputs forward the SA token.
+func (v *ClusterLogForwarderValidator) validateSAUsage(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	clf, ok := obj.(*obs.ClusterLogForwarder)
+	if !ok {
+		return nil, fmt.Errorf("expected ClusterLogForwarder, got %T", obj)
+	}
+
+	saName := clf.Spec.ServiceAccount.Name
+	if saName == "" {
+		return nil, nil
+	}
+
+	req, err := admission.RequestFromContext(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get admission request from context: %w", err)
+	}
+
+	sar := &authorizationapi.SubjectAccessReview{
+		Spec: authorizationapi.SubjectAccessReviewSpec{
+			User:   req.UserInfo.Username,
+			Groups: req.UserInfo.Groups,
+			UID:    req.UserInfo.UID,
+			ResourceAttributes: &authorizationapi.ResourceAttributes{
+				Group:     "",
+				Verb:      "use",
+				Resource:  "serviceaccounts",
+				Name:      saName,
+				Namespace: clf.Namespace,
+			},
+		},
+	}
+
+	if err := v.Client.Create(ctx, sar); err != nil {
+		return nil, fmt.Errorf("failed to check service account usage permission: %w", err)
+	}
+
+	if !sar.Status.Allowed {
+		return nil, fmt.Errorf("user %q is not authorized to use service account %q in namespace %q", req.UserInfo.Username, saName, clf.Namespace)
+	}
+
+	return nil, nil
+}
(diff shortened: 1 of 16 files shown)


vparfonov:
@coderabbitai review

vparfonov:
@coderabbitai review

internal/webhook/clusterlogforwarder_webhook.go
```diff
@@ -0,0 +1,132 @@
...
+	return nil, nil
+}
+
+func (v *ClusterLogForwarderValidator) validateSAUsage(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	clf, ok := obj.(*obs.ClusterLogForwarder)
+	if !ok {
+		return nil, fmt.Errorf("expected ClusterLogForwarder, got %T", obj)
+	}
+
+	if !internalobs.Outputs(clf.Spec.Outputs).NeedServiceAccountToken() {
```
jcantrill: We care about the SA usage regardless of the output since it can be granted SCC permissions and mounted to any workload

internal/constants/annotations.go
```diff
@@ -13,6 +13,8 @@ const (
 	// CLO's default log level for vector is `warn`: https://issues.redhat.com/browse/LOG-3435
 	AnnotationVectorLogLevel = "observability.openshift.io/log-level"
 
+	AnnotationModifier = "observability.openshift.io/modifier"
```
jcantrill: Given this is serialized userinfo, consider renaming this to maybe "resource-owner" or maybe "serviceaccount-owner".  Our main concern is abuse of the serviceaccount and we wish to verify the user who declared it can be used by the CLF is aware and approves of the elevated privileges it has

internal/webhook/clusterlogforwarder_webhook.go
```diff
@@ -0,0 +1,132 @@
...
+
+	if clf.Annotations == nil {
+		clf.Annotations = make(map[string]string)
+	}
+	clf.Annotations[constants.AnnotationModifier] = string(userInfoJSON)
+
+	return nil
+}
+
+func (v *ClusterLogForwarderValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
```
jcantrill, line 31: Introducing this webhook will allow us to move most all our "post create validations" here in future.

internal/webhook/clusterlogforwarder_webhook.go
```diff
@@ -0,0 +1,132 @@
...
+	if !ok {
+		return fmt.Errorf("expected ClusterLogForwarder, got %T", obj)
+	}
+
+	req, err := admission.RequestFromContext(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get admission request from context: %w", err)
+	}
+
+	userInfoJSON, err := json.Marshal(req.UserInfo)
```
jcantrill: We may have an internal method or methods that wrapper this library... internal/util/json?

internal/webhook/clusterlogforwarder_webhook.go
```diff
@@ -0,0 +1,132 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get admission request from context: %w", err)
+	}
+
+	saName := clf.Spec.ServiceAccount.Name
+	if saName == "" {
+		return nil, nil
+	}
+
+	sar := &authorizationapi.SubjectAccessReview{
```
jcantrill, line 62: Do we already have internal/runtime and buiders that do this from adding SAR to metric requests?

internal/webhook/clusterlogforwarder_webhook.go
```diff
@@ -0,0 +1,132 @@
...
+	clf, ok := obj.(*obs.ClusterLogForwarder)
+	if !ok {
+		return nil, fmt.Errorf("expected ClusterLogForwarder, got %T", obj)
+	}
+
+	if !internalobs.Outputs(clf.Spec.Outputs).NeedServiceAccountToken() {
+		return nil, nil
+	}
+
+	req, err := admission.RequestFromContext(ctx)
```
jcantrill, line 57: Why do we get userinfo here but also add it to an annotation in the Default function?  Do we know if this "user" is the API server or the actual person who created the CLF?

internal/validations/observability/validate_permissions.go
```diff
@@ -151,6 +164,42 @@ func gatherPipelineInputs(clf obs.ClusterLogForwarder) (sets.String, bool) {
 	return *inputTypes, noOfReceivers > 0
 }
 
+// validateModifierCanUseSA checks that the user who last modified the CLF has permission to use the referenced SA.
+// Legacy CLFs created before the webhook was deployed will not have the modifier annotation.
```
jcantrill: This validation runs post admission.  The webhook runs on 'create' and 'update' which I don't believe either would run on upgrade of the operator.  I think if we went this route it would be acceptible to validate permission on update and reject; this is something we should be able to note in the release notes.  I'm not certain any of the following would be required then because it would already be validated in the webhook

internal/webhook/clusterlogforwarder_webhook.go
vparfonov: - Mutating (Default): Stores the user identity as the `resource-owner` annotation for recording who last created/modified the CLF. This is persisted on the resource.
- Validating (validateSAUsage): Reads UserInfo directly from the `AdmissionReview` to perform the SAR check. It does NOT read the annotation.
The `AdmissionReview.Request.UserInfo` is the real authenticated user (the person who ran `oc` command), not the API server's identity.

internal/validations/observability/validate_permissions.go
vparfonov: removed the reconciler checking

internal/webhook/clusterlogforwarder_webhook.go
vparfonov: done

internal/constants/annotations.go
vparfonov: "resource-owner" is good one.
@jcantrill if we remove check on reconcile, is it make sense to record user id who last created/modified the CLF in annotation?

internal/webhook/clusterlogforwarder_webhook.go
vparfonov: My main concern is that `non-cluster-admin` CLF creates will break without new RBAC. The webhook requires `use` on the referenced `ServiceAccount`. So anyone who is not cluster-admin and without a role granting `use` will start getting `not authorized` **after upgrade**.
@jcantrill

internal/constants/annotations.go
vparfonov: I removed checking on reconcile

internal/validations/observability/validate_permissions.go
vparfonov: My main concern is that non-cluster-admin CLF creates will break without new RBAC. The webhook requires use on the referenced ServiceAccount. So anyone who is not cluster-admin and without a role granting use will start getting not authorized after upgrade.
Release notes **must** cover this, probably with a _breaking changes mark_.

@jcantrill

vparfonov: @r2d2rnd Maybe you should know about this too and it would be interesting to hear your opinion

PR #3385, opened 2026-08-04
fix(collector): only declare hostPath volumes when needed (LOG-9713)

vparfonov:
### Description

Only declare hostPath volumes when their input sources are configured.

The collector pod spec unconditionally declares `hostPath` volumes for `audit API`, `journal` and `OVN` logs even when those inputs are not enabled. 

This fix moves the volume declarations inside the same input-source conditions that control the `volumeMounts`, so app-only collectors don't declare unnecessary `hostPath` access.

/cc @Clee2691 
/assign @jcantrill 

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA: https://redhat.atlassian.net/browse/LOG-9713
- Enhancement proposal:

### DIFF
--- a/internal/collector/collector.go
+++ b/internal/collector/collector.go
@@ -167,16 +167,39 @@ func (f *Factory) NewPodSpec(trustedCABundle *v1.ConfigMap, spec obs.ClusterLogF
 	}
 
 	if f.isDaemonset {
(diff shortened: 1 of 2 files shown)


jcantrill (commented):
/lgtm
/approve

PR #3406, opened 2026-08-17
[release-6.5] fix(lfme): set -secureMetrics=false on exporter daemonset

Clee2691:
### Description

This PR sets the `-secureMetrics` arg on the `LogFileMetricExporter` container  to `false`.

/cc @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://redhat.atlassian.net/browse/LOG-9872

### DIFF
--- a/internal/metrics/logfilemetricexporter/factory.go
+++ b/internal/metrics/logfilemetricexporter/factory.go
@@ -114,7 +114,7 @@ func newLogMetricsExporterContainer(exporter loggingv1a1.LogFileMetricExporter, 
 	}
 	exporterContainer.Command = []string{"/bin/bash"}
 	exporterContainer.Args = []string{"-c",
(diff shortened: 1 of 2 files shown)


config/rbac/kustomization.yaml
```diff
@@ -21,4 +21,4 @@ resources:
 # More info: https://book.kubebuilder.io/reference/metrics.html
 #- metrics_auth_role.yaml
 #- metrics_auth_role_binding.yaml
```
vparfonov, line 23: do we need uncoment this ?
https://github.com/openshift/cluster-logging-operator/blob/release-6.6/config/rbac/kustomization.yaml#L23

Clee2691: No, these are for the collector metrics auth, not the LFME metrics auth. These changes are scoped to the LFME only, not the collector.

Clee2691: not relevant anymore

PR #3429, opened 2026-08-25
fix(collector): check resource existence before deletion

Clee2691:
### Description
Avoid issuing `DELETE` API calls against non-existent deployments and daemonsets during reconciliation. 

Previously, `Remove()` and `RemoveDeployment()` would blindly call `Delete`, causing the API server
to return `404`s every reconcile cycle and inflating audit log noise. Now check with Get first - return early if NotFound, only `Delete` if the resource actually exists.

/cc @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://redhat.atlassian.net/browse/LOG-9810

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of collector workloads when resources are absent or require removal.
  * Prevented unnecessary delete operations for missing DaemonSets and Deployments.
  * Added clearer error handling when workload lookup or deletion fails.
  * Ensured stale collector resources are removed using the appropriate workload type.
  * Improved reliability when resources disappear during cleanup, avoiding unnecessary failures.

### DIFF
--- a/internal/collector/collector_test.go
+++ b/internal/collector/collector_test.go
@@ -1,8 +1,10 @@
 package collector
 
 import (
+	"context"
 	"os"
 	"path"
+	"sync/atomic"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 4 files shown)


jcantrill:
```
Verification Summary

PR: cluster-logging-operator#3429 — fix(collector): check resource existence before deletion
JIRA: LOG-9810
Cluster: api.d3cru-uytx7-os2.6hoh.p3.openshiftapps.com (ROSA, OCP 4.21.32)
Operator version: built from PR HEAD d2cf00de9, image tag pr-3429 (deployed via make clean deploy)

Test Setup

The bug: CLO issues a DELETE deployments/<name> every reconcile cycle even though the collector runs as a DaemonSet, so the API server returns a 404 each time (audit-log noise). Reproduced by deploying a Vector DaemonSet collector via a ClusterLogForwarder — this is the exact path (RemoveStaleWorkload → RemoveDeployment) that blindly deleted the non-existent Deployment on every reconcile.

Steps Performed

1. Built and deployed the operator from PR #3429 (make clean deploy); confirmed pod running on image :pr-3429.
2. Created SA log-collector + collection RBAC and a daemonset-based ClusterLogForwarder (instance).
3. Confirmed the collector deploys as a DaemonSet with no Deployment present — the stale-workload removal target.
4. Forced 3 reconcile cycles and inspected operator logs for delete activity/errors.
5. Ran the PR's unit tests, which deterministically count DELETE API calls (the audit-log symptom cannot be observed directly — ROSA hides the control plane / master node audit logs).

Results

┌─────┬───────────────────────────────────────────────────────────────┬────────┐
│  #  │                             Test                              │ Result │
├─────┼───────────────────────────────────────────────────────────────┼────────┤
│ 1   │ PR operator deploys and runs from PR image                    │ PASS   │
├─────┼───────────────────────────────────────────────────────────────┼────────┤
│ 2   │ Collector deploys as DaemonSet; no Deployment exists          │ PASS   │
├─────┼───────────────────────────────────────────────────────────────┼────────┤
│ 3   │ CLF reconciles to Ready=True across 3 forced reconciles       │ PASS   │
├─────┼───────────────────────────────────────────────────────────────┼────────┤
│ 4   │ No delete-related errors in operator logs                     │ PASS   │
├─────┼───────────────────────────────────────────────────────────────┼────────┤
│ 5   │ Unit test: 0 DELETE calls when deployment absent              │ PASS   │
├─────┼───────────────────────────────────────────────────────────────┼────────┤
│ 6   │ Unit test: 1 DELETE when deployment exists                    │ PASS   │
├─────┼───────────────────────────────────────────────────────────────┼────────┤
│ 7   │ Unit test: no error when deleted mid-flight (Get→Delete race) │ PASS   │
└─────┴───────────────────────────────────────────────────────────────┴────────┘

Conclusion

Verified. The fix issues a Get before Delete and returns early on NotFound, eliminating the repeated DELETE deployments/<name> → 404 audit events described in LOG-9810. Unit tests deterministically confirm zero DELETE calls when the workload is absent (the exact expected result in the ticket), and the fixed operator reconciles a real DaemonSet-based collector cleanly with no delete errors. Direct audit-log confirmation was not possible because this ROSA cluster does not expose the control-plane/master audit logs; the unit-test delete-count assertion covers that symptom deterministically instead.
```
/label verified

Issue #3430, opened 2026-08-26
Allow the `parse` filter to merge parsed JSON fields into the event root


GuillerVks:
### Describe the bug / compatibility gap

This issue concerns migration from a legacy OpenShift logging collector to OpenShift Logging with Vector and the `observability.openshift.io/v1` `ClusterLogForwarder` API. The behavior was reproduced with OpenShift Logging 6.2.12.

A common legacy ingestion contract is:

- Plain-text application logs remain in `message`.
- Valid JSON application logs are parsed and their fields are indexed at the event root.

The current `parse` filter always places valid JSON under `structured` and removes `message`. There is no supported option to select the event root as the parse target.

This changes existing field paths, for example:

```text
description           -> structured.description
log_level             -> structured.log_level
```

For established logging deployments, this can create incompatible schemas for the same sources before and after migration. Existing mappings, searches, dashboards, alerts, detection rules, and downstream consumers may depend on the root-level fields.

We would like to contribute an implementation, but because this changes the public API we are opening this issue first to agree on the expected API and collision semantics.

### Environment

- OpenShift Logging Operator: `6.2.12`
- ClusterLogForwarder API: `observability.openshift.io/v1`
- Collector: Vector
- Output type used for reproduction: Elasticsearch

The `ClusterLogForwarder` reports `Authorized=True`, `Valid=True`, and `Ready=True`.

### Relevant generated configuration

The operator generates this transformation for the application pipeline:

```vrl
if .log_source == "container" {
    parsed, err = parse_json(.message)
    if err == null {
        .structured = parsed
        del(.message)
    }
}
```

### Expected behavior

The `parse` filter should optionally support merging a parsed JSON object into the event root while retaining the current behavior as the default for backward compatibility.

The option should also define:

- Whether the original `message` is preserved after successful parsing.
- How collisions with collector-owned fields are handled.
- What happens when the JSON payload is valid but is not an object.
- How parse and collision failures are exposed.

A safe collision behavior should preserve collector metadata such as `@timestamp`, `kubernetes`, `openshift`, `log_type`, and `log_source` rather than allowing application JSON to overwrite it.

The exact API shape should follow maintainer guidance. Conceptually, the required capabilities are equivalent to selecting the root as the parse target, preserving `message`, and preserving existing fields on collision.

### Actual behavior

Given this application log:

```json
{"description":"JSON test trace","log_level":"INFO","sequence":1}
```

The current filter produces fields under `structured`:

```json
{
[...]

PR #3431, opened 2026-08-26
Add root target options to the parse filter

GuillerVks:
### Description

Add optional configuration to the `parse` filter for merging parsed JSON objects into the event root. Existing record fields take precedence on collision, valid non-object JSON remains in `message`, and `preserveMessage` can retain the original JSON text.

The existing behavior remains unchanged when `parse` configuration is omitted: parsed JSON is stored under `structured` and the original message is removed.

This is an initial implementation for discussion. The API shape and collision semantics can be adjusted based on maintainer feedback.

The change includes generated CRD, CSV and API documentation updates, unit coverage for generated VRL, and functional coverage for root fields, collisions, message retention, and non-object JSON.

/cc vparfonov
/assign jcantrill

### Links

- Depending on PR(s):
- GitHub issue: Related: #3430
- JIRA:
- Enhancement proposal:


jcantrill:
The behavior of the parse filter as-is was intentionally designed to normalize log records so the various log types would have a more consistent envelope which is ViaQ; there are some variances to this because of the history of the project.  Given the intentional design decision, we do not intend to accept this PR.

In general, there are plans in motion to move toward OpenTelemetry and OTLP which means we are reducing any work related to modifying the ViaQ data model

GuillerVks:
Thank you for the clarification. While I understand the decision and the direction toward OpenTelemetry, I do not believe this is an acceptable outcome for existing customers who have relied on this previously supported behavior in production environments.

Changes of this nature should be accompanied by a supported migration path or a compatible alternative to minimize the impact on existing deployments and integrations.

As this PR is not expected to move forward, we will continue the discussion through the appropriate customer support channels to address the migration challenges and impact on existing customers.

jcantrill:
Closing the PR as WONTDO for the reasons identified above.

> I do not believe this is an acceptable outcome for existing customers who have relied on this previously supported behavior in production environments.

For posterity, Logging 6.0.0 was released in Sept. 2024 and there have been 6 additional releases since then.   I would not consider this change a "surprise" given the intent described above.  Logging 6.x intentionally introduced a new API to address  various inconsistencies that were introduced over time from the previous API.

PR #3440, opened 2026-08-28
fix(vector): fix internal_metrics typo and set sane 30s scrape interval

saule1508:
### Description

The Vector collector DaemonSet consistently consumes 4–6 GiB of RAM per node (and up to 10 GiB) on our cluster. By looking at the prometheus metrics, the source of the memory consumption is the component internal_metrics which generates upward of 3Mib/s (compared to 0.2 MiB/s forthe application logs).

This is because the generated `vector.toml` completely omits `scrape_interval_secs`, forcing Vector to poll its entire registry on its internal 1-second default

This PR corrects the typo in the key name (ScrapeIntervalSeconds instead of ScrapIntervalSeconds) and in the serialization tags (secs instead of seconds) and sets a sane, explicit default of 30 seconds inside the generator.

/cc jcantrill
/assign alanconway

### Links

- Depending on PR(s):
- GitHub issue:
- JIRA:
- Enhancement proposal:

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the internal metrics scrape interval setting name.
  * Internal metrics now scrape every 30 seconds by default across generated configurations.
  * Prometheus output authentication explicitly uses the GET HTTP method.

### DIFF
--- a/internal/generator/vector/api/config_test.go
+++ b/internal/generator/vector/api/config_test.go
@@ -26,8 +26,8 @@ type = "file"
 path = "/var/run/ocp-collector/secrets"
 
 [sources.internal_metrics]
+scrape_interval_secs = 30
 type = "internal_metrics"
-scrape_interval_seconds = 2
 
 [transforms.bar]
 inputs = ["internal_metrics"]
@@ -54,8 +54,8 @@ secret:
     path: "/var/run/ocp-collector/secrets"
 sources:
   internal_metrics:
+    scrape_interval_secs: 30
     type: "internal_metrics"
-    scrape_interval_seconds: 2
 transforms:
   bar:
     inputs: ["internal_metrics"]
--- a/internal/generator/vector/api/sources/internal_metrics.go
+++ b/internal/generator/vector/api/sources/internal_metrics.go
@@ -5,8 +5,8 @@ import (
 )
 
 type InternalMetrics struct {
-	Type                 types.SourceType `json:"type" yaml:"type" toml:"type"`
-	ScrapIntervalSeconds uint             `json:"scrape_interval_seconds,omitempty" yaml:"scrape_interval_seconds,omitempty" toml:"scrape_interval_seconds,omitempty"`
+	Type                  types.SourceType `json:"type" yaml:"type" toml:"type"`
+	ScrapeIntervalSeconds uint             `json:"scrape_interval_secs,omitempty" yaml:"scrape_interval_secs,omitempty" toml:"scrape_interval_secs,omitempty"`
 }
 
 func (i *InternalMetrics) SourceType() types.SourceType {
@@ -15,6 +15,7 @@ func (i *InternalMetrics) SourceType() types.SourceType {
 
 func NewInternalMetrics() types.Source {
 	return &InternalMetrics{
-		Type: types.SourceTypeInternalMetrics,
+		Type:                  types.SourceTypeInternalMetrics,
+		ScrapeIntervalSeconds: 30,
 	}
 }
--- a/internal/generator/vector/conf/complex.toml
+++ b/internal/generator/vector/conf/complex.toml
@@ -13,6 +13,7 @@ type = "directory"
 path = "/var/run/ocp-collector/secrets"
 
 [sources.internal_metrics]
+scrape_interval_secs = 30
 type = "internal_metrics"
 
 [sources.input_audit_host]
--- a/internal/generator/vector/conf/complex_http_receiver.toml
+++ b/internal/generator/vector/conf/complex_http_receiver.toml
@@ -13,6 +13,7 @@ type = "directory"
 path = "/var/run/ocp-collector/secrets"
 
 [sources.internal_metrics]
+scrape_interval_secs = 30
 type = "internal_metrics"
 
 [sources.input_audit_host]
--- a/internal/generator/vector/conf/container.toml
+++ b/internal/generator/vector/conf/container.toml
@@ -14,6 +14,7 @@ type = "directory"
 path = "/var/run/ocp-collector/secrets"
 
 [sources.internal_metrics]
+scrape_interval_secs = 30
 type = "internal_metrics"
 
 [sources.input_myinfra_container]


vparfonov:
@jcantrill fixing name of config parameter is totally correct [scrape_interval_secs](https://vector.dev/docs/reference/configuration/sources/internal_metrics/#scrape_interval_secs), but not sure about value in `30 seconds`

saule1508:
Hi @vparfonov, thank you for the comment ! Currently the scrape_interval_secs is omitted, and it default to one second (https://vector.dev/docs/reference/configuration/sources/internal_metrics/) but due to the volume of metrics computed in our openshift cluster collector nodes (which is not a specially big cluster) this generates massive memory consumption.
I think 30 seconds make more sense because anyway prometheus is going to scrape every 30 seconds in our environment and very likely in most environments.

jcantrill:
@saule1508 do you have a graphic or other image to identifies the memory savings after this change?

saule1508:
@jcantrill Spent the day measuring this properly. Test cluster, Logging 6.6, 7 collectors, external Loki, worker nodes have 128 CPU cores.

On this PR specifically:

vector_internal_metrics_cardinality = 359 per collector, so at the 1s default each collector emits ∼359 events/s.

* sum(rate(vector_component_received_bytes_total{component_id="internal_metrics"}[5m])): 2.7 MB/s → 105 KB/s (∼26×) after setting scrape_interval_secs = 30.

For scale, that 2.7 MB/s was more than all real log sources on the same collectors combined — application, audit and infrastructure inputs are each in the 100–500 KB/s range.
image


* sum(rate(vector_component_sent_bytes_total{component_type="prometheus_exporter"}[5m])) is unchanged across the change.

So the collector was producing ∼26× more metric snapshots than anything consumed, each additionally cloned through the generated add_nodename_to_metric remap. Our ServiceMonitors (instance, minimal-instance, telemetry-instance) all inherit the platform-default 30s scrape interval.

Correction to my original report: I opened this attributing our collector memory usage to internal metrics. I believe that was mostly wrong. Our actual memory problem is that collector memory scales with node core count, not log volume — ∼810 MB on 8-core nodes vs 5–6.5 GB on 128-core nodes in the same cluster, with memory.peak hitting 9.7 GiB against a 10 GiB limit.

Setting VECTOR_THREADS=8 dropped those 128-core collectors from ∼6 GB to under 1 GB. I'll report that separately once I've confirmed the mechanism — I suspect it's simply that the collector container has no CPU limit (cpu.max = max), so Vector sizes its runtime to all 128 host cores, in which case it may be fixable via spec.collector.resources.limits.cpu with no code change. Testing that tomorrow.

I do see a memory reduction after adding the scrape_interval_secs = 30, not as compelling as the one from the vector_threads. I will do a proper test and post a chart tomorrow.

I am not sure the value of 30s is the best, since the default scape interval is 30s it might make sense, or could be a bit lower, maybe 15s ? I can test other values if it helps. Will work on it tomorrow

jcantrill:
> Setting VECTOR_THREADS=8 dropped those 128-core collectors from ∼6 GB to under 1 GB. I'll report that separately once I've confirmed the mechanism — I suspect it's simply that the collector container has no CPU limit (cpu.max = max), so Vector sizes its runtime to all 128 host cores, in which case it may be fixable via spec.collector.resources.limits.cpu with no code change. Testing that tomorrow.

If the CLF defines no collector requests/limits, the operator defaults to the following values:
```
	DefaultRequestMemory = resource.MustParse("64Mi")
	DefaultRequestCpu    = resource.MustParse("500m")
	DefaultLimitMemory   = resource.MustParse("2048Mi")
	DefaultLimitCpu      = resource.MustParse("6000m")
```
This means it should at most consume 6 cores at 2G each.  If you provide any values, it takes what is spec'd.  Vector is agressive and will consume as much is available.  If the default's are not being respected then that is a bug and should be logged accordingly

saule1508:
@jcantrill 

Following up on your comment about default limits, I have found strong evidence of a bug where the default CPU limit is not being applied.

On a test cluster, I examined a collector pod running on a 128-cores worker node.

* The spec has no CPU limit

oc describe command shows:

```
 oc describe pod -n openshift-logging instance-4wkgq
 ...
Command:
      sh
    Args:
      /usr/bin/run-vector.sh
    State:          Running
      Started:      Mon, 31 Aug 2026 12:41:59 +0000
    Ready:          True
    Restart Count:  4
    Limits:
      memory:  10Gi
    Requests:
      cpu:     200m
      memory:  2Gi

````

* But vector uses all cores

but when I look into the container at the number of tasks I have 177, so it will be 128 threads for sure

```
[root@instance-4wkgq bin]# ls /proc/1/task/ | wc -l
177
```

This confirms that the operator is not enforcing the expected default CPU limit of 6000m, causing Vector to consume excessive resources on high-core nodes.

* workaround and solution

I applied the following limit (via terraform)

```
      collector = {
        resources = {
          limits = {
            memory = "10Gi"
            cpu    = "8"
          }
          requests = {
            cpu    = "200m"
            memory = "2Gi"
          }
        }
      }
```

and now the threads are only 36 and the memory dropped to 500MB, in the collector pod:

```
[root@instance-nm8cg bin]# ls /proc/1/task/ | wc -l
36
[root@instance-nm8cg bin]# grep VmRSS /proc/1/status
VmRSS:    449756 kB
```

**Comment by saule1508** [comment]
@jcantrill do you want me to create a new issue regarding the default cpu limit not being respected ? I can also look at the code to see if I can find the error ?
At last I think I have solve my issue of memory by explicitly setting the cpu limit, which is great

**Comment by jcantrill** [comment]
> I have found strong evidence of a bug where the default CPU limit is not being applied.

What does the spec resources of the DaemonSet say?  The operator creates the DS and the information is passed accordingly to the container.  I can not make any judgement without inspection of:

* the CLF.spec.collector.resourcces
* the daemonset.yaml

The information about does identify the point you are trying to make but it is not the entire story

**Comment by jcantrill** [comment]
Tested locally with ClusterLogForwarder spec:
```
  spec:
    managementState: Managed
    outputs:
    - lokiStack:
        authentication:
          token:
            from: serviceAccount
        target:
          name: lokistack-dev
          namespace: openshift-logging
      name: loki-stack
      tls:
        ca:
          configMapName: openshift-service-ca.crt
          key: service-ca.crt
      type: lokiStack
    pipelines:
    - inputRefs:
      - application
      name: app
      outputRefs:
      - loki-stack
    serviceAccount:
      name: logcollector
```
```
$ oc describe ds/collector 
Name:           collector
. . .
Labels:         app.kubernetes.io/component=collector
                app.kubernetes.io/instance=collector
                app.kubernetes.io/managed-by=cluster-logging-operator
                app.kubernetes.io/name=vector
                app.kubernetes.io/part-of=cluster-logging
                app.kubernetes.io/version=6.6.0
                vector.dev/exclude=true
. . .
  Containers:
   collector:
    Limits:
      cpu:     6
      memory:  2Gi
    Requests:
      cpu:     500m
      memory:  64Mi
    Environment:
      COLLECTOR_CONF_HASH:      1dab91a8a88ed489db858b95cbd5dbf5
      K8S_NODE_NAME:             (v1:spec.nodeName)

```
If CLF specs no resources then the defaults are applied.  If the CLF specs ANY resources then the values provided in the spec are honored as-is

PR #3449, opened 2026-09-01
test(multilineexception): add functional test for incorrectly concatenates independent JSON log records.

vparfonov:
### Description

Add test verifying that independent log records are NOT merged when the first line contains an exception keyword.
This reproduces the production scenario where a container emits single-line JSON logs, one log has a exception-related term (e.g. SecretConfigException) in its message field, and subsequent normal logs are greedily consumed by the catch-all continuation rule.
    
The test sends 3 independent messages:
1. `"SecretConfigException: ..."` - triggers exception detection
2. `"Request completed in 235ms"` - regular log
3. `"Processing batch 42 records"` - regular log
    
Expected: 3 separate log events in the output
Buggy behavior: all 3 merged into 1 event
    

/cc @Clee2691 
/assign @jcantrill 

### Links

- Depending on PR(s): https://github.com/ViaQ/vector/pull/305
- GitHub issue:
- JIRA: https://redhat.atlassian.net/browse/LOG-9998
- Enhancement proposal:

### DIFF
--- a/test/functional/filters/multilineexception/multiline_exception_test.go
+++ b/test/functional/filters/multilineexception/multiline_exception_test.go
@@ -134,4 +134,42 @@ created by main.main
 		}),
 	)
 
+	// LOG-9963: Independent log records where the first line contains an exception
+	// keyword must NOT cause subsequent unrelated lines to be merged.
+	// Reproduces the production scenario: a container emits single-line JSON logs;
+	// one log has an exception-related term (e.g. SecretConfigException) in its
+	// message field, and subsequent normal logs are greedily consumed by the
+	// catch-all continuation rule.
+	It("should not merge independent log records when first line contains an exception keyword", func() {
+		testruntime.NewClusterLogForwarderBuilder(framework.Forwarder).
+			FromInput(obs.InputTypeApplication).
+			WithMultilineErrorDetectionFilter().
+			ToHttpOutput()
+
+		Expect(framework.Deploy()).To(BeNil())
+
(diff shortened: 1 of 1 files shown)


test/functional/filters/multilineexception/multiline_exception_test.go
```diff
@@ -134,4 +134,47 @@ created by main.main
...
+
+		var buffer []string
+		for _, msg := range messages {
+			crioLine := functional.NewCRIOLogMessageWithStream(timestamp, constants.STDOUT, msg, false)
+			buffer = append(buffer, crioLine)
+		}
+		appNamespace = framework.Pod.Namespace
+		Expect(framework.WriteMessagesToNamespace(strings.Join(buffer, "\n"), appNamespace, 1)).To(Succeed())
+
+		for _, output := range framework.Forwarder.Spec.Outputs {
```
jcantrill, line 167: Don't need to spin through all the outputs if we know there is only one

test/functional/filters/multilineexception/multiline_exception_test.go
```diff
@@ -134,4 +134,47 @@ created by main.main
...
+			outputType := output.Type
+			raw, err := framework.ReadRawApplicationLogsFrom(string(outputType))
```
jcantrill, line 169: We don't use this often enough but the raw log read and parsing are captured in "ReadApplicationLogsFrom" and can replace those here

vparfonov:
> /test functional-target

this fail depends on https://github.com/ViaQ/vector/pull/305

PR #3451, opened 2026-09-02
feat(elasticsearch): enhance Elasticsearch output to support multiple endpoints

Clee2691:
### Description

This PR enhances the `ElasticSearch` output to support multiple endpoints, allowing Vector to distribute events across multiple endpoints using P2C (Power of Two Choices) load balancing with automatic failover. 

The `ClusterLogForwarder` API currently only supports a single URL for Elasticsearch outputs via the embedded `URLSpec` struct, preventing users from configuring multi-node Elasticsearch clusters for high availability and load distribution.

/cc @vparfonov 
/assign @jcantrill 

### Links
- JIRA: https://redhat.atlassian.net/browse/LOG-9994
- Design Spec: https://github.com/rhobs/redhat-openshift-logging-workspace/pull/15

## Summary by CodeRabbit

- **New Features**
  - Elasticsearch outputs now support multiple endpoints for load balancing and automatic failover.
  - Configure endpoints alone or combine them with an optional primary URL.
  - Added support for multi-node Elasticsearch output scenarios.

- **Bug Fixes**
  - URL and TLS validation now checks every configured Elasticsearch endpoint.
  - Network policy and port handling include all configured endpoints.

- **Documentation**
  - Updated Elasticsearch output documentation and schemas with endpoint configuration, precedence, failover behavior, and expanded TLS profile details.


test/functional/outputs/elasticsearch/forward_to_elasticsearch_test.go
```diff
@@ -319,4 +320,81 @@ var _ = Describe("[Functional][Outputs][ElasticSearch] Logforwarding to ElasticS
...
+		// the combined log count. This avoids a sequential 5-minute timeout on a
+		// node that received zero logs due to non-deterministic load balancing.
+		getLogsFromAllEndpoints := func() int {
+			var logs1, logs2 []string
+			var wg sync.WaitGroup
+
+			queryNode := func(result *[]string, nodeName, index string, opts ...functional.Option) {
+				defer wg.Done()
+				defer GinkgoRecover()
+				*result, _ = framework.GetLogsFromElasticSearchIndex(nodeName, index, opts...)
```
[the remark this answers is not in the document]
Clee2691: Not a real issue because the point of multiple endpoints is that Vector can send to one or both depending on availability. This would be a test for Vector itself

api/observability/v1/output_types.go
```diff
@@ -676,8 +676,32 @@ type ElasticsearchTuningSpec struct {
 	Compression string `json:"compression,omitempty"`
 }
 
+// EndpointURL is a URL to an Elasticsearch endpoint.
+// +kubebuilder:validation:XValidation:rule="isURL(self)", message="invalid URL"
+type EndpointURL string
+
+// Elasticsearch provides optional extra properties for `type: elasticsearch`
+//
+// +kubebuilder:validation:XValidation:rule="has(self.url) || self.endpoints.size() > 0", message="URL or endpoints required"
```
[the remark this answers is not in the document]
Clee2691: fixed by removing `self == ''` from the URL field validtion

jcantrill:
blocked by https://github.com/openshift/cluster-logging-operator/pull/3458 with minio fix

internal/generator/vector/output/elasticsearch/elasticsearch.go
```diff
@@ -19,6 +19,17 @@ import (
 	"github.com/openshift/cluster-logging-operator/internal/utils"
 )
 
+func mergeEndpoints(es *obs.Elasticsearch) []string {
```
jcantrill, line 22: Consider sorting the outcome so we don't thrash the config on various reconciliations, or maybe it is better form to sort in the constructor...

internal/network/ports_test.go
```diff
@@ -137,6 +137,36 @@ var _ = Describe("Network Ports", func() {
...
+				obs.OutputSpec{
+					Type: obs.OutputTypeElasticsearch,
+					Elasticsearch: &obs.Elasticsearch{
+						Endpoints: []obs.EndpointURL{
+							"https://es1.example.com:9200",
+							"https://es2.example.com:9300",
+						},
+					},
+				},
+				makeTCPPorts(9200, 9300),
```
jcantrill, line 154: Maybe this should sort the ports to ensure the config always is consistent

Clee2691: This doesn't matter here because the test uses `ConsistOf` which doesn't care about ordering

internal/generator/vector/output/elasticsearch/elasticsearch.go
Clee2691: This won't happen because these are string slices. The outcome is the same regardless. It is not a map.

jcantrill: What if the order is returned differently from etcd? Are we guaranteed how it goes in is how it comes out?

internal/network/ports_test.go
jcantrill: I'm thinking not of the test but any object that might change because of order change.  if it doesn't go into a config then it probably does not matter

jcantrill:
@Clee2691 finding while validating:

 

> ⚠️ CEL error message for the "neither provided" case leaks a CEL evaluation artifact. When endpoints is absent, the rule has(self.url) || self.endpoints.size() > 0 evaluates self.endpoints on a missing key, so the user sees no such key: endpoints evaluating rule: URL or endpoints required instead of a clean URL or endpoints required. It still correctly rejects the CR — this is a message-quality issue only. A guarded rule (has(self.url) || (has(self.endpoints) && self.endpoints.size() > 0)) would surface the intended message alone. Worth a heads-up since the acceptance criteria call out this exact validation.

internal/network/ports_test.go
Clee2691: This function is just a test function. It isn't used to generate an actual config

internal/generator/vector/output/elasticsearch/elasticsearch.go
Clee2691: The way the endpoints are "compiled" is through a slice. The order of a slice stays the same and the way it processes the items in the slice will be the same, it will never be out of order.

It would only return out of order elements if the endpoints API was a set or a map, which it isn't because it is an `Endpoint` slice.

The only way the order would be different is if the user explicitly changes the order in the API itself

